Issue #110, opened 2018-07-11
try/catch new WebSocket to bubble errors


jacogr:
https://github.com/polkadot-js/api/blob/master/packages/api-provider/src/ws/connect.ts#L13

Currently when the construction fails, the UI just has a blank page.

https://github.com/polkadot-js/apps/issues/132

PR #114, opened 2018-07-11
Trap exceptions on WebSocket create

jacogr:
Closes https://github.com/polkadot-js/api/issues/110


jacogr:
Will get https://github.com/polkadot-js/apps/issues/132 into a state where at least the UI is displayed, instead of a blank page. (It does not work around the issue, nothing can really be done atm)

PR #116, opened 2018-07-23
Revert params: null for jsonrpc (fixed  in Node)

jacogr:


jacogr:
Ref https://github.com/paritytech/jsonrpc/pull/293

Issue #111, opened 2018-07-11
Class-ify provider, HTTP & Ws


jacogr:
Convert to ES6 classes in the same fashion as https://github.com/polkadot-js/client/pull/132

https://github.com/polkadot-js/api/blob/master/packages/api-provider/src/http/index.ts
https://github.com/polkadot-js/api/blob/master/packages/api-provider/src/ws/index.ts

Not the only 2 places in the API layer, but a start. (Bonus convert all the closures with self)

PR #119, opened 2018-08-02
WIP: Add example implementation of @polkadot/api. Errors encountered

ltfschoen:
Errors shown after running `yarn; yarn run build`:

api-example/src/index.ts:1:17 - error TS2307: Cannot find module '@polkadot/api'.

1 import Api from '@polkadot/api';
                  ~~~~~~~~~~~~~~~

api-example/src/index.ts:6:18 - error TS2350: Only a void function can be called with the 'new' keyword.

6 const provider = new HttpProvider('http://127.0.0.1:9933');
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

api-example/src/index.ts:6:18 - error TS7009: 'new' expression, whose target lacks a construct signature, implicitly has an 'any' type.

6 const provider = new HttpProvider('http://127.0.0.1:9933');
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

api-example/src/index.ts:11:10 - error TS7006: Parameter 'header' implicitly has an 'any' type.

11   .then((header) => console.log(header))
            ~~~~~~

api-example/src/index.ts:12:11 - error TS7006: Parameter 'error' implicitly has an 'any' type.

12   .catch((error) => console.error(error));
             ~~~~~


jacogr:
So going through step by step, the errors gives you exactly what you need to resolve -

- `api-example/src/index.ts:1:17 - error TS2307: Cannot find module '@polkadot/api'.`

It cannot find the api - so the module resolution is wrong. So it needs to be added, in exactly the same way you did api-example here. (It is not there since nothing actually depended on it previously)

- api-example/src/index.ts:6:18 - error TS2350: Only a void function can be called with the 'new' keyword.

Seems the README is completely wrong and not updated between earlier changes. Looking at HttpProvider, it is not a class, so it should just be `const provider = httpProvider('http://127.0.0.1:9933')`. It relates to this https://github.com/polkadot-js/api/issues/111 (It should be a class). So Update, fix README or we need the conversion and all the downstream changes.

- api-example/src/index.ts:11:10 - error TS7006: Parameter 'header' implicitly has an 'any' type.

No type has been defined. (This is TS, it doesn't like unknowns)

- api-example/src/index.ts:12:11 - error TS7006: Parameter 'error' implicitly has an 'any' type.

As previous one.

ltfschoen:
Thanks I'll do #111 before updating this branch

PR #139, opened 2018-08-18
feat: Add support for InputNumber in ui-app.

ltfschoen:
Required for polkadot-js/apps PR #144, specifically commit https://github.com/polkadot-js/apps/pull/221/commits/f2da789f3654a60ba1f8a4d4819c770cb7121c1e

Note: I understand we need to remove poc-1, but have left it in since may be performed all at once in a subsequent commit

### DIFF
--- a/packages/type-extrinsics/src/democracy.ts
+++ b/packages/type-extrinsics/src/democracy.ts
@@ -12,7 +12,7 @@ const propose: CreateItemOptions = {
   description: 'Propose',
   params: [
     param('proposal', 'Proposal'),
-    param('value', 'Balance')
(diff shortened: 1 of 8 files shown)


packages/type-extrinsics/src/democracy.ts
```diff
@@ -12,7 +12,7 @@ const propose: CreateItemOptions = {
   description: 'Propose',
   params: [
     param('proposal', 'Proposal'),
-    param('value', 'Balance')
+    param('value', 'InputNumber')
```
jacogr: No. The Type is `Balance`, the component in the UI may be something else.

PR #153, opened 2018-09-14
Api & RxApi now created via new

jacogr:
basically 

```js
import Api from '@polkadot/api';
import RxApi from '@polkadot/api-rx';

const api = new Api();
const rx = new RxApi();
```


jacogr:
cc @chevdor

PR #157, opened 2018-09-17
Use & build gitbook from @polkadot/dev

jacogr:
- Included gitbook packages & build in the base (so available everywhere) - https://github.com/polkadot-js/dev/pull/140
- The above now also builds gitbook (when book.json is available) and copies the generated HTML to /docs (Where GH pages expects it)
- Exposed docs folder on GH pages via repo settings, once this is merged, it _should_ build, copy and be available

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 
 # @polkadot/api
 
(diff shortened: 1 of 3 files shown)


jacogr:
cc @chevdor

Issue #158, opened 2018-09-17
Gitbook does not render nested md files


jacogr:
So everything looks nice, e.g. https://polkadot.js.org/api/

However, from https://polkadot.js.org/api/api/ scrolling down to Classes and then clicking on `Api` does not quite work... Basically it seems like this https://github.com/polkadot-js/api/tree/master/docs/api/classes does not result in a corresponding `api.html` file generated from `api.md`. (While all the others at higher levels seem to be doing fine)

Since we copy across to docs in the build script, I also too a look at my local `_book` folder, same there, it is missing. :(

jacogr:
cc @chevdor 

I did ping @amaurymartiny he couldn't spot anything out of the ordinary either...

jacogr:
The issue is that those files are not included in SUMMARY.md, they are just linked. So gitbook does not convert them.

There is an open issue for this, and without it being closed, it seems from my vantage point that we -

- either need to have all those files added to the SUMMARY.md file (and if this is a manual process, it seems like a non-starter)
- not do gitbook for now (as nice as it is from a "o, cool" perspective)
- ... any other option?

At this point, unless it is a quick fix, not going to push for this at all. With https://github.com/polkadot-js/api/issues/154 in-place we have a first pass good enough version.

jacogr:
Amaury indicated that the `mdEngine: 'gitbook'` (in typedoc.js) renders via gitbook and automagically generates the SUMMARY.md files. However, in my case -

1. Removed/renamed book.json (it just complains about includes markdown)
2. Added the above config

And it generates those files. All empty... Comparing the config against his stuff where it does work. Thus far, no differences.

jacogr:
Progress... this config does seem to render...

```
module.exports = {
  exclude: ['**/*spec.ts'],
  excludeExternals: true,
  excludePrivate: true,
  hideGenerator: true,
  includes: './docs/guides',
  out: 'docs',
  module: 'commonjs',
  mdEngine: 'gitbook',
  stripInternal: 'true',
  theme: 'markdown'
}
```

Issue #154, opened 2018-09-17
API documentation does not render correctly


jacogr:
https://github.com/polkadot-js/api/blob/master/docs/api/classes/_index_.api.md

Seems like there is a missing "```" somewhere. Example doesn't render as example, text renders as, well, text.

And a link to the gitbook pages - https://polkadot.js.org/api/api/classes/_index_.api.html and this (looks) a bit worse https://polkadot.js.org/api/api-provider/classes/_ws_index_.wsprovider.html

jacogr:
cc @chevdor

jacogr:
It _may_ be this https://github.com/polkadot-js/api/blob/master/packages/api/src/index.ts#L23

(However, we are closing them on L30, so clueless)

jacogr:
Re-opened, the linked changes are not changing the source, only the generated doc. To make it stick, the actual TS docs i the code is to be changed.

jacogr:
Stirpped the javascript & backticks from the source (as part of the poc-3 refactoring in #161), the example tag still does not render properly. Next step, trying to indent?

PR #159, opened 2018-09-19
Gitbook rendering updates

jacogr:
Ok, @chevdor is going to kill me...

- First off, this Closes https://github.com/polkadot-js/api/issues/158
- API documentation for classes now are rendering correctly
- Removed the type-* from the doc build with a .nodoc file

Now for the long version -

- Had to move all the docs/*.adoc to docs/*.md - gitbook does not include linked *.md files when using the adoc plugin, so all links to the api, api-provider, etc. *.md files were broken
- Even at this point, the nested SUMMARY is an issue, so had to manually add the Api, {Ws,Http}Provider and ApiRx classes to get them to link

What this does not do -

- No fix for #154 atm, it still shows up incorrectly


typedoc.js
```diff
@@ -1,19 +1,16 @@
...
-  stripInternal: 'true',
-  theme: 'markdown',
-  ignoreCompilerErrors: true,
   moduleResolution: 'node',
-  suppressExcessPropertyErrors: true,
-  suppressImplicitAnyIndexErrors: true
-}
+  mdEngine: 'gitbook',
+  stripInternal: 'true',
+  theme: 'markdown'
```
amaurymartiny: In package.json, you pass `--theme default --out docs/html`, which overwrites options here.

jacogr: Ahh, package.json is actually not used. (But I didn't want to remove it ... yet, it helps when you can actually test HTML from typedoc as well) 

This is a whole sprawling thing, effectively what is actually called is the build-docs script, which is shared - https://github.com/polkadot-js/dev/blob/master/packages/dev/scripts/polkadot-dev-build-docs.sh

(Hence build calling build-ts & build-docs)

amaurymartiny: Ah, right, this is the obscure part I missed.

jacogr: Obscure is the right word, yes.

Issue #161, opened 2018-09-19
Rework codecs & extrinsics for poc-3 compatibility


jacogr:
I cannot put it more succinctly :)

- Remote extrinsics & storage (the latter will still contain some, e.g. `:code`, basically all these https://github.com/paritytech/substrate/pull/764)
- Pull storage, extrinsics and events (this is new) from state_getMeta
- Fixup encoders and do it properly - currently they are all-over, some in primitives, some in params
- Make everything work....
- ... I missed some stuff

cc @amaurymartiny :)

amaurymartiny:
Yeah, I'll work on this.

jacogr:
This _could_ be useful - https://github.com/polkadot-js/api/pull/149

(No need to use anything in anything in there, but it is worth a look - it makes some attempt for saner Json/u8a encoding/decoding. Either way, the work here will replace the above, so it will never be merged as-is, nor make much progress form where it is. Feel free to adapt, copy or completely ignore as a science experiment)

jacogr:
New extrinsics format for poc-3 - https://github.com/paritytech/wiki/blob/master/Extrinsic.md

The basics are the same, version header added, order somewhat changed. More importantly, it is not documented.

amaurymartiny:
Putting here to have a written record of what's been discussed on Riot. It must be done in the following order.

1. Add a static `decode` method on all types, which takes any input of any type, and decodes it correctly to the type's base type (i.e. for `Text`, `decode()` returns a string).
    - The constructor will just be: `super(MyType.decode(value));`
2. Remove `fromJSON` and `fromU8a` from all types, create them only with constructor.
3. Remove `Base` class, let all types extend directly JS types (e.g. Moment extends Date, Text extends String...)
4. Replace `Base` with a Codec interface, which all types need to implement, which basically is
```javascript
interface Codec {
  static decode(input: any): any;
  encodedLength: number;
  toHex(): string;
  toJSON(): any;
  toU8a(): Uint8Array;
}
```

jacogr:
Add toString() - this is quite useful in display (UI only uses this - although I suppose it doesn't need to be on the interface since it is part of base "Object")

amaurymartiny:
@ltfschoen It's not a one branch effort, Jaco and I have been creating small PRs all the way to slowly refactor all stuff, see all the linked PRs above.

All PRs that are related to this refactoring have a "Related #161" message.

jacogr:
Closing. Feature-complete.

PR #162, opened 2018-09-23
Decode runtime metadata

jacogr:
Fails on CI since e2e is active. Currently being tested with e2e -

`yarn test packages/api-provider/test/e2e.basics.test.js`

While it decodes fully, there are a couple of issues with the implementation as it stands from a usability perspective.

- ...


packages/api-format/src/codec/U16.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright 2017-2018 @polkadot/api-format authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import BN from 'bn.js';
+
+import BaseNumber from './base/Number';
+
+export default class U16 extends BaseNumber {
```
jacogr: Only have u16 implemented here since it is all we need for this specific one. (u8, u32, u64, u128 is basically exactly the same as this one, just with the bitlength updated)

packages/api-format/src/codec/base/Array.ts
```diff
@@ -0,0 +1,100 @@
...
+      const raw = new this._Type().fromU8a(input.subarray(offset));
+
+      this.raw.push(raw as T);
+      offset += raw.byteLength();
+    }
+
+    return this;
+  }
+
+  map <O> (fn: (entry: T, index?: number) => O): Array<O> {
```
jacogr: We implement the base Array functions, length, map, filter, reduce, forEach so this class can be used like an array. So our own Array-like. (Probably need some more, so it can be used an an iterator)

packages/api-format/src/codec/base/ArrayCompact.ts
```diff
@@ -0,0 +1,14 @@
...
+// of the ISC license. See the LICENSE file for details.
+
+import { Base } from '../types';
+
+import LengthCompact from './LengthCompact';
+import BaseArray from './Array';
+
+export default class CodecArrayCompact <T extends Base<any>> extends BaseArray<T> {
+  constructor (Type: { new(): T }, value: Array<T> = [] as Array<T>) {
+    super(Type, value, LengthCompact);
```
jacogr: As per StringCompact comment, this will become the default.

packages/api-format/src/codec/base/Enum.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright 2017-2018 @polkadot/api-format authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Base } from '../types';
+
+export default class CodecEnum implements Base<number> {
```
jacogr: This takes a byte and maps it to a number. We actually really should do a TS-like enum impl for these. Not sure how that looks though. Probably pass enum as type and then do some magic?

packages/api-format/src/codec/base/EnumType.ts
```diff
@@ -0,0 +1,47 @@
+// Copyright 2017-2018 @polkadot/api-format authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Base } from '../types';
+
+export default class CodecEnumType implements Base<Base<any>> {
```
jacogr: This is something that based on an enum, it decodes to a different type. The same comments on the enum applies, but it really should probably actually include enum instead of re-doing the logic. (composition?)

packages/api-format/src/codec/base/LengthCompact.spec.js
```diff
@@ -0,0 +1,23 @@
+// Copyright 2017-2018 @polkadot/api-format authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import LengthCompact from './LengthCompact';
+
+describe('Compact', () => {
```
jacogr: TODO. Port the Rust tests...

packages/api-format/src/codec/base/Option.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright 2017-2018 @polkadot/api-format authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Base } from '../types';
+
+export default class CodecOption implements Base<Base<any> | null> {
```
jacogr: Based on the length of the first byte, it determines if something follows or not. If 0, nothing follows, if 1, we have actual data f a specific type.

packages/api-format/src/output.ts
```diff
@@ -15,6 +15,28 @@ import isNull from '@polkadot/util/is/null';
...
+
+// NOTE Here we are trying to do things a bit differently, i.e. more in line with where we think things
+// could go with the primitive classes. (Good testbed anyway for the decoding, although maybe a bit messy
+// with duplication). Overall, bit hackly since it shoehorns in a potential new implementation into an
+// original way of doing things
+
+const metaDecode = (input: Array<number>) => {
+  const metadata = new Metadata();
+
+  try {
```
jacogr: try/catch block here for coding debugging. To be removed.

packages/api-provider/test/e2e.basics.test.js
```diff
@@ -0,0 +1,62 @@
...
+        console.error('extrinsics', extrinsics);
+      })
+      .catch((error) => {
+        console.error(error);
+
+        throw error;
+      });
+  });
+
+  it('retrieves the wasm metadata', () => {
```
jacogr: This is where we test :)

packages/api-format/src/codec/base/Length.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export default class CodecLength implements Base<BN> {
+  raw: BN;
+
+  constructor (value: BN | number = new BN(0)) {
+    this.raw = bnToBn(value);
+  }
+
+  byteLength (): number {
+    return 4;
```
jacogr: u32 LE, always 4 bytes.

packages/api-format/src/codec/base/Number.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  toU8a (): Uint8Array {
+    return bnToU8a(this.raw, this.bitLength, true);
+  }
+
+  toString (): string {
+    return bnToHex(this.raw, this.bitLength);
+  }
+
+  toBn (): BN {
```
jacogr: We need some functions get the get values out, toBn and toNumber does that. (We don't really want to encourage access to raw directly, it should be moved to _raw and made at the very least protected)

packages/api-format/src/codec/base/Array.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  constructor (Type: { new(value?: any): T }, value: Array<any> = [] as Array<any>) {
+    this._length = new Length(value.length);
+    this._Type = Type;
+    this.raw = value.map((entry) =>
+      new this._Type(entry)
+    );
+  }
+
+  static with <O extends Base<any>> (Type: { new(value?: any): O }): { new(value?: any): CodecArray<O> } {
```
jacogr: Shortcut function so we can have less noise when actually constructing these inline

packages/api-format/src/codec/String.ts
```diff
@@ -0,0 +1,58 @@
...
+  private _length: Length;
+
+  raw: string;
+
+  constructor (value: string = '') {
+    this._length = new Length(value.length);
+    this.raw = value;
+  }
+
+  byteLength (): number {
```
jacogr: As with Arrays, we don't have length, in any of the classes. (Well, we do have in String and Array). `byteLength` has the total length including the length of the prefixes, where length is just the number of elements. (So in coding use byteLength, when working with data, just a straight length is fine)

packages/api-format/src/codec/Metadata.ts
```diff
@@ -0,0 +1,290 @@
...
+import CodecStruct from './base/Struct';
+import String from './String';
+import U16 from './U16';
+
+class EventMetadata extends CodecStruct<{
+  name: String,
+  arguments: CodecArray<String>,
+  documentation: CodecArray<String>
+}> {
+  constructor () {
```
jacogr: I really tried to not have to pass through the type and a structure that defines the classes (that will be used upon construction). I failed at my first attempt. So there is room for improvement, since what gets passed in the constructor is the same (more-or-less) as the generic type.

packages/api-format/src/codec/Metadata.ts
```diff
@@ -0,0 +1,291 @@
...
+
+// Decodes the runtime metadata as passed through from the `state_getMetadata` call.
+export default class RuntimeMetadata extends CodecStruct<{
+  outerEvent: OuterEventMetadata,
+  modules: CodecArray<RuntimeModuleMetadata>
+}> {
+  constructor (value?: any) {
+    super({
+      outerEvent: OuterEventMetadata,
+      modules: CodecArray.with(RuntimeModuleMetadata)
```
jacogr: CodecArray.with are helper functions that just makes it simpler to define an array (and friends) in-place without having to do yet another class definition for a nested type.

packages/api-format/src/codec/sample.json
```diff
@@ -0,0 +1,2199 @@
...
+            "documentation": [
+              " The public proposals. Unsorted."
+            ]
+          },
+          {
+            "name": "DepositOf",
+            "modifier": 0,
+            "type": {
+              "key": "PropIndex",
+              "value": "(T::Balance, Vec<T::AccountId>)"
```
jacogr: This effective translates to -

```
{
  balance: Balance,
  vecAvvountId: Array<AccountId>
}
```

or rather as a CodecStruct -

```
extends class CodecStruct<{
  balance: Balance,
  vecAccountId: CodecArray<AccountId>
}> {
  constructor (value?: any) {
    super({
      balance: Balance,
      vecAccountId: CodecArray.with(AccountId)
    });
  }
}
```

And sadly, looking at the above it is pretty straight forward to generate, however... since we stll have the issue earlier with the getters not being automagically defined, it is a bit of an issue. Definition is one thing, but we really don't want people to go `instance.raw.balance.toNumber()` when they can do `instance.balance.toNumber()` - if no other way, maybe trim `raw` to something like `v`, i.e. shorter.

Dunno, just thinking through.

packages/api-format/src/codec/sample.json
```diff
@@ -0,0 +1,2199 @@
...
+            },
+            "documentation": [
+              " Those who have locked a deposit."
+            ]
+          },
+          {
+            "name": "LaunchPeriod",
+            "modifier": 2,
+            "type": "T::BlockNumber",
```
jacogr: These `T::Type` stuff is driving me a bit batty. The Rust basically bleeds through. I'm in favour of stripping any `T::` stuff and just not having it at all. (Especially with events not having them and even having things duplicated as `T::`. And then even in the modules, there are some without `T`, and really `T::AccountId` === `AccountId` for the API purposes)

packages/api-format/src/codec/sample.json
```diff
@@ -0,0 +1,2199 @@
...
+              " Get the voters for the current proposal."
+            ]
+          },
+          {
+            "name": "VoteOf",
+            "modifier": 0,
+            "type": {
+              "key": "(ReferendumIndex, T::AccountId)",
+              "value": "bool"
+            },
```
jacogr: Where these `type: { key: ..., value: ... }` are found, it is actually a `Map`. so key becomes the param that is passed, and value is the result type, i.e. type.

jacogr: Oooh, and I picked an interesting one - the map key is actually a struct as well (it is ok, when we call it gets encoded, so ends up just being a u8a)

packages/api-format/src/codec/Metadata.ts
jacogr: Somewhere there must be a solution between key and lookup types, to not pass the type in, but just the struct in the constructor - https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#keyof-and-lookup-types

packages/api-codec/src/base/Array.ts
```diff
@@ -0,0 +1,117 @@
...
+
+import CodecBase from './Base';
+import Length from './LengthCompact';
+
+// This manages codec arrays. Intrernally it keeps track of the length (as decoded) and allows
+// construction with the passed `Type` in the constructor. It aims to be an array-like structure,
+// i.e. while it wraps an array, it provides a `length` property to get the actual length, `at(index)`
+// to retrieve a specific item. Additionally the helper functions `map`, `filter`, `forEach` and
+// `reduce` is exposed on the interface.
+export default class CodecArray <
```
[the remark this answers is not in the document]
jacogr: I would love that. No idea how to implement it though. (It would make it much easier since we won't need to do the map, filter, reduce, etc.

packages/api-codec/src/base/Array.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  filter (fn: (entry: T, index?: number) => any): Array<T> {
+    return this.raw.filter(fn);
+  }
+
+  forEach (fn: (entry: T, index?: number) => any): any {
+    return this.raw.forEach(fn);
+  }
+
+  fromJSON (input: any): CodecArray<T> {
```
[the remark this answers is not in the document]
jacogr: Agreed actually. However the reason I went down this route is twofold -

1. For the stuff where we need to know the actual type, it is probably easier constructing with the class and then where we from it can be used.

2. ... erm, I started with it that way around. However TS basically doesn't allow statics in interfaces (where I started) and it became one big hack to try and work around this.

Not 100% crazy about the above, happy to see it changed.

jacogr:

PR #164, opened 2018-09-24
Rewrite type-storage

amaurymartiny:
1st run at rewriting type-storage.

The user-facing interface will not change:
- `api.state.getStorage(storage.balances.freeBalance, myAddr)`
- `api.state.getStorage(storage.timestamp.now))`

where `storage.balances.freeBalance` is a function, with some metadata. So:
- `storage.balances.freeBalance(myAddr)` will return the correct key
- `storage.balances.freeBalance.{documentation,type}` will return the desired value.

Encoding/Decoding is 100% handled by api-codec, so that type-storage basically only concats and hashes when relevant.

TODO:
- [ ] Find a clean way to get the type of the argument (myAddr above)
- [ ] Make tests pass


packages/api-codec/src/Metadata.ts
```diff
@@ -194,7 +191,14 @@ class StorageFunctionType extends CodecEnumType<Type | StorageFunctionType$Map>
   }
 }
 
-class StorageFunctionMetadata extends CodecStruct {
+export interface IStorageFunctionMetadata {
```
amaurymartiny: Really not satisfied with this, looks too java-like. But I needed StorageFunctionMetadata in type-storage, without `byteLenght` and `from*`

packages/type-storage/src/utils/createFunction.ts
```diff
@@ -0,0 +1,62 @@
...
+    // - storage.balances.freeBalance(address)
+    // - storage.timestamp.blockPeriod()
+    func = (arg?: any) => {
+      if (functionMetadata.type.isMap) {
+        if (!arg) {
+          throw new Error(`${functionMetadata.name} expects one argument.`);
+        }
+
+        // TODO: Ideal syntax
+        const argType = new functionMetadata.type.asMap.key();
```
amaurymartiny: Need String->Class mapping here.

packages/type-storage/src/utils/createFunction.ts
```diff
@@ -0,0 +1,62 @@
...
+        return xxhash(u8aConcat(prefix.toU8a(), argType.from(arg).toU8a()), 128);
```
amaurymartiny: Ideally we would want:
- `storage.balances.freeBalance('5ab.....')`
- and `storage.balances.freeBalance([/* equivalent in u8a */])`

to return the same result.

I think a `from` fucntion in CodecBasc that does `if typeof arg === ... then fromU8a else fromJSON` would do the trick.

packages/type-storage/src/types.d.ts
```diff
@@ -2,22 +2,21 @@
...
-  parachains: Section<Storages, any, any>,
-  session: Section<Storages, any, any>,
-  staking: Section<Storages, any, any>,
-  system: Section<Storages, any, any>,
-  timestamp: Section<Storages, any, any>
-};
+type WellKnownKeys = keyof typeof wellKnownKeys;
 
-export type Storage$Sections = keyof Storages;
+export interface StorageFunction extends IStorageFunctionMetadata {
```
jacogr: I'm guessing this could also be (so we don't define the `IStorageFunctionMetadata`, something like -

```
import { StorageFunctionMetadata } from '@polkadot/api-codec/Metadata;

export interface StorageFunction {
  (...args: any[]): any;
  [K in keyof StorageFunctionMetadata]: StorageFunctionMetadata[K];
}
```

(Nothing wrong here, just giving options in case we don't want to define the same thing twice, the class basically has all the values anyway. Really dunno what is best-practice, but I know that I've had interface fatigue a lot in these...)

packages/type-storage/src/utils/createFunction.ts
jacogr: The constructor of `Address` should take care of this case, it takes both u8a & string as params. So `new Address('5C62W7ELLAAfix9LYrcx5smtcffbhvThkM5x7xfMeYXCt72s')` and `new Address([1, 2, 3, 4, 5, 6, 7, 8, 1, 2, 3, 4, 5, 6, 7, 8, 1, 2, 3, 4, 5, 6, 7, 8, 1, 2, 3, 4, 5, 6, 7, 8])` will yields the same.

packages/api-codec/src/Metadata.ts
jacogr: Left a suggestion for you there...

packages/type-storage/src/utils/createFunction.ts
```diff
@@ -0,0 +1,64 @@
...
+    // - storage.timestamp.blockPeriod()
+    func = (arg?: any) => {
+      if (functionMetadata.type.isMap) {
+        if (!arg) {
+          throw new Error(`${functionMetadata.name} expects one argument.`);
+        }
+
+        const argType = functionMetadata.type.asMap.key.toString(); // The argument type, as string
+        const ArgType = CodecTypes[argType as CodecType]; // The argument type, as Class
+        return xxhash(u8aConcat(prefix.toU8a(), new ArgType(arg).toU8a()), 128);
```
amaurymartiny: @jacogr Any idea how to fix this?
```
Cannot use 'new' with an expression whose type lacks a call or construct signature.
```

amaurymartiny:
Will add tests now.

packages/type-storage/src/wellKnownKeys.ts
```diff
@@ -0,0 +1,55 @@
...
+// Small helper function to factorize code on this page.
+const createWKKFunction = (prefix: String, functionMetadata: WKKMetadata) =>
+  // TODO The expected 2nd argument is a StorageFunctionMetadata, but we actually only need the fields `documentation`
+  // and `type` of it, so in order to not input other fields (byteLength, fromJSON...) we lazily cast as any.
+  createFunction(prefix, { ...functionMetadata, modifier: new StorageFunctionModifier().fromJSON(0) } as any, {
+    isUnhashed: true
+  });
+
+export const code = createWKKFunction(new String(':code'), {
+  documentation: new CodecArray(String, ['Wasm code of the runtime.']),
```
jacogr: In my JS tests this was an issue.... since we have String & String. In my case this defaulted to the JS version. All worked fine until I wanted to do stuff on the actual string :(

Need a new name for `String`, as it stands, everywhere I use it, it is now called `CodecString`

jacogr: Maybe we should call it `Text`? (Don't quite like word prefixes unless they serve a purpose, it becomes muddy... Single character prefixes, well, too Java-like maybe)

jacogr: And while we are at it, Array -> Vector?

jacogr:

packages/type-storage/src/wellKnownKeys.ts
```diff
@@ -0,0 +1,56 @@
...
+  // TODO The expected 2nd argument is a StorageFunctionMetadata, but we actually only need the fields `documentation`
+  // and `type` of it, so in order to not input other fields (byteLength, fromJSON...) we lazily cast as any.
+  createFunction(prefix, { ...functionMetadata, modifier: new StorageFunctionModifier().fromJSON(0) } as any, {
+    isUnhashed: true
+  });
+
+export const code = createWKKFunction(new U8a(':code'), {
+  documentation: new Vector(Text, ['Wasm code of the runtime.']),
+  name: new Text('code'),
+  type: new StorageFunctionType().fromJSON('Vec<u8>')
```
amaurymartiny: How do you create a new StorageFunctionType? A plain type.

jacogr: Well, we should just be able to do `new StorageFunctionType('Vec<u8>')`... However these bloody typed enums are way-over complicated and that doesn't seem to do the trick. :(

jacogr: ... in this case, the enum needs to be set and then the value, so it would be standing on one leg, scratching your ears while humming... something like...

```
// init, it by default sets to "Plain"
const type = new StorageFunctionType();

// now set the value on the wrapped type
type.raw.fromJSON('Bytes');
```

(Use `Bytes` here, the types returned are now also aliassed that way, athough I'm guessing the above will do exactly the same actually - if not, it should)

The above is ugly, well, the whole EnumType just feels horrible.

packages/type-storage/src/fromMetadata.spec.js
```diff
@@ -0,0 +1,25 @@
+import Metadata from '@polkadot/api-codec/Metadata';
+import json from '@polkadot/api-codec/Metadata.rpc.json';
+
+import { fromMetadata } from './fromMetadata';
+
+// Use the pre-generated metadata
+const metadata = new Metadata().fromU8a(Uint8Array.from(json.result));
```
jacogr: You can just use `new Metadata().fromJSON(json.result)`

The `fromJSON` should always allow decoding on the RPCs. Initially it was not the case, but it has been fixed. (i.e. you would normally expect a structure to operate on a, well, json object, in this case it is a number so the codec should handle it.)

packages/api-provider/src/mock/mocks.ts
```diff
@@ -2,7 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
 
-// FIME: This file is way too long and way too messy
+// FIXME: This file is way too long and way too messy
```
jacogr: Still true :)

packages/api-provider/src/mock/mocks.ts
```diff
@@ -20,6 +20,7 @@ import randomAsU8a from '@polkadot/util-crypto/random/asU8a';
 import testKeyring from '@polkadot/util-keyring/testing';
 
 const keyring = testKeyring();
+// FIXME No more .public
```
jacogr: As discussed, once this is in, I'll start mopping up.

packages/type-storage/src/fromMetadata.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { createFunction } from './utils/createFunction';
+import { ModuleStorage, Storage } from './types';
+
+/**
+ * Sets the 1st letter of a string to lowercase.
+ *
+ * @param s - The string to lower first letter.
+ */
+function lowerFirstLetter (s: string) {
```
jacogr: We don't have a util/string/* section, but this would nicely fit in there. (Maybe just mark for moving). (I'm meaning in the common repo)

packages/type-storage/src/fromMetadata.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Extend a storage object with the storage modules & module functions present
+ * in the metadata.
+ *
+ * @param storage - A storage object to be extended.
+ * @param metadata - The metadata to extend the storage object against.
+ */
+export const fromMetadata = (storage: Storage, metadata: Metadata) => {
+  metadata.modules.forEach(moduleMetadata => {
```
jacogr: (nitpick) I know all the cool kids do it, but not a huge fan of the naked arguments. (So really have steered clear of them in all the polkadot-js repos)

packages/type-storage/src/key/index.ts
```diff
@@ -13,17 +12,14 @@ import formatParams from './params';
...
-export default function bindKey <T> ({ isUnhashed, key, params }: SectionItem<T>): Keygen {
+export default function bindKey<T> ({ isUnhashed, key, params }: SectionItem<T>): Keygen {
   const prefix = u8aFromString(key);
 
   return (...keyParams: Array<Storage$Key$Value>): Uint8Array => {
-    const postfix = keyParams.length !== 0
-      ? u8aConcat.apply(null, formatParams(params, keyParams))
-      : new Uint8Array([]);
+    const postfix =
+      keyParams.length !== 0 ? u8aConcat.apply(null, formatParams(params, keyParams)) : new Uint8Array([]);
```
jacogr: The extra whitespace doesn't cost anything and reads a bit better, so suggesting -

```
const postfix = keyParams.length !== 0
  ? u8aConcat.apply(null, formatParams(params, keyParams))
  : new Uint8Array([]);
```

Just a bit easier to see option a & option b in one glance

packages/type-storage/src/key/index.ts
```diff
@@ -13,17 +12,14 @@ import formatParams from './params';
...
     const prefixedKey = u8aConcat(prefix, postfix);
 
-    return isUnhashed
-      ? prefixedKey
-      : xxhash(prefixedKey, 128);
+    return isUnhashed ? prefixedKey : xxhash(prefixedKey, 128);
```
jacogr: Same as above with the spacing.

packages/type-storage/src/wellKnownKeys.ts
```diff
@@ -0,0 +1,56 @@
...
+  documentation: Vector<Text>;
+  name: Text;
+  type: StorageFunctionType;
+}
+
+// Small helper function to factorize code on this page.
+const createWKKFunction = (prefix: Text | U8a, functionMetadata: WKKMetadata) =>
+  // TODO The expected 2nd argument is a StorageFunctionMetadata, but we actually only need the fields `documentation`
+  // and `type` of it, so in order to not input other fields (byteLength, fromJSON...) we lazily cast as any.
+  createFunction(prefix, { ...functionMetadata, modifier: new StorageFunctionModifier().fromJSON(0) } as any, {
```
jacogr: Would prefer some whitespace here, i.e. key per line, it is getting a bit unwieldly

packages/type-storage/src/fromMetadata.ts
amaurymartiny: "naked argument" -> You mean adding explicit type here?

packages/type-storage/src/fromMetadata.ts
amaurymartiny: OK, will put a comment and put it there on my free time.

packages/type-storage/src/key/index.ts
amaurymartiny: OK. While I agree here specifically, I also believe code formatting is subjective.

Small proposal: go prettier the whole way, 80 chars/line, and my code formatting === your code formatting everywhere.

packages/type-storage/src/fromMetadata.ts
jacogr: "free time" :)

packages/api-provider/src/mock/mocks.ts
```diff
@@ -36,31 +37,30 @@ function makeBlockHeader (prevNumber: BN) {
     },
     extrinsicsRoot: randomAsU8a(),
     number: blockNumber,
-    parentHash: prevNumber.eqn(-1)
-      ? new Uint8Array(32)
-      : bnToU8a(prevNumber, 256, false),
+    parentHash: prevNumber.eqn(-1) ? new Uint8Array(32) : bnToU8a(prevNumber, 256, false),
```
jacogr: (nitpick) spacing, option per line.

packages/type-storage/src/storage.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2017-2018 @polkadot/storage authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Storage } from './types';
+import * as wellKnownKeys from './wellKnownKeys';
+
+export const storage: Storage = {
+  wellKnownKeys // Prefill storage with wellKnownKeys, as not returned by state_getMetadata
```
jacogr: Naming - I think it will probably be better to call these something like `substrate` or `runtime` (since system is taken).

amaurymartiny: Runtime created confusion with Metadata.ts types (also called Runtime*), so I preferred substrate.

Another idea is to have `storage.code`, `storage.authorityCount` etc, since they're not really in a module.

PR #166, opened 2018-09-25
Expand types & tests

jacogr:
- Loads of additional tests
- AuthorityId
- MisbehaviorReport
- SignedBlock
- ValidatorPrefs
- Expand EnumType to handle indexes starting at non-0 - hence the need for actual enums...
- Expand Struct to handle anonymous JSON structures
- Add index.ts inside codec and root (commented around exports not there)

Additional FIXMEs marked in the code -

- Would need to check the (currently) unused AccountIndex with actual node tests and reading the code - I believe it is being done wrong with the index extraction. (i.e. 1 byte is actually 2 bytes, it is probably more in line with the compact length encoding approach)


jacogr:

packages/api-codec/src/Bft.ts
```diff
@@ -0,0 +1,92 @@
...
+    return this.raw.hash as Hash;
+  }
+
+  get signature (): Signature {
+    return this.raw.signature as Signature;
+  }
+}
+
+export type JustificationStruct = {
+  round_number?: UInt | BN | number | string,
```
amaurymartiny: `UInt | BN | number | string` shows very often, create a type for this?

packages/api-codec/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+// NOTE Nonce is renamed to Index
+export { default as Index } from './Nonce';
+export { default as PropIndex } from './PropIndex';
+export { default as ProposalIndex } from './ProposalIndex';
+export { default as RawAddress } from './RawAddress';
+export { default as ReferendumIndex } from './ReferendumIndex';
+export { default as Signature } from './Signature';
+export { default as SignedBlock } from './SignedBlock';
+// NOTE Text is currently only used internally, export not needed
+// export { default as Text } from './Text';
```
amaurymartiny: I need it in type-storage: for the well known keys, I manually type the documentation of those storage functions.

However I can also do `import Text from '@polkadot/api-codec/Text';`

packages/api-codec/src/codec/Struct.ts
```diff
@@ -6,59 +6,83 @@ import u8aConcat from '@polkadot/util/u8a/concat';
...
 // it needs to decoded in the specific defined order.
-//
-// TODO:
-//   - Check the constructor, something is really, really wrong with the way the defs are used
 export default class Struct <
   S = { [index: string]: { new(value?: any): Base } },
   T = { [K in keyof S]: Base },
-  V = { [K in keyof S]: any }
+  V = { [K in keyof S]: any },
+  E = { [K in keyof S]: string }
```
amaurymartiny: Can you provide an example in the comments on what each type STVE represents?

packages/api-codec/src/index.ts
jacogr: Yeap, I was thinking about that. Text is a funny one since it sounds cool, but apart from docs, actually never used - (apart from you on the API). So general API users should never need this one... Well, never say never...

packages/api-codec/src/codec/Struct.ts
jacogr: Done, pushing.

packages/api-codec/src/Bft.ts
jacogr: Done, pushing. (Both U8 & UInt)

jacogr:
Related https://github.com/polkadot-js/api/issues/161

PR #169, opened 2018-09-26
Round out metadata types

jacogr:
Types -

- Gas (u64)
- Moment (Date, per-second accuracy)
- NewAccountOutcome (enum)
- Origin (cannot instantiate, placeholder)
- ParachainId (u64, Polkadot)
- Permill (and Perbill as a bonus)
- RuntimeVersion (new from rpc)
- StorageData, StorageKey & StorageChangeSet
- VoteThreshold (enum)

Helpers -

- Tuple, `(type, type, type)` implementations, handled as a Struct, with different JSON encoding
- Enum & EnumType, now takes `Array<type>` or `{ [index: number]: type }` as definitions
- Struct can now encode/decode with different keys for json

With this we should now have full coverage of all the types from metadata. Additionally, with the addition of Storage* the currently exposed jsonrpc types are catered for.

FIXME -

- With the new Tuple, need to go through everything again and use Tuple where applicable. (It is not an issue for now, but could be on the Node.js client where the toJSON encoding is actually used)

### DIFF
--- a/packages/api-codec/src/Metadata.ts
+++ b/packages/api-codec/src/Metadata.ts
@@ -174,14 +174,11 @@ export class StorageFunctionType$Map extends Struct {
 }
 
 export class StorageFunctionType extends EnumType<Type | StorageFunctionType$Map> {
-  // NOTE Since this is called dynamically, we may have empty values here
   constructor (index?: number, value?: any) {
     super([
       Type,
       StorageFunctionType$Map
-    ], ['Plain', 'Map']);
-
-    this.setValue(index, value);
+    ], index, value);
   }
 
   get isMap (): boolean {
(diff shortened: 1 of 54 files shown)


packages/api-codec/src/Metadata.ts
```diff
@@ -283,11 +283,6 @@ export default class RuntimeMetadata extends Struct {
     ) as RuntimeMetadata;
   }
 
-  // FIXME Really not crazy about having to manually add all the getters. Preferably it should
-  // be done automagically in the actual Struct - however what is really important here
-  // here is that we should nbot lose the autocompletion and checking that TS gives us. So if
-  // we have to choose between the 2, manual defs it would have to be.
-
```
jacogr: Well, that boat has sailed (for now). Added it mostly everywhere we use structs. It is not _that_ bad.

jacogr:
Related https://github.com/polkadot-js/api/issues/161

PR #171, opened 2018-09-27
API formatting for inputs & outputs

jacogr:
Related #161


packages/api-codec/src/codec/EnumType.ts
```diff
@@ -52,7 +52,14 @@ export default class EnumType <T> extends Base<Base<T>> {
     return this;
   }
 
-  setValue (index?: number, value?: any): void {
+  setValue (index?: | EnumType<T> | number, value?: any): void {
+    if (index instanceof EnumType) {
```
jacogr: Not a fan of this... it would have been fine if it was value, however value/index are most probably used wrongly, index is the actual value, value is well, the constructor value.

packages/type-storage/src/utils/createFunction.ts
```diff
@@ -29,36 +29,37 @@ export function createFunction (
...
   options: CreateItemOptions = {}
-) {
-  let func: any; // The return value, it's a function that has metadata.
+): StorageFunction {
+  const meta = new StorageFunctionMetadata(functionMetadata);
+  const storageFn = meta as StorageFunction;
 
   if (options.isUnhashed) {
-    func = () => prefix.toU8a();
+    storageFn.create = () => prefix.toU8a();
```
jacogr: Sorry :( To get line 33 done, i.e. create a new instance from another one, had to do this. The assignment, i.e, Object.assign as removed doesn't actually do the trick...

jest.config.js
```diff
@@ -2,7 +2,7 @@ const config = require('@polkadot/dev/config/jest');
 
 module.exports = Object.assign({}, config, {
   moduleNameMapper: {
-    '@polkadot/api-(codec|format|jsonrpc|provider|rx)(.*)$': '<rootDir>/packages/api-$1/src/$2',
+    '@polkadot/api-(codec|jsonrpc|provider|rx)(.*)$': '<rootDir>/packages/api-$1/src/$2',
```
jacogr: API format is now not used in API anymore, so be like all the cool kids and remove a lot of line of code...

packages/api-codec/src/StorageKey.ts
```diff
@@ -2,8 +2,45 @@
...
+}
+
+type StorageFunctionWithArg = [StorageFunction, any];
+
+const DEFAULT_TYPE = new StorageFunctionType(0, 'Bytes');
+
+// A representation of a storage key (typically hashed) in the system. It can be constructed
+// by passing in a raw key or a StorageFunction with (optional) arguments.
 export default class StorageKey extends Bytes {
+  private _outputType: StorageFunctionType;
```
jacogr: Not 100% sure on this atm - basically the StorageKey carries around the resulting type. Then when a result comes back, the formatter can jump in and grab it.

packages/api-codec/src/StorageKey.ts
```diff
@@ -2,8 +2,67 @@
...
 
+import { AnyU8a } from './types';
+
+import isFunction from '@polkadot/util/is/function';
+
 import Bytes from './Bytes';
+import { StorageFunctionMetadata } from './Metadata';
+
+export interface StorageFunction extends StorageFunctionMetadata {
+  create (arg?: any): StorageKey;
```
jacogr: Ok, check this - it is moved in from storage (and storage imports this), however... instead of `create (arg?: any): StorageKey` it is now, well, that.

The issue is in type-storage in the create function, I needed to create a new instance since the Object.assign we had for docs/type didn't actually work as expected. The stuff just didn't get pulled through. So now create it as a clone, then cast and assign the create.

I know it is uglier... However, we probably don't expect people to call this themselves...

Any suggestions welcome (when looking at type-storage/createFunction)

packages/api-codec/src/Metadata.ts
```diff
@@ -17,12 +19,12 @@ import U16 from './U16';
 // end and work up. (Just so we don't use before definition)
 
 export class EventMetadata extends Struct {
-  constructor () {
+  constructor (value?: any) {
```
jacogr: We can now instantiate all of these, it aligns with all the other codec classes apart for just being useful

packages/api-codec/src/StorageKey.ts
```diff
@@ -2,8 +2,67 @@
...
+function isStorageFunction (value?: any): value is StorageFunction {
+  return value && isFunction(value.create);
+}
+
+// A representation of a storage key (typically hashed) in the system. It can be constructed
+// by passing in a raw key or a StorageFunction with (optional) arguments.
 export default class StorageKey extends Bytes {
+  private _outputType: string | null;
+
+  constructor (value: AnyU8a | StorageKey | StorageFunction | StorageFunctionWithArg = new Uint8Array(), ...args: Array<any>) {
```
jacogr: Not sure we should actually allow AnyU8a, it is useful, but if we funnel towards using StorageFunction as a constant, probably not the best idea to have too many options.

packages/api-codec/src/StorageKey.ts
jacogr: Alternative would be something like -

```
export interface StorageFunction {
  (arg?: any): StorageKey;
  meta: StorageFunctionMetadata
}
```

There are benefits, i.e. can actually skip the new Metadata we do in createFunction. Also keeps the cleanliness of using the code should creation be needed.

packages/api/src/formatting.spec.js
```diff
@@ -2,17 +2,17 @@
...
-import storage from '@polkadot/storage';
+import storage from '@polkadot/storage/testing';
 
 import Api from './index';
 
 const ADDR_ONE = '5EhmTa7fL6SdjgKXo9g6hetR6nHnRAmrtisoGFWEESjzECtY';
 const ADDR_TWO = '5FjbQAHSptRU4Q7Py6zDsRnE7UkpZikJVg5nm1tmu3WuZXKd';
 
-const ENC_ONE = '0x11cf1094db4db43356b7787c3e59c39f';
+const ENC_ONE = '0xe541e84609054b9504b8316ea839b366';
```
jacogr: Obviously still need to double-check these against a running node with e2e tests, i.e. check the balance of Alice with a `--dev` chain. If something horribly large, it works, otherwise, well, we are missing a trick.

packages/api-codec/src/codec/Base.ts
```diff
@@ -2,6 +2,10 @@
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
 
+import logger from '@polkadot/util/logger';
+
+export const l = logger('codec');
```
amaurymartiny: Is this needed?

jacogr:
@amaurymartiny Yes, please... I don't like it, but I built it up bit by bit as needed so it ended up evolving that way. Certainly not the be-all and end-all, and I did it under duress, I hate those ;) At least the logic is follow-able, so certainly ripe for making it proper.

packages/api-codec/src/codec/Base.ts
jacogr: Only for debug logging. I continue adding and removing the bloody logs, so the `l.debug(() => )` ends up as a noop in normal mode and as a bonus when running tests it spits out the result.

jacogr: Currently Struct & createType use it. (Should have added it earlier actually, I continued committing console.log changes.

PR #172, opened 2018-09-27
`Base` as an interface instead of a class

amaurymartiny:
Proposal: remove the Base class, use an interface `Codec<T>` instead, that all api-codec types **needs** to implement (+/- similar to a rust trait).

In particular, use as much as possible existing JavaScript classes (Array, Map, ...) and create child classes that implement `Codec<T>`.

Types:
- [x] Vector: extends Array. Syntax is much simpler (no `.get(1)`, simply `[1]`, remove `this.raw`)
- [ ] Struct: could probably extend Map
- [ ] maybe others

Note: Tests don't change at all, cf `*.spec.js`, only syntax simplifications.

@jacogr If this is something that's interesting, I'll have a look more into it tomorrow.


jacogr:
Related #161

jacogr:
One additional thing, for UInt, probably want to extends BN (When going for BigInt, probably need to extends that). I'm currently doing a lot of stuff such as -

```
          new BlockNumber(
            sessionCurrentIndex.toBn()
              .sub(eraLastLengthChange.toBn())
              .mod(sessionsPerEra.toBn())
              .mul(sessionLength.toBn())
              .add(sessionBlockProgress.toBn())
          )
```

(in the observable API - they are all BlockNumber, but have a conversion for each and then a final wrap)

For the time-being, will probably just fake add/sub/mul (a couple of them), that way when we extend properly, there is no change, just thing being more effective)

amaurymartiny:
Things I'm not 100% satisfied:

1. `fromJSON` and `fromU8a`: most of the JS types (like BN, Uint8Array...) don't allow you to replace the value as is, so where before we had `this.raw = ...`, now I cannot say `this = ...`. The best way I have done right now is returning a new instance. As put in the comments, we could go forward and remove all `from*`.
2. `a.byteLength` vs `a.byteLength()`: BN has `byteLength()` method, and Uint8Array has `byteLength` property. If we remove BN, we could go with `get byteLength() {...}` everywhere, for consistency.
3. Tuple should extend Vector instead of Struct.

Still TODO:
- [ ] Option
- [ ] EnumType
- [ ] Address (extends `AccountId | AccountIndex`)

packages/types/src/Moment.ts
```diff
@@ -42,21 +43,24 @@ export default class Moment extends Base<Date> {
   }
 
   fromJSON (input: any): Moment {
-    this.raw = Moment.decode(input);
-
-    return this;
+    // FIXME this returns a new Object unfortunately, can't "replace" current value
```
jacogr: Remove from, as per your other PR.

packages/types/src/Moment.ts
```diff
@@ -7,16 +7,17 @@ import bnToBn from '@polkadot/util/bn/toBn';
...
+import { Codec } from './types';
 
 const BITLENGTH = 64;
 
 // A wrapper around seconds/timestamps. Internally the representation only has
 // second precicion (aligning with Rust), so any numbers passed an/out are always
 // per-second. For any encoding/decoding the 1000 multiplier would be applied to
 // get it in line with JavaScript formats
-export default class Moment extends Base<Date> {
+export default class Moment extends Date implements Codec<Moment> {
```
jacogr: Is this correct, i.e. `Codec<Moment>` (e.g. Bool above extends built-in JS `Boolean` class, one of the 2 is wrong)

packages/types/src/Text.ts
```diff
@@ -2,62 +2,56 @@
...
     const [offset, length] = Compact.decode(input);
-
-    this.raw = u8aToUtf8(input.subarray(offset, offset + length.toNumber()));
-
-    return this;
+    // FIXME this returns a new Object unfortunately, can't "replace" current value
+    // Two solutions:
+    // - either use static
+    // - or completely remove from*, and force to use constructor
+    return new Text(u8aToUtf8(input.subarray(offset, offset + length.toNumber())));
```
jacogr: Yeap, the other PR needs to go first. (This will just create too much messy-ness)

packages/types/src/codec/Enum.spec.ts
```diff
@@ -4,6 +4,11 @@
 
 import Enum from './Enum';
 
+enum TestEnum {
```
jacogr: We could actually go full-hog ts specs now. (Previously it was problematic since I fed it garbage data - either way, not needed right now)

packages/types/src/codec/U8aFixed.ts
```diff
@@ -2,43 +2,44 @@
...
 export default class U8aFixed extends U8a {
-  protected _bitLength: number;
+  protected _bitLength: BitLength;
 
   constructor (value: AnyU8a = new Uint8Array(), bitLength: BitLength = 256) {
-    super(value);
+    super(
+      value instanceof U8aFixed
+        ? value
+        : U8aFixed.trimLength(toU8a(value), bitLength / 8)
```
jacogr: Just check current master as well, I had some hoops to jump through - well, if not corrects, tests do catch this one.

packages/types/src/codec/UInt.ts
```diff
@@ -24,13 +22,13 @@ type BitLength = 8 | 16 | 32 | 64 | 128 | 256;
...
 //   - Apart from encoding/decoding we don't actuall keep check on the sizes, is this good enough?
-export default class UInt extends Base<BN> {
+export default class UInt extends BN implements Codec<UInt> {
   private _bitLength: BitLength;
   private _isHexJson: boolean;
 
   constructor (value: AnyNumber = 0, bitLength: BitLength = 64, isHexJson: boolean = true) {
     super(
-      UInt.decode(value)
+      UInt.decode(value).toNumber()
```
jacogr: Cannot use toNumber here - JS numbers are limited to 53 bits (on uints) - rather toString() it into a decimal representation.

packages/types/src/types.d.ts
```diff
@@ -11,4 +11,16 @@ export type AnyNumber = UInt | BN | Uint8Array | number | string;
 
 export type AnyU8a = U8a | Uint8Array | Array<number> | string;
 
+export interface Codec<T> {
+  byteLength(): number;
```
jacogr: We could, to get past the BN vs Array issue, call this `encodedLength`, which is actually exactly what it is. (As a property/getter)

amaurymartiny:
This is a clusterfuck of merge conflicts. Will open a new one once everything else is ready.

PR #173, opened 2018-09-28
Rewrite type-extrinsics

amaurymartiny:
API: `extrinsics.timestamp.set(...)` -> Returns correctly encoded `Extrinsic`.

TODO:
- [x] Add camelCase: `extrinsics.balances.set_balance` -> `extrinsics.balances.setBalance`
- [x] Remove type-params
- [x] Remove type-primitives

In another PR:
- Add encode with publicKey
- Add sign with private key

Notes:
- skipped all failing tests in `type-primitives`
- `codec/` and `legacy` folders should be removed asap


jacogr:
Related #161

packages/type-extrinsics/src/fromMetadata.spec.js
```diff
@@ -0,0 +1,34 @@
...
+
+  it('should return a value if the storage function does not expect an argument', () => {
+    expect(() => newExtrinsics.balances.setBalance('5C62W7ELLAAfix9LYrcx5smtcffbhvThkM5x7xfMeYXCt72s', 2, 3)).not.toThrow();
+  });
+
+  it('should return the correct storage key', () => {
+    expect(newExtrinsics.timestamp.set([10101]).toU8a()).toEqual(
+      new Uint8Array([
+        // index
+        // 3, 0, // FIXME This was taken from the old tests, should work, but doesn't
```
jacogr: It is length encoded. The 40 is the length. Passing isBare = true should give the result you expect.

packages/type-extrinsics/src/utils/createExtrinsic.ts
```diff
@@ -0,0 +1,51 @@
...
+  const expectedArgs = meta.arguments.filter((arg) => arg.type.toString() !== 'Origin');
+
+  extrinsicFn = (...args: any[]): Extrinsic => {
+    if (expectedArgs.length.valueOf() !== args.length) {
+      throw new Error(`Extrinsic ${prefix}.${name} expects ${expectedArgs.length.valueOf()} arguments, got ${args.length}.`);
+    }
+
+    return new Extrinsic(
+      u8aConcat(
+        new Uint8Array([index, meta.id.toU8a()[1]]),
```
amaurymartiny: Not sure if that's the cleanest way

jacogr: Well, we get to 3,0... so... Just not sure this is u16 (not sure why it is not a u8) and the toU8a() encodes to little-endian, so the value we are actually interested in is in position 0?

i.e.

`0x0004` -> `Uint8Array([4, 0])`

I would actually just `toNumber()` it there. It is supposed to fit in a byte, if it goes over 255 we have bigger issues.

packages/type-extrinsics/src/utils/createExtrinsic.ts
```diff
@@ -0,0 +1,51 @@
...
+      throw new Error(`Extrinsic ${prefix}.${name} expects ${expectedArgs.length.valueOf()} arguments, got ${args.length}.`);
+    }
+
+    return new Extrinsic(
+      u8aConcat(
+        new Uint8Array([index, meta.id.toNumber()]),
+        ...expectedArgs.map((argument, index) => {
+          const type = argument.type.toString(); // Argument type, as string
+
+          return createType(type, args[index]).toU8a(true);
```
jacogr: This should be encoded? In the test you to `toU8a(true)`, this should be ready for submittion.

PR #175, opened 2018-09-28
Pull in observable API from UI

jacogr:
Related #161

### DIFF
--- a/packages/api-rx/src/types.d.ts
+++ b/packages/api-rx/src/types.d.ts
@@ -12,10 +12,11 @@ export type RxApiInterface$Section = {
   [index: string]: RxApiInterface$Method
 };
 
-type RxApiInterface$Keys = {
-  readonly [key in keyof typeof interfaces]: RxApiInterface$Section;
-};
+export type RxApiInterface = {
+  readonly author: RxApiInterface$Section;
+  readonly chain: RxApiInterface$Section;
+  readonly state: RxApiInterface$Section;
+  readonly system: RxApiInterface$Section;
 
-export type RxApiInterface = RxApiInterface$Keys & {
   isConnected: () => BehaviorSubject<boolean>
 }
--- a/packages/api-rxe/src/index.ts
+++ b/packages/api-rxe/src/index.ts
@@ -1,0 +1,589 @@
+// Copyright 2017-2018 @polkadot/ui-react-rx authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { RxApiInterface, RxApiInterface$Method } from '@polkadot/api-rx/types';
+import { Method } from '@polkadot/jsonrpc/types';
+import { RxBalance, RxBalanceMap, RxFees, RxProposal, RxProposalDeposits, RxReferendum, KeyWithParams, KeyWithoutParams, RxReferendumVote } from './types';
+
+import BN from 'bn.js';
+import { EMPTY, Observable, combineLatest } from 'rxjs';
+import { switchMap, defaultIfEmpty, map } from 'rxjs/operators';
+// FIXME We need to import the create from runtime and actually build it up
+import storage from '@polkadot/storage/testing';
+import assert from '@polkadot/util/assert';
+import isUndefined from '@polkadot/util/is/undefined';
+
+type OptBN = BN | undefined;
+type OptDate = Date | undefined;
+type MapFn<R, T> = (combined: R) => T;
+
+type ResultReferendum = [BN, ExtrinsicDecoded, number];
+type ResultProposal = [BN, ExtrinsicDecoded, string];
+type ResultProposalDeposits = [BN, Array<string>];
+
+const defaultMapFn = (result: any): any =>
+  result;
+
+export default class ObservableApi {
+  private api: RxApiInterface;
+
+  constructor (api: RxApiInterface) {
+    this.api = api;
+  }
+
+  private combine = <T, R> (observables: Array<Observable<any>>, mapfn: MapFn<R, T> = defaultMapFn): Observable<T> => {
+    return combineLatest(...observables).pipe(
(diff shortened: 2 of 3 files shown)


jacogr:
Ok, pretty happy (?) with where it as a phase I. I had to limit the intent a bit and stuck with the existing naming instead of (as a first try), renaming everything along the way as well. So it basically just imports the new names and types.

As it stands, going to just use it as-is in the UI and see what pops up. Once working as expected, can marge and actually use the package. (Just from a speed of use perspective, sticking with the in-UI updated version with these large number of changes is a bit more prudent)

packages/api-codec/src/Metadata.ts
```diff
@@ -285,9 +287,9 @@ export default class RuntimeMetadata extends Struct {
 
   // We receive this as an Array<number> in the JSON output from the Node. Convert
   // to u8a and use the fromU8a to do the actual parsing
-  fromJSON (input: Array<number>): RuntimeMetadata {
+  fromJSON (input: Uint8Array | string | Array<number>): RuntimeMetadata {
```
jacogr: The actual JSON has changed, it now comes in not as an `Array<number>`, but rather as a hex string. Cater for all.

packages/api-codec/src/codec/Compact.ts
```diff
@@ -0,0 +1,64 @@
...
+//     xx xx xx 00
+// (2**6 ... 2**14 - 1)  (u8, u16)  low LH high
+//     yL yL yL 01 / yH yH yH yL
+// (2**14 ... 2**30 - 1)  (u16, u32)  low LMMH high
+//     zL zL zL 10 / zM zM zM zL / zM zM zM zM / zH zH zH zM
+// (2**30 ... 2**536 - 1)  (u32, u64, u128, U256, U512, U520) straight LE-encoded
+//     nn nn nn 11 [ / zz zz zz zz ]{4 + n}
+//
+// Note: we use *LOW BITS* of the LSB in LE encoding to encode the 2 bit key.
+export default class Compact {
```
jacogr: We pulled this encoding (basically just the encode/decode functions) from Length - parity-codec is looking to use this for all numbers, so make it share-able as "utility" functions. (Naming for the encoding now also aligns with what we have in parity-codec)

packages/api-codec/src/codec/Length.ts
```diff
@@ -6,83 +6,39 @@ import { AnyNumber } from '../types';
...
 // FIXME Not crazy about the use of Length. If we look at Vector, Bytes & String,
 // the implementations are basically the same. Vector is slightly different since
 // it iterates over the length. We need a cleaner way where we have less copied
 // code and a more generic implementation around the use of Length. Looking at
 // Array or Struct, the same type of wrapper would be useful here.
 export default class Length extends Base<BN> {
-  constructor (value: AnyNumber = new BN(0)) {
+  private _bitLength: BitLength;
+
+  constructor (value: AnyNumber = new BN(0), bitLength: BitLength = 32) {
```
jacogr: Bit out of scope for this PR, so not changing things here... but since we have pulled the encoding logic into Compact, we would actually get rid of this use - rather, just call Compact.encode() with the length directly in the variable length uses instead of replying on Length. (Which as per the FIXME has always been a bit of a funny step-child)

packages/api-observable/src/Base.ts
```diff
@@ -0,0 +1,72 @@
...
+import { StorageFunction } from '@polkadot/api-codec/StorageKey';
+import assert from '@polkadot/util/assert';
+import isUndefined from '@polkadot/util/is/undefined';
+
+type MapFn<R, T> = (combined: R) => T;
+
+const defaultMapFn = (result: any): any =>
+  result;
+
+export default class ApiBase {
```
jacogr: Split out the implementation from the UI -

- Base - just the basics, i.e. a nice wrapper for raw calls and storage
- Calls - Anything that maps to an actual call on the API, i.e. getBlock
- Queries - Anything that just simply wraps a storage query
- Combined - Multiple observables combined into a single result

packages/api-observable/src/Queries.ts
```diff
@@ -0,0 +1,207 @@
...
+      .rawStorage(storage.staking.intentions)
+      .pipe(
+        // @ts-ignore After upgrade to 6.3.2
+        map((intentions: Array<AccountId> = []) =>
+          intentions
+        )
+      );
+  }
+
+  stakingFreeBalanceOf = (address: AccountId | string): Observable<Balance | undefined> => {
```
jacogr: These need to be renamed - kept it as-is for the initial PR, i.e. just takin what is there and moving it to the correct storage area. (Actually started off renaming all, but in reality would first need to get the UI working and then start renaming)

packages/api-observable/src/classes.ts
```diff
@@ -0,0 +1,73 @@
+// Copyright 2017-2018 @polkadot/ui-observable authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import BN from 'bn.js';
+import { AccountId, Balance, BlockNumber, PropIndex, Proposal, ReferendumIndex, VoteThreshold } from '@polkadot/api-codec/index';
```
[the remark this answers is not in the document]
jacogr: Habit. In the UI (I think) the tsconfig path mapping is a bit wonky with some packages. (Just the regex not being quite as good as it should be). Here it is actually not needed. (Looking at tsconfig, quite simple here). For end-users it is never required.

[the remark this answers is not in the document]
jacogr: Nope, just checked, it still is needed - I think it is a tsconfig regex change, i.e. changing it to `@polkadot/api-codec` and it complains.

jacogr:
100% with you on the documentation - and something else sorely missing here is the testing. (But first get it working and named properly.)

PR #180, opened 2018-10-01
UncheckedExtrinsic and encode it

amaurymartiny:
Usage:

```javascript
const a = extrinsics.timestamp.set(...); // returns unsigned UncheckedMortalExtrinsic
const b = a.sign(keypair); // returns signed UncheckedMortalExtrinsic

author_submitExtrinsic(a.toU8a()); // works, inherent sent
author_submitExtrinsic(b.toU8a()); // works, transaction sent
```


packages/api-codec/src/ExtrinsicIndex.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright 2017-2018 @polkadot/api-codec authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import U16 from './U16';
+
+export default class ExtrinsicIndex extends U16 {
```
amaurymartiny: Is ExtrinsicIndex U16 or U32? From old tests it seems like U16, but [here](https://github.com/paritytech/substrate/search?utf8=%E2%9C%93&q=extrinsic_Index&type=) it seems U32.

packages/type-extrinsics/src/SignableExtrinsic.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { Extrinsic, Index, UncheckedExtrinsic } from '@polkadot/api-codec/index';
+import { KeyringPair } from '@polkadot/util-keyring/types';
+import u8aConcat from '@polkadot/util/u8a/concat';
+
+export default class SignableExtrinsic extends UncheckedExtrinsic {
+  sign (sender: KeyringPair, nonce: Index): Extrinsic {
+    // Prepend sender's publicKey and nonce
+    const message = u8aConcat(
+      sender.publicKey(),
```
amaurymartiny: We need a `sender.address` that returns an Address.

packages/type-extrinsics/src/SignableExtrinsic.ts
```diff
@@ -0,0 +1,28 @@
...
+      nonce.toU8a(),
+      this.toU8a(true)
+    );
+
+    const signature = sender.sign(message);
+
+    return new Extrinsic(
+      u8aConcat(
+        new Uint8Array([255]), // FIXME works, but why?
```
amaurymartiny: ideally `sender.address().prefixKey()`

packages/api-codec/src/ExtrinsicIndex.ts
jacogr: Seems like u32 based on what you linked. The issue with the old tests are that the length encoding is now completely different, so there are some hoops to jump through to get it going or converted - so e2e is probably easier (initial) tests.

Basically, say you had a Bytes of 16 bytes in length (`0123456789abcdef`), previously it would have been

```
[
// length
16, 0, 0, 0,
// bytes
0, 1, 2, 3, 4, 5, 6, 7, 8, 9, a, b, c, d, e, f
]
```

now

```
[
// length (encoded with Compact
16 << 2,
// bytes
0, 1, 2, 3, 4, 5, 6, 7, 8, 9, a, b, c, d, e, f
]
```

(and then when it is 2 bytes or 4 bytes, it gets weirder...)

packages/type-extrinsics/src/SignableExtrinsic.ts
jacogr: I was afraid of that. And got a snif of that in the UI today.Basically we probably need to move coded to the common repo then, which screws us up a bit. (since deps on repos are `dev` -> `common` -> `api` -> `apps/client`). As a stopgap,

```
new Address(sender.publicKey())
```

Not against doing the move, only at this point it is easier having it in here...

packages/type-extrinsics/src/SignableExtrinsic.ts
jacogr: It is the prefix for AccountId. So if you had it as `Address` and then used the `toU8a()` it would have added the prefix. Maybe there is a much-up with AccountId as well - I know that AccountIndex is not correct, it itself should add the prefix.

jacogr: TL;DR Making it an Address and encoding it will add the prefix (Should be Address, so it handles both AccountId & AccountIndex)

packages/type-extrinsics/src/Descriptor.ts
```diff
@@ -0,0 +1,51 @@
...
+    // If the extrinsic function has an argument of type `Origin`, we ignore it
+    // FIXME should be `arg.type !== Origin`, but doesn't work...
+    const expectedArgs = this._meta.arguments.filter((arg) => arg.type.toString() !== 'Origin');
+
+    return u8aConcat(
+      this._index.toU8a(),
+      ...expectedArgs.map((argument, index) => {
+        const type = argument.type.toString(); // Argument type, as string
+
+        return createType(type, this._args[index]).toU8a();
```
jacogr: Ummmm, we should probably allow Text in here as well. (Missed that usecase completely)

packages/type-extrinsics/src/UncheckedMortalExtrinsic.ts
```diff
@@ -0,0 +1,96 @@
...
+    // 8 bytes: The Transaction Index of the signing account (number of signed transactions from the account preceeding this one).
+    // 2 bytes: The Transaction Era (detailed below).
+    let signatureInformation;
+    if (!isSigned) {
+      // Nothing if it's an inherent.
+      signatureInformation = new Uint8Array();
+    } else {
+      // Address.
+      // @ts-ignore keyring -> "Object is possibly undefined", no it's not
+      const address = new Address(keyring.publicKey());
```
jacogr: Ummm, yes, this is weird, the above check for `isUndefined` should be trapping these.

packages/type-extrinsics/src/UncheckedMortalExtrinsic.ts
```diff
@@ -0,0 +1,96 @@
...
+      descriptor.toU8a()
+    );
+  }
+
+  get isSigned () {
+    return this._isSigned;
+  }
+
+  sign (keyring: KeyringPair, nonce: Index) {
+    this._isSigned = true;
```
jacogr: Maybe first sign, then set the flag (just in case it throws). So just swap these around.

packages/type-extrinsics/test/e2e/transfer.spec.js
```diff
@@ -0,0 +1,36 @@
...
+
+  beforeAll(() => {
+    // check other e2e tests here...
+    api = new Api(new Ws('ws://127.0.0.1:9944'));
+  });
+
+  it('makes a transfer for an inherent', () => {
+    const inherent = extrinsics.balances.transfer(keyring.bob.publicKey(), 12345);
+
+    return api.author.submitExtrinsic(inherent.toU8a()).then(console.log);
```
jacogr: We should not need to `toU8a` here. (Hopefully it just accepts the input. If not, we can just adjust the jsonrpc definition to accept this type.)

packages/types/src/ExtrinsicIndex.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright 2017-2018 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import U32 from './U32';
+
+export default class ExtrinsicIndex extends U32 {
```
jacogr: u16? (2 bytes)

PR #181, opened 2018-10-01
[WIP]: author extrinsicUpdate method

yjkimjunior:
This won't work yet but trying to unblock [#218](https://github.com/polkadot-js/apps/issues/218) and [#304](https://github.com/polkadot-js/apps/issues/304)

* need to squash commits :p


jacogr:
Cool. Nice. 

API is in a bit of flux atm, it actually only supports poc-3 - i.e. running substrate master with --dev mode. (Which is just an issue for probably 2 weeks). I screwed up a bit, should have has a poc-3 branch like with the UI.

jacogr:
Related #161

packages/type-jsonrpc/src/author.ts
```diff
@@ -22,8 +22,18 @@ const submitExtrinsic: MethodOpt = {
   type: 'Hash'
 };
 
+const extrinsicUpdate: MethodOpt = {
+  isSigned: true,
+  description: 'Subscribe and watch an extrinsic',
+  params: [
+    createParam('metadata', 'subscriber', 'Bytes')
```
jacogr: You should not need the first 2 params, only the last one. (The first 2 are internal for subscriptions)

jacogr: See e.g. `state_storage` or `chain_newHead` (also subscription APIs). Additionally, you need the `subscription` param, once again look at the above 2. (Yes, it is a bit hairy, any cleanups appreciated)

packages/type-jsonrpc/src/author.ts
```diff
@@ -22,8 +22,22 @@ const submitExtrinsic: MethodOpt = {
...
+  isSigned: true,
+  isSubscription: true,
+  subscribe: [
+    'author_submitAndWatchExtrinsic',
+    'author_unwatchExtrinsic'
+  ],
+  params: [
+    createParam('Bytes')
+  ],
+  type: 'ExtrinsicUpdate'
```
jacogr: Now it gets tricky -

1. Match the type here to the Rust code, it should be `Status` (since it is RPC, we could name it slightly more descriptive, e.g `ExtrinsicStatus` - the first version . is a bit "generic") - https://github.com/paritytech/substrate/blob/master/core/rpc/src/author/mod.rs#L67 (without the generic params)
2. Implement the actual type class inside `api-codec` - since it is a typed enum, it is a bit more tricky. Basically you need `extends EnumType<class | class | class>` where `class` is the actual classes. (These can be classes inside the `ExtrinsicStatus` file, e.g. see https://github.com/polkadot-js/api/blob/master/packages/api-codec/src/MisbehaviorReport.ts#L57)

You had to pick a difficult one ;) (Metadata also has an EnumType, that one is probably be a bit cleaner since it doesn't have the enum overrides, i.e. it starts at index 0, is the above report it starts at 0x11 which just makes things a bit more complicated)

packages/types/src/Status.ts
```diff
@@ -0,0 +1,27 @@
+import { AnyU8a } from './types';
+
+import blake2Asu8a from '@polkadot/util-crypto/black2/asU8a';
```
ltfschoen: i think it should be should be `blake2` not `black2`

jacogr:
Closed in #264 - one thing outstanding, we need to add the method to api-observable for use in the UI

PR #186, opened 2018-10-02
Update Address/AccountIndex coding

jacogr:
- Move AccountIndex prefix encoding to AccountIndex itself
- Fix AccountIndex to 1 byte addresses are actually 1 byte
- Address only adds prefix for AccountId (0xff)


jacogr:
Related #161

PR #200, opened 2018-10-04
Small fixes in operation

jacogr:
- query block hash from chain, not state
- Bytes.toJSON() should include the actual length

### DIFF
--- a/packages/types/src/Bytes.ts
+++ b/packages/types/src/Bytes.ts
@@ -3,6 +3,7 @@
 // of the ISC license. See the LICENSE file for details.
 
 import u8aConcat from '@polkadot/util/u8a/concat';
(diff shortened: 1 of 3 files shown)


packages/types/src/Bytes.ts
```diff
@@ -19,6 +20,14 @@ export default class Bytes extends U8a {
     return this.length + Compact.encode(this.length).length;
   }
 
+  toHex (): string {
```
amaurymartiny: Not sure if you need it in the UI, but in #172 we could add toHex/fromHex to all types?

jacogr: That would actually be nice - it currently is only on the U8a ones, but that probably will make sense to all for the sake of consitency.

PR #203, opened 2018-10-04
fix: Fixes #154 Modify TypeDocs that change the source to resolve rendering correctly

### LINKED ISSUE #154: API documentation does not render correctly

ltfschoen:
* Add TypeDoc/TSDoc labels

* Use HTML <BR> element in TypeDoc to generate new lines even to achieve a newline between lines of code

* Use HTML `<PRE><CODE>` combination in TypeDoc for multiple line code generation

* Add missing TypeDocs for HTTP `constructor` and `send` for HTTP and WS

* Remove unnecessary assignment of `= true` default value in TypeDoc @param value and clean up text alignment

* Verified that it displays correctly in development environment by running `yarn; yarn polkadot-dev-build-docs; yarn gitbook serve`, and then navigating to localhost:4000

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -14,12 +14,17 @@ import ExtError from '@polkadot/util/ext/error';
 import isFunction from '@polkadot/util/is/function';
 
 /**
+ * @name Api
+ * @summary The API may use a HTTP or WebSockets provider.
+ * @description It allows for querying a Polkadot Client Node.
  * @example
+ * <BR><PRE><CODE>
  * import Api from '@polkadot/api';
  * import WsProvider from '@polkadot/api-provider/ws';
- *
+ * <BR>
  * const provider = new WsProvider('http://127.0.0.1:9944');
  * const api = new Api(provider);
+ * </CODE></PRE>
  */
 export default class Api implements ApiInterface {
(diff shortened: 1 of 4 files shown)


packages/api/src/index.ts
```diff
@@ -46,14 +51,17 @@ export default class Api implements ApiInterface {
 
   /**
    * @name signature
-   * @signature jsonrpcSignature (method: InterfaceMethodDefinition): string
+   * @signature `jsonrpcSignature (method: InterfaceMethodDefinition): string`
```
jacogr: This could be whacked. I actually saw it a day or two ago and then got ditracted. (It was for my lil jsdoc extension that built the docs on flowtype - when it still worked)

ltfschoen: I've pushed commit c9bc957 that removes this line

PR #207, opened 2018-10-05
fix: Fixes #204. Adds coloured syntax highlighting to TypeDoc code examples

### LINKED ISSUE #204: Coloured syntax highlighting missing in code snippets generated from TypeDocs in API docs
Currently the examples that are generated from Markdown include coloured syntax highlighting as shown below
screen shot 2018-10-04 at 18 53 32

However the API Docs themselves that are generated from TypeDoc/TSDoc labels in code comments sometimes also contain an `@example` code snippet, however it does not contain any coloured syntax highlighting. 

screen shot 2018-10-04 at 18 56 16

The approach used to achieve generation of multi-line code snippets including spaces between lines of code was to wrap the code with `<PRE><CODE> ... </CODE></PRE>` and to add `<BR>` between lines of code where a newline was desired. However it does not colour the code based on the syntax used.

One possible solution could be to use [Google Code Prettify](https://github.com/google/code-prettify), which according to the usage instructions simply requires applying their CSS class to one of the elements i.e. `<PRE class="prettyprint">` or `<CODE class="prettyprint">`, and somehow loading their JS file in the root HTML document (or just injecting their CDN script) that's generated by building with Gitbooks  (i.e. `<script src="https://cdn.rawgit.com/google/code-prettify/master/loader/run_prettify.js"></script>`). However I tried following this approach, but it did not work, and may be because CSS classnames may not be added to HTML elements that are used in TypeDocs.

ltfschoen:
* Add Gitbook plugin Sunlight Highlighter https://www.npmjs.com/package/gitbook-plugin-sunlight-highlighter

* Still requires use of <BR> HTML element to shift code onto a new line

* Example use of Sunlight Highlighter from their README `javascript+theme:gitbook+lineNumbers:true+lineNumberStart:100` (in this example it uses their gitbook theme with line numbers shown and line numbers starting from 100)

* Updated book.json to add the Sunlight Highlighter plugin, and explicitely configured it to use the gitbook theme by default

* Example below shows enabling line numbers only on TypeDoc example code that is longer (i.e. 8 lines long) using ````javascript+lineNumbers:true+lineNumberStart:1`

screen shot 2018-10-05 at 08 09 45

* Example below shows coloured syntax being used just ````javascript`

screen shot 2018-10-05 at 08 08 13

### DIFF
--- a/packages/api-provider/src/ws/index.ts
+++ b/packages/api-provider/src/ws/index.ts
@@ -180,7 +180,7 @@ export default class WsProvider extends E3.EventEmitter implements WSProviderInt
    * @example
    * <BR>
    *
-   * ```javascript+lineNumbers:true+lineNumberStart:1
+   * ```javascript
    * const provider = new WsProvider('ws://127.0.0.1:9944');
    * const api = createApi(provider);
    *
--- a/packages/api-provider/src/ws/index.ts
+++ b/packages/api-provider/src/ws/index.ts
@@ -43,13 +43,15 @@ interface WSProviderInterface extends ProviderInterface {
  * @description Unlike the [[HttpProvider]], it does support subscriptions and allows
  * listening to events such as new blocks or balance changes.
  * @example
- * <BR><PRE><CODE>
+ * <BR>
+ *
+ * ```javascript+lineNumbers:false
  * import createApi from '@polkadot/api';
  * import WsProvider from '@polkadot/api-provider/ws';
- * <BR>
+ *
  * const provider = new WsProvider('ws://127.0.0.1:9944');
  * const api = createApi(provider);
- * </CODE></PRE>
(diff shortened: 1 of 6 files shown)


packages/api-provider/src/ws/index.ts
```diff
@@ -176,16 +178,18 @@ export default class WsProvider extends E3.EventEmitter implements WSProviderInt
    * @return {Promise<number>}                     Promise resolving to the dd of the subscription you can use with [[unsubscribe]].
    *
    * @example
-   * <BR><PRE><CODE>
+   * <BR>
+   *
+   * ```javascript+lineNumbers:true+lineNumberStart:1
```
jacogr: We are not using this consistently, it should be either with or without.

PR #212, opened 2018-10-05
Mortal -> Extrinsic, expand for decoding start

jacogr:
Related #161 

- Extrinsic is not decodable (eras still an issue)
- Extrinsic now extends from Struct, wrapping components


packages/types/src/MethodIndex.spec.js
```diff
@@ -2,17 +2,17 @@
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
 
-import CallIndex from './CallIndex';
+import MethodIndex from './MethodIndex';
 
 describe('CallIndex', () => {
```
amaurymartiny: Here should be MethodIndex

PR #217, opened 2018-10-05
Intelligent constructor

amaurymartiny:
Effort to remove all `fromJSON` and `fromU8a` from types, and initialize them solely via constructor. Their constructor should be intelligent enough to handle all kinds of inputs (incl. of course JSON and Uint8Array inputs).

For consistency, I propose that this decoding/input handling be done in the static method `decode`. The constructor will just call this `decode` method.

Related #161.

Done in this PR:
- [x] Struct
- [x] Text
- [x] Vector
- [x] Uint
- [x] Enum
- [x] Option
- [x] Ua8
- [ ] EnumType
- [ ] Extrinsic


jacogr:
On-board with the `static decode`, we have been doing this de-facto anyway. (It helps to build in some hoops before passing stuff to `super`)

packages/types/src/Text.ts
```diff
@@ -18,14 +22,31 @@ import Compact from './codec/Compact';
...
 
+  static decode (input: any): string {
+    if (isString(input)) {
+      return input;
+    } else if (input instanceof Text) {
+      return input.raw;
+    } else if (input instanceof Uint8Array) {
+      const [offset, length] = Compact.decode(input);
+      return u8aToUtf8(input.subarray(offset, offset + length.toNumber()));
+    } else if (Array.isArray(input)) {
```
jacogr: I would propose removing Array until needed. As it stands we don't have anything coming in anymore as `[...]`

packages/types/src/codec/Struct.ts
```diff
@@ -43,30 +46,54 @@ export default class Struct <
...
-        // the [key in keyof T] was added, the idea is to provide better checks, which
-        // does backfire here, but works externally.)
-        raw[key] = new Types[key](
-          isTuple && Array.isArray(value)
-            ? value[index]
-            // @ts-ignore as above
-            : value[key]
-        );
+        if (value instanceof Uint8Array) {
+          const [compactLength, typeLength] = Compact.decode(value.subarray(currentIndex));
```
jacogr: The length needs to be decoded by the child itself, not this one.

packages/types/src/codec/Struct.ts
```diff
@@ -43,30 +46,54 @@ export default class Struct <
       }, {} as E);
   }
 
-  static decode <S, V, T> (Types: S, value: V | Array<any>, isTuple: boolean): T {
+  static decode<S, V, T> (Types: S, value: V | Array<any> | AnyU8a, isTuple: boolean): T {
```
jacogr: Erk, the isTuple is a hold-over, I don't believe is is used anymore, Tuples do their own thing. (It was pre-changes when stuff still came in as `Array<number>`)

PR #219, opened 2018-10-06
Add Compact{16,32,64,128,256}

### LINKED ISSUE #197: Add Compact{8,16,32,64,128,256} coders
- Compact to extend UInt as a base
- Drop-in replacement for UInt with compact encoding
- Adjust Type to turn `Compact<u32>` into `Compact32` (along with 8, 16, 64, 128, 256)

jacogr:
- Closes #197 
- Related #161


packages/types/src/Bytes.ts
```diff
@@ -16,11 +16,11 @@ export default class Bytes extends U8a {
   }
 
   byteLength (): number {
-    return this.length + Compact.encode(this.length).length;
+    return this.length + Compact.encodeU8a(this.length, DEFAULT_LENGTH_BITS).length;
   }
 
   fromU8a (input: Uint8Array): Bytes {
-    const [offset, length] = Compact.decode(input);
+    const [offset, length] = Compact.decodeU8a(input, DEFAULT_LENGTH_BITS);
```
amaurymartiny: Instead of adding `DEFAULT_LENGTH_BITS` as 2nd arg everywhere, you could add a default value for this 2nd arg in the function definition?

jacogr: It was that way originally, but it creates issues and bugs that _look_ correct.

i.e. for numbers, you want the default to probaly be 64, but for lengths it needs to be 32. So made it explicit everywhere. (The issue is not an issue yet, but will be - all numbers will be compact encoded in the system, so want to make sure we pass the param through and not just forget about it.)

Issue #176, opened 2018-09-30
Rework api & rpc packages


jacogr:
Rename the actual RPC wrappers -

- api -> rpc-core
- api-rx -> rpc-rx
- api-provider -> rpc-provider

Then for the actual high-level API -

- api-observable -> api

In this `api-codec` seems like the odd man out, I would make the published package `@polkadot/codec`

@amaurymartiny Thoughts?

amaurymartiny:
How we name internally our folders doesn't actually matter to me, the important thing is the `@polkadot/*` npm packages name.

So I'm thinking
- `api`-> `@polkadot/rpc` or `@polkadot/api`
- `api-rx`->`@polkadot/rx` : This one shouldn't be used at all
- `api-provider`->`@polkadot/provider`
- `api-codec`->`@polkadot/codec` or `@polkadot/types` : users don't need to know we do codec, but need to know these are the the types they actually need to use.
- `api-observable` -> `@polkadot/observables`
- ??? -> `@polkadot/api`: We don't have this one yet (because the UI doesn't need it), but I would like to see a simple-to-use api library that is promised-based. Ideally it would just import `@polkadot/observables` and run `.toPromise()` everywhere, because Rx is just a programming pattern among others, while Promises are standard. And this one would be called `@polkadot/api`.

jacogr:
- `@polkadot/rpc` - happy
- `@polkadot/rpc-rx` - maybe not externally used, but it does have some critical function
- `@polkadot/provider` - happy
- `@polkadot/types` - happy (full-circle, primitives was types, then became primitives + some coding, then replaced, now back)
- `@polkadot/api-rx` - feels simpler than typing `observables`
- `@polkadot/api` - wraps above, just with `.toPromise()`?

jacogr:
Thought about the .toPromise() thing - it is actually an issue, since most of the use is actually queries and these are for anything non-trivial subscription-based. Ror that your options are -

- observables
- (current) api package

Just went through what use we actually have out there - 2 datapoints only, Polkabot & UI. So not convinced that the .toPromise() will add any value. Happy to re-visit when we have actual requests and rather focus on making what is there bullet-proof.

jacogr:
@amaurymartiny So thinking about this (expanding the above). There is the promise & rx-based versions. 

Storage -

- rx-based queries via chain.storage, which is a subscription (updates are pushed as they come in)
- promise-base queries via chain.getStorage, which is just a single value

So effectively -

[...]

PR #220, opened 2018-10-06
First stage renames, rpc-{core,provider,rx}

jacogr:
- Part of https://github.com/polkadot-js/api/issues/176
- Related #161

Items -

- Remove generated docs so they can be re-generated with new names
- Rename api -> rpc-core, api-provider -> rpc-provider & api-rx -> rpc-rx
- Update examples to just use the new packages (they still need to be poc-3 friendly)

Currently it does not touch api-observable (yet). Still some thinking around that one and the api-* namespace required.

index ts polkadot workspace 2018-10-06 16-59-49


jacogr:
I just don't like the removed lines/files here. Going to revert those - it needs to be cleaned, but this way is just horrible. (CI needs to do it, otherwise it is just a road to having a hiding down the road)

PR #222, opened 2018-10-08
fix: Fixes #221. Exclude tests files in build directory. Only use those in src

### LINKED ISSUE #221: Duplicate tests since build directory also being included in Jest testing
When I run the tests with `yarn run test`, in the summary it shows that it's testing duplicate copies of the same *.test.js files that are both in the src/ and build/ directories of each package. See screenshot below:

screen shot 2018-10-08 at 11 40 22

Also, when I run the tests with `yarn run test`, a couple of the tests were failing due to the error ` The name `@polkadot/_____` was looked up in the Haste module map. It cannot be resolved, because there exists several different files, or packages, that provide a module for that particular name and platform. The platform is generic (no extension). You must delete or blacklist files until there remains only one of these`. See screenshot below.
screen shot 2018-10-08 at 11 41 28

ltfschoen:
* See modulePathIgnorePatterns: https://jestjs.io/docs/en/configuration.html#modulepathignorepatterns-array-string


ltfschoen:
I've added a commit that incorporates the renaming that occurred in PR #220.

ltfschoen:
@jacogr I've added a commit with api/build included in the list

PR #225, opened 2018-10-08
Lowercase bool type

jacogr:


packages/types/src/index.ts
```diff
@@ -16,7 +16,8 @@ export { default as Balance } from './Balance';
 // export * from './Bft;
 export { default as Block } from './Block';
 export { default as BlockNumber } from './BlockNumber';
-export { default as Bool } from './Bool';
+// NOTE We are exporting Bool as bool to match with Rust
```
amaurymartiny: Do we need to follow Rust types' case?

These are JS classes after all, imo we should follow JS guidelines above Rust code-similarity.

jacogr:
I sunno, maybe we should just sat screw-it and stick to JS naming. I dunno anymore. One the one hand great to be close, on teh other hand we have deviated a lot already by cleaning up.

Issue #226, opened 2018-10-08
Duplicate types PropIndex and ProposalIndex


ltfschoen:
These two files appear to be duplicate of each other:

https://github.com/polkadot-js/api/blob/master/packages/types/src/PropIndex.ts

https://github.com/polkadot-js/api/blob/master/packages/types/src/ProposalIndex.ts

jacogr:
Yeap, the naming sucks - however they are meant to be different types. The issue originates in Rust actually, remember we just have a mapping 1-to-1 with the Rust types so we can interpret the metadata. 

Specially asked about this one, cannot just alias the 2 to 1 (although the current are both u32)

From the code -

- PropIndex - public proposal index
- ProposalIndex - council proposal index

jacogr:
Update to above, here is where the possible alias is marked - https://github.com/polkadot-js/api/blob/master/packages/types/src/Type.ts#L64

(However, since these are distinct types, cannot quite execute on that - have to keep an eye on Rust and see if the situation improves)

PR #236, opened 2018-10-10
fix: Fixes #235 TypeDoc updates consistent with renaming

### LINKED ISSUE #235: TypeDocs to be updated to be consistent first stage of renaming
Need to update the TypeDocs to accommodate the renaming that occurred in PR #220

ltfschoen:
* Add link to parity-codec repo

* Move non-type dependency import with non-type dependencies

* Update Readme's to use Yarn instead of NPM install

* Update type-jsonrpc Readme. Fix typo. Refer to TSLint not Flow

* Refer to rpc-core as Rpc instead of Api

* Update rpc-rx files to refer to it as RpcRx (instead of RxApi or just Api)

* Refer to rpc-core as rpc-core (instead of as api) in the rpc-rx Readme

* Fix name in license header and rename Api to Rpc for rpc-core e2e tests

* Update Readme of rpc-core.
    
    * Update name to be @polkadot/rpc-core instead of @polkadot/api
    
    * Refer to @polkadot/rpc-core as `Rpc`
    
    * Highlight that given block header hash should be a 0x-prefixed hex string with length of 64 (not only length of 10 as shown in the example)
    
    * Showcase more information that you can obtain from the header object (i.e. best block number, parentHash, stateRoot, extrinsicsRoot, and digest)
    
    * Use `header.raw` instead of just `header` to access properties of the header object
    
    * Update the example to console.log the subscription ID
    
    * Add missing semicolon

* Update api-observable Readme to refer to rpc-core, previously called api

* Clarifies difference between PropIndex and ProposalIndex from Issue #226 in comments

* Remove disused mockHttp and mockWs that are duplicates of those in rpc-provider

* Change license from ui-observable to api-observable

* Remove the guide file that serves no purpose, along with where it's referred to in the `includes` section of the typedoc.js

* Remove docs/api-rx folder/files since replaced with docs/rpc-rx

* Remove docs/api-provider folder/files since replaced with docs/rpc-provider

* Remove docs/api folder/files since replaced with docs/rpc-core


packages/rpc-core/README.md
```diff
@@ -64,11 +68,12 @@ api.chain
       console.error('error:', error);
     }
 
-    console.log(`best #${header.number.toString()}`);
+    console.log(`best #${header.raw.number.toString()}`);
```
jacogr: This should be `header.blockNumber` rather. (As far as possible, should not encourage access to .raw)

PR #251, opened 2018-10-11
Add api/promise

jacogr:


packages/api/src/promise/SubmittableExtrinsic.ts
```diff
@@ -0,0 +1,34 @@
...
+import { AnyNumber, AnyU8a } from '@polkadot/types/types';
+import { ApiPromiseInterface } from './types';
+
+import { Extrinsic, Hash } from '@polkadot/types/index';
+
+export default class SubmittableExtrinsic extends Extrinsic {
+  private _api: ApiPromiseInterface;
+
+  constructor (api: ApiPromiseInterface, extrinsic: Extrinsic) {
+    super({
```
amaurymartiny: `super(extrinsic)` might magically work.

packages/api/src/promise/index.ts
```diff
@@ -0,0 +1,117 @@
...
+   * import Api from '@polkadot/api/promise';
+   *
+   * Api.create().then(async (api) => {
+   *   const timestamp = await api.st.timestamp.now();
+   *
+   *   console.log(`lastest block timestamp ${timestamp}`);
+   * });
+   * ```
+   */
+  static create (wsProvider?: WsProvider): Promise<ApiPromise> {
```
ltfschoen: could we update the example to show how to use the optional `wsProvider` parameter.
i think it'd be used as follows:
```
   * ```javascript
   * import Api from '@polkadot/api/promise';
   * import Ws from '@polkadot/rpc-provider/ws';
   *
   * const wsProvider = new Ws('ws://127.0.0.1:9944');
   * 
   * Api.create(wsProvider).then(async (api) => {
   *   const timestamp = await api.st.timestamp.now();
   *
   *   console.log(`lastest block timestamp ${timestamp}`);
   * });
   * ```
```

jacogr: 100% right. The next PR trying to make a start at actual docs. (All these need to go into examples)

PR #252, opened 2018-10-11
Intelligent constructor (Part 2)

amaurymartiny:
Follows #217.

All types have their `decode` method.

Notable changes:
- Bytes (therefore StorageKey) and Extrinsic are now **always** length-prefixed, updated tests.
- Since the JSON is not length-prefixed, I added a FIXME in their decode method to manually add this length for now is the value is a hex

Again, I didn't touch fromJSON/fromU8a, so should work.


packages/type-storage/src/utils/createFunction.ts
```diff
@@ -59,13 +60,19 @@ export default function createFunction (
...
         u8aConcat(
           u8aFromUtf8(`${section.toString()} ${method.toString()}`),
           createType(type, arg).toU8a(true)
         ),
         128
       );
+
+      // StorageKey is a Bytes, so is length-prefixed
+      return u8aConcat(
+        Compact.encodeU8a(methodAndArgs.length, DEFAULT_LENGTH_BITS),
```
jacogr: As discussed, not sure about this at all. If I look at what works, it is just the hex, with no length -

`freeBalance('5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ')` -> `0xce3f3d8f09e3411403f5ca59d042a40e` 16 bytes, no length prefix. (That works as expected on the UI)

jacogr: ... Then again, once passed to storageKey it will whack the length.

PR #257, opened 2018-10-11
Storage key are not length-prefixed anymore

amaurymartiny:
Also: now `storage.balances.freeBalance('5GwPuAgYgP6q58uWTXp4uSg6FwfzQv9HfFZwAFEREUrQjCvy')` returns a StorageKey instead of a U8a.

### DIFF
--- a/packages/type-storage/src/utils/createFunction.ts
+++ b/packages/type-storage/src/utils/createFunction.ts
@@ -18,6 +18,18 @@ export interface CreateItemOptions {
 }
 
 /**
+ * Prepend a Uint8Array with its compact length.
+ *
+ * @param u8a - The Uint8Array to be prefixed
+ */
+function addLengthPrefix (u8a: Uint8Array): Uint8Array {
+  return u8aConcat(
+    Compact.encodeU8a(u8a.length, DEFAULT_LENGTH_BITS),
+    u8a
+  );
+}
+
+/**
  * From the schema of a function in the module's storage, generate the function
  * that will return the correct storage key.
  *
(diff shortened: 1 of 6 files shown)


packages/types/src/KeyValue.spec.js
```diff
@@ -5,11 +5,11 @@
 import KeyValue, { KeyValueOption } from './KeyValue';
 
 describe('KeyValue', () => {
-  it('decodes KeyValue from u8a', () => {
+  it.skip('decodes KeyValue from u8a', () => {
```
amaurymartiny: If StorageKey is not length-prefixed, there's not way we can decode from u8a. Fortunately we don't need the decoding for KeyValue.

amaurymartiny:
Alright. So if input or output is U8a, then it's length-prefixed. If it's in hex, then it's not length-prefixed.

packages/type-storage/src/utils/createFunction.ts
```diff
@@ -17,6 +17,18 @@ export interface CreateItemOptions {
   method?: string;
 }
 
+/**
+ * Prepend a Uint8Array with its compact length.
+ *
+ * @param u8a - The Uint8Array to be prefixed
+ */
+function addLengthPrefix (u8a: Uint8Array): Uint8Array {
```
amaurymartiny: Not sure where to put this, we use a couple of times

jacogr: Compact?

amaurymartiny:
Related #161

PR #263, opened 2018-10-12
Round out ApiPromise examples

jacogr:


docs/examples/promise/07_transfer_dots/index.js
```diff
@@ -0,0 +1,36 @@
...
+  // retrieve the nonce for Alice, used to sign the transaction
+  const aliceNonce = await api.st.system.accountNonce(alice.address());
+
+  // Create a extrinsic, transferring 12345 units to Bob. We can also create,
+  // sign and send in one operation (as per the samples in the Api documentation),
+  // here we split it out for the sake of readability
+  const transfer = api.tx.balances.transfer(BOB_ADDR, 12345);
+
+  // sign the transaction using our account
+  transfer.sign(alice, aliceNonce);
```
amaurymartiny: Maybe sign should already take care of nonce? Not sure if it should go here, on in the combined api-observable.

jacogr: It it does return a promise, yes. Currently it just returns the observable. We had this chat a couple of days ago, not sure which way is best - I would love it to be optional as well. For now, not yet...

PR #264, opened 2018-10-12
Align all RPC methods with Rust

jacogr:


jacogr:
Update: Just completing e2e for updateExtrinsic tests and checking our favourite Metadata

jacogr:
Ok, so the tests took a bit of a turn -

- had to update EnumType to properly decode with fromJSON
- Added tests for actual extrinsicUpdate (e2e & rpc with decoding)
- Made `.send()` in ApiPromise subscriptable (optional) as well
- Had to do proper checks for optional args in rpc-core

jacogr:
cc @yjkimjunior I copied your extrinsicUpdate and got it working with e2e tests as well. (No issues on your side at all). So this will close your https://github.com/polkadot-js/api/pull/181

PR #255, opened 2018-10-11
feat: Generate Methods in API docs. Contains bugs. See detailed comment

ltfschoen:
* Run:
```
./node_modules/.bin/tsc packages/types/src/Metadata-md.ts; node packages/types/src/Metadata-md.js;
yarn run build; yarn gitbook serve
```
* It should generate docs/METHODS.md (see Bugs below) 
* It should be viewable at http://localhost:4000/METHODS.html (see Bugs below)

Bugs:
1) Gitbooks generates METHODS.html from the METHODS.md, and there's a link to it in the SUMMARY.md file, but is not formatting it correctly.

2) It generates the following error for some reason all of a sudden. Initially it was working correctly.

```
/Users/me/code/src/polkadot-js/api/packages/types/src/codec/Compact.js:27
var MAX_U8 = new bn_js_1["default"](2).pow(new bn_js_1["default"](8 - 2)).subn(1);
             ^

TypeError: bn_js_1.default is not a constructor
    at Object.<anonymous> (/Users/scon/code/src/polkadot-js/api/packages/types/src/codec/Compact.js:27:14)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/scon/code/src/polkadot-js/api/packages/types/src/codec/Vector.js:22:17)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
```

PR #249, opened 2018-10-11
Generate md from metadata

jacogr:
It generates markdown from metadata - which is actually great to include somewhere in the docs. 

Implementation wise, well.. "when you have a hammer..." - it is done as a spec (???!). Why? To get it going fast without having to worry about babel-something-something and typescript-something-something.

It needs a lot of polish, some nice intros, formatting, etc. If useful, let it in and work from this, if not copy and/or ignore. (I do believe we need a link to "base methods in substrate" so people have a place to start - this is meant to provide that)

Output (as printed to console) -

# Metadata
    
    
    
... 

## Storage

...

### system
accountNonce(`AccountId`): `Index`

extrinsicCount(): `u32`

blockHash(`BlockNumber`): `Hash`

extrinsicData(`u32`): `Bytes`

randomSeed(): `Hash`

number(): `BlockNumber`
  The current block number being processed. Set by `execute_block`.
parentHash(): `Hash`

extrinsicsRoot(): `Hash`

digest(): `Digest`

events(): `Vec<EventRecord>`

### consensus
originalAuthorities(): `Vec<SessionKey>`

### balances
totalIssuance(): `Balance`
  The total amount of stake on the system.
existentialDeposit(): `Balance`
  The minimum amount allowed to keep an account open.
reclaimRebate(): `Balance`
  The amount credited to a destination's account whose index was reclaimed.
transferFee(): `Balance`
  The fee required to make a transfer.
creationFee(): `Balance`
  The fee required to create an account. At least as big as ReclaimRebate.
nextEnumSet(): `AccountIndex`
  The next free enumeration set.
enumSet(`AccountIndex`): `Vec<AccountId>`
  The enumeration sets.
freeBalance(`AccountId`): `Balance`
  The 'free' balance of a given account.   This is the only balance that matters in terms of most operations on tokens. It is  alone used to determine the balance when in the contract execution environment. When this  balance falls below the value of `ExistentialDeposit`, then the 'current account' is  deleted: specifically `FreeBalance`. Furthermore, `OnFreeBalanceZero` callback  is invoked, giving a chance to external modules to cleanup data associated with  the deleted account.   `system::AccountNonce` is also deleted if `ReservedBalance` is also zero (it also gets  collapsed to zero if it ever becomes less than `ExistentialDeposit`.
reservedBalance(`AccountId`): `Balance`
[...]

Issue #193, opened 2018-10-03
Document new poc-3 era API


jacogr:
TODOs (will be expanded)

- adapt current samples to be poc-3 compliant
- jsonrpc needs all methods documented
- storage needs full documentation on use
  - [ ] static interface
  - [ ] dynamic interface
  - [ ] actual use
- extrinsics needs full documentation on use
  - [ ] static interface
  - [ ] dynamic interface
  - [ ] actual use

Potentially -

~~api-observable needs full documentation for all the functions (Will be renamed to @polkadot/api)~~ (not advocating use, examples or docs - only UI use, to be revamped)

jacogr:
cc @amaurymartiny Please fee free to adjust the above, it needs a lot of detail.

ltfschoen:
PROGRESS UPDATE (please correct me if i'm wrong)

- [x] adapt current samples to be poc-3 compliant
  - **STATUS** - Interfaces sections have been including in the menu in the [API documentation portal](https://polkadot.js.org/api/) including [RPC](https://polkadot.js.org/api/METHODS_RPC.html), [Chain state (runtime)](https://polkadot.js.org/api/METHODS_STORAGE.html), [Extrinsics (runtime)](https://polkadot.js.org/api/METHODS_EXTRINSICS.html), and [Events (runtime)](https://polkadot.js.org/api/METHODS_EVENTS.html) that reflect those available in the Substrate UI https://polkadot.js.org/apps/next/ that contains PoC-3 updates from the polkadot-js/apps "next" branch
- [x] jsonrpc needs all methods documented
  - **STATUS** - See [RPC](https://polkadot.js.org/api/METHODS_RPC.html)
- storage needs full documentation on use
  - [x] static interface
    - **STATUS** - See [Chain state (runtime)](https://polkadot.js.org/api/METHODS_STORAGE.html)
  - [ ] dynamic interface
    - **STATUS** - **What are the expectations here?**
  - [ ] actual use
    - **STATUS** - See examples https://polkadot.js.org/api/examples/promise/, and https://polkadot.js.org/api/examples/rx/
- extrinsics needs full documentation on use
  - [x] static interface
    - **STATUS** - See [Extrinsics (runtime)](https://polkadot.js.org/api/METHODS_EXTRINSICS.html)
  - [ ] dynamic interface
    - **STATUS** - As above
  - [ ] actual use
[...]

PR #265, opened 2018-10-12
docs: Relates to #193. Generate Substrate Interface methods in API docs for RPC, Storage, and Extrinsics from Metadata

ltfschoen:
Note that each commit has detailed commit comments.

* Summary of Steps to preview the changes in this PR:
  * Generate Markdown files before running and previewing Gitbook at localhost:4000/api
    ```
    yarn run build;
    yarn gitbook serve;
    ```
  * Important Note: When you run `yarn run build`, it runs `yarn run build:methodsdoc` first before the rest of the commands now. **Consider adding `yarn run build:methodsdoc` as the first script in .travis.yml**

  * Screenshots below show where you can click in UI to view the Substrate Interface method pages:

screen shot 2018-10-12 at 14 58 15
 
screen shot 2018-10-12 at 14 59 17

  * Generate decoded JSON file (packages/types/src/Metadata.decoded.json) of the Metadata in the Metadata.rpc.ts file.
    ```
    yarn run build:decodedjson
    ```
    
* Allows generating a docs/METHODS_STORAGE.md and docs/METHODS_EXTRINSICS.md files by running `yarn run build:methodsdoc`
      * Note: The script runs `node packages/types/src/MetadataMdWrapper.js`, which uses Babel.js to transpile the TypeScript file MetadataMd.ts into JavaScript as MetadataMd.js by using the babel-plugin-module-resolver dependency that has been added to package.json. It runs MetadataMd.js, which parses a hard-coded copy of a Metadata response from the Substrate Node (and also appends the contents of METHODS_STORAGE_SUBSTRATE.md to the bottom of METHODS_STORAGE.md, see later comments for details), and then generates the two aforementioned Markdown files to be included in the Gitbook.
    
* Added links that work when you go to http://localhost:4000/api/ and click the "api" link on the side menu
    * Note: This was achieved by first adding the links in packages/api/README.md (i.e. `[Extrinsics (Substrate Interface)](../METHODS_EXTRINSICS.md)`). Note that the links could also be `METHODS_EXTRINSICS.html` and it would still work. It is VERY important to add links to these Markdown files in docs/SUMMARY.md (i.e. `[Storage](METHODS_STORAGE.md)`), as Gitbook generates HTML files in the same docs/ directory (i.e. it generates docs/METHODS_STORAGE.html) for any links provided in that file, and when you run `yarn run build; yarn gitbook serve` it will put a copy of those HTML files in the _book directory so it can be accessed in the browser.
   
* Manually reads METHODS_STORAGE_SUBSTRATE.md that contains Storage Substrate (storage.substrate) methods, and appends them to the Storage Methods at the same time as METHODS_STORAGE.md is generated (since storage.substrate methods are not available from Metadata)
    
* Consistent styling with:
  * Description at top of Substrate Interface pages for RPC, Extrinsics, and Storage
  * ▸ symbol before each method
  * bold method names
  * adding Summary label and only showing the summary label if a summary exists
  * adding a horizonal line between each section of methods
    
* Fix by adding new line after description when there is one in Extrinscis and Storage

* Add script to generate decoded JSON file of the Metadata. Remove test that was there before.
  * Running `yarn run build:decodedjson` will generate packages/types/src/Metadata.decoded.json from the Metadata in the Metadata.rpc.ts file.

* Replaces PR #249 and #255 from jg-meta-md from jg-meta-md-luke branches


docs/METHODS_RPC.md
```diff
@@ -0,0 +1,107 @@
+## RPC
+
+_The following RPC methods are not part of the default Substrate runtime. They are hard-coded in Substrate._
```
jacogr: Ok, this is a bit wrong. The RPCs are never exposed by the runtime, rather they are provided by substrate itself. So these are the Remote Calls that are available by default. (It allows you to interact with the actual node, query and submit)

docs/SUMMARY.md
```diff
@@ -7,6 +7,11 @@
   - [HttpProvider](rpc-provider/classes/_http_index_.httpprovider.md)
   - [WsProvider](rpc-provider/classes/_ws_index_.wsprovider.md)
 
+## Substrate Interfaces
```
jacogr: Rather call this `Runtime Node Interfaces`

packages/api/README.md
```diff
@@ -7,8 +7,12 @@ The API wrappers provide a standard interface for use -
 - A static `.create(<optional WsProvider>)` that returns an API istance when connected, decorated and ready-to use
 - The above is just a wrapper for `new Api(<optional WsProvider>) `, exposing the `isReady` getter
 - `api.rpc.<section>.<method>` provides access to actual RPC calls, be it for queries, submission or retrieving chain information
+  - [RPC (Substrate Interface)](../METHODS_RPC.md)
```
jacogr: Here we can remove the "Substrate part"

packages/types/src/METHODS_STORAGE_SUBSTRATE.md
```diff
@@ -0,0 +1,26 @@
+
+---
+
+### substrate
+
+_Well known keys that are not returned by state.getMetadata()_
```
jacogr: Makes sense, but rather don't refer to the call, just say "these are keys that are always available to the runtime implementation". (It is like the RPC calls, i.e. always there, not dynamic)

packages/types/src/Metadata.spec.js
```diff
@@ -1,19 +0,0 @@
-// Copyright 2017-2018 @polkadot/types authors & contributors
```
jacogr: Rather put this one back - it has benefit running as part of the test suite, I'd rather not lose it. (Even the logging is useful since it shows that it works and then the unique types as well. So just even visually it helps a _lot_ in debugging the Metadata)

packages/types/src/MetadataJsonDecode.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright 2017-2018 @polkadot/types authors & contributors
```
jacogr: Maybe move all of these under `scripts`, so it is not in the root?

packages/types/src/MetadataJsonDecodeWrapper.js
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
```
jacogr: Just move.

packages/types/src/MetadataMd.ts
```diff
@@ -0,0 +1,99 @@
+// Copyright 2017-2018 @polkadot/types authors & contributors
```
jacogr: Just move it.

packages/types/src/MetadataMdWrapper.js
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
```
jacogr: Just move it.

docs/METHODS_RPC.md
```diff
@@ -0,0 +1,107 @@
...
+▸ **getHeader**(hash: `Hash`): `Header`
+- **summary**: Retrieves the header for a specific block
+
+▸ **getRuntimeVersion**(): `RuntimeVersion`
+- **summary**: Get the runtime version
+
+▸ **getRuntimeVersionAt**(hash: `Hash`): `RuntimeVersion`
+- **summary**: Get the runtime version at a specific block
+
+▸ **newHead().subscribe**(`<CALLBACK>`): `Header`
```
jacogr: Generate all these as well via script, it is outdated from what is in master atm.

docs/SUMMARY.md
ltfschoen: If RPCs are never exposed by the runtime, should RPC be under the "Runtime Node Interfaces" heading?

jacogr: They are expose by the Node. So it is basically -

- RPC
- Chain state (runtime)
- Extrinsics (runtime)

So I would denote the latter 2 with "(runtime)" 

In that case, the heading is misleading, you are right. (Would just remove "substrate", so basically just "Interfaces" works here as well)

ltfschoen:
@jacogr just doing one more commit so the RPC page shows subscribe/unsubscribe

ltfschoen:
@jacogr I've pushed the last commits

PR #270, opened 2018-10-12
Do not use fromU8a/fromJSON anywhere

amaurymartiny:
- I left the fromU8a and fromJSON just to be sure.
- The most important change is removing `fromJSON` in `packages/rpc-core/src/index.ts`, putting it back should make the UI work like before.
- `new Metadata(fromRPC)` works, so that's pretty cool

**IMPORTANT:** While I tried to think of all possible scenarios, there's a high risk this will break the UI, because of unthought cases in the `decode` functions.


packages/rpc-core/src/index.ts
```diff
@@ -177,7 +177,7 @@ export default class Rpc implements RpcInterface {
   }
 
   private formatOutput (method: RpcMethod, params: Array<Base>, result?: any): Base | Array<Base | null | undefined> {
-    const base = createType(method.type as string).fromJSON(result);
+    const base = createType(method.type as string, result);
```
amaurymartiny: This is the critical line.

packages/types/src/codec/EnumType.ts
```diff
@@ -18,14 +22,15 @@ type TypesDef = {
...
 //   - It should rather probably extend Enum instead of copying code
 //   - There doesn't actually seem to be a way to get to the actual determined/wrapped value
-export default class EnumType <T> extends Base<Base<T>> {
+export default class EnumType<T> extends Base<Base<T>> {
   private _Types: TypesArray;
   private _index: number;
   private _indexes: Array<number>;
 
-  constructor (def: TypesDef, index?: number | EnumType<T>, value?: any) {
+  constructor (def: TypesDef, value?: any, index?: number | EnumType<T>) {
```
amaurymartiny: Note: Changed the order of value & index.

packages/types/src/Bft.spec.js
```diff
@@ -36,7 +36,7 @@ describe('BftHashSignature', () => {
 
 describe('Justification', () => {
   const just = new Justification({
-    round: 32,
+    round_number: 32, // We get round_number from JSON, and change it to round
```
jacogr: I don't want to construct with `new Justification({ round_number: ... })`, would rather use either `round` or `roundNumber` (which is actually correct). So it sadly needs to cater for renames :(

packages/types/src/Metadata.ts
```diff
@@ -324,7 +327,19 @@ export default class RuntimeMetadata extends Struct {
...
+    }, RuntimeMetadata.decodeMetadata(value));
+  }
+
+  static decodeMetadata (value: any): object | Uint8Array {
+    if (isHex(value)) {
+      return RuntimeMetadata.decodeMetadata(hexToU8a(value));
+    } else if (isU8a(value)) {
+      // this does not look correct, metadata now has a length prefix. Strip, move on.
+      const [offset] = Compact.decodeU8a(value, DEFAULT_LENGTH_BITS);
+      return value.subarray(offset);
```
jacogr: (Nitpick) extra blank line before this.

packages/types/src/Metadata.ts
```diff
@@ -324,7 +327,19 @@ export default class RuntimeMetadata extends Struct {
...
+    }
+    // Decode as normal struct
```
jacogr: (Nitpick) extra blank line before this.

packages/types/src/codec/EnumType.ts
```diff
@@ -18,14 +22,15 @@ type TypesDef = {
...
+    const decoded = EnumType.decodeEnumType(def, value, index);
```
jacogr: (Nitpick) extra blank line after this.

packages/types/src/codec/Option.ts
```diff
@@ -7,6 +7,8 @@ import isU8a from '@polkadot/util/is/u8a';
 import isUndefined from '@polkadot/util/is/undefined';
 
 import Base from './Base';
+import { Constructor } from '../types';
+import Null from '../Null';
```
jacogr: Whoohoo! We found another use for `Null` :)

packages/types/src/codec/Option.ts
```diff
@@ -15,27 +17,30 @@ import Base from './Base';
...
 
     this._isEmpty = isNull(value) || isUndefined(value);
   }
 
-  static decodeOption<O> (Type: { new(value?: any): Base<O> }, value?: any): Base<O> {
+  static decodeOption<O> (Type: Constructor<Base<O>>, value?: any): Base {
     if (isU8a(value)) {
+      if (value[0] === 0) {
+        return new Null();
+      }
```
jacogr: (nitpick) Extra blank line after.

packages/types/src/Bft.spec.js
```diff
@@ -36,7 +36,7 @@ describe('BftHashSignature', () => {
 
 describe('Justification', () => {
   const just = new Justification({
-    round_number: 32, // We get round_number from JSON, and change it to round
+    round: 32, // We get `round_number` from JSON, but both `round` and `round_number` works
```
jacogr: Out of scope here, but will make a round of these and just make sure that the all the places where we use this the `camelCase(input)` matches, so we don't have funnies like `round_number` -> `round`, rather `round_number` -> `roundNumber`. (You suggested getting rid of the maps, that will allow it - well... still needed on toJSON.)

amaurymartiny:
In favor of #294.

PR #273, opened 2018-10-12
Dangling st -> query renames

jacogr:


packages/api/src/Base.ts
```diff
@@ -93,45 +93,45 @@ export default abstract class ApiBase<R, S, E> implements ApiBaseInterface<R, S,
   }
 
   /**
-   * @description Contains all the raw rpc sections and their subsequent methods in the API as defined by the jsonrpc interface definitions. Unlike the dynamic `api.st` and `api.tx` sections, these methods are fixed (although extensible with node upgrades) and not determined by the runtime.
+   * @description Contains all the chain state modules and their subsequent methods in the API. These are attached dynamically from the runtime metadata.
```
jacogr: Looks horrible, but just swapped rpc & query (alphabetical, however have kept the sections together, so runtime* still goes before this)

packages/api/test/e2e/promise-queries.spec.js
```diff
@@ -22,7 +22,7 @@ describe.skip('e2e queries', () => {
     expect(api.runtimeMetadata).toBeDefined();
     expect(api.runtimeVersion).toBeDefined();
     expect(api.rpc).toBeDefined();
-    expect(api.st).toBeDefined();
+    expect(api.query).toBeDefined();
```
jacogr: Since these are only in the e2e tests, I missed them. We need tests ASAP :(

PR #287, opened 2018-10-14
fix: Change Readme contents to reflect implementation

ltfschoen:
I had a closer look at the code to see what Node Interface methods we are wrapping in @polkadot/api-observable, and which Node Interface methods we are wrapping in @polkadot/rpc-core, and I think these changes are appropriate.

The link to the package @polkadot/rpc-rx that we are using appears to have been accidently removed previously, so I've added it back again.

Note importantly that @polkadot/api _contains_ @polkadot/api/rx, whereas @polkadot/rpc-rx is a different separate package

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,10 @@ This library provides a clean wrapper around all the methods exposed by a Polkad
 The API is split up into a number of internal packages -
 
 - [@polkadot/api](packages/api/) The API library, providing both Promise and RxJS Observable-based interfaces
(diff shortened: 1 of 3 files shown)


jacogr:
Yes, api-observable needs to -

- start using api/rx
- remove all the single-use methods, only contain wrappers (i.e. all the stuff where combine is used)

api-rx, like api-core needs to be on this list, but it one of those things we probably don't want use of directly unless people know what they are doing - they only take RPC calls and either make them promises or observables.

PR #292, opened 2018-10-14
refactor: Miscellaneous changes for consistency

ltfschoen:
* Split Promise Examples 02 and 03 into both ApiPromise and ApiRx versions

* Update Promise Examples 05 and 07 to use Keyring Testpairs

* Add Capitalised comments.

* Add Node Interface comments 

* Fix typos

* Fix incorrect variable name

* Use Yarn instead of Npm in guides

* Explain how `ApiRx.create()` is equivalent to `new ApiRx().isReady`

* Fix example since should be connected instead of duplicate `disconnect`. Add missing on `ready` to example

* Update Typedocs examples to match Promises examples

* Rename Api to ApiPromise or ApiRx depending on what's imported

* Rename `provider` to `wsProvider` for consistency

* Add missing semicolons

* Change WS and HTTP ports in tests to the ones we usually use

* Remove unnecessary blank lines


docs/examples/promise/03_listen_to_balance_change/index.js
```diff
@@ -1,35 +1,5 @@
-// Import the Api
-const { ApiPromise } = require('@polkadot/api');
+const exampleApiPromise = require('./exampleApiPromise').default;
+const exampleApiRx = require('./exampleApiRx').default;
```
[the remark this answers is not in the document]
ltfschoen: Ok, so the Rx examples should be in the examples/rx directory:
  * examples/promise/03_listen_to_balance_change/index.js
  * examples/rx/03_listen_to_balance_change/index.js

So at https://polkadot.js.org/api/, the menu would display:

EXAMPLES
  **ApiPromise**
    Simple connect
    Listen to blocks
    Listen to balance change

  **ApiRx**
    Simple connect
    Listen to blocks
    Listen to balance change

docs/examples/promise/02_listen_to_blocks/exampleApiRx.js
```diff
@@ -0,0 +1,15 @@
+// Import the API Rx
```
jacogr: This is not promise - rxjs should go into /example/rxjs. Revert the naming changes.

Each sample is a runnable, installable file. Look at package.json, you can go to samples, do `yarn` and then `yarn start`

docs/examples/promise/02_listen_to_blocks/index.js
```diff
@@ -9,14 +9,15 @@ async function main () {
...
   });
 
   // Id for the subscription, we can cleanup and unsubscribe via
   // `api.chain.newHead.unsubscribe(subscriptionId)`
-  console.log(`subsciptionId: ${subscriptionId}`);
+  console.log(`subscriptionId: ${subscriptionId}`);
 }
 
-main().catch(console.error);
+main().catch(console.error).finally(_ => process.exit());
```
jacogr: Like the README said, listens until Ctrl-C. (If we do this it will just grab the first header, not show them as they come in. So revert to original, it is correct.)

docs/examples/promise/03_listen_to_balance_change/index.js
```diff
@@ -32,4 +33,4 @@ async function main () {
   });
 }
 
-main().catch(console.error);
+main().catch(console.error).finally(_ => process.exit());
```
jacogr: Same comment as above - here it actually said, send something and see the balance change. Doing this will; exit immediately with only one result. (Before it has changed)

docs/examples/promise/README.md
```diff
@@ -24,3 +24,14 @@ Some of the examples use the following accounts:
 Those accounts are easy to add if you don't have/see them. The seed of Alice's account is `Alice␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣␣` and the seed of Bob is... well you guess...
 
 NOTE: Note the spaces padding Alice's key up to 32 chars.
+
+The address and public key of the test accounts may also be obtained as follows:
```
jacogr: I would leave this out - we can document it in keyring. (testing is really internal-use, i.e. it started as "just for testing", i.e. *.spec.js)

docs/examples/rx/02_listen_to_blocks/index.js
```diff
@@ -0,0 +1,17 @@
...
+  // Here we pass the (optional) provider
+  ApiRx.create(wsProvider).subscribe((api) => {
+    // Use the RPC Node Interface
+    api.rpc.subscribeNewHead().subscribe((header) => {
+      console.log(`best #${header.blockNumber.toNumber()}`);
+    });
+  });
+}
+
+main().catch(console.error).finally(_ => process.exit());
```
jacogr: No exit here, until Ctrl-C

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -0,0 +1,20 @@
...
+  // Here we pass the (optional) provider
+  ApiRx.create(wsProvider).subscribe((api) => {
+    // Use the Storage chain state (runtime) Node Interface.
+    api.query.balances.freeBalance(Alice).subscribe((balance) => {
+      console.log('new balance', balance);
+    });
+  });
+}
+
+main().catch(console.error).finally(_ => process.exit());
```
jacogr: No exit here.

jacogr:
@amaurymartiny Help ^^^ :) (Amaury is def. the resident guru)

amaurymartiny:
(untested)

```javascript
  const subscriptionApiRx = ApiRx.create()
    .pipe(
      switchMap((api) =>
        combineLatest([
          api.query.system.accountNonce(Alice),
          api.query.timestamp.blockPeriod(),
          api.query.session.validators()
            .pipe(
              switchMap((validators) => {
                // Retrieve the balances for all validators
                return combineLatest(
                  ...validators.map((authorityId) =>
                    api.query.balances.freeBalance(authorityId)
                      .pipe(
                        map(((balance) => ({
                          address: authorityId.toString(),
                          balance: balance.toString()
                        })))
                      )
                  )
                );
              })
            )
        ])
      )
    )
    .subscribe(([accountNonce, blockPeriod, validatorBalances]) => {
      console.log(`accountNonce(${Alice}) ${accountNonce} ` +
        `with blockPeriod ${blockPeriod.toNumber()} seconds`);
      console.log('validators', validatorBalances);
    });
```

Notes about rxjs:
- no nested subscribes, that can always be transformed into flatMap/switchMap
- no console.logs inside Observable definition, those should be pure functions. Only log inside subscribes
- for clarity, the `api.query.session.validators().pipe(...)` part could be extracted and put somewhere else, called e.g. `validatorsWithBalance`

docs/examples/promise/02_listen_to_blocks/index.js
```diff
@@ -9,14 +9,20 @@ async function main () {
...
   const subscriptionId = await api.rpc.chain.newHead((header) => {
     console.log(`best #${header.blockNumber}`);
   });
 
   // Id for the subscription, we can cleanup and unsubscribe via
   // `api.chain.newHead.unsubscribe(subscriptionId)`
-  console.log(`subsciptionId: ${subscriptionId}`);
+  console.log(`subscriptionId: ${subscriptionId}`);
+
+  setTimeout(() => {
```
jacogr: Ok, so we will never see more than 1 or 2 blocks? Not quite the intent of the example at all :(

docs/examples/promise/03_listen_to_balance_change/index.js
```diff
@@ -8,20 +8,22 @@ async function main () {
...
-  // Retrieve the initial balance. Since the call has no callback, it is simply a promise
-  // that resolves to the current on-chain value
+  // Retrieve the initial balance. Since the call has no callback, it is simply a
+  // promise that resolves to the current on-chain value
   let previous = await api.query.balances.freeBalance(Alice);
 
-  console.log(`${Alice} has a ${previous} balance`);
-  console.log(`You may leave this example running and start example 06 or transfer any value to ${Alice}`);
+  console.log(`${Alice} has a previous balance of ${previous || '???'}`);
+  console.log(`You may leave this example running and start example 06` +
```
jacogr: Just keep it single instead of mixing concat with template.

docs/examples/rx/01_simple_connect/index.js
```diff
@@ -0,0 +1,40 @@
...
+    .subscribe(([chain, name, version]) => {
+      console.log(`You are connected to chain #${chain.toString()} ` +
+                  `with node name #${name.toString()} ` +
+                  `and node version v#${version.toString()}`);
+    });
+
+  // Id of the subscription
+  console.log(`subscriptionApiRx: ${subscriptionApiRx}`);
+
+  setTimeout(() => {
```
jacogr: This could unsubscribe before you have the info, i.e. software was not running. Rather unsubscribe after you have the results.

docs/examples/rx/02_listen_to_blocks/index.js
```diff
@@ -0,0 +1,25 @@
...
+    // Use the RPC Node Interface
+    api.rpc.subscribeNewHead().subscribe((header) => {
+      console.log(`best #${header.blockNumber.toNumber()}`);
+    });
+  });
+
+  // Id of the subscription
+  console.log(`subscriptionApiRx: ${subscriptionApiRx}`);
+
+  setTimeout(() => {
```
jacogr: This will unsubscribe, blocks will stop coming it, however the app won't exit. The idea is to run until the user exits. The node will remove the subscription on exit, so no need to here.

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+async function main () {
+  let previous = null;
+
+  // Here we pass the (optional) provider
+  ApiRx.create(wsProvider).subscribe((api) => {
+    // Here we subscribe to any balance changes and update the on-screen value.
+    // Use the Storage chain state (runtime) Node Interface.
+    const subscriptionIdBalance = api.query.balances.freeBalance(Alice).subscribe((current) => {
+      console.log(`${Alice} has a previous balance of ${previous || '???'}`);
```
jacogr: This line will display each time the balance changes along with the next one.

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+      previous = current;
+
+      console.log(`Balance of ${Alice}: ${current}, ${change} change`);
+    });
+
+    // Id of the subscription
+    console.log(`subscriptionApiRx: ${subscriptionIdBalance}`);
+
+    setTimeout(() => {
```
jacogr: It will now only monitor for 5s, which is probably not even enough time to fire up the UI.

docs/examples/rx/05_read_storage/index.js
```diff
@@ -0,0 +1,55 @@
...
+const { ApiRx } = require('@polkadot/api');
+
+// Our address for Alice on the dev chain
+const Alice = '5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ';
+
+async function main () {
+  // Create the API and wait until ready using default provider.
+  // Subscribe to the basic chain state/storage queries
+  // Use the Storage chain state (runtime) Node Interface.
+  const subscriptionApiRx = ApiRx.create()
```
jacogr: This example is completely un-followable.

- Api.Rx.create().toPromise()
- validators as seperate function

The idea is not to be overly verbose, but at the same time you shouldn't need a PhD in optimisation to be able to follow the code. As examples things should be clear and understandable.

docs/examples/rx/05_read_storage/index.js
```diff
@@ -0,0 +1,55 @@
...
+    .subscribe(([accountNonce, blockPeriod, validatorBalances]) => {
+      console.log(`accountNonce(${Alice}) ${accountNonce} ` +
+        `with blockPeriod ${blockPeriod.toNumber()} seconds`);
+      console.log('validators', validatorBalances);
+    });
+
+  // Id of the subscription
+  console.log(`subscriptionApiRx: ${subscriptionApiRx}`);
+
+  setTimeout(() => {
```
jacogr: Once again, we stop before we may have had values.

docs/examples/promise/03_listen_to_balance_change/index.js
ltfschoen: Ok, but the reason I used concat is because the if there are more than ~100 characters per line (with browser window width as wide as possible) it clips the contents and enables horizontal scroll bar, so the reader of the embedded code snippet has to scroll to the right to read the rest of it.

jacogr: Fair reason.

docs/examples/promise/02_listen_to_blocks/index.js
jacogr: I believe we just keep it running until the user exits. No need to add delays, we are trying to show how things work and allow people to experiment. Plenty of other "unsub" examples.

ltfschoen:
@jacogr I believe I've addressed your review comments. I've got almost all the examples to work. Except for the following:

1) In [docs/examples/rx/02_listen_to_blocks/index.js](https://github.com/polkadot-js/api/blob/428e3a4f6a8c98c73441156d824b62774c1b5aa1/docs/examples/rx/02_listen_to_blocks/index.js) when I use the following code, if i unsubscribe, and confirm it was unsubscribed by calling .closed on the subscription which returns true, the problem I encounter is that it still keeps receiving blocks and console.log'ing them
```
const { ApiRx } = require('@polkadot/api');
const { WsProvider } = require('@polkadot/rpc-provider');

const wsProvider = new WsProvider('ws://127.0.0.1:9944');

async function main () {
  // Here we pass the (optional) provider
  const subscriptionApiRx = ApiRx.create(wsProvider).subscribe((api) => {
    // Use the RPC Node Interface
    api.rpc.chain.subscribeNewHead().subscribe((header) => {
      console.log(`Chain is at best block #${header.blockNumber.toNumber()}`);
    });
  });

  setTimeout(() => {
    // Cleanup and unsubscribe from the subscription
    subscriptionApiRx.unsubscribe();
    console.log(`Unsubscribing successful? ${subscriptionApiRx.closed}`);
  }, 10000);
}

main().catch(console.error);
```

So I ended up changing the code to gracefully exit and only then unsubscribe after the user presses CTRL+C to trigger the interrupt signal. But doing so adds a lot of extra code that deviates from the purpose of the example, and although it kills the program so unsubscribing doesn't matter anyway, I can still only assume that unsubscribe doesn't work properly (as discovered and mentioned earlier)

2) In [docs/examples/rx/03_listen_to_balance_change/index.js](https://github.com/polkadot-js/api/blob/f85fb147722cec0c9d9cc6e0d7edce47b16afe11/docs/examples/rx/03_listen_to_balance_change/index.js), when i run it shows my last balance in the console and suggests that I transfer DOTs to Alice to see it update. So I and have a Substrate Node running in the background with `./target/debug/substrate --dev --ws-port 9944`. If I then run polkadot-js/apps using the "next" branch, and try to transfer funds to Alice from any of the test accounts, I can't, because in the UI all the test accounts like Alice have a balance of 0

3) In [docs/examples/rx/07_transfer_dots/index.js](https://github.com/polkadot-js/api/blob/428e3a4f6a8c98c73441156d824b62774c1b5aa1/docs/examples/rx/07_transfer_dots/index.js) It only functions running `transfer` and `sign`, but I'm not sure why it passes a hash value of 34 to the subscribe block so it outputs `submitted transfer 12345 to Bob with hash 34` in the terminal. I've been unable to get `.send` to work, so I've commented it out. 
If I uncomment the `.send` block of code, after a few seconds running it outputs the following error in the terminal
```
(node:1001) UnhandledPromiseRejectionWarning: Error: [-32603]: Unknown error occured
    at checkError (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/rpc-provider/coder/json/decodeResponse.js:20:9)
    at decodeResponse (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/rpc-provider/coder/json/decodeResponse.js:28:3)
    at Object.decodeResponse (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/rpc-provider/coder/json/index.js:24:62)
    at WsProvider.onSocketMessageResult.response [as onSocketMessageResult] (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/rpc-provider/ws/index.js:100:35)
    at W3CWebSocket.WsProvider.onSocketMessage.message [as onmessage] (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/rpc-provider/ws/index.js:84:61)
    at W3CWebSocket._dispatchEvent [as dispatchEvent] (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/yaeti/lib/EventTarget.js:107:17)
    at W3CWebSocket.onMessage (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/websocket/lib/W3CWebSocket.js:234:14)
    at WebSocketConnection.<anonymous> (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/websocket/lib/W3CWebSocket.js:205:19)
    at WebSocketConnection.emit (events.js:182:13)
    at WebSocketConnection.processFrame (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/websocket/lib/WebSocketConnection.js:554:26)
(node:1001) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:1001) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

docs/examples/rx/07_transfer_dots/index.js
```diff
@@ -0,0 +1,59 @@
...
+      switchMap((api) =>
+        combineLatest([
+          // Retrieve the nonce for Alice, to be used to sign the transaction.
+          api.query.system.accountNonce(alice.address()),
+          api.rpc.chain.getBlockHash(0)
+            .pipe(
+              // Pipe nonce into transfer.
+              // Use the Extrinsics (runtime) Node Interface.
+              switchMap((aliceNonce, blockHash) => {
+                return combineLatest(
```
amaurymartiny: No need for combineLatest, here, just return `api.tx.balances.transfer()....`

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -0,0 +1,38 @@
...
+const wsProvider = new WsProvider('ws://127.0.0.1:9944');
+
+// Known account we want to use (available on dev chain, with funds)
+const Alice = '5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ';
+
+async function main () {
+  let previous = null;
+
+  // Here we pass the (optional) provider
+  ApiRx.create(wsProvider).subscribe((api) => {
```
amaurymartiny: Same thing, avoid nested subscribes.

```javascript
ApiRx
  .create(wsProvider)
  .pipe(
    switchMap((api)=> api.query.balances.freeBalance(Alice))
  ) 
```

If you want to add some console.logs, add either in subscribe, or on the `tap` operator.

ltfschoen:
@amaurymartiny I've pushed commits that i believe address your review comments

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -11,14 +13,20 @@ async function main () {
...
+      // Use the Storage chain state (runtime) Node Interface.
+      switchMap((api) => api.query.balances.freeBalance(Alice))
+    )
+    .subscribe((current) => {
+      console.log(`${Alice} has ${previous || '???'} previous balance`);
+      console.log(`You may leave this example running and start example 06 ` +
+                  `or transfer any value to Alice at ${Alice}`);
+
       // Calculate the delta
       const change = current.sub(previous);
```
amaurymartiny: There's a way to not have the external `previous` variable, using the rxjs `scan` operator. But that might create confusion for beginners, so let's leave it like this.

docs/examples/rx/01_simple_connect/index.js
```diff
@@ -0,0 +1,40 @@
+// Required imports
+const { combineLatest } = require('rxjs');
+const { switchMap } = require('rxjs/operators');
+const { ApiRx } = require('@polkadot/api');
+const { WsProvider } = require('@polkadot/rpc-provider');
+
+// Initialise the provider to connect to the local node
+const wsProvider = new WsProvider('ws://127.0.0.1:9944');
+
+async function main () {
```
amaurymartiny: Doesn't need to be async.

docs/examples/rx/02_listen_to_blocks/index.js
```diff
@@ -0,0 +1,40 @@
+// Import the API Rx
+const { ApiRx } = require('@polkadot/api');
+const { WsProvider } = require('@polkadot/rpc-provider');
+
+const wsProvider = new WsProvider('ws://127.0.0.1:9944');
+
+async function main () {
```
amaurymartiny: idem

docs/examples/rx/02_listen_to_blocks/index.js
```diff
@@ -0,0 +1,40 @@
...
+  // Here we pass the (optional) provider
+  const subscriptionApiRx = ApiRx.create(wsProvider).subscribe((api) => {
```
amaurymartiny: switchMap

docs/examples/rx/05_read_storage/index.js
```diff
@@ -0,0 +1,62 @@
...
+      ).subscribe(([accountNonce, blockPeriod, validatorBalances]) => {
+        console.log(`Account Nonce of ${accountNonce} for Alice with address ${Alice}`);
+        console.log(`Block Period: ${blockPeriod.toNumber()} seconds`);
+        console.log(`Validator Balances: ${JSON.stringify(validatorBalances, null, 2)}`);
+
+        const isReceived = !!accountNonce && !!blockPeriod && !!validatorBalances;
+
+        // Cleanup and unsubscribe from the subscription if all data received
+        if (isReceived) {
+          subscriptionApiRx.unsubscribe();
```
amaurymartiny: If you use `first()` operator, you don't need to manually unsubscribe.

There's a subtle difference though:
- if the node behaves correctly, then the two versions are identical
- however if the node errors, and one of the value is `undefined/null/false`, the subtle difference is that:
  - your version will not unsubscribe, and will wait for a next value. So potentially multiple console.logs.
  - the `first()` version will unsubscribe after the 1st value. So one console.log, will some values equal to `undefined/null/false`

docs/examples/rx/03_listen_to_balance_change/index.js
ltfschoen: I'll add that as a "challenge"

docs/examples/rx/01_simple_connect/index.js
ltfschoen: if i get rid of `async` it causes error:
```
/Users/me/code/src/polkadot-js/api/docs/examples/rx/01_simple_connect/index.js:40
main().catch(console.error);
      ^

TypeError: Cannot read property 'catch' of undefined
```

docs/examples/rx/02_listen_to_blocks/index.js
ltfschoen: removing `async` here causes same error as above

docs/examples/rx/01_simple_connect/index.js
amaurymartiny: Only `main()` then at the end? imo if there are errors, they should be handled rxjs-ly. cc @jacogr

docs/examples/rx/05_read_storage/index.js
ltfschoen: i've added this as a "challenge" in the Readme

ltfschoen:
@jacogr I've removed the unsubscribe from example promises 02

docs/examples/rx/05_read_storage/index.js
ltfschoen: Maybe another option in this example is to replace `combineLatest` with `forkJoin`, since it only takes action when a response has been received for all requests https://www.learnrxjs.io/operators/combination/forkjoin.html

docs/examples/rx/01_simple_connect/index.js
ltfschoen: if I change it from `async function main () {` to `function main () {`, and change `await ApiRx.create()` to `ApiRx.create()`, and change `main().catch(console.error)` to `main()` they work

ltfschoen:
@jacogr @amaurymartiny I've pushed a commit that:
* updates dependencies used by examples
* removes the unsubscribe code from rx/02_listen_to_blocks
* when i tried running promises/07_transfer_dots and rx/07_transfer_dots with latest dependencies,, it now causes the following error when `api.tx.balances.transfer` is called (previously this worked and an error only occurred when `.sign` was called, as previously mentioned in this thread):

```
/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/rxjs/internal/util/hostReportError.js:4
    setTimeout(function () { throw err; });
                             ^

ExtError: Unable to determine type from '::Type'
    at assert (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/util/assert.js:44:9)
    at getTypeClass (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/types/codec/createType.js:148:20)
    at Method.filterOrigin.reduce (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/types/Method.js:136:49)
    at Array.reduce (<anonymous>)
    at Function.getArgsDef (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/types/Method.js:132:38)
    at Function.decodeMethod (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/types/Method.js:92:30)
    at new Method (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/types/Method.js:45:28)
    at extrinsicFn (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/extrinsics/utils/createUnchecked.js:32:15)
    at Object.decorated [as transfer] (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/node_modules/@polkadot/api/rx/index.js:207:76)
    at SwitchMapSubscriber.api.rpc.chain.getBlockHash.pipe.switchMap [as project] (/Users/me/code/src/polkadot-js/api/docs/examples/rx/07_transfer_dots/index.js:36:20)
```

jacogr:
1. The e2e examples are skipped on CI since it doesn't have a node running - un-skipping will make it fail on CI each and every time
2. See this example - added an equivalent example for RX (as was the case in Promise) and it works - https://github.com/polkadot-js/api/pull/356 (both cases - obviously when un-skipped :))

jacogr:

ltfschoen:
@jacogr 
* I've merged latest from master
* added tests to docs/examples/promises examples including 01, 02, 03, and 05. they work when running a substrate node.
* updated README.md with the `yarn run test` command. updated package.json so it points to the jest config file so that it runs the tests in the docs/examples/ directory too
* removed usage of wsProvider in the rx/03 example since it was preventing it from working
* i've removed the following examples: promises/05 & 07, and rx/05 & 07. i've included the error messages in the commit. the 05 examples weren't working because `api.query.session` is no longer available, and the 07 examples weren't working because of the `ExtError: Unable to determine type from '::Type'` error. these examples have been excluded from this PR, but can be included in a future PR once things are stable and we're not encountering those errors anymore (which i suspect may be after Substrate 1.0 beta is released..?

ltfschoen:
@jacogr I've made updates and it all appears to be working correctly now

ltfschoen:
Sorry, what's stale? Should I not have moved 04_generate_account and made it an example again? Or something else?

jacogr:
GitHub itself state `jacogr dismissed their stale review 3 hours ago` - so the review itself is stale (as it was done more than a month ago now). Any review dismissal requires a comment to be mad as to the reason, i.e. "comments addressed", "re-reviewed" or anything else. 

So making the comment "Stale" on a stale review seems apt. Stale review with "stale" comment.

README.md
```diff
@@ -48,3 +48,21 @@ To get started -
 5. Build the API Docs, via `yarn run build`
 6. Ready! Now you can launch the API Docs, via `yarn gitbook serve`
 7. Access the API Docs via [http://localhost:4000](http://localhost:4000)
+
+## tests
+
+Run a Substrate chain node:
+```bash
+curl https://getsubstrate.io -sSf | bash;
```
jacogr: Until getsubstarte actually works as expected, we need to point to master. (This approach is actually what made you have a horror of a time here, I would not like the repeat that with others)

README.md
```diff
@@ -48,3 +48,21 @@ To get started -
...
+substrate --dev purge-chain --dev;
+substrate --dev --telemetry-url ws://telemetry.polkadot.io:1024
```
jacogr: Additionally, dev chains, no telemetry, just the defaults, nothing else.

docs/README.md
```diff
@@ -15,6 +15,7 @@ You can find the Polkadot repositories at:
 - https://github.com/polkadot-js
 - https://github.com/paritytech/polkadot
 - https://github.com/paritytech/substrate
+- https://github.com/paritytech/substrate-node-template
```
jacogr: Rather remove, this belongs on the wiki or elsewhere.

docs/examples/promise/01_simple_connect/index.spec.js
```diff
@@ -0,0 +1,32 @@
+// Copyright 2017-2018 @polkadot/api authors & contributors
```
jacogr: Remove the tests please. The apps itself are the tests. (It just adds clutter and the whole idea behind the samples is to be lean-and-mean)

docs/examples/promise/02_listen_to_blocks/index.js
```diff
@@ -9,14 +9,12 @@ async function main () {
...
   // are found, the call itself returns a promise with a subscription that can be
-  // used to unsubscribe from the newHead subscription
+  // used to unsubscribe from the newHead subscription. We show how
+  // to unsubscribe from a subscription in subsequent examples.
+  // Use the RPC Node Interface
   const subscriptionId = await api.rpc.chain.subscribeNewHead((header) => {
-    console.log(`best #${header.blockNumber}`);
+    console.log(`Chain is at best block #${header.blockNumber}`);
   });
-
```
jacogr: Bring back. We are showing what is available, even if not used elsewhere.

docs/examples/promise/02_listen_to_blocks/index.spec.js
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017-2018 @polkadot/api authors & contributors
```
jacogr: Remove.

docs/examples/promise/03_listen_to_balance_change/index.js
```diff
@@ -8,27 +10,31 @@ async function main () {
   // Create an await for the API
   const api = await ApiPromise.create();
 
-  // Retrieve the initial balance. Since the call has no callback, it is simply a promise
-  // that resolves to the current on-chain value
-  let previous = await api.query.balances.freeBalance(Alice);
+  // Retrieve the current on-chain balance. The call has a callback since WebSockets 
+  // subscriptions require a handler.
+  let previous = await api.query.balances.freeBalance(Alice, (previous) => {
+    return new BN(previous);
```
jacogr: Nope. Balance already extends `BN`. So this is not very useful and just confuses. We don't need to import `BN` and/or used it at all.

docs/examples/promise/03_listen_to_balance_change/index.js
```diff
@@ -8,27 +10,31 @@ async function main () {
...
+  console.log(`You may leave this example running and start example 07 to transfer DOTs ` +
+              `to Alice at ${Alice}`);
 
-  // Here we subscribe to any balance changes and update the on-screen value
+  // Here we subscribe to any balance changes and update the on-screen value.
+  // Use the Storage chain state (runtime) Node Interface.
   api.query.balances.freeBalance(Alice, (current) => {
     // Calculate the delta
-    const change = current.sub(previous);
+    const change = current.sub(new BN(previous));
```
jacogr: Revert. The original was correct.

docs/examples/promise/03_listen_to_balance_change/index.spec.js
```diff
@@ -0,0 +1,30 @@
+// Copyright 2017-2018 @polkadot/api authors & contributors
```
jacogr: Remove.

docs/examples/promise/04_generate_account/README.md
```diff
@@ -0,0 +1,5 @@
+# Generate accounts
```
jacogr: This example does not belong here at all, it is not API functionality - we don't even import the API in the example at all. Remove.

docs/examples/promise/04_generate_account/index.js
```diff
@@ -0,0 +1,32 @@
+const crypto = require('crypto');
```
jacogr: Remove.

docs/examples/promise/04_generate_account/package.json
```diff
@@ -0,0 +1,20 @@
+{
```
jacogr: Remove.

docs/examples/promise/05_read_storage/index.html
```diff
@@ -361,6 +361,90 @@
             
         </li>
     
+        <li class="chapter " data-level="3.2" data-path="../../rx/">
```
jacogr: Not sure why this is in the PR, it is generated. Should be removed and CI should just do its job.

docs/examples/promise/05_read_storage/index.html
```diff
@@ -407,7 +491,7 @@ <h1>
                                 <section class="normal markdown-section">
                                 
                                 <h1 id="read-storage">Read storage</h1>
-<p>Many important variables are available through the storage API. This example shows how to call a few of those APIs.</p>
+<p>Many important variables are available through the storage API. This example shows how to call a few of those APIs using ApiPromise.</p>
```
jacogr: Revert, CI generated.

docs/examples/promise/05_read_storage/index.spec.js
```diff
@@ -0,0 +1,57 @@
+// Copyright 2017-2018 @polkadot/api authors & contributors
```
jacogr: Remove.

docs/examples/promise/07_transfer_dots/index.html
```diff
@@ -361,6 +361,90 @@
             
         </li>
     
+        <li class="chapter " data-level="3.2" data-path="../../rx/">
```
jacogr: CI-generated, no need to check it in.

docs/examples/promise/07_transfer_dots/index.html
```diff
@@ -407,11 +491,11 @@ <h1>
                                 <section class="normal markdown-section">
                                 
                                 <h1 id="tranfer-dots">Tranfer DOTs</h1>
-<p>This sample shows how to create a transaction to make a transfer from one an account to another.</p>
+<p>This sample shows how to create a transaction to make a transfer from one an account to another using ApiPromise.</p>
```
jacogr: CI-generated, remove.

docs/examples/promise/07_transfer_dots/index.js
```diff
@@ -1,7 +1,7 @@
 // Import the API, Keyring and some utility functions
 const { ApiPromise } = require('@polkadot/api');
 const { Keyring } = require('@polkadot/keyring');
-const { stringToU8a } = require('@polkadot/util');
+const { stringShorten, stringToU8a } = require('@polkadot/util');
```
jacogr: Not sure why we are doing this. The text changes helps, the extra "show off what utils we have" doesn't really. So keep to what it was.

docs/examples/promise/08_system_events/index.spec.js
```diff
@@ -0,0 +1,22 @@
+// Copyright 2017-2018 @polkadot/api authors & contributors
```
jacogr: Remove.

docs/examples/promise/README.md
```diff
@@ -1,19 +1,25 @@
...
-Here you will find a list of examples that takes you through the basics of connecting to a local node, retrieving data from the Node and chain and execute transactions on the chain. It uses the [[ApiPromise]] interface.
+Here you will find a list of examples that takes you through the basics of connecting to a local node, retrieving data from the Node and chain and execute transactions on the chain. It uses the ApiPromise interface.
 
 ## Prerequisites
 
-For the following examples, you need a local node. It is usually convenient testing with:
-
-```
-substrate --dev
+For the following examples, you need to install and run a local Substrate chain node.
```
jacogr: As above, no getsubstrate, the Substrate wiki should take care of how to run. (We really don't want to re-wrok the world)

docs/examples/rx/02_listen_to_blocks/index.html
```diff
@@ -0,0 +1,597 @@
+
```
jacogr: Remove. CI-generated.

docs/examples/rx/03_listen_to_balance_change/index.html
```diff
@@ -0,0 +1,618 @@
+
```
jacogr: Remove, Ci-generated.

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -0,0 +1,48 @@
+const BN = require('bn.js');
```
jacogr: Unneeded.

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -0,0 +1,48 @@
...
+  // We avoid having nested `.subscribe` blocks. Use `.subscribe` or
+  // `.tap` blocks for debugging with console.log
+  ApiRx
+    .create()
+    .pipe(
+      // Here we subscribe to any balance changes and update the on-screen value.
+      // Use the Storage chain state (runtime) Node Interface.
+      switchMap((api) => {
+        return api.query.balances.freeBalance(Alice, (previous) => {
+          return new BN(previous);
```
jacogr: Unneeded, it is already a `BN`.

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -0,0 +1,48 @@
...
+        });
+      })
+    )
+    .subscribe((current) => {
+      console.log(`Alice's account ${Alice} had a previous balance of: ${previous || '???'}`);
+      console.log(`You may leave this example running and start example 07 to transfer DOTs ` +
+                  `to Alice at ${Alice}`);
+
+      // Calculate the delta
+      const change = current.sub(new BN(previous));
```
jacogr: Unneeded, already a BN.

docs/examples/rx/05_read_storage/index.html
```diff
@@ -0,0 +1,636 @@
+
```
jacogr: Remove, CI-generated.

docs/examples/promise/05_read_storage/index.html
jacogr: Remove, CI-generated.

ltfschoen:
@jacogr I've addressed your latest review comments. I encountered the following issues:

1) **docs/examples/rx/07_transfer_dots:**
uses `ApiRx` so the response from calling `api.tx.balances.transfer(...).sign(...).send()` is the extrinsic status of `Finalised` since `.send()`'s signature is `(method) SubmittableExtrinsic.send(): Observable<ExtrinsicStatus>`, however I want it to return the extrinsic's hash (not the status of the extrinsic) so the example works the same as the **docs/examples/promises/07_transfer_dots** example where we use `ApiPromise` so `api.tx.balances.transfer(...).sign(...).send()` instead returns the hash of the extrinsic since `.send()`'s signature in `ApiPromise` is `(method) SubmittableExtrinsic.send(statusCb?: (status: ExtrinsicStatus) => any): Promise<Hash>`, which is different

2) **docs/examples/rx/03_listen_to_balance_change:**
it works, but not exactly the same as **docs/examples/rx/03_listen_to_balance_change:**, as it shows all the console.log's for each balance change, instead of just the console.log that show's how much it changed by.

I tried changing it to the following:
```
--- a/docs/examples/rx/03_listen_to_balance_change/index.js
+++ b/docs/examples/rx/03_listen_to_balance_change/index.js
@@ -6,28 +6,26 @@ const { ApiRx } = require('@polkadot/api');
 // Known account we want to use (available on dev chain, with funds)
 const Alice = '5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ';
 
-function main () {
-  let previous = undefined;
+async function main () {
+  // Instantiate the API via Promise
+  const api = await ApiRx.create().toPromise();
+
+  let previous = await api.query.balances.freeBalance(Alice);
+
+  console.log(`Alice's account ${Alice} had a previous balance of: ${previous || '???'}`);
+  console.log(`You may leave this example running and start example 07 to transfer DOTs ` +
+              `to Alice at ${Alice}`);
 
   // Here we pass the (optional) provider
   // We avoid having nested `.subscribe` blocks. Use `.subscribe` or
   // `.tap` blocks for debugging with console.log
-  ApiRx
-    .create()
+  api
     .pipe(
       // Here we subscribe to any balance changes and update the on-screen value.
       // Use the Storage chain state (runtime) Node Interface.
       switchMap((api) => api.query.balances.freeBalance(Alice))
     )
     .subscribe((current) => {
-      previous = previous || current;
-
-      // FIXME - refactor like the ApiPromise where only the balance change console.log's
-      // are shown when Alice's balance changes, instead of also showing the below console.log's too.
-      console.log(`Alice's account ${Alice} had a previous balance of: ${previous || '???'}`);
-      console.log(`You may leave this example running and start example 07 to transfer DOTs ` +
-                  `to Alice at ${Alice}`);
-
       // Calculate the delta
       const change = current.sub(previous);
       ...
```

But i get the following error and it doesn't return the value of the previous balance:
```
Alice's account 5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ had a previous balance of: [object Object]
You may leave this example running and start example 07 to transfer DOTs to Alice at 5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaDtZ
(node:41977) UnhandledPromiseRejectionWarning: TypeError: api.pipe is not a function
    at main (/Users/scon/code/src/polkadot-js/api/docs/examples/rx/03_listen_to_balance_change/index.js:23:6)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

docs/examples/rx/07_transfer_dots/index.js
```diff
@@ -0,0 +1,49 @@
+const { combineLatest } = require('rxjs');
+const {  first, switchMap } = require('rxjs/operators');
```
jacogr: Errant space before `first`

docs/examples/rx/07_transfer_dots/index.js
```diff
@@ -0,0 +1,49 @@
...
+          // Create an extrinsic, transferring 12345 units to Bob.
+          .transfer(BOB_ADDR, 12345)
+          // Sign the transaction using our account keypair, nonce,
+          // and optionally the block hash
+          .sign(alice, aliceNonce)
+          // Send the transaction (optional status callback)
+          .send();
+      })
+    )
+    // FIXME - return the transfer hash instead of the status like when using ApiPromise
```
jacogr: No fixme's in the examples :) The behaviour here is fine.

package.json
```diff
@@ -20,7 +20,7 @@
     "check": "tslint --project . && tsc --noEmit --pretty",
     "clean": "polkadot-dev-clean-build",
     "postinstall": "polkadot-dev-yarn-only",
-    "test": "jest --coverage"
+    "test": "jest --coverage --config='jest.config.js'"
```
jacogr: Why the extra config here?

packages/api/src/promise/index.ts
```diff
@@ -30,76 +30,7 @@ import SubmittableExtrinsic from './SubmittableExtrinsic';
...
- *     // send the transaction (optional status callback)
- *     .send((status) => {
- *       console.log(`current status ${status.type}`);
- *     })
- *     // retrieve the submitted extrinsic hash
- *     .then((hash) => {
- *       console.log(`submitted with hash ${hash}`);
- *     });
- * });
+ * @example See [Polkadot-JS Api Promise Examples](https://polkadot.js.org/api/examples/promise/)
```
jacogr: Keep the usage in here. It is great for an overview instead of clicking through.

packages/api/src/promise/index.ts
```diff
@@ -114,9 +45,9 @@ export default class ApiPromise extends ApiBase<Rpc, QueryableStorage, Submittab
    * <BR>
    *
    * ```javascript
-   * import Api from '@polkadot/api/promise';
+   * import ApiPromise from '@polkadot/api/promise';
```
jacogr: Keep as it was.

packages/api/src/rx/index.ts
```diff
@@ -28,93 +28,13 @@ const l = logger('api-rx');
...
- *          .sign(keyring.alice, nonce)
- *          // send the transaction
- *          .send()
- *      )
- *   )
- *   // subscribe to overall result
- *   .subscribe((hash) => {
- *     console.log(`submitted with hash ${hash}`);
- *   });
+ * @example See [Polkadot-JS Api Rx Examples](https://polkadot.js.org/api/examples/rx/)
```
jacogr: As above, keep simple usage in here.

packages/api/src/rx/index.ts
```diff
@@ -129,11 +49,12 @@ export default class ApiRx extends ApiBase<RpcRx, QueryableStorage, SubmittableE
    * <BR>
    *
    * ```javascript
-   * import Api from '@polkadot/api/rx';
+   * import ApiRx from '@polkadot/api/rx';
```
jacogr: Keep as it was.

packages/api/test/e2e/promise-queries.spec.js
```diff
@@ -4,15 +4,15 @@
 
 import testingPairs from '@polkadot/keyring/testingPairs';
 
-import Api from '../../src/promise';
+import ApiPromise from '../../src/promise';
```
jacogr: Revert.

packages/api/test/e2e/rx-queries.spec.js
```diff
@@ -4,15 +4,15 @@
 
 import testingPairs from '@polkadot/keyring/testingPairs';
 
-import Api from '../../src/rx';
+import ApiRx from '../../src/rx';
```
jacogr: Revert.

packages/rpc-provider/src/http/index.ts
```diff
@@ -26,11 +26,11 @@ const l = logger('api-http');
  * <BR>
  *
  * ```javascript
- * import Api from '@polkadot/api/promise';
+ * import ApiPromise from '@polkadot/api/promise';
```
jacogr: Revert.

packages/rpc-provider/src/ws/index.ts
```diff
@@ -48,11 +48,11 @@ const l = logger('api-ws');
  * <BR>
  *
  * ```javascript
- * import Api from '@polkadot/api/promise';
+ * import ApiPromise from '@polkadot/api/promise';
```
jacogr: Revert.

packages/rpc-provider/test/mockHttp.ts
```diff
@@ -4,7 +4,7 @@
 
 import nock from 'nock';
 
-const TEST_HTTP_URL = 'http://localhost:9944';
+const TEST_HTTP_URL = 'http://localhost:9933';
```
jacogr: Should not overlap with the actual node endpoints. (So changing it is correct, changing it ot an overlapping port is not)

packages/rpc-provider/test/mockWs.ts
```diff
@@ -4,7 +4,7 @@
 
 import { Server } from 'mock-socket';
 
-const TEST_WS_URL = 'ws://localhost:9955';
+const TEST_WS_URL = 'ws://localhost:9944';
```
jacogr: Revert, no overlap.

packages/rpc-rx/src/index.ts
```diff
@@ -30,8 +30,8 @@ type CachedMap = {
  * import RpcRx from '@polkadot/rpc-rx';
  * import WsProvider from '@polkadot/rpc-provider/ws';
  *
- * const provider = new WsProvider('http://127.0.0.1:9944');
- * const api = new RpcRx(provider);
+ * const wsProvider = new WsProvider('http://127.0.0.1:9944');
```
jacogr: As it was.

package.json
ltfschoen: i'll remove it.  i had added it so it would also run the tests that i added in the docs/examples subdirectories, but those tests have all been removed

ltfschoen: i'll remove it.  i had added it so it would also run the tests that i added in the docs/examples subdirectories, but those tests have all been removed

ltfschoen:
@jacogr i've pushed commits that address the review comments

docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -0,0 +1,46 @@
...
+    .create()
+    .pipe(
+      // Here we subscribe to any balance changes and update the on-screen value.
+      // Use the Storage chain state (runtime) Node Interface.
+      switchMap((api) => api.query.balances.freeBalance(Alice))
+    )
+    .subscribe((current) => {
+      previous = previous || current;
+
+      // FIXME - refactor like the ApiPromise where only the balance change console.log's
```
jacogr: Not convinced we actually want FIXMEs in the examples at all. It should just be simple and working.

packages/api/README.md
```diff
@@ -31,18 +32,19 @@ Due to this dynamic nature, this API departs from traditional APIs which only ha
...
 ```
-npm install --save @polkadot/api
+yarn add @polkadot/api
 ```
 
 Subscribing to blocks via Promise-based API -
 
 ```javascript
-import { ApiPromise } from '@polkadot/api';
+import Api from '@polkadot/api/promise';
```
jacogr: Any reason this is not the original import? (This has been done in a couple of places now). This example use is actually great since it shows the 2 side-by-side, so being a bit more explicit as it was is a bit cleaner and clearer.

packages/api/src/Base.ts
```diff
@@ -178,13 +178,17 @@ export default abstract class ApiBase<R, S, E> implements ApiBaseInterface<R, S,
    * <BR>
    *
    * ```javascript
-   * api.on('disconnected', () => {
+   * api.on('connected', () => {
```
jacogr: +1

packages/api/src/promise/index.ts
```diff
@@ -36,70 +36,98 @@ import SubmittableExtrinsic from './SubmittableExtrinsic';
  * <BR>
  *
  * ```javascript
- * import ApiPromise from '@polkadot/api/promise';
+ * import Api from '@polkadot/api/promise';
```
jacogr: Not 100% sure why it has been changed?

packages/api/src/promise/index.ts
```diff
@@ -36,70 +36,98 @@ import SubmittableExtrinsic from './SubmittableExtrinsic';
...
+ * main().catch(console.error);
  * ```
  * <BR>
  *
  * Submitting a transaction -
  * <BR>
  *
  * ```javascript
- * import ApiPromise from '@polkadot/api/promise';
+ * import Api from '@polkadot/api/promise';
```
jacogr: I would just remove this and revert to the original here. IT is a bit overwhelming as an in-code comment now. (Examples are great for making these valid). So rather -

- revert to original
- import keyring (can just do the test one here, with an appropriate comment)
- leave rest as-is

packages/api/src/rx/index.ts
```diff
@@ -77,12 +84,15 @@ const l = logger('api-rx');
  *       ])
  *   )
  *   .subscribe(([blockPeriod, timestamp]) => {
+ *     // Convert the timestamps from type `Moment` into seconds
+ *     const blockPeriodSeconds = moment(blockPeriod).unix();
+ *     const timestampSeconds = moment(timestamp).unix();
  *     const elapsed = last
- *       ? `, ${timestamp.toNumber() - last}s since last`
+ *       ? `, ${timestampSeconds - last}s since last`
```
jacogr: +1 Ahhh, yes, this was pre-change to extending date. (Maybe use `.getTime()` above, it is more along the line what 90% out there use.

packages/api/src/rx/index.ts
```diff
@@ -91,30 +101,49 @@ const l = logger('api-rx');
...
- *          .sign(keyring.alice, nonce)
- *          // send the transaction
- *          .send()
- *      )
- *   )
- *   // subscribe to overall result
- *   .subscribe((hash) => {
- *     console.log(`submitted with hash ${hash}`);
- *   });
+ * const ALICE_SEED = 'Alice'.padEnd(32, ' ');
```
jacogr: As per the previous in-code example, too unwieldly now :(

packages/api/README.md
ltfschoen: I just changed the naming for consistency based on this review comment https://github.com/polkadot-js/api/pull/292#discussion_r238654837.

I agree that it would be better to show more explicitly here though since users can compare them side-by side.

I've pushed a commit reverting back

jacogr:
Closing, replaced by #600 (which tracks current interfaces closer)

PR #296, opened 2018-10-15
Expand SignedBlock tests

jacogr:


packages/types/src/SignedBlock.spec.js
```diff
@@ -2,11 +2,20 @@
...
 
+import extrinsics from '@polkadot/extrinsics/static';
+
+import Method from './Method';
 import rpc from './SignedBlock.rpc.json';
 import SignedBlock from './SignedBlock';
 
 describe('SignedBlock', () => {
-  const block = new SignedBlock().fromJSON(rpc.result);
+  let block;
```
amaurymartiny: This really makes a difference?

jacogr: It is a bit more inefficient. Just wanted to keep it together in beforeEach

PR #314, opened 2018-10-18
Cleanup util-crypto & keyring imports

jacogr:
- Import via `const { whatEver, somethingElse } from '@polkadot/util-crypto'`
- Pull in new util with stringToU8a
- Cleanup masses of imports in types, i.e. import via `const { ... } from '@polkadot/util'`


packages/rpc-provider/src/mock/mocks.ts
```diff
@@ -14,7 +14,7 @@ import storage from '@polkadot/storage/static';
 import Header from '@polkadot/types/Header';
 import bnToU8a from '@polkadot/util/bn/toU8a';
 import u8aToHex from '@polkadot/util/u8a/toHex';
-import randomAsU8a from '@polkadot/util-crypto/random/asU8a';
+import { randomAsU8a } from '@polkadot/util-crypto';
```
jacogr: Looking at this (not for this PR, but to be done in util-crypto, we probably want alias for `random` -> `randomAsU8a` and `xxhash` -> `xxhashAsU8a` and `blake2` -> `blake2AsU8a` (As the defaults, the `*AsU8a` can still stay, but it would make it cleaner)

PR #315, opened 2018-10-18
Cleanup imports from @polkadot/util

jacogr:


packages/type-jsonrpc/src/create/method.ts
```diff
@@ -4,7 +4,7 @@
 
 import { RpcMethodOpt, RpcMethod } from '../types';
 
-import isUndefined from '@polkadot/util/is/undefined';
+import { isUndefined } from '@polkadot/util';
```
amaurymartiny: Relative imports before absolute imports is uncommon.

packages/type-extrinsics/src/fromMetadata.ts
```diff
@@ -2,10 +2,11 @@
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
 
-import camelCase from '@polkadot/util/string/camelCase';
-import Metadata, { RuntimeModuleMetadata } from '@polkadot/types/Metadata';
 import { Extrinsics, ModuleExtrinsics } from './types';
 
+import { stringCamelCase } from '@polkadot/util';
+import Metadata, { RuntimeModuleMetadata } from '@polkadot/types/Metadata';
```
amaurymartiny: alphabetical order

packages/type-jsonrpc/src/create/method.ts
jacogr: It is a types split above - and yes, with those with the new types things have become _very_ muddy

PR #321, opened 2018-10-18
class Struct extends Map

amaurymartiny:
Related #161.

Instead of doing all types at once, let's do one by one, so that it's an easier transition. During this transition, I still added `this.raw = this;` in the Struct class.

This is one of the bigger ones. Breaking changes:
- `Struct.keys()` returns an iterator, not an array
- `Struct.values()` returns an iterator, not an array

TODO:
- [x] I'll just update @polkadot/dev for the jest testMatch, so that I remove that from this PR


packages/types/src/Method.ts
```diff
@@ -147,19 +147,20 @@ export default class Method extends Struct {
   }
 
   get args (): Array<Base> {
-    return (this.getAtIndex(1) as Struct<Base>).values();
+    // FIXME This should return a Struct instead of an Array
+    return [...(this.get('args') as Struct).values()];
```
amaurymartiny: @jacogr I think in the UI you use `method.args()` as an array. So Map or Array?

jacogr: Prefer array. Plus it is actually closer to `.values()`

jacogr:
@amaurymartiny dev PR merged.

packages/api-observable/src/classes.ts
```diff
@@ -16,15 +16,15 @@ export class RxProposal extends Struct.with({ id: PropIndex, proposal: Proposal,
   }
 
   get address (): AccountId {
-    return this.raw.address as AccountId;
+    return this.get('address') as AccountId;
```
jacogr: +1

I def prefer this - less magic.

packages/types/src/codec/Struct.ts
```diff
@@ -54,54 +77,40 @@ export default class Struct<
...
         } else if (isObject(value)) {
-          // @ts-ignore FIXME Ok, something weird is going on here or I just don't get it...
-          // it works, so ignore the checker, although it drives me batty. (It started when
-          // the [key in keyof T] was added, the idea is to provide better checks, which
-          // does backfire here, but works externally.)
-          raw[key] = new Types[key](
-            // @ts-ignore FIXME
-            value[jsonKey]
-          );
+          raw[key] = value[jsonKey as string] instanceof Types[key]
```
jacogr: Nice! I really tried to fix that :)

PR #324, opened 2018-10-19
fix: Fixes #320. Removes build folders/files from Git tracking. Adds Gitignore matching polkadot-js/common

### LINKED ISSUE #320: Remove Git tracked files that are built by Gitbooks and update Gitignore so works in both Polkadot-JS Api and Common
In the development environment, we do not want changes to be tracked by Git for files that are generated by running `yarn run build; yarn gitbook serve`. We only want Git to track the files that are user-generated

Update Gitignore so works in both Polkadot-JS Api and Common

ltfschoen:
* Removes from Git tracking all files that are generated when you run `yarn run build; yarn gitbook serve`. These files were removed from Git tracking by running `git rm --cache <filename>` on each of them

* Updates .gitignore file so it's the same as in polkadot-js/common (see this PR into polkadot-js/common https://github.com/polkadot-js/common/pull/231), and only permits relevant files


ltfschoen:
I'm convinced this isn't going to work. It doesn't render the Gitbook correctly.

PR #327, opened 2018-10-19
 Simple subscription results combinator

jacogr:


amaurymartiny:
Works well, but I'm not sure it's my favorite api. Esp the `.next()` reminds me of iterators.

Most intuitive:

```javascript
api.combine([
  cb => api.getSomething1(..., cb),
  cb => api.getSomething2(cb),
], ([value1, value2]) => {...})
```

jacogr:
Happy to change naming to something that does not confuse. The combine([]) approach came to mind first, however I'm not sure how to work around the optional callback in those cases, i.e. when callback is specified it is a subscription, if not it is a promise.

PR #328, opened 2018-10-19
Decode logs, without events

amaurymartiny:


packages/types/src/Digest.ts
```diff
@@ -0,0 +1,42 @@
...
+// FIXME Once we have Events, have a look on how to decode them
+export class Other extends Null {
+}
+
+export class DigestItem extends EnumType<AuthoritiesChange | ChangesTrieRoot | Other> {
+  constructor (value: any) {
+    super({
+      0: AuthoritiesChange,
+      1: ChangesTrieRoot,
+      2: Other
```
jacogr: I'm not 100% on the order here, see https://github.com/paritytech/substrate/blob/master/core/sr-primitives/src/generic/digest.rs#L87

amaurymartiny: AH, correct

PR #332, opened 2018-10-20
refactor: Fixes #329. Convert tests to TypeScript

### LINKED ISSUE #329: Convert all tests to TypeScript
Right now they are in `.spec.js`, for historical reasons.

ltfschoen:
* Git renames .test.js to .test.ts. Used `find ./packages -type f -name "*.spec.js" | grep -v build` to list all the files in the packages/ directory that had the .test.js extension, then used a spreadsheet with `git mv` column "A", a column "B" with the list from the above script, and a column "C" with the same as column "B" but renamed with .test.ts at the end of each row, and a column "D" where each row just had `;`. Then copy/paste to text file in Sublime Text 2 and replaced the spaces with single spaces, and ` ;` with just ``. Later I found that not all the .spec.js files in nested directories were detected by the above command, so they were renamed manually.
* Changes from `require` to ES6 `import`
* Fix TSLint errors resulting from extra .ts files (checked when running `yarn run build; yarn run check`
* **IMPORTANT NOTES**:
  * Many of the types have been made `any` (bad practice) where I am unsure what specific type to use or where interfaces are involved. Help is required. 
  * "WIP" commit https://github.com/polkadot-js/api/pull/332/commits/1b0e61a16afa755c300f482f34f017facfb57340 requires help. I am having difficulty closeing-out remaining TSLint errors.


packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -9,7 +9,7 @@ import Api from '../../src/promise';
 const keyring = testingPairs();
 
 describe.skip('e2e queries', () => {
-  let api;
+  let api: any;
```
jacogr: Unless we are constructing a specific mocked API (e.g. see https://github.com/polkadot-js/api/blob/master/packages/api/src/promise/SubmittableExtrinsic.spec.ts#L17), this should not be any. I believe it sort-of defats the purpose just renaming with a blanket instead of using TS.

packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -35,7 +35,7 @@ describe.skip('e2e queries', () => {
   });
 
   it('subscribes to queries', (done) => {
-    api.rpc.chain.subscribeNewHead((error, header) => {
+    api.rpc.chain.subscribeNewHead((error: any, header: any) => {
```
jacogr: This should be adapted, interface has changed. (Not part of this PR, nor changed here, but as it stands the e2e test is broken)

packages/api/test/e2e/rx-transfer.spec.ts
```diff
@@ -27,7 +27,7 @@ describe.skip('e2e transfer', () => {
             .send()
         )
       )
-      .subscribe((hash) => {
+      .subscribe((hash: Uint8Array) => {
```
jacogr: This is not a Uint8Array, it is a `Hash`

packages/rpc-core/src/formatting.spec.ts
```diff
@@ -13,8 +15,8 @@ const ENC_ONE = '0x24988a82b3bbaae58680bc9de5dea5aa';
 const ENC_TWO = '0xdb190750267d8849c35316ec9d342111';
 
 describe('formatting', () => {
-  let api;
-  let provider;
+  let api: any;
+  let provider: any;
```
jacogr: Here provider is any, i.e. mocked out. So correct.

packages/rpc-core/src/formatting.spec.ts
```diff
@@ -13,8 +15,8 @@ const ENC_ONE = '0x24988a82b3bbaae58680bc9de5dea5aa';
 const ENC_TWO = '0xdb190750267d8849c35316ec9d342111';
 
 describe('formatting', () => {
-  let api;
-  let provider;
+  let api: any;
```
jacogr: `api: Api`

jacogr:

PR #331, opened 2018-10-20
WIP: Examples using Promises/RxJs for for Deploy Smart Contract and Change Runtime Code

ltfschoen:
* See FIXME's in the code where errors or unexpected results are occurring
* Refer to PR #292

PR #334, opened 2018-10-21
Make Extrinsic compatible with Method

jacogr:
i.e. for eg `democracy.propose` we pass in Extrinsic from `api.tx.consensus.setCode` - as passed into a Method, we add the ability to extract the methodIndex and args, so we can use it in construction.

(We don't have this issue in the UI atm since it does not use `api.tx`, however these changes makes the attached e2e test work and avoid things like https://github.com/polkadot-js/api/pull/331/files#diff-d1fcbdf3a9f9eac6ad272b51b4e10d6eR27)

### DIFF
--- a/packages/types/src/Method.ts
+++ b/packages/types/src/Method.ts
@@ -80,7 +80,7 @@ export default class Method extends Struct {
       value.methodIndex &&
       value.args
     ) {
-      // destructure value, we only pass args/methodsIndex out
+      // destructure value, we only pass args/methodIndex out
       const { args, methodIndex } = value;
 
       // Get the correct callIndex
--- a/packages/types/src/Method.ts
+++ b/packages/types/src/Method.ts
@@ -80,10 +80,13 @@ export default class Method extends Struct {
       value.methodIndex &&
       value.args
     ) {
+      // destructure value, we only pass args/methodsIndex out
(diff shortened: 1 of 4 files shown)


packages/types/src/Method.ts
```diff
@@ -80,21 +80,29 @@ export default class Method extends Struct {
...
       value.args
     ) {
+      // destructure value, we only pass args/methodIndex out
+      const { args, methodIndex } = value;
+
       // Get the correct callIndex
-      const callIndex = value.methodIndex instanceof MethodIndex
-        ? value.methodIndex.callIndex
-        : value.methodIndex;
+      const callIndex = methodIndex instanceof MethodIndex
```
jacogr: Not for this PR. But I'm really wondering if Method should not just get rid of MethodIndex and just encapsulate the callIndex value in here. The whole methodIndex/callIndex dance is just getting cumbersome and adds some complexity and confusion. (It really is never seperated from Method itself, so it is overkill)

jacogr:
cc @ltfschoen This branch will solve your "cannot propose issues", it is an oversight in the API handling with Method/Extrinsic, where Method cannot instantiate from an Extrinsic like it should.

jacogr:
Replaced by https://github.com/polkadot-js/api/pull/335

PR #337, opened 2018-10-22
Address type cleanups

jacogr:
- When decoding from number, AccountIndex allows only <= 0xef for 1 byte
- Add constants for AccountIndex "magic"
- Expand Address tests, cleanup into decoding/encoding along with AccountId


packages/types/src/AccountId.spec.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright 2017-2018 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import AccountId from './AccountId';
+import U8a from './codec/U8a';
+
+describe('AccountId', () => {
+  describe('decoding', () => {
```
jacogr: Think these shows up as new (instead of just renames), since -

- Added types
- Added describe encoding/decoding sections (and categorised the tests in them)
- Moved testEncode/Decode into the relevent describe blocks

amaurymartiny: Yeah, much cleaner.

PR #340, opened 2018-10-24
Auto re-subscribe upon re-connect

### LINKED ISSUE #75: Re-connection should re-create subscriptions
Currently when the connection is lost (i.e. node restarted), the API will reconnect. However in this case the old/previous subscripts are still active and not updated - basically new subscriptions will work, old ones won't get data resulting in a half-working app.

So...

1. Keep a list of subscriptions,
2. On re-connect, unsubscribe from what we have (the node may be the same one with an internet hiccup) and then re-do the subscription fresh
3. The issue comes in with ids... currently subscriptions are mapped through on the Id from the node, in this case these probably needs to be a mapping layer that takes "id the app knows it at" -> "id the node knows it at"

Some through needed still.

jacogr:
Closes #75


jacogr:
Tested via an (modified) example for block subscriptions, does what it says on the tin.

PR #349, opened 2018-10-25
Add EventRecord & PairOf

jacogr:


packages/types/src/codec/PairOf.ts
```diff
@@ -0,0 +1,36 @@
...
+// PairOf defines a general structure that creates a Tuple pair of a specific type. These values
+// are accessible via the first & second getters
+export default class PairOf<
+  T extends Base
+> extends Tuple<{ first: Constructor<T>, second: Constructor<T> }> {
+  static with<
+    S extends ConstructorDef = { [index: string]: Constructor<Base> }
+  > (Type: S): Constructor<Tuple<S>>;
+
+  static with<T extends Base> (Type: Constructor<T>): Constructor<PairOf<T>> {
```
amaurymartiny: Not strictly overloading here. But as long as it's used internally I approve this.

jacogr: The issue with use comes in here - https://github.com/polkadot-js/api/pull/349/files#diff-a86daab9b4c7e3dcd4eb7d91f8feb064R143

(note the `any` cast, otherwise it complains bitterly)

Well, yes, best we can do for now - not overly excited about it either, but the lesser of evils.

jacogr:
Related #161

PR #350, opened 2018-10-26
Event definitions

jacogr:


packages/types/src/Event.ts
```diff
@@ -0,0 +1,149 @@
...
+import Base from './codec/Base';
+import Struct from './codec/Struct';
+import Tuple from './codec/Tuple';
+import U8aFixed from './codec/U8aFixed';
+import { TypeDef, getTypeClass, getTypeDef } from './codec/createType';
+import Metadata from './Metadata';
+
+const EventTypes: { [index: string]: Constructor<EventData> } = {};
+
+// // length 0x10 = 0b100000 = 0b100 << 1 = 4
```
jacogr: Yes, just here until the tests are done :) My debugging efforts to track.

packages/types/src/Event.ts
```diff
@@ -0,0 +1,109 @@
...
+    super(Types, value);
+
+    this._typeDef = typeDef;
+  }
+
+  get typeDef (): Array<TypeDef> {
+    return this._typeDef;
+  }
+
+  toHex (): string {
```
amaurymartiny: Already defined in Struct.

packages/types/src/Event.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  get data (): EventData {
+    return this.get('data') as EventData;
+  }
+
+  get index (): EventIndex {
+    return this.get('index') as EventIndex;
+  }
+
+  toHex (): string {
```
amaurymartiny: same

packages/types/src/Address.ts
```diff
@@ -31,37 +31,36 @@ export default class Address extends Base<AccountId | AccountIndex> {
...
+      } else if (value[0] === 0xff) {
         return new AccountId(value.subarray(1));
       }
-      return value.length === 32
-        ? new AccountId(value)
-        : new AccountIndex(value);
+
+      const [offset, length] = AccountIndex.readLength(value);
+
+      return new AccountIndex(value.subarray(offset, offset + length), length);
```
amaurymartiny: I still think that the 2nd argument (maxLength) is not needed.

```
// // event: balances.NewAccount(AccountId, AccountIndex, NewAccountOutcome)
// 0100
// 37e027d776cd005c12bcf6722421374a9037167a0ceaf918f341c4ad68d54e59
// 0700000000 // AccountIndex+NewAccountOutcome
```

You pass in `new AccountIndex([07, 00, 00, 00, 00])` (without 2nd arg). That ones decodes it, it finds out that the length is 4 bytes, its encodedLength is 4. Then move on, NewAccountOutcome, 1 byte, and everything works.

packages/types/src/Event.ts
jacogr: Grrr, added it for debugging actually. Thanks.

packages/types/src/Address.ts
jacogr: Yes, however - AccountIndex it can be up to 8 bytes in length. If not added, it would trim those, e.g. 

https://github.com/polkadot-js/api/blob/master/packages/types/src/AccountIndex.spec.ts#L49

https://github.com/polkadot-js/api/blob/master/packages/types/src/Address.spec.ts#L71

PR #375, opened 2018-11-05
docs: Fixes #374. Remove FIXME. Add example retrieving Method instance data

### LINKED ISSUE #374: Method instance args getter should return Array not a Struct so remove the FIXME
If we create a new instance of the `Method` class, it is understood based on discussion with @amaurymartiny that the current functionality of the `args` getter is ok, and the FIXME as shown [here](https://github.com/polkadot-js/api/blob/master/packages/types/src/Method.ts#L170) is no longer required.

For example, when you go to http://localhost:3000/#/extrinsics and click Submit Transaction, and you want to obtain the recipientId (dest) and amount (value) from the prop `this.props.value.extrinsic` that passed down to polkadot-js/apps/packages/ui-signer/src/Extrinsic.tsx, then 

```
const methodInstance = new Method(extrinsic, extrinsic.meta);

// WRONG - do not use `.raw`. 
// See https://github.com/polkadot-js/apps/pull/389#discussion_r230735365
// const recipientId = methodInstance.args[0].raw.toString();
// const amount = methodInstance.args[1].raw.toString();

// WRONG - suggested in the FIXME, but not preferred approach
// const recipientId = methodInstance.args.get('dest').toString();
// const amount = methodInstance.args.get('value').toString();

// RIGHT
const recipientId = methodInstance.get('args').get('dest').toString();
const amount = methodInstance.get('args').get('value').toString();  // or .toBn().toString()
```

Proposed action is to remove the following [FIXME](https://github.com/polkadot-js/api/blob/master/packages/types/src/Method.ts#L170)
```
get args (): Array<Base> {
    // FIXME This should return a Struct instead of an Array
    return [...(this.get('args') as Struct).values()];
  }
```

ltfschoen:


packages/types/src/Method.ts
```diff
@@ -66,6 +66,25 @@ export default class Method extends Struct {
...
+   *
+   * @example
+   * <BR>
+   *
+   * ```javascript
+   * import { Method } from '@polkadot/types';
+   *
+   * const methodInstance = new Method(extrinsic, extrinsic.meta);
+   *
+   * // WRONG - never access `.raw` outside of the @polkadot/types package
```
amaurymartiny: IMHO this should not go into the docs. The end-user docs should never mention `.raw`.

ltfschoen: @amaurymartiny I've pushed a commit to remove mention of `.raw` in user-facing docs

packages/types/src/Method.ts
```diff
@@ -66,6 +66,21 @@ export default class Method extends Struct {
...
+   * @example
+   * <BR>
+   *
+   * ```javascript
+   * import { Method } from '@polkadot/types';
+   *
+   * const methodInstance = new Method(extrinsic, extrinsic.meta);
+   *
+   * // Obtain recipient id and amount for an extrinsic
+   * const recipientId = methodInstance.get('args').get('dest').toString();
```
jacogr: Ok, so `args` is a getter, `dest` is a member of the struct, so 

```
const recipientId = methodInstance.args.get('dest').toString();
...
```

jacogr: ... and I could be wrong, actually `args` returns an array, so it needs to be indexed, it is not a struct wrapping a map. So correct would be -

```
const recipientId = methodInstance.args[0].toString();
````

If you are doing this, you should know exactly what the method should contain. (I am assuming this is a transfer)

Comment (jacogr) [reply packages/types/src/Method.ts]: ... and yes, I am wrong, it is the `args` (which is actually only used internally) vs the `args` getter. Naming here is not quite 100% :(

Comment (jacogr) [reply packages/types/src/Method.ts]: See https://github.com/polkadot-js/api/issues/374#issuecomment-436151539

I would have expected (if we know it is a transfer), that this is -

```
const recipentId = methodInstance.args.dest.toString();
```

If we dunno, it is a different kettle of fish.

Comment (amaurymartiny) [reply packages/types/src/Method.ts]: The cleanest is:
- Remove FIXME
- In Method.ts, `get args()` returns a Map
- (optional) In Method.ts, add `get argsArray()` that returns the array
- `const recipientId = methodInstance.args.get('dest').toString();`

Comment (ltfschoen) [reply packages/types/src/Method.ts]: is the following approach on the right track?:
```
  get args () {
    return {
      get: (keyLookup) => {
        for (const [k, v] of this.get('args').entries()) {
          if (k === keyLookup) {
            return v;
          }
        }

        return undefined;
      },
      toArray: () => {
        return this.get('args').toArray();
      }
    };
    // // FIXME This should return a Struct instead of an Array
    // return [...this.get('args').values()];
  }
```

Comment (amaurymartiny) [reply packages/types/src/Method.ts]: No, it should be as simple as:
```
get args () {
  return this.get('args');
}
```
This.get('args') returns a Struct, so already has `get (keyLookup)`, `toArray` etc.

amaurymartiny: I can have a look in person with you

ltfschoen: Ok, hopefully tomorrow as i'm working from home today, i've picked up a bad cold and my nose is dripping like a hose

jacogr: That is just weird :) The code _looks_ equivalent.

jacogr:

jacogr:
Closing, not willing to break the API this close before 1.0

PR #382, opened 2018-11-07
Allow construction of all types via U8a

jacogr:
This is needed to work around the storage creation - basically we go from a `Bytes` -> `any` type. This works wonderfully, except in the case of `Bytes`. Here it passes an` Uint8Array` (output from `Option`) to the constructor and loses all length. We cannot pass the length so then everything else will break.

So the solution is 2-fold -

- Allow all types to also construct via `U8a` (all support `Uint8Array` already, so it is a small hop to get her)
- Allow `Option` to actually return the `Base<any>` value - this is actually correct, since we return `Base` everywhere else
- Now when converting from `StorageData` (which is just `Bytes`) to anything else, the `U8a` creation path is followed

It is short-term though - once `U8a` extends `Uint8Array` properly, this will mostly be reverted.

Additionally, now we have explicitly added the `Uint8Array` (and `U8a`) to all constructors - although not specified, it is actually used that way already.

### DIFF
--- a/packages/rpc-core/src/index.ts
+++ b/packages/rpc-core/src/index.ts
@@ -182,7 +182,7 @@ export default class Rpc implements RpcInterface {
       // outputType that we have specified. Fallback to Data on nothing
       const type = (params[0] as StorageKey).outputType || 'Data';
 
-      return createType(type, base.raw);
+      return createType(type, base);
     } else if (method.type === 'StorageChangeSet') {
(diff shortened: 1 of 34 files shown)


amaurymartiny:
It might have been easier to do U8a extends Uint8Array (I know, my job...), and in this is case `isU8a(some instance of U8a)` will return true, which avoid the additional ifs.

I can clean that up once i come back to this.

PR #392, opened 2018-11-12
U8a extends Uint8Array

amaurymartiny:
related to #161

### DIFF
--- a/packages/types/src/AccountId.ts
+++ b/packages/types/src/AccountId.ts
@@ -7,7 +7,6 @@ import { AnyU8a } from './types';
 import { decodeAddress, encodeAddress } from '@polkadot/keyring';
 import { hexToU8a, isHex, isString, isU8a, u8aToU8a } from '@polkadot/util';
 
-import U8a from './codec/U8a';
 import U8aFixed from './codec/U8aFixed';
 
 // A wrapper around an AccountId/PublicKey representation. Since we are dealing with
--- a/packages/types/src/AccountId.ts
+++ b/packages/types/src/AccountId.ts
@@ -28,9 +28,9 @@ export default class AccountId extends U8aFixed {
     if (isU8a(value) || Array.isArray(value)) {
       return u8aToU8a(value);
     } else if (isHex(value)) {
-      return hexToU8a(value);
+      return hexToU8a(value.toString());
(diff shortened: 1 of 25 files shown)


packages/types/src/AccountId.ts
```diff
@@ -25,23 +24,23 @@ export default class AccountId extends U8aFixed {
     return encodeAddress(value);
   }
 
-  static decodeAccountId (value: AnyU8a): Uint8Array {
-    if (value instanceof U8a) {
-      return value.raw;
-    } else if (isU8a(value) || Array.isArray(value)) {
+  private static decodeAccountId (value: any): Uint8Array {
```
jacogr: Can we keep this explicit, rather than `any`?

packages/types/src/Bytes.ts
```diff
@@ -46,10 +42,10 @@ export default class Bytes extends U8a {
 
   toU8a (isBare?: boolean): Uint8Array {
     return isBare
-      ? this.raw
+      ? super.toU8a(isBare)
```
jacogr: Shouldn't this one also be `this` like below? (Same thing, bug may help in consistency?)

packages/types/src/codec/U8aFixed.ts
```diff
@@ -11,25 +13,23 @@ type BitLength = 8 | 16 | 32 | 64 | 128 | 256 | 512;
...
+    );
   }
 
-  private _trimLength (): number {
-    return this._bitLength / 8;
-  }
+  private static decodeU8aFixed (value: AnyU8a, bitLength: BitLength = 256): AnyU8a {
+    if (isU8a(value)) {
+      return value.subarray(0, bitLength / 8);
+    } else if (Array.isArray(value) || isBuffer(value) || isString(value)) {
```
jacogr: Do we need Buffer in here? (We generally don't use them anywhere else)

packages/types/src/Bytes.ts
jacogr: No, I mean the following -

```
toU8a (isBare?: boolean): Uint8Array {
  return isBare
    ? this
    : u8aConcat(
        Compact.encodeU8a(this.length, DEFAULT_LENGTH_BITS),	       
        this
      );
}
```

PR #395, opened 2018-11-13
Override subarray, returning via Uint8Array.from

jacogr:


packages/types/src/AccountId.spec.ts
```diff
@@ -2,8 +2,12 @@
 // This software may be modified and distributed under the terms
 // of the ISC license. See the LICENSE file for details.
 
-import AccountId from './AccountId';
 import U8a from './codec/U8a';
+import Vector from './codec/Vector';
+import jsonVec from './json/AccountIdVec.001.json';
+import AccountId from './AccountId';
+import { setAddressPrefix } from '@polkadot/keyring';
```
amaurymartiny: (unsubtantial) reorder imports

packages/types/src/codec/U8a.spec.ts
```diff
@@ -49,6 +49,6 @@ describe('U8a', () => {
   });
 
   it('implements subarray correctly', () => {
-    expect(u8a.subarray(1, 3).toString()).toEqual('0x0203');
+    expect(u8a.subarray(1, 3).toString()).toEqual('2,3');
```
amaurymartiny: Actually here, would be nicer to remove the toString, and toEqual(Uint8Array.from())

PR #396, opened 2018-11-13
UInt extends BN

amaurymartiny:
related #161.


packages/types/src/codec/UInt.ts
```diff
@@ -30,24 +29,26 @@ export default class UInt extends Base<BN> {
...
-      return new BN(value, 10);
+      return new BN(value, 10).toString();
     }
 
-    return bnToBn(value);
+    return bnToBn(value).toString();
   }
 
-  get bitLength (): UIntBitLength {
+  bitLength (): UIntBitLength {
```
jacogr: Q: Do we actually use/rely on `bitLength` anywhere? (Or is is just there as a "could be useful"). The reason for asking, is that I for the life of me cannot remember why we have it.

packages/types/src/Moment.ts
```diff
@@ -27,8 +27,6 @@ export default class Moment extends Base<Date> {
       return value.raw;
     } else if (value instanceof Date) {
       return new Date(Math.ceil(value.getTime() / 1000) * 1000);
-    } else if (value instanceof UInt) {
```
jacogr: In removing this, we don't seem to cater for `BN` construction anymore? Should we not have a `isBn` path as well?

packages/types/src/codec/Compact.ts
```diff
@@ -147,6 +147,6 @@ export default class Compact extends Base<UInt | Moment> {
   }
 
   toU8a (isBare?: boolean): Uint8Array {
-    return Compact.encodeU8a(this.raw.toBn(), this.bitLength);
+    return Compact.encodeU8a(this.raw.toBn(), this.bitLength());
```
jacogr: Ahh... ok, to my question above, here it is used. (I should have probably left a note in there somewhere as to why we have `bitLength`)

packages/types/src/Moment.ts
```diff
@@ -40,7 +38,7 @@ export default class Moment extends Base<Date> {
     );
   }
 
-  get bitLength (): UIntBitLength {
+  bitLength (): UIntBitLength {
```
jacogr: Is there a reason why these are now not getters? Additionally, I believe in `U8aFixed` we still have it as a getter?

packages/types/src/codec/UInt.ts
```diff
@@ -66,84 +67,10 @@ export default class UInt extends Base<BN> {
...
-    return bnToU8a(this.raw, this._bitLength, true);
-  }
-
-  toString (): string {
-    return this.raw.toString();
+    return bnToU8a(this, this._bitLength, true);
   }
 
   toBn (): BN {
-    return this.raw;
```
jacogr: Yay! Lotsa stuff gone. :D

packages/types/src/Moment.ts
amaurymartiny: The return at the bottom of the function handles the BN case.

packages/types/src/Moment.ts
amaurymartiny: I would love to have bitLength everywhere as getter too, but on BN, bitLength is a method, so on Moment it needs to be a method too.

I'll change bitLenght on U8aFixed to method too, for consistency.

amaurymartiny: Alternatively, @ts-ignore in Uint, but we stray away from BN

jacogr: No, rather keep it as a method, it is cleaner.

PR #400, opened 2018-11-14
Moment extends Date

amaurymartiny:


packages/types/src/Moment.spec.ts
```diff
@@ -14,7 +14,7 @@ describe('Moment', () => {
         expect(new Moment(input)[toJSON ? 'toJSON' : 'toString']()).toBe(expected);
       });
 
-    testDecode('Date', new Date(1537968546280), '2018-09-26T13:29:07.000Z');
+    testDecode('Date', new Date(1537968546280), 'Wed Sep 26 2018 15:29:07 GMT+0200 (Central European Summer Time)');
```
amaurymartiny: So:
- if we keep the Date's default toString, it will show like this (i.e. with timezone, not sure if this test actually passes in other timezones)
- or we can overwrite toString in Moment to return toIsoString, but users might be confused.

I went with the 1st solution

amaurymartiny: Travis answered my question, it doesn't pass in other timezones. I'll put toISOString then

amaurymartiny: Travis answered me: tests don't pass in other timezones. Only checking the day in tests.

packages/types/src/codec/Compact.spec.ts
```diff
@@ -120,8 +120,8 @@ describe('Compact', () => {
 
     it('constructs properly via number as Moment', () => {
       expect(
-        new Compact(Moment, 1537968546).toString()
-      ).toEqual('2018-09-26T13:29:06.000Z');
+        new Compact(Moment, 1537968546).toString().startsWith('Wed Sep 26 2018') // The time depends on the timezone this test is run in
```
jacogr: I suppose the tests could have used `toISOString`

amaurymartiny: No, toISOString doesn't exist on Compact.

PR #403, opened 2018-11-15
Simplify `Base`

amaurymartiny:
- All types in `types` implement the Codec interface
- Base now only means "this class holds some other kind of value". Unfortunately, we won't get rid of it, imo the following classes need Base:
    - Address
    - Enum, EnumType, Option (although they could all subclass one same generic Enum)
    - Vector, if we don't manage to extend Array
- I think there's some way to let Compact and Set not extend Base
- Option.isEmpty -> Option.isNone, also added Option.isSome, to follow Rust.
- `protected raw`, so nobody external can access it.


packages/types/src/Metadata.ts
```diff
@@ -314,11 +313,10 @@ export class RuntimeModuleMetadata extends Struct {
   }
 
   get storage (): StorageMetadata | undefined {
-    const value = (this.get('storage') as Option<StorageMetadata>).value;
-
-    return value
-      ? value.raw
-      : value;
+    // FIXME This should actually return an Option<StorageMetadata>
```
amaurymartiny: To be 100% correct, here, it should return `Option<StorageMetadata>` instead of `StorageMetadata | undefined`.

I'm pretty sure that this will only be used internally, so I would recommend to put `Option<StorageMetadata>`, and we will write `someMetadataModule.storage.unwrap().functions...` (or if .isSome else...). Just want to double check @jacogr

jacogr: Only internally. So we can break it in whichever way we seem fit.

jacogr:
On the Base issue, it is actually horrible. (And maybe it is just one impl. thereof). I stumbled across this today -

https://github.com/polkadot-js/common/pull/251/files#diff-6435c166f95a4ef0453f9a679ef8ebecR114

Maybe just for the enum, but since the thing returns a Base-something, I have no other way to actually get to the internals. (And that _will_ break relatively soon)

packages/types/src/codec/EnumType.spec.js
```diff
@@ -33,4 +33,15 @@ describe('Struct', () => {
...
+
+  it('allows accessing the type and value', () => {
+    const text = new Text('foo');
+    const enumType = new EnumType(
+      [Text, U32],
+      { Text: text }
+    );
+
+    expect(enumType.type).toBe('Text');
+    expect(enumType.value).toEqual(text);
```
amaurymartiny: @jacogr I added a test here, you should be able to access the .raw value like this. Why it returns a Base on the trie is a big unknown for me, I can have a look once it's more stabilized.

PR #409, opened 2018-11-21
Use compact* from util for encoding

jacogr:


packages/type-extrinsics/package.json
```diff
@@ -11,6 +11,6 @@
   "license": "Apache-2.0",
   "dependencies": {
     "@babel/runtime": "^7.1.5",
-    "@polkadot/util": "^0.32.21"
+    "@polkadot/util": "^0.32.23"
```
amaurymartiny: Shouldn't @polkadot/types be here? Idem in other package.jsons.

jacogr: Yeap. The same thing that got us in trouble the first time around. Will add to storage as well.

amaurymartiny: And i think also in api, api-observable, rpc-core, rpc-rx etc

PR #411, opened 2018-11-22
Update with Grandpa types

jacogr:


jacogr:
Moving to WIP until the Substrate macros return all values again. There _could_ be more types missing since some modules are now skipped.

Issue #406, opened 2018-11-20
Compiler is not throwing all errors


amaurymartiny:
Repro:
- replace this line https://github.com/polkadot-js/api/blob/master/packages/api/src/promise/types.d.ts#L10
with `import { Codec } from '@polkadot/types/codec';`
- `tsc --noEmit`

Expected:
Compiler should throw, because the import does not exist.

Actual:
It passes.

jacogr:
With @chevdor 's project, I toggled this line for some additional errors (in his project). It may need to explicitly set? https://github.com/polkadot-js/dev/blob/master/packages/dev/config/tsconfig.json#L25

It _could_ be related, not 100%, but worth looking into the options which is mostly left as-is.

mn-io:
I'd like to add that I have to use the types with following tslint option:
"esModuleInterop": true

otherwise following error will be thrown:
.../node_modules/@polkadot/storage/types.d.ts
(7,8): Module '".../node_modules/@types/bn.js/index"' has no default export.

jacogr:
Seems to be `*.d.ts` files only - acting stupid in `*.ts` files does get caught.

jacogr:
Ok, some playing (yesterday)  - it is only in the `*.d.ts` files this happens. The `*.ts` files are fine. Chatted to @amaurymartiny so the suggestion is t do it right and rename these to `*.ts` (which is basically all repos)

PR #421, opened 2018-11-29
Move types.d.ts -> types.ts

jacogr:
Closes https://github.com/polkadot-js/api/issues/406

(There has been quite a few lil nigglies in there)


jacogr:
Just doing the same in common and will include the bump of that package in this PR as well

Issue #419, opened 2018-11-29
Struct.with to automatically define getters for properties


jacogr:
Something like

```
Object.defineProperty(this, 'propName', {
  enumerable: true,
  writable: false,
  get: () => this.get('propName'); // or via index
});
```

(Obviously looping through the structure provided with the names)

PR #425, opened 2018-11-29
Define properties on Struct.with automatically

jacogr:
Closes https://github.com/polkadot-js/api/issues/419

Sorry for the "revert revert" commit - I pushed this one to master by mistake, then reverted, then branched, then reverted the revert. So messy and fat-fingered.

### DIFF
--- a/packages/types/src/codec/Struct.ts
+++ b/packages/types/src/codec/Struct.ts
@@ -111,6 +111,13 @@ export default class Struct<
     return class extends Struct<S> {
       constructor (value?: any, jsonMap?: Map<keyof S, string>) {
         super(Types, value, jsonMap);
+
+        (Object.keys(Types) as Array<keyof S>).forEach((key, index) => {
+          Object.defineProperty(this, key, {
+            enumerable: true,
+            get: () => this.getAtIndex(index)
+          });
+        });
       }
     };
   }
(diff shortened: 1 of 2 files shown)


packages/types/src/codec/Struct.ts
```diff
@@ -111,6 +111,13 @@ export default class Struct<
     return class extends Struct<S> {
       constructor (value?: any, jsonMap?: Map<keyof S, string>) {
         super(Types, value, jsonMap);
+
+        (Object.keys(Types) as Array<keyof S>).forEach((key, index) => {
+          Object.defineProperty(this, key, {
+            enumerable: true,
+            get: () => this.getAtIndex(index)
```
amaurymartiny: `this.get(key)` is faster, we don't do an .toArray() each time.

packages/types/src/codec/Struct.ts
```diff
@@ -111,6 +111,13 @@ export default class Struct<
     return class extends Struct<S> {
       constructor (value?: any, jsonMap?: Map<keyof S, string>) {
         super(Types, value, jsonMap);
+
+        (Object.keys(Types) as Array<keyof S>).forEach((key, index) => {
+          Object.defineProperty(this, key, {
```
amaurymartiny: Does this by any chance add TS typings on the struct?

packages/types/src/codec/Struct.ts
jacogr: Yes, actually had it as `get(key)` originally, will revert. Thanks.

packages/types/src/codec/Struct.ts
jacogr: Not 100% sure what you mean - if it is available, nope, look at the test, had to any to extract it :(

PR #431, opened 2018-12-03
Document getters and functions for all types (static missing)

### LINKED ISSUE #376: Document type classes in @polkadot/types
We need to start linking through to these, the first order of business is to document the basics in them -

- name, description
- getters
- static methods should be explicitly excluded from the docs (i.e. `U8a.decodeU8a` should not be included, since these are meant to be only used internally, neither any other statics in the classes). There is one exception to this blanket ban on static docs - the `Compact` methods to manage (read, write, add) lengths are utility functions and should be in.
- Do not document (exclude explicitly) functions that should be replaced by extending - specifically here we have the `add`, `sub`, `eq`, `isZero` on the `UInt` classes - since that will extend `BN` (soon-ish), don't put any focus on them.

In short -

- docs should be useful
- it should be linkable from other places, e.g. doing an in-doc `[[ MyFantasticType ]]` should link through to the `MyFantasticType` class (this way we can use it from e.g. the api classes)
- it should focus on the descriptions and getters
- it should not have internal-only methods not fields (no fields at all)

jacogr:
Closes #376

### DIFF
--- a/packages/types/src/Null.ts
+++ b/packages/types/src/Null.ts
@@ -10,10 +10,16 @@ import { Codec } from './types';
  * Implements a type that does not contain anything (apart from `null`)
  */
 export default class Null implements Codec {
+  /**
+   * @description The length of the value when encoded as a Uint8Array
+   */
   get encodedLength (): number {
     return 0;
   }
 
+  /**
+   * @description Returns a hex string representation of the value
+   */
   toHex (): string {
     return '0x';
   }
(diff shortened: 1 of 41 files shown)


packages/types/src/Null.ts
```diff
@@ -10,10 +10,16 @@ import { Codec } from './types';
...
 export default class Null implements Codec {
+  /**
+   * @description The length of the value when encoded as a Uint8Array
+   */
   get encodedLength (): number {
     return 0;
   }
 
+  /**
+   * @description Returns a hex string representation of the value
```
amaurymartiny: does @inheritdoc work here? In which case you don't need to put the same desc everywhere.

PR #433, opened 2018-12-03
API cleanups, circular deps

jacogr:


packages/types/src/Origin.ts
```diff
@@ -2,14 +2,18 @@
...
+import Null from './Null';
+
 /**
  * @name Origin
  * @description
  * Where Origin occurs, it should be ignored as an internal-only value, so it should
  * never actually be constructed
  */
-export default class Origin {
+export default class Origin extends Null {
```
jacogr: Done so `Origin` actually is a `Codec`

Issue #438, opened 2018-12-04
Type documentation links are not working


jacogr:
From https://polkadot.js.org/api/types/

- Metadata (and it looks the same as others) is not creating a link
- Other links (e.g. AccountId), try to load the .md file instead of the generated HTML - well, actually HTML is not generated

Need a solution that does not entail adding all these to SUMMARY.md

ltfschoen:
What do you think about this hack https://github.com/GitbookIO/gitbook/issues/659#issuecomment-418074102?
For all the links that we have in the README files of the subdirectories like polkadot-js/api/docs/types/README.md, we'd make duplicates of all the markdown files that they link to in the subdirectory polkadot-js/api/docs/hidden/, and then we'd list them all at the bottom of the main polkadot-js/api/docs/SUMMARY.md, but they wouldn't need to be named i.e.
```
[](hidden/types/classes/_accountid_.accountid.md)
[](hidden/types/classes/_accountindex_.accountindex.md)
...
```
and then we'd add the following CSS to polkadot-js/api/docs/styles/website.css, so that all the links that we've added to the main SUMMARY.md file that link to the hidden/ subdirectory are hidden and don't appear in the navigation page, but because they are added as links in the main SUMMARY.md file, the corresponding .html files will be generated from the .md files in the subdirectories so the links work.
```
.summary .chapter[data-path^="hidden/"] {
  display: none;
}
```

Note: It looks like this Gitbook issue has been around for a while, these are the two issues I found:
* https://github.com/GitbookIO/gitbook/issues/81
* https://github.com/GitbookIO/gitbook/issues/659

jacogr:
Yes, I saw the hack - It would mean we need to add the types to 3 places - types/README, types/index.js and docs/SUMMARY - but sadly I see no other option. So as much as it sucks, it is probably the _only_ solution.

jacogr:
Just a quick one - I would not move the md to hidden, rather make the docs/types/classes folder hidden from SUMMARY.md - that should hide it there and have it in README

PR #445, opened 2018-12-06
Fix types documentation links

jacogr:
Closes https://github.com/polkadot-js/api/issues/438


docs/SUMMARY.md
```diff
@@ -24,3 +24,123 @@
   - [Read chain state](examples/promise/05_read_storage/README.md)
   - [Make a transfer](examples/promise/07_transfer_dots/README.md)
   - [Display system events](examples/promise/08_system_events/README.md)
+
+<!-- These are hidden, allowing HTML generation but does not appear in the sidebar -->
+
+- [hidden](types/classes/_accountid_.accountid.md)
```
jacogr: This is  beyond horrible, but it is what it is - see the linked issue that related back to "best practices" for these types of hacks. (There are plugins to generate the SUMMARY.md automagically, but we are actually crafting these to only have specific items in the index)

docs/README.md
```diff
@@ -2,11 +2,9 @@
 
 You can query a node and interact with the Polkadot network using Javascript. This guide describes the Polkadot Javascript API.
 
-{% include "SUMMARY.md" %}
```
jacogr: Removed since the "hidden" override doesn't apply here. And since it is in the sidebar via SUMMARY.md, we are not losing navigation

docs/styles/website.css
```diff
@@ -1,3 +1,8 @@
+/* Exclude the items add to the SUMMARY.md root */
+.summary .chapter[data-path*="types/classes/"] {
```
jacogr: Hidden and tested to hide.

PR #454, opened 2018-12-10
Improve events

xlc:
This allows to detect if ws connection failed (because node is not running)


jacogr:
Not 100% sure why CI doesn't report back, looked at CI and everything seems all-ok there. (And the one build did pass)

PR #456, opened 2018-12-10
Update types parsing to latest metadata

amaurymartiny:


packages/types/src/Metadata/latest.json
```diff
@@ -1473,7 +1473,7 @@
             "name": "VoteOf",
             "modifier": 0,
             "type": {
-              "key": "(ReferendumIndex, AccountId)",
+              "key": "( ReferendumIndex , AccountId )",
```
jacogr: This one is weird - I'm assuming it relates to the same issue, but basically above we have all spaces in tuples whacked, here they are added. (Thank heavens we have trims in createType)

packages/types/src/Type.ts
```diff
@@ -172,12 +172,16 @@ export default class Type extends Text {
...
+        // `T :: AccountId` -> `T::AccountId` https://github.com/paritytech/substrate/issues/1244
+        .replace(/\s+::\s+/g, '::')
         // anything `T::<type>` to end up as `<type>`
         .replace(/T::/g, '')
         // `system::` with `` - basically we find `<T as system::Trait>`
         .replace(/system::/g, '')
         // replace `<T as Trait>::` (possibly sanitised just above)
         .replace(/<T as Trait>::/g, '')
+        // `<TasTrait>::type` -> `type` https://github.com/paritytech/substrate/issues/1244
+        .replace(/<TasTrait>::/g, '')
```
jacogr: Could have possibly gone for the same `\s+` hackery as above with `T::` here

amaurymartiny: Prefer to leave it as 2 lines, the 1st one is necessary, the 2nd is a bug fix

PR #463, opened 2018-12-12
Simplify findClosing (maybe?)

amaurymartiny:
Using a stack instead of depth. Slightly less performant since it takes more memory, so if it doesn't fix the codeclimate issue then I'll close it.


amaurymartiny:
Meh, "No new, fixed, or changed issues."

PR #464, opened 2018-12-12
Allow Struct to take Map (and Struct) in constructor

jacogr:


packages/types/src/codec/Struct.ts
```diff
@@ -74,6 +74,14 @@ export default class Struct<
       return {} as T;
     }
 
+    return Struct.decodeStructFromObject(Types, value, jsonMap);
+  }
+
+  private static decodeStructFromObject<
```
jacogr: Still a horror, but at least now all the horror bits are together :)

PR #466, opened 2018-12-12
Add types from latest metadata

amaurymartiny:
- Add 2 new types
- Since the metadata generated by substrate has inconsistent whitespaces, I decided to remove them all in our metadata.


packages/types/src/Metadata/latest.json
```diff
@@ -1452,12 +1490,34 @@
             "modifier": 0,
             "type": {
               "key": "ReferendumIndex",
-              "value": "(BlockNumber,Proposal,VoteThreshold)"
+              "value": "(ReferendumInfo)"
```
amaurymartiny: This one is a weird one, should simply be ReferendumInfo?

jacogr: It comes from 

```
/// Information concerning any given referendum.
		pub ReferendumInfoOf get(referendum_info): map ReferendumIndex => Option<(ReferendumInfo<T::BlockNumber, T::Proposal>)>;
```

It think if you look at the original, it would make sense why the Tuple was there. But here Tuple and no-Tuple are equivalent. But yes, weird.

packages/types/src/Type.ts
```diff
@@ -172,15 +174,13 @@ export default class Type extends Text {
   private static _removeTraits (): Mapper {
     return (value: string): string => {
       return value
-        // `T :: AccountId` -> `T::AccountId` https://github.com/paritytech/substrate/issues/1244
-        .replace(/\s+::\s+/g, '::')
+        // remove all whitespaces
```
jacogr: The hammer approach :)

amaurymartiny: Hope it works... If there are whitespaces, they should be handled before (e.g. `as HasCompact`)

packages/types/src/Metadata/latest.json
amaurymartiny: Hmm, so actually, here it should be `Option<(ReferendumInfo)>`

packages/types/src/codec/createType.ts
```diff
@@ -94,6 +95,9 @@ export function getTypeDef (_type: Text | string): TypeDef {
   } else if (startingWith(type, 'Compact<', '>')) {
     value.info = TypeDefInfo.Compact;
     value.sub = getTypeDef(subType);
+  } else if (startingWith(type, 'Option<', '>')) {
```
jacogr: I believe we are missing a "from the info, create the type" around line 117 in this file.

packages/types/src/Type.ts
jacogr: I don't think it will have any negative impact at all. As long as we track the different options, we are fine. (And we still do that)

packages/types/src/Metadata/latest.json
jacogr: Since it is storage, it doesn't make much difference, they all come back as either `null` or the value. However, it may point to something funny in the parsing and/or option just stripped out (I think the original was like that - Option was actually just ignored in `Type.ts`).

Issue #475, opened 2018-12-15
api.query at a specific block


jacogr:
Currently api.query will perform queries (one-shot) leaving out the optional head hash. It is useful to do historic queries as well where it is added. 

Suggesting adding ‘at’ that will take the hash params. So basically current we can only do

```
api.query.balances.freeBalanceOf(Alice)
```

which performs the query for the latest block. An extension to this could be

```
api.query.balances.freeBalanceOf.at(hash, Alice) 
```

which will perform at a specific block. The reason for the `at` is that it would be simple to determine which route (i.e. which underlying RPC call) to take. It is a bit tougher and messier to try and adjust the calls and determine intent by just using the current endpoints, i.e. it is a one-time query, a subscription query or a one-time-at query. At the same time we still have control of which params goes where. Additionally can be added without breaking old use cases. 

(Suggestions and comments obviously welcome on approach)

jacogr:
The reason for the ‘at’ is that it would be simple to determine which route (underlying RPC call) to take. And have control of which params goes where. Additionally can be added without breaking old use cases.

PR #477, opened 2018-12-15
Add ability to do api.query at a specific hash

jacogr:
- Closes https://github.com/polkadot-js/api/issues/475
- Add `api.query.<section>.<method>.at(hash, arg?)` to do a one-shot query at a specific hash

### DIFF
--- a/packages/api/src/promise/types.ts
+++ b/packages/api/src/promise/types.ts
@@ -2,8 +2,8 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { Codec } from '@polkadot/types/types';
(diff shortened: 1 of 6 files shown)


packages/api/src/promise/types.ts
```diff
@@ -13,6 +14,7 @@ import SubmittableExtrinsic from './SubmittableExtrinsic';
 
 export interface QueryableStorageFunction extends StorageFunction {
   (arg?: any): Promise<Codec | null | undefined>;
+  at: (hash: Hash, arg?: any) => Promise<Codec | null | undefined>;
```
jacogr: Erk, missing the same definition for RxJS - will add as well.

PR #483, opened 2018-12-17
Factorize some u8a decoding logic

amaurymartiny:
Hopefully codeclimate will be slightly more happy

Also laying some ground work for the default values PR.

### DIFF
--- a/packages/types/src/codec/utils/decodeU8a.ts
+++ b/packages/types/src/codec/utils/decodeU8a.ts
@@ -15,9 +15,8 @@ export default function decodeU8a (u8a: Uint8Array, types: Constructor[]): Codec
   if (!types.length) {
     return [];
   }
-  const Type = types.shift();
+  const Type = types.shift() as Constructor;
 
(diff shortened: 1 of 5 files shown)


packages/types/src/codec/utils/decodeU8a.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @param u8a - The u8a to decode.
+ * @param types - The array of Constructor to decode the U8a against.
+ */
+export default function decodeU8a (u8a: Uint8Array, types: Constructor[]): Codec[] {
+  if (!types.length) {
+    return [];
+  }
+
+  const Type = types.shift() as Constructor;
```
jacogr: Wouldn't it make more sense to reduce across the types Array instead of modifying it?

More a nitpick, but it could lead to issues where the caller expects his input not to be modified.

amaurymartiny: Agreed

Issue #359, opened 2018-10-29
Update static Metadata to use latest Compact version


jacogr:
- As found in https://github.com/polkadot-js/api/pull/358
- Replace hex in `Metadata.rpc.ts` with hex from `json/Metadata.003.json` (Grandpa, added in https://github.com/polkadot-js/api/pull/411)
- Update tests with actual correct parameters and outputs

PR #484, opened 2018-12-17
Add support for default values

### LINKED ISSUE #232: Return default values as specified in meta
Current when storage entry is empty (null), we don't automagically set the default value. So for instance, we are doing this - https://github.com/polkadot-js/api/pull/231/files#diff-2e3ff95c7b73d14c1a8b2fb9c44197a9R143

### LINKED ISSUE #474: Support default values in Metadata when it lands
- Description and tracking issue https://github.com/paritytech/substrate/issues/1256
- Implementation in https://github.com/paritytech/substrate/pull/1264

(Assigning @amaurymartiny since he has been looking at it and playing with it already)

amaurymartiny:
- Fixes #232 
- Closes https://github.com/polkadot-js/api/issues/359
- Closes #474

Emeric's PR isn't quite finished yet, so this PR will hang here for a little while.

TODO:
- [x] Wait for LockPeriod and Vote types to be added.
- [x] Update latest.json with latest decoded metadata


amaurymartiny:
Yes, I'll think about a way to do that.

Issue #478, opened 2018-12-15
api.tx.send with subscription to provide associated events


jacogr:
(Needs https://github.com/polkadot-js/api/issues/475)

- submitAndWatch
- track status events
- upon completion (non-Finalization), end
- upon Finalization, track events in next steps
- get block, check for extrinsic inclusion (against data)
- if found, pull system.events for that block
- return events as part of the subscription, i.e. callback becomes `(status: ExtsinsicStatus, events?: Events)` as opposed to the current `(status: ExtinsicStatus)`

Promise here is easy, I'm not convinced about the RxJS version, i.e. don't want to change the signature to `[ExtrinsicStatus, Vector]` there - not 100% sure how to deal with allowing `.subscribe((status: ExtsinsicStatus, events?: Events) => {...})` there

jacogr:
Alternatively - only return ExitrinsicStatus, add another type (e.g. FinalisedBlock) that has the events embedded in the type.

PR #493, opened 2018-12-19
Retrieve events as part of send

jacogr:
- Related to https://github.com/polkadot-js/api/issues/478
- Only promise atm, to see what it actually needs to do
- Confirmed working (against branch with Tomasz's status fix in)

TODO:

- Once we have "derive infrastructure", adapt the logic to that
- Once happy, adapt to RxJs & api-observable - this rather is more-or-less how it looks instead of "everything done"


packages/api/src/promise/SubmittableExtrinsic.ts
```diff
@@ -22,17 +23,40 @@ export default class SubmittableExtrinsic extends Extrinsic {
...
   }
 
+  private checkStatus (statusCb: (result: SendResult) => any): (status: ExtrinsicStatus) => Promise<void> {
+    return async (status: ExtrinsicStatus): Promise<void> => {
+      let events: Array<any> | undefined = undefined;
+
+      if (status.type === 'Finalised') {
+        const blockHash = status.value as Hash;
+        const { block: { extrinsics } }: SignedBlock = await this._api.rpc.chain.getBlock(blockHash);
+        const allEvents: Array<EventRecord> = await this._api.query.system.events.at(blockHash) as any;
```
amaurymartiny: Why as any?

packages/api/src/promise/SubmittableExtrinsic.ts
```diff
@@ -22,17 +23,40 @@ export default class SubmittableExtrinsic extends Extrinsic {
...
+
+        const index = extrinsics
+          .map((ext) => ext.hash.toHex())
+          .indexOf(this.hash.toHex());
+
+        if (index !== -1) {
+          events = allEvents.filter(({ phase }) =>
+            phase.type === 'ApplyExtrinsic' && (phase.value as u32).eqn(index)
```
amaurymartiny: Why not only `(phase.value as u32).eqn(index)`

packages/api/src/promise/SubmittableExtrinsic.ts
jacogr: the `at() as Array<Type>` here complains - for some obscure reason.

packages/api/src/promise/SubmittableExtrinsic.ts
jacogr: in `Finalisation` we don't have the value, e.g. https://github.com/polkadot-js/api/blob/master/packages/types/src/EventRecord.ts#L24 it is only on the `ApplyExtrinsic`

packages/api/src/promise/SubmittableExtrinsic.ts
jacogr: Ahhh, that may be the issue, that fact I completely missed - ok, will retry with that approach before merging.

jacogr: Brackets be damned -

```
Conversion of type 'Codec | null | undefined' to type 'EventRecord[]' may be a mistake because neither type sufficiently overlaps with the other.
```

So it is since the return of at does not include Array. (Casting to Vector will probably do it)

PR #509, opened 2018-12-21
Let static use the RPC response json directly

### LINKED ISSUE #506: Remove static.ts
This file: https://github.com/polkadot-js/api/blob/master/packages/types/src/Metadata/static.ts

Its content is exactly the same as latest.substrate.json, we don't need them both.

amaurymartiny:
closes #506


jacogr:
Ok, we just need to copy & bump dev, currently the *.json files are not included in npm, so this will fail from the registry -  around here somewhere https://github.com/polkadot-js/dev/blob/master/packages/dev/scripts/polkadot-dev-build-ts.sh#L28

PR #530, opened 2019-01-04
Remove api-observable

jacogr:


jacogr:
Yeap, that is actually only the relevant stuff - the rest are 2017-2018 -> 2017-2019 and removing the whole folder.

PR #540, opened 2019-01-07
Add derive functions

amaurymartiny:
Add `api-derive` package, get host derived functions.

TODO in this PR:
- [x] Add 3-4 more functions, esp. those that are combined (i.e. a derived one that calls another derived one). I don't want to overload this PR either.
- [x] Add memoization

### DIFF
--- a/packages/api-derive/src/chain/bestNumber.ts
+++ b/packages/api-derive/src/chain/bestNumber.ts
@@ -1,0 +1,17 @@
+// Copyright 2017-2019 @polkadot/ui-react-rx authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import ApiRx from '@polkadot/api/rx';
+import { BlockNumber, Header } from '@polkadot/types/index';
+import { Observable } from 'rxjs';
+import { filter, map } from 'rxjs/operators';
+
+export function bestNumber (api: ApiRx) {
+  return (): Observable<BlockNumber> =>
+    api.rpc.chain.subscribeNewHead()
+      .pipe(
+        filter((header: Header) => header && !!header.blockNumber),
+        map(({ blockNumber }: Header) => blockNumber)
+      );
+}
--- a/packages/api-derive/src/chain/bestNumber.ts
+++ b/packages/api-derive/src/chain/bestNumber.ts
@@ -2,10 +2,10 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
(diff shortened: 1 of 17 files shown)


packages/api-derive/src/chain/bestNumber.ts
```diff
@@ -0,0 +1,20 @@
...
+import ApiRx from '@polkadot/api/rx';
+import { BlockNumber, Header } from '@polkadot/types/index';
+import { Observable } from 'rxjs';
+import { filter, map } from 'rxjs/operators';
+
+/**
+ * Get the latest block number.
+ */
+export function bestNumber (api: ApiRx) {
+  return (): Observable<BlockNumber> =>
```
amaurymartiny: I went this way so that it's a pure currified function, should be easy to memoize. However this won't work that well with typedoc.

jacogr:
I'll have the apps stuff merged soon-ish (just testing the rx change in there). That should make the existing derived functions accessible.

packages/api-derive/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  [P in keyof T]: ReturnType<T[P]>
+};
+
+export interface Derive {
+  chain: ReturnTypes<typeof chain>;
+  session: ReturnTypes<typeof session>;
+}
+
+export default function decorateDerive (api: ApiRx): Derive {
+  const derive: Partial<Derive> = {};
```
jacogr: This could just be `const derive = {} as Derive` (and would allow you to get rid of the cast on return)

packages/api-derive/src/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  Object.keys(functions).forEach((sectionName: string) => {
+    const section = functions[sectionName as keyof Derive];
+    derive[sectionName as keyof Derive] = Object.keys(section).reduce((result, methodName) => {
+      // Create cache for the section_method function
+      // @ts-ignore No idea how to make this work...
```
jacogr: Yeah... good luck with that...

packages/api-derive/src/session/sessionProgress.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright 2017-2019 @polkadot/ui-react-rx authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import ApiRx from '@polkadot/api/rx';
```
jacogr: We generally have `@polkadot/*` as the last imports elsewhere. I'm _not_ against this, since it is alphabetical. *sigh* Maybe should just follow this everywhere (and then be able to use order plugins as well)

PR #547, opened 2019-01-08
Support and return default values

### LINKED ISSUE #510: Decode the default values using actual type
The current implementation of metadata decoding decodes the `default` field of metadata storage functions assuming that they are `vec<u8>`, which they are.

But the `vec<u8>` actually represents something, so an additional layer of decoding needs to be done. Basically decode this `vec<u8>` using the `type` of the storage.

This additional decoding has been commented on purpose for now:
https://github.com/polkadot-js/api/blob/master/packages/types/src/Metadata/Modules.ts#L221-L230. It should be enabled when needed.

jacogr:
Closes #510


jacogr:
Ok, I would need to find those. I have done e2e tests and everything _seemed_ fine - not 100% how to get things comprehensive here.

amaurymartiny:
I think it was Proposal. Or anything that inflectively needed Metadata

amaurymartiny:
Idea: as a test, we could add that for all defaults in Metadata, `new CorrespondingType(default_value)` should not throw?

PR #557, opened 2019-01-09
Publishreplay(1) for all derive functions

amaurymartiny:


packages/api-derive/src/util/drr.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright 2015-2018 Parity Technologies (UK) Ltd.
```
jacogr: erk :)

PR #566, opened 2019-01-10
Refactor api-promise/rx

### LINKED ISSUE #559: ApiPromise to use ApiRx internally
Started in https://github.com/polkadot-js/api/pull/539, complete the loop

amaurymartiny:
- Goes against #560.
- Closes #559 

The idea is that:
- ApiBase holds one `_rpcBase/_rpcRx` (these 2 should be combined together).
- ApiBase decorates `this.rpc`, `this.query` and `this.extrinsic` using `_rpxRx` (so internally using Observables)
- But the return type of `this.rpc.*`, `this.query.*` and `this.extrinsic.*` is generic, called OnCall
- ApiBase has an abstract method called `onCall()`, which is basically takes in an Observable, and returns an OnCall

So for impl:
- ApiRx extends ApiBase, `OnCall=Observable<Codec>`, and `onCall()` simply returns the Observable
- ApiPromise extends ApiBase, `OnCall=Promise<Codec> | UnsubFunction`, and `onCall()` does either a `.toPromise()`, or a `.subscribe(cb)`

Good stuff:
- no need to decorate twice
- ApiRx and ApiPromise are much light, and it's easy to extend to other interfaces (like generators)

Bad stuff:
- Does it work?

TODO:
- [x] Make extrinsics work
- [x] Add derive in ApiBase


jacogr:
Yes! Much better than my attempts at making it sane - I thought this needed a completely fresh pair of eyes...

amaurymartiny:
After playing around with this a bit more, I don't think it's actually possible to make this method work.

The reason is:
- api-derive needs ApiRx.
- So if I instantiate ApiPromise, I **need** to instantiate ApiRx
- So there are actually only 2 easy ways: ApiPromise extends ApiRx (hard to do imo), or ApiPromise holds ApiRx. After digging this method I constantly came back to this point.

Replaced by #585.

Issue #588, opened 2019-01-15
Split indices from balances in api.derive.* endpoints


jacogr:
- Once https://github.com/paritytech/substrate/pull/1404 is available
- move the account <-> id mappings into api.derive.indices
- fallback to balances (for now) when indices not available (do check against endpoints for calls, detect which version to use)

jacogr:
It has been merged into substrate master.

PR #596, opened 2019-01-17
Allow for balance/indices lookups

jacogr:
- Closes https://github.com/polkadot-js/api/issues/588
- Closes https://github.com/polkadot-js/api/issues/597
- split out balances/indices (tested as backwards compatible with old & new)
- add derive.accounts.* for lookups
- unintended bonus - cater for new lookup metadata types


jacogr:
Just need to update tests - compact now in transfer, so have some juggling.

PR #608, opened 2019-01-20
fix peerId mapping

xlc:
it will be good if the mapping can be auto inferred (e.g. by help of `_.camelCase`)


xlc:
Do you mean there are cases where the output is not snake case? If that's the case I think it make sense to fix it on the substrate side to ensure the naming consistency, instead of this case-by-case naming mapping which is just error-prone.

jacogr:
Yes, there is a mixture of `snake_case` and `camelCase`. (Generally the JSON, historically is `camelCase` which is done with a serde flag, but it is not applied consistently and has been ignored for newer RPC changes.)

Issue #569, opened 2019-01-11
Docs: Add examples for ApiRx to docs and fix existing examples


Stefie:
- [x] Add a chapter for ApiRx examples like https://polkadot.js.org/api/examples/promise/
- [x] Go through examples in https://polkadot.js.org/api/api/classes/_rx_index_.apirx.html and update them to work with current API, polkadot and substrate versions

PR #616, opened 2019-01-22
[Docs] Go through and update existing ApiRx examples

Stefie:
Last fixes for #569


Stefie:
That's fine. I just accidentally pushed the last commit here

PR #623, opened 2019-01-23
[Docs] Re-applied lost changes from PR #616

Stefie:
Update existing broken ApiRx examples
- add missing imports
- replace subscribe inside subscribe syntax  with switchMap
- smaller fixes

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -206,8 +206,8 @@ export default abstract class ApiBase<OnCall> implements ApiBaseInterface<OnCall
    * api.tx.balances
    *   .transfer(<recipientId>, <balance>)
    *   .sign(<keyPair>, <accountNonce>, <blockHash (optional)>)
-   *   .send((status) => {
-   *     console.log('tx status', status);
+   *   .send(({status}) => {
+   *     console.log('tx status', status.asFinalised.toHex());
    *   });
    * ```
    */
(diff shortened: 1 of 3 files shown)


packages/api/src/Base.ts
```diff
@@ -206,8 +206,8 @@ export default abstract class ApiBase<OnCall> implements ApiBaseInterface<OnCall
    * api.tx.balances
    *   .transfer(<recipientId>, <balance>)
    *   .sign(<keyPair>, <accountNonce>, <blockHash (optional)>)
-   *   .send((status) => {
-   *     console.log('tx status', status);
+   *   .send(({status}) => {
```
jacogr: I would almost go `signAndSend` here, keeps it simple.

jacogr: For 99% of the cases `signAndSend` does the trick. The only area where it is problematic, is sending bulk transactions in a loop since you need to increment the nonce between them.

Stefie: Ah, the problem I ran into....  
In my case I just did 2 different transactions from the same account within the 4sec block period and the first one never got through. That's not such an uncommon use-case.

So with signAndSend we can only make one transfer per account per block period. I think that's a little unfortunate.

Stefie: (Or at least worth documenting)

Stefie: Nevermind... I just realized that we have the same behaviour with sign().send() because also here we're only getting a different accountNonce with the next block update

PR #630, opened 2019-01-23
Add query.<sec>.<met>.{size, hash}(arg? any)

jacogr:
Additional storage helpers to retrieve the hash of a entry value and the size of the entry value

Where I am not happy with this is the fact that size returns `U64` (and it is defined as `Codec`) and hash returns `Hash` (once again defined as `Codec`)


jacogr:
Yeap, I saw that, hence the comment.

PR #631, opened 2019-01-23
Revert "Bubble isSubscription from rpc-rx (#627)"

jacogr:
This reverts commit 767a1978d03e0b4634c001f08b8d5ce8372f7f96.

### DIFF
--- a/packages/api/src/promise/index.ts
+++ b/packages/api/src/promise/index.ts
@@ -14,10 +14,7 @@ import { isFunction } from '@polkadot/util';
 import ApiBase from '../Base';
 import Combinator, { CombinatorCallback, CombinatorFunction } from './Combinator';
 
-interface RxFn {
-  (...params: Array<any>): Observable<Codec | undefined | null>;
-  isSubscription: boolean;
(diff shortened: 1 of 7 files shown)


packages/api/src/promise/index.ts
```diff
@@ -196,8 +193,8 @@ export default class ApiPromise extends ApiBase<OnCall> implements ApiPromiseInt
     };
   }
 
-  protected onCall (method: RxFn, params: Array<any>, isSubscription: boolean = false): OnCall {
-    if (!params || params.length === 0 || !isSubscription) {
+  protected onCall (method: RxFn, params: Array<any>, isSubscription?: boolean): OnCall {
+    if (!params || params.length === 0 || isSubscription === false) {
```
jacogr: Ok, I believe it is here - since we don't actually pass `isSubscription` through everywhere, it defaults to `true` - basically what I'm seeing is that stuff doesn't actually create subscriptions. (And yes, this was an after-the-tests "o, that looks good", erm, _improvement_)

Issue #590, opened 2019-01-15
Support new metadata formatting (when available)


jacogr:
- When this gets merged - https://github.com/paritytech/substrate/pull/1428
- This breaks backwards compat, so treat with care (and needs a changelog entry)

PR #632, opened 2019-01-24
Support new versioned metadata format

jacogr:
Closes https://github.com/polkadot-js/api/issues/590


packages/types/src/Metadata/v1/index.ts
```diff
@@ -0,0 +1,14 @@
...
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import Struct from '../../codec/Struct';
+import MagicNumber from './MagicNumber';
+
+export default class MetadataV1 extends Struct {
+  constructor (value?: any) {
+    super({
+      magicNumber: MagicNumber
```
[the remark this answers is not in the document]
jacogr: Yeap... basically if you look at the current, it is basically a proxy parser - try one, then the other. All the types passed through are still for V0. What I will probably do as a first stab is to 

- get the parsing for both formats going (this is where the magic number is used to detect)
- define the V1 types
- have a conversion tool from V1 -> v0 - so basically we still use V0-structures everywhere

Then once we have these rolled out can remove support for V0 (`Null` in the enum), remove the `try/catch` on parsing (all these are one-liners) and then get all the storage, extrinsics, etc. to use the V1 format natively

The latter part, since I need to support both formats and need a conversion tool, will leave out from this PR - it will just add more moving parts.

packages/api/src/Base.ts
```diff
@@ -300,8 +300,8 @@ export default abstract class ApiBase<OnCall> implements ApiBaseInterface<OnCall
       this._runtimeVersion = await this._rpcBase.chain.getRuntimeVersion();
       this._genesisHash = await this._rpcBase.chain.getBlockHash(0);
 
-      const extrinsics = extrinsicsFromMeta(this.runtimeMetadata);
-      const storage = storageFromMeta(this.runtimeMetadata);
+      const extrinsics = extrinsicsFromMeta(this.runtimeMetadata.asV0);
```
amaurymartiny: Question: if i understand correctly, each time we call .asV0, it will recalculate v1ToV0 function (a bit heavy), which might not be super optimized.

Though if v0 is going away soon, I prefer not to forget about this optimization.

jacogr: There is a cache for the v1 -> v0 conversion - https://github.com/polkadot-js/api/pull/632/files#diff-a208134f9b85707425c723f3705ebb29R86

jacogr: I think this type of "convert once" thing will probably follow-around each time we have a bump.

amaurymartiny: Ah correct

PR #638, opened 2019-01-26
derive.chain.* allows for non-available session

jacogr:
Detect defaults in override so the call returns even without `query.session` - should address issues such at this - https://github.com/polkadot-js/apps/issues/651

Basically for `chain.{getHeader,subscribeNewHead}` we do an existence check for `query.storage`, if not available, we return an empty array observable for validators instead of failing.


jacogr:
Small enough, merging it to have a full-check across the UI with the API

PR #650, opened 2019-01-29
Codec types (call definitions in API)

jacogr:
- Closes https://github.com/polkadot-js/api/issues/649

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -8,7 +8,7 @@ import { Codec, CodecArg, CodecCallback } from '@polkadot/types/types';
 import { RxResult } from './rx/types';
 import {
   ApiBaseInterface, ApiInterface$Rx, ApiInterface$Events, ApiOptions,
-  DecoratedRpc, DecoratedRpc$Section,
+  DecoratedRpc, DecoratedRpc$Method, DecoratedRpc$Section,
   Derive, DeriveSection,
   HashResult, U64Result,
   OnCallDefinition, OnCallFunction,
@@ -267,7 +267,7 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
     return this;
   }
 
-  protected abstract onCall (method: OnCallFunction<RxResult, RxResult>, params?: Array<CodecArg>, callback?: CodecCallback): CodecResult | SubscriptionResult;
+  protected abstract onCall (method: OnCallFunction<RxResult, RxResult>, params?: Array<CodecArg>, callback?: CodecCallback, needsCallback?: boolean): CodecResult | SubscriptionResult;
 
   private emit (type: ApiInterface$Events, ...args: Array<any>): void {
     this._eventemitter.emit(type, ...args);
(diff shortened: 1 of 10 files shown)


packages/api/src/types.ts
```diff
@@ -5,79 +5,87 @@
...
 }
 
-interface QueryableStorageFunctionBase<OnCall> extends StorageFunction {
-  (arg?: any): OnCall;
-  at: (hash: Uint8Array | string, arg?: any) => OnCall;
-  hash: (arg?: any) => OnCall; // Observable<Hash>
-  key: (arg?: any) => string;
-  size: (arg?: any) => OnCall; // Observable<U64>
+export type HashResult<CodecResult, SubscriptionResult> =
+CodecResult extends Observable<any>
```
amaurymartiny: weird indentation

packages/api/src/rx/index.ts
```diff
@@ -190,7 +190,7 @@ export default class ApiRx extends ApiBase<OnCall> implements ApiRxInterface {
     return this._isReady;
   }
 
-  protected onCall (method: OnCallFunction<Observable<Codec | undefined | null>>, params: Array<any>): OnCall {
+  protected onCall (method: OnCallFunction<RxResult, RxResult>, params: Array<CodecArg> = []): RxResult {
```
amaurymartiny: Since we're assuming everywhere that there's only one param, for consistency, we could only put `param: CodecArg` here?

packages/api/src/Base.ts
```diff
@@ -338,122 +339,115 @@ export default abstract class ApiBase<OnCall> implements ApiBaseInterface<OnCall
...
+  private decorateRpc<C, S> (rpc: RpcRx, onCall: OnCallDefinition<C, S>): DecoratedRpc<C, S> {
     return ['author', 'chain', 'state', 'system'].reduce((result, _sectionName) => {
-      const sectionName = _sectionName as keyof DecoratedRpc<T>;
+      const sectionName = _sectionName as keyof DecoratedRpc<C, S>;
 
       result[sectionName] = Object.keys(rpc[sectionName]).reduce((section, methodName) => {
         // FIXME Find a better way to know if a particular method is a subscription or not
-        const isSubscription = methodName.includes('subscribe');
-        const method = (...params: any[]) => onCall(rpc[sectionName][methodName], params, isSubscription);
+        // const isSubscription = methodName.includes('subscribe');
```
amaurymartiny: if you remove the isSubscription information,`apiPromise.rpc.state.subscribeStorage(key)` will return a Promise.

Or are we waiting for the isSubscription bubbling from rpc-core?

packages/api/src/Base.ts
```diff
@@ -338,122 +339,115 @@ export default abstract class ApiBase<OnCall> implements ApiBaseInterface<OnCall
...
-    const decorated: any = (...args: any): T =>
-      onCall(
-        (...args: any) => this._rpcRx.state
-          .subscribeStorage([[method, args[0]]])
+  private decorateStorageEntry<C, S> (method: StorageFunction, onCall: OnCallDefinition<C, S>): QueryableStorageFunction<C, S> {
+    // These signatures are allowed and exposed here -
+    //   (arg?: CodecArg): CodecResult;
+    //   (arg: CodecArg, callback: CodecCallback): SubscriptionResult;
+    //   (callback: CodecCallback): SubscriptionResult;
+    const decorated = ((...args: Array<any>): C | S => {
```
amaurymartiny: imo `decorated` should have a better type, its type is the type the end-user would see. So exactly the type you put in the comments, but inside an interface

packages/api/src/types.ts
jacogr: Umpf.

packages/api/src/rx/index.ts
jacogr: onCall is used for anything, i.e. only storage has 1 param, derive, rpc, etc. have multiples. (Still need to deal with that is some way in those typings...)

packages/api/src/Base.ts
jacogr: Ummm, you are right.

We actually need a `needsSubscription` flag here. i.e. if set and callback is not provided, throw.

packages/api/src/rx/index.ts
amaurymartiny: Ah, correct.

jacogr: ... in the future, will probably do the `arg: CodecArg, arg: CodecArg, ...` dance everywhere else, here we can forgo it since it is internal. ()

packages/api/src/Base.ts
jacogr: Ok, added a `needsCallback` that does a check in `onCall` (promise), at least we have a runtime error.

Along with this, update RPC to have proper-ish params as well. Derive is till missing, it is all any, any, any

PR #652, opened 2019-01-30
Option unwrapOr helper

jacogr:
cc @jnaviask


jacogr:
Converted to `unwrapOr` (similar to Rust) based on a suggestion by @amaurymartin

Issue #654, opened 2019-01-30
Payload transaction (hash) signing


jacogr:
Related to https://github.com/paritytech/substrate/pull/1624

This needs to be coordinated with the testnet rollouts - it breaks everything.

Issue #626, opened 2019-01-23
Extend extrinsic signAndSend with immortal, era, nonce


jacogr:
I'm suggesting that instead of `sign(pair, nonce, era)` we pass through `sign({ pair, nonce, era })`. (Need to look into this exactly)

Currently we don't allow the nice `signAndSend` shortcut to set the eras at all. So it is driven from that perspective. So `signAndSend(pair)` can still work as-is. `signAndSend({ pair, ... })` can actually pass in `nonce` (optional, can be detected), `era` (optional) and `blockHash` (optional, generis or detected) then as well.

I would almost suggest that the `signAndSend` default era be set - not sure how many blocks to allow. So by default we actually do mortal instead of immortal.

PR #655, opened 2019-01-31
Cater for signature over hash (with version checks)

jacogr:
- Closes https://github.com/polkadot-js/api/issues/654
- Closes https://github.com/polkadot-js/api/issues/626
- Still needs a bump, breaking change in the signatures (for `sign` on `Extrinsic`, `sign` and `signAndSend` overloaded in Submittables)


packages/api/test/e2e/promise-tx.spec.js
```diff
@@ -26,7 +26,31 @@ describe.skip('e2e transactions', () => {
...
+          });
+
+          if (events.length) {
+            done();
+          }
+        }
+      });
+  });
+
+  t('makes a transfer (sign, then send - compat version)', async (done) => {
```
amaurymartiny: it

packages/api/src/SubmittableExtrinsic.ts
```diff
@@ -3,15 +3,25 @@
...
 import filterEvents from './util/filterEvents';
 
+type SignatureOptionsPartial = {
+  blockHash?: AnyU8a,
+  era?: Uint8Array,
+  nonce: AnyNumber,
+  version?: RuntimeVersion
+};
+
+type StatusCb = (result: SubmittableSendResult) => any;
```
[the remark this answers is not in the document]
jacogr: So, the issue I always run into, making the callback async, which then returns a Promise. So then start adding `Promise<void> | void`. Generally for these, really don't care what the recipient does with the info provided or how he/she gets to it - hence `any`.

packages/api/test/e2e/promise-tx.spec.js
jacogr: Cool, testing just started, so actually just picked that one up.

Issue #660, opened 2019-02-01
Add Signer to Api


ianhe8x:
As a dapp, it doesn't always have access to key pair. Key pairs could be managed in a browser extension or hardware wallet. So I propose to introduce signer to api and let it do the signing. This will allow our browser extension to work with the api by exposing a signer into browser's global.

```
export interface Signer {
  sign(extrinsic: Extrinsic, opt: {from: string} & SignatureOptions): Promise<void>;
}
```

signAndSend is the place to use the signer. I prefer to changing the signature to below which looks more neat
```
  signAndSend (options: SignAndSendOptions)
  signAndSend (options: SignAndSendOptions, statusCb: StatusCb)
  signAndSend (options: SignAndSendOptions, statusCb?: StatusCb)

interface SignAndSendOptions extends Partial<SignatureOptions> {
  from: string | KeyringPair
}
```
though to keep backward compatible, it could also be like:
```
signAndSend (account: KeyringPair | string, _options?: Partial<SignatureOptionsPartial> | StatusCb, statusCb?: StatusCb)
```

I would like to hear from you if it's the right direction before submitting my PR.

jacogr:
Fort going to comment on the signatures: they certainly a bit messy atm, we are not quite happy, but it is moving towards being more useful. We just always try to keep backwards compat, so happy with changes that cleans up the code as long as it does not affect downstream users. 

Honestly, have not considered the signers up to now, which is certainly a gap. (The apps UI manages this by putting stuff on a queue when sent)

So happy to look at any suggestions in this area, it is certainly something that needs to be addressed. 

Would almost suggest leaving the existing (for keypair managed node), adding clean sigs for the new signer and then as we go deprecating the current. (Since the current can be done with a new signer interface). Also so keep the sign/send split as separate steps in mind. (It is useful in some cases)

TL;DR good idea, we certainly need support sooner rather than later

jacogr:
We could keep the signature as you suggested as a last point - would just allow AccountId and Address as well. That would work quite well, is nicely compatible and named as well.

jacogr:
Closed in #662

PR #662, opened 2019-02-04
add Signer to api

ianhe8x:
see #660

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -13,7 +13,7 @@ import {
   HashResult, U64Result,
   OnCallDefinition, OnCallFunction,
   QueryableModuleStorage, QueryableStorage, QueryableStorageFunction,
-  SubmittableExtrinsicFunction, SubmittableExtrinsics, SubmittableModuleExtrinsics
+  SubmittableExtrinsicFunction, SubmittableExtrinsics, SubmittableModuleExtrinsics, Signer
 } from './types';
 
 import EventEmitter from 'eventemitter3';
@@ -148,6 +148,13 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
    */
   get type (): ApiType {
     return this._type;
+  }
+
+  /**
+   * @description Set an external signer which will be used to sign extrinsic when account passed in is not KeyringPair
+   */
+  setSigner(signer: Signer) {
+    this._rx.signer = signer;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/api/test/util/SimpleSigner.js
```diff
@@ -0,0 +1,13 @@
+
+export default class SimpleSigner{
```
jacogr: I would suggest renaming this one to `SingleAccountSigner` or just change it slightly to take in a keyring instance and then doing `keyring.getPair(options.from)` (Which is what I assumed the `SimpleSigner` to be, but maybe that is rather a `KeyringSigner` - the above suggested naming I would probably apply here anyway, just to be explicit)

packages/api/src/SubmittableExtrinsic.ts
```diff
@@ -163,25 +163,40 @@ export default class SubmittableExtrinsic<CodecResult, SubscriptionResult> exten
...
+      ).pipe(
+        first(),
+        mergeMap(async (nonce) => {
+          if(isKeyringPair){
+            this.sign(account as KeyringPair, { ...options, nonce });
+          }else{
+            if(!this._api.signer){
+              throw new Error('no signer exists');
+            }
+            await this._api.signer.sign(this, {
```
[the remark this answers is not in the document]
ianhe8x: I agree that making signer interface `sign(extrinsic, address, options)` keeps kind of consistency with signAndSend. And I can try extracting the logic of expanding options into a build function and let's see if it improves.

packages/api/test/util/SimpleSigner.js
ianhe8x: Just a reminder that this class is only supposed to be used in test. I don't think we need to ship a implementation of Signer with api. (From your comments I feel you think this is a exported class.)

packages/api/src/Base.ts
```diff
@@ -101,6 +101,7 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
     this._rpcRx = new RpcRx(options.provider);
     this._rpc = this.decorateRpc(this._rpcRx, this.onCall);
     this._rx.rpc = this.decorateRpc(this._rpcRx, rxOnCall);
+    this._rx.signer = options.signer;
```
amaurymartiny: imo signer has nothing to do with rx, so it shouldn't be in rx. I would propose `this._signer = options.signer` here

ianhe8x: it is more like `_rx` is not a proper name, because it's the api object that is passed to SubmittableExtrinsic.

ianhe8x:
@jacogr when can this pr get merged?

jacogr:
Sorry, they have been heaping up a bit, all of them. (I have been a bottleneck this week)

PR #663, opened 2019-02-05
Alias state_getRuntimeVersion

jacogr:
Introduced in https://github.com/paritytech/substrate/pull/1679

Will only swap usage away from the original (chain_ version), once fully rolled-out, then deprecate and swap usage. (As normally the case with RPC.) 

As it stands, due to https://github.com/paritytech/substrate/issues/1683 either version is not working.

### DIFF
--- a/packages/type-jsonrpc/src/chain.ts
+++ b/packages/type-jsonrpc/src/chain.ts
@@ -40,7 +40,7 @@ const getFinalisedHead: RpcMethodOpt = {
 };
 
 const getRuntimeVersion: RpcMethodOpt = {
-  description: 'Get the runtime version',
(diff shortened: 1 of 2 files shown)


ascjones:
@jacogr can you confirm that this works now with my fix for https://github.com/paritytech/substrate/issues/1683? It's been merged into master.

jacogr:
On quite limited time since I am away this week. However, will pull down tonight, build and take a peek and confirm. (Just cannot do so immediately)

As long as curl works, it will work.

ascjones:
No worries - curl does work so should be good :+1:

jacogr:
@ascjones Commented in the Substrate issue, the jsonrpc fix resurrects the endpoints and have not seen any issues with quick testing on the apps UI, excellent, thank you.

Issue #661, opened 2019-02-03
No easy / clean way to close ws connection and showdown Api object


xlc:
This prevents to node.js program exit cleanly because there is no disconnect / close method so `process.exit` have to be used.

`WSProviderInterface` should have a `disconnect` method to close ws connection

In additional to that, I think following changes can be helpful (not very confident so feel free to ignore those suggestions)

Move `connect` method to `ProviderInterface` and have `disconnect` on there as well

Add `open` and `close` to `ApiBase` method which does connect & setup and disconnect & teardown.

----

Current we are using `provider.websocket.close()` which logs 

```
API-WS: disconnected from ws://127.0.0.1:9944::1000: Normal connection closure
```

jacogr:
Yes. Certainly an oversight.

I believe both needs to be in place - would just prioritise the shutdown before exposing the open explicitly. 

 (With the calling of metadata and setup on open it is a bit more involved to het 100%)

PR #664, opened 2019-02-05
Expose api.disconnect()

jacogr:
Closes https://github.com/polkadot-js/api/issues/661 (disconnect only)

As it stands untested, would need to self-review properly as well.

### DIFF
--- a/packages/rpc-provider/src/http/index.ts
+++ b/packages/rpc-provider/src/http/index.ts
@@ -60,7 +60,7 @@ export default class HttpProvider implements ProviderInterface {
    * @description Manually disconnect from the connection
    */
   disconnect (): void {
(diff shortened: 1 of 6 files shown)


packages/rpc-provider/src/http/index.ts
```diff
@@ -56,6 +56,13 @@ export default class HttpProvider implements ProviderInterface {
     return false;
   }
 
+  /**
+   * @description Manually disconnect from the connection
+   */
+  disconnect (): void {
+    throw new Error('Unimplemented');
```
xlc: I think a nop is fine in the sense there are no ongoing connections to disconnect?
Unless this suppose to disconnect any outgoing requests that have not responded yet?

PR #668, opened 2019-02-07
Add Api{Promise,Rx}.clone() - new instance, new transport

jacogr:
This creates a clone of the underlying transport (new connection) as well as a new API instance that pulls certain values from the original such as runtime info and genesis info. (Assuming the source instance is ready)

This stems from https://github.com/polkadot-js/apps/issues/675 - basically for that use-case we can create an instance that mimics the main, has an own connection and when done with use thereof dispose of it (most probably via https://github.com/polkadot-js/api/pull/664)


jacogr:
I was in 2 minds - originally started with the approach ‘new Api({ ...oldOptions, provider: oldProbider.clone() })’ approach.

 It felt saner to do the provider cloning in the base and have some control to not do the await queries again. (And this approach means that base needs to have some way of extracting info from the original)

PR #672, opened 2019-02-08
[Docs] Add info on how to add custom types to API

Stefie:
Gautam had the problem on wednesday, that his custom types of his custom extrinsics weren't recognized. Solution was provided by @amaurymartiny , I just added an example to the docs.


packages/api/README.md
```diff
@@ -4,7 +4,7 @@ The Polkadot-JS API provides easy-to-use wrappers around JSONRPC calls that flow
 
 The API wrappers provide a standard interface for use -
 
-- A static `.create(<optional WsProvider>)` that returns an API instance when connected, decorated and ready-to use
+- A static `.create(<optional WsProvider>, <optional CustomTypes>)` that returns an API instance when connected, decorated and ready-to use
```
amaurymartiny: There's only one argument to `create`, which is either WsProvider, or ApiOptions. Not sure what's the best way to show it concisely here though

Stefie: This should be `.create({ provider:<optional WsProvider>, types: <optional CustomTypes>})`when both parameters are passed...

Stefie: Ah, see my comment below. I thought you could pass it as an object if you're passing both.Is that intentional or just an open todo?

amaurymartiny: It's mainly for backward-compatibility.

In the beginning, it was `.create(WsProvider)`. Then, we added ApiOptions so that you could either use the old syntax, or pass `.create(ApiOptions)`, where `ApiOptions = {wsProvider: ..., types: ...}`

jacogr: I would just stick with options documentation. (The provider is only for backwards-compat, no need to document that, I'd rather not have it documented to stop use thereof)

Stefie: So, for clarification: can you do:
`create( { wsProvider: <WsProvider>, types: <CustomTypes> } )` ?

https://github.com/polkadot-js/api/blob/b122f0cde099899df2c5e17018d0d4f1efa13d6c/packages/api/src/types.ts#L96

jacogr: 100%

packages/api/README.md
```diff
@@ -4,8 +4,8 @@ The Polkadot-JS API provides easy-to-use wrappers around JSONRPC calls that flow
 
 The API wrappers provide a standard interface for use -
 
-- A static `.create(<optional WsProvider>, <optional CustomTypes>)` that returns an API instance when connected, decorated and ready-to use.
-- The above is just a wrapper for `new Api(<optional WsProvider>) `, exposing the `isReady` getter
+- A static `.create(<optional ApiOptions>)` that returns an API instance when connected, decorated and ready-to use. ApiOptions can include an optional WsProvider and optional custom type definitions `{ wsProvider: <Optional WsProvider>, types: <Optional CustomTypes> }`.
```
jacogr: { provider: WsProvider, ... }` as per https://github.com/polkadot-js/api/blob/master/packages/api/src/types.ts#L98 (also has the signer now, we can ignore `source` since it is basically "internal", only used in cloning)

Stefie: I fixed the `wsProvider` to `provider`

Stefie:
Can we merge this or will there be significant changes in the near future so we rather shouldn't? Is the code example and the description ok now?

Issue #635, opened 2019-01-25
Remove V0 metadata support once rolled out


jacogr:
V1 and V0 support has been added in https://github.com/polkadot-js/api/pull/632 as a result of https://github.com/polkadot-js/api/issues/590. As support rolls out, we can remove V0 support completely.

Testnets need to be updated -

- [ ] Alexander
- [ ] Charred Cherry

The code needs to be adapted to support V1 natively -

- [ ] metadata `asV0` will throw an error
- [ ] `MetadataVersioned` can remove the V0 fallback
- [ ] `v1/toV0` can be removed
- [ ] `v0/*` can be removed - this is accessed from inside v1, so copy the relevant classes only (V1 is much leaner with less wrappers)
- [ ] storage, extrinsics and events to work natively with V1 structures
- [ ] any `asV0` (which throws), to be converted to `asV1`
- [ ] Update MetadataEnum to wrap `EnumType<Null | V1>`

jacogr:
Closing this. We are refactoring metadata via @amaurymartiny - the reality is that since the Alex testnest starts with V0, we will need support for a while longer. So the multiple support is not going away anytime soon, however we need some serious elbow-grease to make the code maintainable.

PR #678, opened 2019-02-11
MetadataV2 (WIP)

xlc:
JS side for https://github.com/paritytech/substrate/pull/1328

### DIFF
--- a/packages/types/src/Metadata/v2/Calls.ts
+++ b/packages/types/src/Metadata/v2/Calls.ts
@@ -38,9 +38,6 @@ export class MetadataCallArg extends Struct {
 export class MetadataCall extends Struct {
   constructor (value?: any) {
     super({
-      // NOTE Just commented out, it is not available in current, but an open question
-      // around the usefulness of supplying this remains (could be re-added)
-      // id: u16,
       name: Text,
       args: Vector.with(MetadataCallArg),
       docs: Vector.with(Text)
--- a/packages/types/src/Metadata/v2/Calls.ts
+++ b/packages/types/src/Metadata/v2/Calls.ts
@@ -1,0 +1,77 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import Struct from '../../codec/Struct';
+import Vector from '../../codec/Vector';
+import Text from '../../Text';
+import { MetadataName } from './MetadataRegistry';
+
+export class MetadataCallArg extends Struct {
+  constructor (value?: any) {
+    super({
+      name: Text,
+      type: MetadataName
+    }, value);
+  }
+
+  /**
+   * @description The argument name
+   */
+  get name (): Text {
+    return this.get('name') as Text;
+  }
+
+  /**
+   * @description The [[MetadataName]]
+   */
+  get type (): MetadataName {
+    return this.get('type') as MetadataName;
+  }
+}
+
+/**
+ * @name MetadataCall
+ * @description
+ * The definition of a call
+ */
+export class MetadataCall extends Struct {
+  constructor (value?: any) {
+    super({
+      // NOTE Just commented out, it is not available in current, but an open question
+      // around the usefulness of supplying this remains (could be re-added)
+      // id: u16,
+      name: Text,
+      args: Vector.with(MetadataCallArg),
+      docs: Vector.with(Text)
+    }, value);
+  }
+
+  /**
+   * @description The [[MetadataCallArg]] for arguments
+   */
+  get args (): Vector<MetadataCallArg> {
+    return this.get('args') as Vector<MetadataCallArg>;
+  }
+
+  /**
+   * @description The [[Text]] documentation
+   */
+  get docs (): Vector<Text> {
+    return this.get('docs') as Vector<Text>;
+  }
+
+  // /**
+  //  * @description The call function id
+  //  */
+  // get id (): u16 {
+  //   return this.get('id') as u16;
+  // }
+
+  /**
+   * @description The call name
+   */
+  get name (): Text {
+    return this.get('name') as Text;
+  }
+}
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -349,7 +349,7 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
         this._genesisHash = this._options.source.genesisHash;
       }
 
-      const extrinsics = extrinsicsFromMeta(this.runtimeMetadata.asV0);
+      const extrinsics = extrinsicsFromMeta(this.runtimeMetadata);
       const storage = storageFromMeta(this.runtimeMetadata.asV0);
 
       this._extrinsics = this.decorateExtrinsics(extrinsics, this.onCall);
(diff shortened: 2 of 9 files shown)


packages/types/src/Metadata/v2/Calls.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * @name MetadataCall
+ * @description
+ * The definition of a call
+ */
+export class MetadataCall extends Struct {
+  constructor (value?: any) {
+    super({
+      // NOTE Just commented out, it is not available in current, but an open question
```
jacogr: You could remove this now, should actually do so in V1 as well - it is always a chicken-and-egg while these things are ongoing, you first want the JS done before the substrate merge so the whole world doesn't break, so left the comment and never came back to just clean up.

xlc:
@jacogr I found out it is very hard to implement this without breaking backward compatibility. I cannot touch any existing types otherwise it may break codec. This means I cannot add new field to existing class, I have to clone whole bunch of the classes and somehow be able to make everything else works with both new and old version.
Any suggestions? Should I just forgot about keeping backward compatibility?

All I want is replace all `Type` with `MetadataName`, but I can't just replace it...

jacogr:
Backwards compat is actually important, the issue is that there are multiple chains running multiple versions and rollout takes time. (Which is the reason we still support V0 as well and have not yet converted the touchpoints to V1 yet - and the latter is much simpler to actually work with.)

So the approach you have taken here, basically using metadata.version to determine which to construct is actually in the right direction. (And actually doing that for V1 first without the conversion will probably smooth the way considerably, so need a bump for https://github.com/polkadot-js/api/issues/635 - at least for the "work natively with" parts needs to go ASAP)

So what I've looked at -

- the metadata parsing, as horrible as it is, would have a bunch of duplication, basically taking the V1 classes and converting/adding the new fields
- So touching the existing types, I'm assuming something like (taking an example), `Address` or `Vector` - what issues do you have here (Just trying to wrap my head around)

jacogr:
Where I do see issues, or artahre where we have base, but knowingly override -

- https://gist.github.com/xlc/3778f339afb9c1f4557ac9391cd14928#file-gistfile1-txt-L4467-L4485 - here is it is important for the JS side to know (when all is said and done) it is an `AccountId` so the parsing and representation is done properly (e.g. custom toString()` - so that one should never map itself to a simple `Vec<u8>` - which is `Bytes` since we actually want to work with a `Uint8Array` here, not a `Vector.with(u8)`)
- https://gist.github.com/xlc/3778f339afb9c1f4557ac9391cd14928#file-gistfile1-txt-L6805-L6820 - Vote is also a funny one, to make it actually usable end-user wise, we have stuff like https://github.com/polkadot-js/api/pull/677
- Good to see that `Address` is a primitive, https://gist.github.com/xlc/3778f339afb9c1f4557ac9391cd14928#file-gistfile1-txt-L5729-L5735, as is contains either the `Ed15519AccountId` which ties to the first item or an `AccountIndex` (haven't looked this one up)

Most of the types though are just, well simple extensions - see eg. `BlockNumber` it really has no reason to exist as a class (And the bulk of the types fits in here since they are just primitives, the couple of exceptions just needs to be kept in mind)

xlc:
One example that I am want to change but can't is this `FunctionArgumentMetadata.type`
I just want to change it to `MetadataName` so this will become `type: arg.type` https://github.com/polkadot-js/api/pull/678/files#diff-602329a265e76eae65c29a236bf8a0cbR64
But I can't change it, I have to duplicated it.
This means `fromMetadataV2` cannot return the same type `ModulesWithMethods`, it have to return `ModulesWithMethodsV2`, and that means a lot of the codebase have to be duplicated and carefully managed.

jacogr:
Yes, the `fromMetadata` in extrinsics and `storage` is going to be issues. Which is why, as horrible as it is, there is the `v1` -> `V0` conversion - that PR got way out of control and was the only way to do it without duplication at that point. 

(Hence me linking the issue above, I would, not expecting you to, get that "working natively with V1"  ASAP, which will at least pave the way somewhat)

The only light at the end of the tunnel is that the duplication can be removed once there is reasonable upgrade confidence, so it doesn't stay around forever.

jacogr:
@amaurymartiny Would like to hear your thoughts as well

xlc:
`Address` have non-standard encoding method so it have to be primitive. I am not sure about the ` AccountId` issue,  it should be in JS side map `Custom("substrate_primitives::authority_id", "Ed25519AuthorityId")` to `AccountId` unless it is also been used on the context that is not AccountId.

Yeah similar to `Vote`, there are few usability issues. e.g. I don't see there is a `Option.None()` so I don't have a good way to return a `None`. I also want a `Option::map` method similar to the Rust one, which I find hard to implement because it needs to access the constructor of the destination type at runtime. so I can't just `maybeValue.map(x => new NewType(x))`, instead it kind have to be maybeValue.map(NewType, x => new NewType(x))`

jacogr:
`AccountId` is the same as `Address` - `Address` is just an `AccountId` or `AccountIndex`, so the custom encoding actually doesn't sit in `Address`, but rather in the wrapped types.

`Option` does have `isNone` and `isSome`, so it is a bit different, but should be _relatively_ equivalent. On the storage side `Option` is actually a useless construct since it will, when the value is `None` always return the default (as specified) or the type default. So it _always_ has an initialised value.

xlc:
That reminds me an issue I had before. Indices module doesn’t make mush sense for my runtime (we have first class support of user issued token so there isn’t a good way to detect if a user balance is zero for all the tokens) so I want to remove it. That means the type Address and AccountId are identical and we don’t have the wrapped type, it is always AccountId. 
So it will be good to avoid those special logics because it make customization hard.

jacogr:
So the differences between the 2 are that `Address` when encoded as `Uint8Array` has a prefix that denotes the type. So basically, assuming you have an (hex) `AccountId` of `0x123456...`

When encoded as `AccountId` it is `[0x12, 0x34, 0x56, ...]`

When encoded as an `Address` it is `[0xff, 0x12, 0x34, 0x56, ...]`

So basically an `Address` acts like an Enum where the first byte indicates the type. When `0xff` it is `AccountId`, when `<0xef` (I think) it is an 1-byte `AccountIndex`, etc. So dropping indices has no actual impact there at all since the values coming in will always have the `0xff` prefix.

xlc:
But that means one byte overhead per account id in a tx which is what I want to avoid. It can never be AccountIndex so there is no need to have the extra magic byte.

jacogr:
Then just use `AccountId` everywhere for the 32-byte version. `Address` is exactly what described above (an Enum-like primitive, barring any from-memory first-byte mismatches). `Address` along with `AccountIndex` was introduced specifically to reduce transaction bloat.

xlc:
Yeah that's want I want to do. But how do I do it? Fork polkadot.js and replace all types that using Address with AccountId?
This one of the goal I would like to achieve on this metadata improvement. The js sdk should be able infer the correct type to use, whether is Address or AccountId, the type information tells it.
If I ever introduce the indices module in a runtime upgrade, the js sdk should still handles it. It will notice the metadata updated, infer that the type of sender is Address instead of AccountId. And if AccountId is passed when Address is expected, the automatic conversion logic kicks in trying to convert it to the correct type, and in this case it knows how to convert AddressId to Address, everything works smoothly.

jacogr:
We don't infer, but we can act on what the metadata provides. Currently an `Address` is defined as `type Address = <Indices as StaticLookup>::Source;`

So if a chain changes that to `type Address = AccountId` and the metadata provides that info, we can do the same mapping and things just work. However we cannot and should not "in the absence of xyz" make magic happen.

xlc:
Close in favor of #712

Issue #683, opened 2019-02-12
Type Text encodedLength is error while string is not simple letter 


aniiantt:
https://github.com/polkadot-js/api/blob/master/packages/types/src/Text.ts#L45
```javascript
  get encodedLength (): number {
    return this.length + Compact.encodeU8a(this.length).length;
  }

  /**
   * @description The length of the value
   */
  get length (): number {
    // only included here since we ignore inherited docs
    return super.length;
  }
```
For example:
```javascript
new Text('中文').length // 2
new Text('中文').encodedLength // 3
new Text('中文').toHex() // 0x18e4b8ade69687
```
so it's encodedLength should be 7 ? but got 3.

It is work well for some change like the following:
```javascript
  get encodedLength() {
    const length = stringToU8a(this.toString()).length;
    return length + Compact.encodeU8a(length).length;
  }
```

jacogr:
Yes, def an issue. Encoded length is the length of the `Uint8Array` so should match with what is in `toU8a`

```js
get encodedLength() {
  return this.toU8a().length;
}
```

jacogr:
If you don't beat me to it, will make a PR with my suggested change and testcases from you in around an hour.

PR #684, opened 2019-02-12
Fix encodedLength for non-ascii strings

jacogr:
Closes https://github.com/polkadot-js/api/issues/683

### DIFF
--- a/packages/types/src/Text.ts
+++ b/packages/types/src/Text.ts
@@ -43,7 +43,7 @@ export default class Text extends String implements Codec {
    * @description The length of the value when encoded as a Uint8Array
    */
   get encodedLength (): number {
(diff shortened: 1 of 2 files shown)


packages/types/src/Text.ts
```diff
@@ -43,7 +43,7 @@ export default class Text extends String implements Codec {
    * @description The length of the value when encoded as a Uint8Array
    */
   get encodedLength (): number {
-    return this.length + Compact.encodeU8a(this.length).length;
+    return this.toU8a().length;
```
xlc: I am slightly worry about the performance impact of this, it should be possible to do it more efficiently

jacogr: It only gets used as part of the checking when decoding to determine offsets. So yes, impact while decoding metadata.

PR #688, opened 2019-02-13
Expose #687 on index

jacogr:
Missed these https://github.com/polkadot-js/api/pull/687


jacogr:
@amaurymartiny One of those random failures in the drr tests I've been speaking about - https://travis-ci.com/polkadot-js/api/builds/100720244 (This time didn't restart it, so at least it can be seen)

Issue #738, opened 2019-02-26
Allow custom Extrinsic implementation


ianhe8x:
Several changes required so that we can use TypeRegistry to replace the default Extrinsic.
1. priority changes: TypeRegistry should have higher priority than default implementation.
2. A Extrinsic interface.
2. SubmittableExtrinsic: change to have a _extrinsic member instead of extends directly.

jacogr:
Discussed in https://github.com/polkadot-js/api/issues/614

jacogr:
Closed in #739

PR #739, opened 2019-02-26
allow custom Extrinsic implementation

ianhe8x:
see #738
Originally part of refactor for Metadata V2.

### DIFF
--- a/packages/api/src/SubmittableExtrinsic.ts
+++ b/packages/api/src/SubmittableExtrinsic.ts
@@ -3,8 +3,9 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { KeyringPair } from '@polkadot/keyring/types';
-import { AccountId, Address, Extrinsic, ExtrinsicStatus, EventRecord, Hash, Index, Method, SignedBlock, Struct, Text, Vector } from '@polkadot/types/index';
-import { Codec, CodecCallback, SignatureOptions } from '@polkadot/types/types';
+import { AccountId, Address, ExtrinsicStatus, EventRecord, Hash, Index, Method, SignedBlock, Struct, Text, Vector } from '@polkadot/types/index';
+import { Codec, CodecCallback, ExtrinsicLike, SignatureOptions } from '@polkadot/types/types';
+import getRegistry from '@polkadot/types/codec/typeRegistry';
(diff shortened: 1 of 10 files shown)


packages/api/src/SubmittableExtrinsic.ts
```diff
@@ -3,8 +3,9 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { KeyringPair } from '@polkadot/keyring/types';
-import { AccountId, Address, Extrinsic, ExtrinsicStatus, EventRecord, Hash, Index, Method, SignedBlock, Struct, Text, Vector } from '@polkadot/types/index';
-import { Codec, CodecCallback, SignatureOptions } from '@polkadot/types/types';
+import { AccountId, Address, ExtrinsicStatus, EventRecord, Hash, Index, Method, SignedBlock, Struct, Text, Vector } from '@polkadot/types/index';
+import { Codec, CodecCallback, ExtrinsicLike, SignatureOptions } from '@polkadot/types/types';
+import getRegistry from '@polkadot/types/codec/typeRegistry';
```
jacogr: Just export `getRegistry` in `@polkadot/codec/index` (so it is available everywhere as part of `@polkadot/types`)

ianhe8x:
The ExtrinsicInterface is the place i'm not quite certain at the moment.` { hash: Hash }` is fine (i guess no parachain would go into that deep to replace Hash with other implementation), but as to `signature` I'm not sure if Codec is too generic.

jacogr:
For `Signature`, I would go for U8a as well, seems reasonable and the lowest-level. (Since it probably should end up with an array)

We probably need to have an `IHash` and `ISignature` for these.

jacogr:
"try to make codeclimate happy" - an impossible task. 

I'll give this a look-over and play with the e2e tests as well.

jacogr:
Looks good code wise. Will just do the e2e tests in the morning. (Sorry for the slight delay on my side getting this in)

ianhe8x:
`makes a transfer, swaps types and then another one` in the e2e test fails, though I tried master branch that testcase fails as well, so i guess that's fine.

jacogr:
That one is “fine” since sr25519 is still unused and there is some ongoing work looking at key derivation, however something broke in keyring so needs a fix soon-ish. (However completely out of scope here)

PR #742, opened 2019-02-26
Pull in vuepress foir docs (as opposed to gitbook)

jacogr:
Getting rid of the security nightmares that an unsupported gitbook let loose on the environment (And quite sane, a number of higher-profile projects have gone this route)

... works locally, let's see what it ends up after a proper CI build


jacogr:
Closed #740

PR #768, opened 2019-03-11
Basic query.section.method.multi impl

jacogr:
The good news, it seems to work. ~~The bad news... not happy with the typings here at all since the RxJS version doesn't have a callback.~~ 

~~And cannot override it, i.e. only add the signature in the Promise version as it stands. Additionally, in the RxJS version the resturn signature is not a subscription, rather is would be an `Array<Codec>` - unless we turn the result into a `Vector`, it will never be a `Codec`~~

@amaurymartiny Thoughts? Did something simple to just see if the rpc-core part indeed works as intended

EDIT: Updated with specific `MultiResult`, basically used in the RxJs version only (Promise has the array callback). TL;DR - more generic params


packages/api/src/Base.ts
```diff
@@ -515,24 +515,30 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
...
-    // FIXME as above...
+    decorated.key = (arg?: CodecArg): string =>
+      u8aToHex(compactStripLength(method(arg))[1]);
+
+    decorated.multi = (args: Array<CodecArg>, callback: CodecArrayCallback): M =>
+      onCall(
+        () => this._rpcRx.state.subscribeStorage(args.map((arg) => [method, arg])),
+        [],
+        callback
+      ) as unknown as M;
```
amaurymartiny: Instead of introducing M, this could be a C imo. And `type RxArrayResult = Observable<Vector<Codec>>;
`

jacogr: The issue with Vector... it is not quite a generic "take in anything" approach, it assumes exactly the same items in the vector, hence taking in a type. A `Tuple` on principle can have anything, but has the same "tell me what will be in it at construction". 

So even just trying to squeeze in something like that [here](https://github.com/polkadot-js/api/blob/master/packages/rpc-core/src/index.ts#L235) creates some ripple-effects down the line. (Caching, pulling older values from existing arrays since not everything is specified, etc.)

PR #773, opened 2019-03-14
Remove /index imports (swap to bare)

jacogr:


amaurymartiny:
nice!

jacogr:
Didn't want this duplication initially, https://github.com/polkadot-js/api/pull/773/files#diff-e5e546dd2eb0351f813d63d1b39dbc48R8 - however not doing it there once and then having to `/index` is just a disaster.

PR #779, opened 2019-03-15
fix: remove cache from api-derive

### LINKED ISSUE #777: derives in ApiPromise always return cached result
if using ApiPromise, it can be easily reproduced by executing `await api.derive.chain.bestNumber()` and `(await api.rpc.chain.getHeader()).blockNumber` repeatedly.

I think the current cache solution works only for rxjs because an observable is a stream, you still can get latest from it even you hold a cached one. but for promise it's one shot thing.

when the observable is toPromised, it will be terminated after first result, makes caching it meaningless.

amaurymartiny:
fixes #777.

I'm removing cached derive functions, similar to https://github.com/polkadot-js/api/issues/777#issuecomment-473188236 (I think Jaco you meant cache instead of ddr in your comment).

```javascript
const f = api.derive.some.thing(param);
const g = api.derive.some.thing(param);

// Before
console.log(f === g); // true

// After this PR
console.log(f === g); // false
```

I think it's easier to reason about, there's less magic. The only functions that need to be memoized are the ones in rpc-core, as we don't want to create multiple WS subscriptions for same function calls.

As for ddr, I think we can leave it, it's quite harmless.


jacogr:
Yeap, I meant cache, pre-morning-coffee comment.

PR #789, opened 2019-03-21
Metadata v2 and Linked Map (WIP)

### LINKED ISSUE #784: Support linked_map
Support `linked_map` which introduced at https://github.com/paritytech/substrate/pull/1763

Needs ability to enumerate the collection.

This is also required to support the new staking methods.

xlc:
Support https://github.com/paritytech/substrate/pull/1763
Closes #784 
Depends on https://github.com/paritytech/substrate/pull/2037

This is going to make #712 harder because all the conflicts...

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -17,7 +17,7 @@ import {
 } from './types';
 
 import EventEmitter from 'eventemitter3';
-import { map } from 'rxjs/operators';
+import { map, switchMap } from 'rxjs/operators';
 import decorateDerive, { Derive as DeriveInterface } from '@polkadot/api-derive';
 import extrinsicsFromMeta from '@polkadot/extrinsics/fromMetadata';
 import RpcBase from '@polkadot/rpc-core';
@@ -535,21 +535,16 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
     // Linked Map support
 
     if (method.headKey) {
-      decorated.head = ((callback?: CodecCallback) => {
-        return onCall(
+      decorated.head = (): C =>
+        onCall(
           (arg: CodecArg) => this._rpcRx.state
-            .subscribeStorage([[method, arg]])
+            .getStorage(arg)
             .pipe(
-              // state_storage returns an array of values, since we have just subscribed to
-              // a single entry, we pull that from the array and return it as-is
-              map((result: Array<Codec>): Codec =>
-                result[0]
-              )
-            ),
-          [method.headKey],
-          callback
+              switchMap(key => this._rpcRx.state.getStorage([method, key]))
+            )
+            ,
+          [method.headKey]
         ) as C;
-      });
     } else {
       decorated.head = () => {
         throw new Error(`${method.name} is not LinkedMap`);
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -17,7 +17,7 @@ import {
 } from './types';
 
 import EventEmitter from 'eventemitter3';
-import { map, switchMap } from 'rxjs/operators';
+import { map, switchMap, tap } from 'rxjs/operators';
 import decorateDerive, { Derive as DeriveInterface } from '@polkadot/api-derive';
 import extrinsicsFromMeta from '@polkadot/extrinsics/fromMetadata';
 import RpcBase from '@polkadot/rpc-core';
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -532,6 +532,30 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
     decorated.key = (arg?: CodecArg): string =>
       u8aToHex(compactStripLength(method(arg))[1]);
 
+    // Linked Map support
+
+    if (method.headKey) {
+      decorated.head = ((callback?: CodecCallback) => {
+        return onCall(
+          (arg: CodecArg) => this._rpcRx.state
+            .subscribeStorage([[method, arg]])
+            .pipe(
+              // state_storage returns an array of values, since we have just subscribed to
+              // a single entry, we pull that from the array and return it as-is
+              map((result: Array<Codec>): Codec =>
+                result[0]
+              )
+            ),
+          [method.headKey],
+          callback
+        ) as C;
+      });
+    } else {
+      decorated.head = () => {
+        throw new Error(`${method.name} is not LinkedMap`);
+      };
+    }
+
     return this.decorateFunctionMeta(method, decorated) as QueryableStorageFunction<C, S>;
   }
 
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -17,19 +17,21 @@ import {
 } from './types';
 
 import EventEmitter from 'eventemitter3';
-import { map, switchMap, tap } from 'rxjs/operators';
+import { map, switchMap } from 'rxjs/operators';
 import decorateDerive, { Derive as DeriveInterface } from '@polkadot/api-derive';
 import extrinsicsFromMeta from '@polkadot/extrinsics/fromMetadata';
 import RpcBase from '@polkadot/rpc-core';
 import RpcRx from '@polkadot/rpc-rx';
 import storageFromMeta from '@polkadot/storage/fromMetadata';
-import { Event, getTypeRegistry, Hash, Metadata, Method, RuntimeVersion } from '@polkadot/types';
+import { Event, getTypeRegistry, Hash, Metadata, Method, RuntimeVersion, Tuple, Option, Vector, Null } from '@polkadot/types';
 import { MethodFunction, ModulesWithMethods } from '@polkadot/types/primitive/Method';
 import { StorageFunction } from '@polkadot/types/primitive/StorageKey';
 import { assert, compactStripLength, isFunction, isObject, isUndefined, logger, u8aToHex } from '@polkadot/util';
 import { cryptoWaitReady } from '@polkadot/util-crypto';
 
 import { createSubmittableExtrinsic, SubmittableExtrinsic } from './SubmittableExtrinsic';
+import { Linkage } from '@polkadot/types/codec/Linkage';
+import { Observable, of } from 'rxjs';
 
 type MetaDecoration = {
   callIndex?: Uint8Array,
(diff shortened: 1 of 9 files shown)


jacogr:
Just browsed through - I am assuming most of the code is copy. In this case we could actually reuse and pull in from v1 where it makes sense. Although, with the structure change it probably sits at the bottom. (And is probably exactly what you did, changed the least amount possible)

Will take a proper look when I am at a PC. 

Thank you.

packages/types/src/Metadata/v2/toV1.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import MetadataV1 from '../v1/Metadata';
+import MetadataV2 from '.';
+
+export default function toV1 (v2: MetadataV2): MetadataV1 {
+  // V2 and V1 have same interface
+  return v2 as any as MetadataV1;
```
xlc: Not sure how safe this is. `instanceof` check may fail unexpectedly. Otherwise I unpack and repack it properly.

packages/types/src/Metadata/v2/Metadata.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+import latestParsed from './latest.substrate.v2.json';
+import rpcData from './static';
+
+describe('MetadataV2', () => {
+  const metadata = new Metadata(rpcData);
+
+  it('decodes latest properly', () => {
+    const str = JSON.stringify(metadata.toJSON());
+
+    // console.error(str);
```
jacogr: These are really useful for manual inspection when having to add new stuff. (It is ok removing it for v0 and V1 since we won't update those anymore, but for the latest would prefer it in). The `str` value actually gets used for visual checks as well. (Yes, it _is_ noisy)

xlc:
I don't think it make sense to have linked double map... Double map will be quite different as there will be two key types, the second key type comes with a configurable hash algorithm (first one is always xx I think).

jacogr:
All fine, keep it simple and to-the-point.

xlc:
Make V0 metadata includes all the new thing is not easy... Later there should be a refactor to upgrade metadata to latest version instead of downgrade to old version.

jacogr:
Yes. https://github.com/polkadot-js/api/issues/635 Have really not been able to get there. The pre-requisites are all done.

packages/types/src/primitive/StorageKey.ts
```diff
@@ -38,10 +39,10 @@ export default class StorageKey extends Bytes {
     if (isFunction(value)) {
       return value();
     } else if (Array.isArray(value)) {
-      const [fn, arg] = value;
+      const [fn, ...arg] = value;
```
xlc: this is not needed for this PR, but needed for double map

packages/api/test/e2e/rx-queries.spec.js
```diff
@@ -50,4 +50,22 @@ describe.skip('e2e queries', () => {
...
+    it('queries state witha key', (done) => {
+      api.query.staking.test(1).subscribe((res) => {
+        console.log('!!', res)
+
+        done();
+      });
+    });
+
+    it.only('queries state head', (done) => {
+      api.query.staking.test.head().subscribe((res) => {
```
xlc: I added 
`pub Test get(test) build(|_config| vec![(1, 2), (3, 4), (5, 6)]): linked_map u32 => u32;` to staking module for testing purpose.
I will change it to use `staking.validators` once I am confident it works

packages/type-storage/src/utils/createFunction.ts
```diff
@@ -51,6 +51,20 @@ export default function createFunction (section: Text | string, method: Text | s
     );
   };
 
+  if (meta.type.isMap && meta.type.asMap.isLinked) {
+    // TODO: there needs some better way to do this
+    const keyHash = new U8a(xxhashAsU8a(`head of ${stringKey}`, 128));
```
xlc: this code is bit messy
`new U8a` is needed here to indicate the data does not have length prefix
I need this `keyFn` function to specify the output format, which is the key type

packages/api/src/Base.ts
```diff
@@ -532,6 +532,25 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
     decorated.key = (arg?: CodecArg): string =>
       u8aToHex(compactStripLength(method(arg))[1]);
 
+    // Linked Map support
+
+    if (method.headKey) {
+      decorated.head = (): C =>
+        onCall(
+          (arg: CodecArg) => this._rpcRx.state
+            .getStorage(arg)
```
xlc: I was trying to implement this `subscribeStorage` but can't get it right.
It needs firstly get storage with head key, which will return the first key, and can be used to fetch first entry

packages/api/src/types.ts
```diff
@@ -59,6 +59,7 @@ export interface QueryableStorageFunctionBase<CodecResult, SubscriptionResult> e
   hash: (arg?: CodecArg) => HashResult<CodecResult, SubscriptionResult>;
   key: (arg?: CodecArg) => string;
   size: (arg?: CodecArg) => U64Result<CodecResult, SubscriptionResult>;
+  head: () => CodecResult;
```
xlc: I am unsure what should be interface for the `enumerate` method. Does it make sense to be `Iterator<CodecResult>`? Also is it helpful to expose lower level methods? Such as get head key, get next key, get previous key.

xlc: Enumerate on client side needs to be very careful because the block head could change while enumerating. In that case I think I need to include the current block head in the Iterator?

packages/api/src/Base.ts
```diff
@@ -532,6 +532,25 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
...
+            .pipe(
+              switchMap(key => this._rpcRx.state.getStorage([method, key]))
```
xlc: this needs to return both the value and the key. but I don't think I can return two values with Observable? Should I return `Tuple` instead?

xlc:
@jacogr have few questions about the interface and implementations.
Also should I move the metadata v2 part into its own PR so it can be merged first?

jacogr:
Yes, we need the v2 stuff first, as it stands it was merged into master - so we currently have breakages :) Love those, btw... it happens :)

packages/api/src/Base.ts
jacogr: Yes, Tuple makes sense so it is still a `Codec`

packages/api/src/types.ts
jacogr: And `Iterator` would make sense. However forcing everybody to iteration themselves is probably not such a good idea. So it would probably make more sense to have the expose function return all the entries. (With an, maybe optional, length limit - so iteration is not done by everybody)

packages/type-storage/src/utils/createFunction.ts
jacogr: You can just use `xxhashAsHex` without the `new U8a()` wrapping.

packages/api/src/Base.ts
jacogr: pleayed with your branch a bit, it works ok for me with `subscribeStorage` - just be aware that it returns an array, so you just need to do something like -

```
        onCall(
          (arg: CodecArg) => this._rpcRx.state
            .subscribeStorage([arg])
            .pipe(
              tap((value) => console.error('head', value)),
              // grab first value of the array 
              switchMap(([key]) => this._rpcRx.state.getStorage([method, key])),
              tap((value) => console.error('first', value))
            ),
          [method.headKey]
        ) as C;
```

amaurymartiny:
Small suggestion: instead of having `.head()`, `.enumerate()` etc on all storage functions (90% of them who aren't even linked maps), could we do `query.staking.validators.asLinkedMap` (will throw if not linked map).

This will return an instance of something like:

```javascript
// I think Map makes sense, it's already an iterator.
// Keys are `StorageKey`, values are `Observable<T>`
class LinkedMap<T> extends Map<StorageKey, T> {
  enumerate: () => Iterator<Observable<T>>; // this should be exactly the same as .entries() or .values() on Map
  head: () => Observable<T>;
  headKey: () => StorageKey; // optional
}
```

would that work? It'll keep things slightly cleaner

packages/api/src/types.ts
jacogr: So externally `head`, `next`, etc. is probably not the best since it add overhead. Just make an `.all` or `.list` where we retrieve all by iterating. (Subscribe to head, subsequent entries can use `getStorage`)

jacogr:
So would suggest after playing - 

- we should return the correct types (or adjust to be) as tuple[option, option] where the first contains the type (this matches up)

- instead of head/list etc. just return the full list when called with no params so no API changes. As it stands it will be ok, but need to revisit since it can be quite heavy on large lists (1000s).

jacogr:
Alexander runtime update has gone through, so the linked metadata should be available there for e2 testing as well. (So it helps with the specific runtime test call in the spec)

xlc:
I was trying to implement the pass empty argument and get all part, and have some issues with what is the expected behaviour with RxJS?
The normal `api.query.staking.validators(alice)` will subscribe storage to a key value, and emit values as it updates.
So doe that means `api.query.staking.validators()` needs to:
- subscribe to head key
- subscribe to the result of head key
- subscribe to each individual items?

There seems like a lot subscription to manage and I don't actually know how to implement it on with the base API object. I am still trying to understand how the `onCall` works.

jacogr:
The enumeration APIs looks correct.

Yes, it is a bit of a nightmare single you basically have a loop and until completed, don't return the Observable. Will see how far you get (in what is left of your day), could also take a look in a short while once I wrapped up some other dangling stuff (i.e. new metadata types on master)

xlc:
For `api.query.staking.validators(address)` it is now return the type `(Value, Linkage<Key>)` instead of just `Value`. This is a breaking change but maybe fine because it only breaks linked map?
Or avoid the breaking change but somehow still expose the `Linkage<Key>` type?

jacogr:
No that is fine. It actually is the right thing based on what the storage actually returns.

jacogr:
... since it actually is a new API and actually not properly supported it is not really "breaking".

xlc:
I think this version works and I will call it a day. Please take a look. Some parts is bit confusing to me and I just change it until it work without understanding fully so there could be issues.

jacogr:
Thank you! Perfect. Will test, took a brief look and it seems correct - will test with Alex.

jacogr:
Small test via this - 

```
describe('alex queries', () => {
  let api;

  beforeAll(async () => {
    api = await Api.create({
      provider: new WsProvider('wss://poc3-rpc.polkadot.io/')
    });

    return api;
  });

  beforeEach(() => {
    jest.setTimeout(30000);
  });

  it('retrieves the list of validators', (done) => {
    api.query.staking.validators((res) => {
      console.log('!!', res.toJSON());

      done();
    });
  });
});
```

Yields this -

```
!! [ { unstakeThreshold: 3, validatorPayment: 0 },
      { unstakeThreshold: 3, validatorPayment: 0 },
      { unstakeThreshold: 3, validatorPayment: 0 },
      { unstakeThreshold: 3, validatorPayment: 0 },
      { unstakeThreshold: 3, validatorPayment: 0 },
      { unstakeThreshold: 3, validatorPayment: 0 },
      { unstakeThreshold: 3, validatorPayment: 0 } ]
```

Only think we seem to miss is the actual validator key

jacogr:
Replaced by https://github.com/polkadot-js/api/pull/801 with this extension - https://github.com/polkadot-js/api/pull/801/commits/fb4416b82c0ede71ddfe7c6116bb844cc8b538b7

Would have loved to merged this one instead, my changes are small.

Issue #776, opened 2019-03-14
Cater for StorageDoubleMap for storage keys


jacogr:
- Implemented in https://github.com/paritytech/substrate/pull/1918
- First used in substrate system events, https://github.com/paritytech/substrate/blob/master/srml/system/src/lib.rs#L331-L345

jacogr:
And this now becomes V3 in the linked PR. (Didn't we _just_ do V2? ;))

jacogr:
Will only revisit this if there are actual substrate PRs that use these types. For now, without proper tests, it will probably need a re-visit anyway if implemented right now (since testing cannot be done properly e2e without assumptions and/or writing sample Rust modules.)

ianhe8x:
@jacogr During our works for supporting Double Map, there are some design decisions need your advice. Before, a StorageFunction can only have one key or none, therefore in [StorageKey.ts](https://github.com/polkadot-js/api/blob/facdbf8f3a4c7e9d528b14e7b03e89ff58b5ed09/packages/types/src/primitive/StorageKey.ts#L13) it is `(arg?: any): Uint8Array;`
Now with double map it adds another possibility. We could make it
```
(): Uint8Array;
(key: any): Uint8Array;
(key1: any, key2: any): Uint8Array;
```
which might be best for UX, but downside is we must also update other places where expects one key at most.
or we don't change interface, and use [key1, key2] for double map. 

We are working towards `[key1, key2]` approach for now unless you suggest otherwise.

jacogr:
Ahh, yes, @amaurymartiny actually commented on that a while back. I would rather try the best-UX version first. From a API pov I think the `[key1, key2]` this is always slightly problematic, and a second choice.

Not 100% sure off-the-top of my head apart from StorageKey and types where the problems may come up. (decoration as well). It is generally relatively simple, the thing with multi-params is that the TS defs just become more and more hairy, i.e in some cases had to do this, excuse my language, crap - https://github.com/polkadot-js/api/blob/master/packages/api/src/types.ts#L112-L116

(Not pretty, but no real other option)

jacogr:
Closing, thanks to the work by @AlexWZ0129

PR #798, opened 2019-03-26
Initial DoubleMap support (support v3 Metadata)

### LINKED ISSUE #799: Metadata load failure should exit early
Currently it seems like the “unknown Metadata version” is swallowed and then use of the API spits out wrong errors. 

I believe it it the fallback to v0 that just parses garbage. Can remove v0 parsing support as a start.

jacogr:
- Part of https://github.com/polkadot-js/api/issues/776
- Closes #799


jacogr:
Yes, I was expecting use, however the only thing that changed running the PR was the version

packages/types/src/Metadata/v3/toV2.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import MetadataV2 from '../v2/Metadata';
+import MetadataV3 from '.';
+
+export default function toV2 (v2: MetadataV3): MetadataV2 {
+  return new MetadataV2(v2);
```
xlc: `DoubleMapType` only exists in v3 so this should fail if trying to convert a metadata includes double to v2.
One way to solve this (without refactor everything to make upgrade instead of downgrade), is duplicate `DoubleMapType` in v2 like how I retrofit `is_linked` in v1

Issue #804, opened 2019-03-27
Keep alive ping


xlc:
(Please correctly me if I am mistaken)

Substrate drops the ws connection if it is idle for a minute. It will be good if there is a way to have polkadot.js sending some ping message in a configurable amount of interval (with a reasonable default) to ensure the connection is alive.

jacogr:
Yes, it does. On the apps I'm sending a health check every 30s (I believe) to do exactly this.

Good idea to incorporate into the base.

jiangfuyao:
Best to slove it from substrate layer. Sometime the api layer closed. The  node WS/RPC service will close too. Need restart by manual. It‘s trouble. So..Best have the substrate add parameter, set automatic close time, or set cancel automatic close.

Issue #802, opened 2019-03-27
Need for subscription on linked maps


jacogr:
(Marking as question since it is not really confirmed and may just be due to me not quite grasping some basics)

Let's start at the top, I believe this is needed: https://github.com/polkadot-js/api/blob/master/packages/api/src/Base.ts#L554 - as new entries get added, the head is updated, so it nicely triggers the reload of all/

However, not convinced about this - https://github.com/polkadot-js/api/blob/master/packages/api/src/Base.ts#L505

The problem is that as the entry changes (say an entry was removed in the middle), I'm not sure how this actually caters for refreshing that part of the list and/or the whole list. Since at on subsequent loads we are basically done, I am not sure how the list gets refreshed. 

TL;DR - Can see how adding a new entry causes a refresh of the list (and can visibly see this with the apps UI), however for stuff that gets removed from the list, cannot quite wrap around the refresh thereof.

This could be related to https://github.com/polkadot-js/apps/issues/857 (however that could very well be a UI bug, things are still WIP and not 100% clean there)

cc @xlc @amaurymartiny  - you may be able to help guide me along the right track. This _looks_ correct, but not sure how/if it works in practice since with these refreshes can be problematic

xlc:
I think the only way to test if it works is have a testing module so we can see how the current code handles all the scenarios.

jacogr:
100% - in this case I just (also) cannot wrap my head around what happens internally (between the subs and RxJs and the conversions at the end) when the previous/next changes.

xlc:
Maybe should provide some diagnostic function to print all the subscriptions to help understanding what is happening.
I am expecting people are going to have leaky subscription issue.

jacogr:
Ok following @xlc with some debugging, there are some gaps, but not surprising since this is very messy. Tracking nominators and entering, removing.

Initial -
```
console.log packages/api/src/Base.ts:604
    head changed

  console.log packages/api/src/Base.ts:554
    list changed: 5GGxQvLetFxUFC3K8rtQxXcVdnMN7gyAkGaMLxaS8Ra1eQNn

  console.log packages/api/src/Base.ts:554
    list changed: 5CqWLoyv1joRsEWEqY1U2UCEHho2uwMGfde3wgGXGEi2ZNtz

  console.log packages/api/src/Base.ts:554
    list changed: 5G9YqP3mR96h7r88DD4n2MVSgG5z4sHD5ooiuZ4cYdqWDF13

[...]

PR #805, opened 2019-03-28
Track linked map changes

jacogr:
- Closes https://github.com/polkadot-js/api/issues/802 (ensuring results for all types of list changes)
- Closes https://github.com/polkadot-js/api/issues/804 (keepalive, needed for the manual tests as they run)
- Split linked-method iteration into own function
- Initial commit https://github.com/polkadot-js/api/pull/805/commits/d9dd70278dff3242f02ce6171ea5a5eb7745578f fixes the updates when the head changes (basically setting `head = key` in the head subscription)
- Use BehaviourSubject so we have a single observable floating around where we can set the value, https://github.com/polkadot-js/api/pull/805/commits/31f8694c75e4b52178e2479051ac233b7eb31c33

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -17,13 +17,13 @@ import {
 } from './types';
 
 import EventEmitter from 'eventemitter3';
-import { map, switchMap } from 'rxjs/operators';
+import { map, switchMap, tap } from 'rxjs/operators';
 import decorateDerive, { Derive as DeriveInterface } from '@polkadot/api-derive';
 import extrinsicsFromMeta from '@polkadot/extrinsics/fromMetadata';
 import RpcBase from '@polkadot/rpc-core';
 import RpcRx from '@polkadot/rpc-rx';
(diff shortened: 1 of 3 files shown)


packages/api/src/Base.ts
```diff
@@ -353,6 +354,14 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
           this._isReady = true;
 
           this.emit('ready', this);
+
+          setInterval(() => {
+            if (this._rpcBase._provider.isConnected()) {
```
xlc: discard the interval the handler means it is now impossible to destroy this api instance.
should have a close/dispose/teardown method to `clearInterval`

jacogr: 100%, still WIP... (Hacked it in for the other tests to work)

jacogr: Should be addressed around here - https://github.com/polkadot-js/api/pull/805/files#diff-95597f67ed68d7dffc27ba1f0501e5b1R336

So `disconnect()` breaks the link, the event from that clears the interval.

Issue #838, opened 2019-04-12
API throws InvalidSignature(0) (Bad signature) for usize function arguments. 


drewstone:
If the function of your Substrate extrinsic contains `usize` arguments, the API seems to throw `InvalidSignature(0)` errors. Changing these to `u32` fixes the issue.

Nonetheless, these errors are caught at the API level before the extrinsics with `usize` function arguments ever touch the chain. Therefore, my intuition is the API cannot parse or serialize `usize` arguments.

jacogr:
@amaurymartiny Would appreciate if you can take a look when you have a gap -https://github.com/polkadot-js/api/blob/master/packages/types/src/primitive/USize.ts 

(That looks innocent enough, the issue however is that it assumes u64 and usize is dependent on platform...)

jacogr:
Seems to be u32 - https://github.com/rustwasm/wasm-bindgen/pull/1448/files#diff-ccdc0411efb64746018c8d6809af3411R86

jacogr:
Closed in #839

PR #839, opened 2019-04-12
USize -> U32

jacogr:
- Possibly a fix for https://github.com/polkadot-js/api/issues/838


xlc:
This makes me wonder, how is this not a compatibility issue between WASM executor and native executor? If usize is used with a value more than u32 max, the result will be different and cause breakage.

jacogr:
@amaurymartiny Can you cast an eye here?

PR #843, opened 2019-04-15
Start supporting double map 

xlc:
We are upgrading our modules to use double map so there will be a set of contributions from us to make double map actually works with polkadot.js.

This doesn't actually support double map, just fixed some crashing error I encountered while working on apps.

### DIFF
--- a/packages/types/src/Metadata/v3/Storage.ts
+++ b/packages/types/src/Metadata/v3/Storage.ts
@@ -125,7 +125,7 @@ export class StorageFunctionType extends EnumType<PlainType | MapType | DoubleMa
    */
   toString (): string {
     if (this.isDoubleMap) {
-      return this.asDoubleMap.toString();
+      return this.asDoubleMap.value.toString();
     }
 
     if (this.isMap) {
--- a/packages/types/src/Metadata/v3/Storage.ts
+++ b/packages/types/src/Metadata/v3/Storage.ts
@@ -11,6 +11,7 @@ import Struct from '../../codec/Struct';
 import Vector from '../../codec/Vector';
 import Bytes from '../../primitive/Bytes';
 import Text from '../../primitive/Text';
+import Type from '../../primitive/Type';
 import {
   MapType,
   PlainType,
@@ -27,25 +28,25 @@ export {
 export class DoubleMapType extends Struct {
   constructor (value?: any) {
     super({
-      key1: Text,
-      key2: Text,
-      value: Text,
+      key1: Type,
+      key2: Type,
+      value: Type,
       key2Hasher: Text
     }, value);
   }
 
   /**
-   * @description The mapped key as [[Text]]
+   * @description The mapped key as [[Type]]
    */
-  get key1 (): Text {
-    return this.get('key1') as Text;
+  get key1 (): Type {
+    return this.get('key1') as Type;
   }
 
   /**
-   * @description The mapped key as [[Text]]
+   * @description The mapped key as [[Type]]
    */
-  get key2 (): Text {
-    return this.get('key2') as Text;
+  get key2 (): Type {
+    return this.get('key2') as Type;
   }
 
   /**
@@ -56,10 +57,10 @@ export class DoubleMapType extends Struct {
   }
 
   /**
-   * @description The mapped key as [[Text]]
+   * @description The mapped key as [[Type]]
    */
-  get value (): Text {
-    return this.get('value') as Text;
+  get value (): Type {
+    return this.get('value') as Type;
   }
 }
 
--- a/packages/types/src/Metadata/v4/Storage.ts
+++ b/packages/types/src/Metadata/v4/Storage.ts
@@ -114,32 +114,32 @@ export class DoubleMapType extends Struct {
   constructor (value?: any) {
     super({
       hasher: StorageHasher,
-      key1: Text,
-      key2: Text,
-      value: Text,
+      key1: Type,
+      key2: Type,
+      value: Type,
       key2Hasher: Text
     }, value);
   }
 
   /**
-   * @description The hashing algorithm used to hash keys, as [[Text]]
+   * @description The hashing algorithm used to hash keys, as [[StorageHasher]]
    */
-  get hasher (): Text {
-    return this.get('hasher') as Text;
+  get hasher (): StorageHasher {
+    return this.get('hasher') as StorageHasher;
   }
 
   /**
-   * @description The mapped key as [[Text]]
+   * @description The mapped key as [[Type]]
    */
-  get key1 (): Text {
-    return this.get('key1') as Text;
+  get key1 (): Type {
+    return this.get('key1') as Type;
   }
 
   /**
-   * @description The mapped key as [[Text]]
+   * @description The mapped key as [[Type]]
    */
-  get key2 (): Text {
-    return this.get('key2') as Text;
+  get key2 (): Type {
+    return this.get('key2') as Type;
   }
 
   /**
@@ -150,10 +150,10 @@ export class DoubleMapType extends Struct {
   }
 
   /**
-   * @description The mapped key as [[Text]]
+   * @description The mapped key as [[Type]]
    */
-  get value (): Text {
-    return this.get('value') as Text;
+  get value (): Type {
+    return this.get('value') as Type;
   }
 }
 
--- a/packages/types/src/Metadata/v3/Metadata.ts
+++ b/packages/types/src/Metadata/v3/Metadata.ts
@@ -109,11 +109,15 @@ export default class MetadataV3 extends Struct implements MetadataInterface {
     return this.modules.map((mod) =>
       mod.storage.isNone
         ? []
-        : mod.storage.unwrap().map((fn) =>
-          fn.type.isMap
-            ? [fn.type.asMap.key.toString(), fn.type.asMap.value.toString()]
-            : [fn.type.asType.toString()]
-        )
+        : mod.storage.unwrap().map((fn) => {
+          if (fn.type.isMap) {
+            return [fn.type.asMap.key.toString(), fn.type.asMap.value.toString()];
+          } else if (fn.type.isDoubleMap) {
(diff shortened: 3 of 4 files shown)


jacogr:
Thank you!

xlc:
The metadata versions is starting getting messy and adding a new feature is not easy because it needs to downgrade from latest version back to v0.

Any plan to refactor so that it is performing upgrade instead of downgrade?

There is v4 (https://github.com/paritytech/substrate/pull/2268) and maybe v5 (https://github.com/paritytech/substrate/pull/2261) upcoming so it is better to do this refactor before they get merged in.

jacogr:
Needs to be done, yes. Agreed that it is sprawling a bit atm. Have not been able to get there myself. 

@amaurymartiny could you take a peek at this PR please - cannot give it the right kind of attention with limited access.

packages/types/src/Metadata/v1/Storage.ts
```diff
@@ -73,21 +74,36 @@ export class MapType extends Struct {
 export class PlainType extends Type {
 }
 
-export class MetadataStorageType extends EnumType<PlainType | MapType> {
+export class MetadataStorageType extends EnumType<PlainType | MapType | DoubleMapType> {
```
amaurymartiny: You can remove this class, and import StorageFunctionType from `v0/modules`

packages/types/src/Metadata/v2/Storage.ts
```diff
@@ -50,21 +51,36 @@ export class MapType extends Struct {
 export class PlainType extends Type {
 }
 
-export class MetadataStorageType extends EnumType<PlainType | MapType> {
+export class MetadataStorageType extends EnumType<PlainType | MapType | DoubleMapType> {
```
amaurymartiny: You can remove this class, and import StorageFunctionType from `v0/modules`

packages/types/src/Metadata/v0/Modules.ts
```diff
@@ -174,21 +213,36 @@ export class MapType extends Struct {
 class PlainType extends Type {
 }
 
-export class StorageFunctionType extends EnumType<PlainType | MapType> {
+export class StorageFunctionType extends EnumType<PlainType | MapType | DoubleMapType> {
```
amaurymartiny: Just to understand your intent here: we agree that v0 doesn't have `DoubleMapType`, but you're still putting it here to be able to factorize code, as this change doesn't break compatibility, correct?

If yes, then see my 2 comments above where we can factorize code a bit more.

packages/types/src/Metadata/v2/Storage.ts
```diff
@@ -11,6 +11,7 @@ import Null from '../../primitive/Null';
 import Text from '../../primitive/Text';
 import Type from '../../primitive/Type';
 import { MetadataStorageModifier } from '../v1/Storage';
+import { DoubleMapType } from '../v0/Modules';
```
amaurymartiny: tiny nit: we tend to order things alphabetically by default, when there's no order natural ordering

amaurymartiny:
Hey @xlc I refactored types/Metadata a bit, so that it's less entangled, but conflicts with this implementation. imo there's only https://github.com/polkadot-js/api/pull/843/files#diff-023179d9a94bd59761a7319a645b1dd0R112 to be added.

wangzhen10:
Hi @amaurymartiny. Thank you for taking the time to review the code. I will change the code based on the suggestions and rebase this branch on the master to fix the conflict.

packages/types/src/Metadata/v0/Storage.ts
```diff
@@ -69,21 +70,36 @@ export class MapType extends Struct {
 export class PlainType extends Type {
 }
 
-export class StorageFunctionType extends EnumType<PlainType | MapType> {
+export class StorageFunctionType extends EnumType<PlainType | MapType | DoubleMapType> {
```
amaurymartiny: There's no DoubleMap in v0, everything that's related to DoubleMap in this file should be deleted

packages/types/src/Metadata/v2/Storage.ts
```diff
@@ -12,6 +12,7 @@ import Bytes from '../../primitive/Bytes';
 import Text from '../../primitive/Text';
 import Type from '../../primitive/Type';
 import { PlainType, StorageFunctionModifier } from '../v1/Storage';
+import { DoubleMapType } from '../v4/Storage';
```
amaurymartiny: Same as above, remove DoubleMap from v2

packages/types/src/Metadata/v3/Storage.ts
```diff
@@ -25,39 +27,39 @@ export {
 export class DoubleMapType extends Struct {
   constructor (value?: any) {
     super({
-      key1: Text,
-      key2: Text,
-      value: Text,
-      key2Hasher: Text
+      key1: Type,
+      key2: Type,
+      value: Type,
```
amaurymartiny: The above 3 are all Text, revert

packages/types/src/Metadata/v3/Storage.ts
```diff
@@ -25,39 +27,39 @@ export {
...
+      key2Hasher: StorageHasher
```
amaurymartiny: Unfortunately key2Hasher is still a Text, revert. Please see https://github.com/paritytech/substrate/blob/master/srml/metadata/src/lib.rs

packages/types/src/Metadata/v4/Storage.ts
```diff
@@ -112,46 +112,46 @@ export class DoubleMapType extends Struct {
...
     super({
       hasher: StorageHasher,
-      key1: Text,
-      key2: Text,
-      value: Text,
-      key2Hasher: Text
+      key1: Type,
+      key2: Type,
+      value: Type,
+      key2Hasher: StorageHasher
```
amaurymartiny: same as above

packages/types/src/Metadata/v4/Storage.ts
```diff
@@ -112,46 +112,46 @@ export class DoubleMapType extends Struct {
...
   }
 
   /**
-   * @description The hashing algorithm used to hash keys, as [[Text]]
+   * @description The hashing algorithm used to hash keys, as [[StorageHasher]]
    */
-  get hasher (): Text {
-    return this.get('hasher') as Text;
+  get hasher (): StorageHasher {
+    return this.get('hasher') as StorageHasher;
```
amaurymartiny: Thanks! That was my mistake

packages/types/src/Metadata/v3/Storage.ts
xlc: those 3 need to be Type otherwise it won’t do the type preprocess step and causing issues. e.g. T::Balance need the type process step to convert to Balance then can be found in the type registry

amaurymartiny: Ah yeah, you're totally right

packages/types/src/Metadata/v3/Storage.ts
wangzhen10: Thank you for the review. When using `Text`, if the type of the key has trait prefix like ’T::AssetId’, I got an error ‘Unknown types found, no types for T::AccountId’. Using `Type` fix the issue. I guess it’s because Type does some format optimise when decoding (such as removing all the trait prefixes `Type._removeTraits()`).

packages/types/src/Metadata/v0/Storage.ts
wangzhen10: I got ‘FATAL: Unable to initialize the API:  Object.values(...)[index] is not a constructor’ error at EnumType.ts:109:14 without having DoubleMapType here. Added DoubleMapType for fixing this issue.
Here `private static createValue (def: TypesDef, index: number = 0, value?: any): Decoded {
    return {
      index,
      value: new (Object.values(def)[index])(value)
    };
  }
` index is 2 but def only have MapType and PlainType. Call stack says, it's because of `const extrinsics = extrinsicsFromMeta(this.runtimeMetadata.asV0);`

packages/types/src/Metadata/v3/Storage.ts
wangzhen10: Changed type of key2Hasher to Text. Thanks for the suggestion.

packages/types/src/Metadata/v0/Storage.ts
amaurymartiny: Ah, I suspected this would happen...

I think there's a small refactoring to do, have a look at point 1 in https://github.com/polkadot-js/api/issues/858#issue-437173145:
- create v0/ and v4/ folders in type-extrinsic
- move current fromMetadata to fromMetadataV0
- fromMetadata should call fromMetadataV0 if it's v0-3
- in v4/, create a fromMetadataV4, which correctly creates extrinsics from v4

Have a look at type-storage, type-extrinsics should follow the same pattern, and this would allow adding hacks like adding DoubleMap in v0

AlexWZ0129: Thank you @amaurymartiny. I will try doing the refactoring following your guide.

ianhe8x: @AlexWZ0129 I've done similar stuffs, you can reuse code from my refactor branch.

AlexWZ0129: @ianhe8x Okay. Thank you, Ian!

packages/types/src/Metadata/v2/Storage.ts
ianhe8x: Regarding the ordering and spacing rules, can we have them in tslint?

amaurymartiny: That's a good idea. But I think since TypeScript is officially moving towards eslint, we'd like to use eslint everywhere in the mid-term, [this](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md) looks for for our case.

But now in this "transition" period before eslint is officially supported, i propose not to overbloat tslint.json, and to manually enforce this few rules.

In any case, you can create an issue on https://github.com/polkadot-js/dev

AlexWZ0129:
> lgtm 👍

Thanks @amaurymartiny !

Issue #580, opened 2019-01-14
Get rid of Globals and Singletons(Method injection, EventTypes injection, TypeRegistry)


ianhe8x:
I believe you all agree that Metadata is a block scoped thing. But now in the code metadata and things derived from metadata are stored in the global scope. It brings conflict when, for example, feature of querying storage from specified block is added into api. Because metadata could be changed since that block and api is still trying to decode with the latest metadata.

Is there any plan to refactor that part? Or maybe I can try contribute that, even though it looks quite a large task.

jacogr:
You are right that this is a non-trivial exercise. 

It practice it _should_ be relatively safe, since new events, methods, ... should generally be added as non-replacements, i.e. they don't clobber the indexes of previously defined values. (So new metadata can decode older blocks - breaking metadata in a backwards incompatible manner is generally a less than good idea, unless you do a chain refresh at that point). 

... Following this works great for chains such as Alexander and Charred Cherry, but probably not for any generic substrate chains that can be added where the same "don't break userspace" approach may not be in place.

It has been in the back of my mind for a long while, solving this elegantly, i.e. make sure you can decode anything new using the latest metadata (refresh transparently as it changes) vs. querying old blocks (where the issue would show up if non-backwards-compatible changes were made), I don't quite have an answer for.

jacogr:
This one is critical though - https://github.com/polkadot-js/api/issues/449 (There is the subscribe metadata functions, we are not using it so we do fall behind on updates - so we break forward-compatibility)

xlc:
Regards to compatibility with different runtime versions, I had some thoughts: https://github.com/paritytech/substrate/issues/1378

The idea solution to me is:
Implement codec metadata: https://github.com/paritytech/substrate/pull/1328
No more type registry is required as the metadata itself is enough to decode the types.
Trivial changes are no longer breaking (https://github.com/paritytech/substrate/pull/1373). How a value should be decoded is handled by the type metadata. The decoded result remain the same.
Adding fields (currently breaking) change can be automatically handled as long as the field is the last one.
Removing fields can also be automatically handled as long as the old field is optional / nullable.
Simple type change (e.g. U32 to U46) can be handled atomically, in the end they are all just BN.

It also enable feature detection and more advanced generic interface by knowing the exact structure of the runtime methods.

Similar idea can be applied to smart contract as well.

All I need is someone help me fix the HasCompact issue in my PR...

jacogr:
[...]

PR #847, opened 2019-04-18
refactor type system (a rough idea)

ianhe8x:
related to #843 and #580 
Here's a rough idea of the refactor.
Goals:
1. decouple the code from a specified Metadata version.
2. change global injection to api scope
3. [Optional] type name resolve improvement, resolve duplicate name across runtimes properly.

Approach
For Goal 1):
1. change type using classes from MetadataV0 to interface.
2. types-storage and types-extrinsic shall handle Metadata of different version with specified logic.
For Goal 2):
1. whereever call createType or global Injection now call typeRegistry. 
2. all places `new Type()` shall be wrapped by `TypeRegistry.withTypeRegistry()` then inside can use TypeRegistry's static property to get access to the api-scope TypeRegistry instance.
For Goal 3):
1. after all type resolving are restrained in TypeRegistry, we should be able to come up with a solution.


packages/types/src/type/Event.ts
```diff
@@ -86,18 +85,18 @@ export class EventIndex extends U8aFixed {
 export default class Event extends Struct {
   // Currently we _only_ decode from Uint8Array, since we expect it to
   // be used via EventRecord
-  constructor (_value?: Uint8Array) {
-    const { DataType, value } = Event.decodeEvent(_value);
+  constructor (_value: Uint8Array = new Uint8Array(), typeRegistry: TypeRegistry) {
```
amaurymartiny: Yes, exactly the idea Jaco and I had some time ago: dependency injection via constructor. Teh tricky part is to figure out how to pass in `typeRegistry` to this constructor.

If i'm not mistaken, we instantiate this class here: https://github.com/polkadot-js/api/blob/master/packages/rpc-core/src/index.ts#L182. Which means that we somehow need to pass the typeRegistry to rpc-core too. I believe it's ultimately the correct way to go. We basically need to have `typeRegistry` be passed around in a lot of function arguments.

amaurymartiny:
I think Goal 1 can be done in a separate PR, and afaict this PR is addressing Goal 2

ianhe8x:
at first the idea is just pass typeRegistry in constructor, but after starting it, I feel this approach not very ideal regarding
1. before we assume the first arg of constructor is always `value`, now we might have to add another that second arg is typeRegistry, even in where typeRegistry is not be needed.
2. it also means we will have to pass typeRegistry through a deep call stack to where it's used.

The use of static holder of typeRegistry is a tricky but effective way to avoid a large scale of changing in signature.

amaurymartiny:
> The use of static holder of typeRegistry is a tricky but effective way to avoid a large scale of changing in signature.

How it this different than the current solution of a singleton? It just _seems_ more "organized" since it's in a class, but it's still a singleton

ianhe8x:
it is api scoped, check https://github.com/polkadot-js/api/blob/b0270056242b089aac9c37bc247c87377ff9b811/packages/type-extrinsics/src/fromMetadata.spec.ts#L12
as long as the entry use withTypeRegistry and providing a instance of registry, the static one will be replaced by that with the call lifecycle, thanks for the fact js is single-threaded, we don't need to worry about thread safety.

ianhe8x:
in Struct.ts I put a protected typeRegistry member, so after an instance is created, if it has further need to access to typeregister, it can use the instance one since the static one will not be available at that moment.

ianhe8x:
Finally get all e2e tests working on my local node. I will start fixing other unit tests and clean the codes. Need advice @amaurymartiny @jacogr .

packages/types/src/codec/TypeRegistry.ts
```diff
@@ -0,0 +1,319 @@
...
+  static get TYPE_REGISTRY (): TypeRegistry {
+    assert(TypeRegistry._TYPE_REGISTRY, 'TYPE_REGISTRY is undefined, make sure it is wrapped in withRegistry()');
+    return TypeRegistry._TYPE_REGISTRY as TypeRegistry;
+  }
+
+  static withRegistry (typeRegistry: TypeRegistry, wrapFn: (...args: any[]) => any) {
+    TypeRegistry.TY_STACK.push(TypeRegistry._TYPE_REGISTRY);
+    TypeRegistry._TYPE_REGISTRY = typeRegistry;
+    const ret = wrapFn();
+    if (isPromise(ret)) {
```
ianhe8x: Supporting promise is probably a bad idea, the stack could be messed if two promises are resolved in a wrong order. will remove promise support later.

amaurymartiny:
If I understood you correctly:

`withRegistry(api.typeRegistry, () => f(arg1, arg2))` will:
  1. save current value of `TypeRegistry._TYPE_REGISTRY` to the stack
  2. temporarily set the singleton `TypeRegistry._TYPE_REGISTRY` to `api.typeRegistry`
  3. `f` will be called, and `f` can access the singleton `TypeRegistry.TYPE_REGISTRY` in its execution lifetime
  4. Once `f` is finished, we pop `TypeRegistry.TYPE_REGISTRY` to the value before

TBH I find this **really** hacky.

I would propose to see the following:

```diff
- withRegistry(api.typeRegistry, () => f(arg1, arg2))
+ withRegistry(api.typeRegistry, (registry) => f(arg1, arg2, registry))
```
and we need to refactor everywhere that uses the singleton. Reason: pure functions, easier to maintain, dependencies are clearly stated.

ianhe8x:
Well, if we pass typeRegistry all the way down to where it's needed, there's no need for withRegistry().
I tried and am really not convinced that passing it through is a better idea than this one.
Another way might work, which is instead of `export class `, we export `(typeRegistry)=>class xxx`. so we can still keep constructor as before, in additionally have an api scope typeRegistry access. But a large scale of signature changes are also inevitable.

amaurymartiny:
> Well, if we pass typeRegistry all the way down to where it's needed, there's no need for withRegistry().

> But a large scale of signature changes are also inevitable.

Agreed with both. TBH thinking long-term, I think that's the cleanest way to go.

ianhe8x:
To do that, we have to make it a convention that second parameter in all Codec's constructor is typeregister, even for types who don't need it. Otherwise it would be hard for Tuple, Vector and othet boxing types to decide whether and how to pass typeregister. And then for types don't have value as first parameter we need to use it like new Type(undefined, typeregister). Feels terrible and could cause confusion

jacogr:
There is a _LOT_ of work here. My 2 satoshis -

I'm certainly very cognisant on the ripple effects elsewhere, that is not inside our control, i.e. https://github.com/hicommonwealth/edgeware-node-types If it was all internally, I would be much more confident instead of us forcing updates everywhere (which is dead work is we ever get to the elusive #1328)

Make no mistake, we need to break the global link and have it local to API, but at the same time I feel we are solving a something that we can solve in cleaner ways down the line if we didn't have any external types being injected.

Issue #848, opened 2019-04-18
Rework polkadot/types overview docs page


Stefie:
Add short description to every type on https://polkadot.js.org/api/types/ so that people looking for information don't need to visit every type detail page to get an idea what a specific type is/ can/ should be used for.

Related to https://github.com/polkadot-js/apps/issues/1017

Stefie:
Related to https://github.com/polkadot-js/apps/issues/1017

PR #849, opened 2019-04-18
Add short description to all types on overview page

Stefie:
Adresses #848
Main changes on: https://polkadot.js.org/api/types/

The descriptions are mainly taken from the `@description` section of each type.

@amaurymartiny  / @jacogr Feedback to the added types would be nice. Specifically if they weren't added to the overview page on purpose and therefore should be removed again

**Types added to overview:**
Codec Types:
- AbstractArray
- VectorAny

Substrate types:
- AccountInfo 
- Seal
- DigestItem 
- Consensus
- AuthoritiesChange 
- Extrinsics 
- BftAuthoritySignature 
- BftHashSignature
- BftAtReport
- BftProposeOutOfTurn
- MisbehaviorKind 
- LocKPeriods 
- KeyValueOption 
- NonceCompact 
- RewardDestination
- NextAuthority

**Types removed from overview:**
- ReferendumInfoExtended (derived)

I removed the whole 'Derived types' section because `ReferendumInfoExtended` doesn't seem to have a type declaration anymore and `HeaderExtended` was then the only type listet in derive so I moved it to 'Substrate types'

### DIFF
--- a/packages/types/README.md
+++ b/packages/types/README.md
@@ -112,34 +112,36 @@ These custom types implement specific types that are found as part of the Substr
 | [[MisbehaviorKind]] | An [[EnumType]] containing a Bft misbehaviour |
 | [[MisbehaviorReport]] | A Misbehaviour report of [[MisbehavioirKind]] against a specific [[AuthorityId]] |
 | [[NewAccountOutcome]] | Enum to track the outcome for creation of an [[AccountId]] |
-| [[Nonce]] | Value1 |
-| [[Origin]] | Value1 |
-| [[ParaId]] | Value1 |
-| [[Perbill]] | Value1 |
-| [[Permill]] | Value1 |
-| [[PrefabWasmModule]] | Value1 |
-| [[PropIndex]] | Value1 |
-| [[Proposal]] | Value1 |
-| [[ProposalIndex]] | Value1 |
-| [[ReferendumIndex]] | Value1 |
+| [[NextAuthority]] | The next authority available as [[SessionKey]] |
+| [[Nonce]] | The Nonce or number of transactions sent by a specific account |
+| [[NonceCompact]] | The Compact<Nonce> or number of transactions sent by a specific account |
+| [[Origin]] | Where Origin occurs, it should be ignored as an internal-only value |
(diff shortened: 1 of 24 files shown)


packages/types/README.md
```diff
@@ -1,23 +1,165 @@
...
+| [[Nonce]] | The Nonce or number of transactions sent by a specific account |
+| [[NonceCompact]] | The Compact<Nonce> or number of transactions sent by a specific account |
+| [[Origin]] | Where Origin occurs, it should be ignored as an internal-only value |
+| [[ParaId]] | Identifier for a deployed parachain implemented as a [[U32]] |
+| [[Perbill]] | Parts per billion (see also [[Permill]]) |
+| [[Permill]] | Parts per million (See also [[Perbill]]) |
+| [[PrefabWasmModule]] | Struct to encode the vesting schedule of an individual account |
+| [[PropIndex]] | An increasing number that represents a specific council proposal index in the system |
+| [[Proposal]] | A proposal in the system. It just extends [[Method]] (Proposal = Call in Rust) |
+| [[ProposalIndex]] | An increasing number that represents a specific council proposal index in the system |
```
Stefie: `PropIndex` and `ProposalIndex` seem to do the exact same thing. Is there a reason to have both or can we delete one of the two?

amaurymartiny: One is in democracy, the other in treasury, so unfortunately we need them both for now

Stefie: Ah, ok. It just looked like an oversight, but that explains it.

amaurymartiny:
Ok to merge as-is, but I have 2 small styling complaints:
- there's a top blank line in the table
- the `[[]]` it not that pretty, maybe a hyperlink?

Screenshot 2019-04-23 at 12 45 51

Stefie:
> there's a top blank line in the table
That's unfortunately inevitable if I'm only using markdown and want a table without a header row. I didn't want to turn it into html table because that would have turned into a massive mess

> the [[]] it not that pretty, maybe a hyperlink?
Agreed, but I just left it as it was. This Syntax turns into a link in the actual VuePress docs and we're using this `[[]]` everywhere in the `.md` files

Issue #829, opened 2019-04-08
Allow api to take pre-bundled metadata information


xlc:
Currently every new instance of Api will query metadata information and then it will become usable.
This adds a significantly initialization time which is not ideal. Image someone wants to develop a serverless lambda with polkadot.js.
It will be good if the constructor of Api takes a metadata json object and if it presents, use it instead of fetch from server.
May optionally still fetch from server and print a warning if mismatch found.

jacogr:
As stand-along instances, I believe this goes against the grain - we are already not quite updating metadata when it changes on-chain, which is an issue. (e.g. https://github.com/polkadot-js/api/issues/449) 

If you are running multiple instances inside a single container, the `.clone()` allows you to pass through metadata.

xlc:
Outdated metadata should not cause issue. Of cause this is up to runtime developer to ensure backward compatibility. But for trivial upgrades, like add new module add new method/storage/event, it should preserve backward compatibility.

This also make it possible to create & sign transaction offline.

KarishmaBothara:
Plan to resolve the above issue as. 
In the API constructor (packages/api/src/Base.ts) if we can pass hex string of runtimeMetaData and its related runtimeVersion as part of ApiOptions and pass it via init method to loadMeta function, where we check if the runtimeVersion passed is same as **this._rpcBase.chain.getRuntimeVersion()** if yes then we assign this.runtimeMetaData to the one passed by user or else get it from the chain i.e


```
private async loadMeta (runtimeVersion, rpcdata ): Promise<boolean> {
   if (!this._options.source || !this._options.source._isReady) {
      this._runtimeVersion = await this._rpcBase.chain.getRuntimeVersion();
      const metadata = new Metadata(rpcdata).asV0;
      this._runtimeMetadata =  this._runtimeVersion === runtimeVersion ?  metadata : await this._rpcBase.state.getMetadata();
   }
}

```
Request your feedback

xlc:
It may help a bit but does not fundamentally solve my issue.

The current API requires a message roundtrip before it is useable. It can take few seconds on mobile network. Replace it with another simple API only reduce the time a bit but won't help with the long latency in mobile app.

And it doesn't help for the offline usage.

jacogr:
Don't misread my comments above as "this can't work or shouldn't get it", rather read it as "this is a very dangerous and bad idea if you don't know what you are doing".

So what does the API does on init, so you only seem to cover one part -

- load meta, used for decoding
[...]

PR #850, opened 2019-04-23
Api pre-bundled metadata

KarishmaBothara:
#829


packages/api/src/Base.ts
```diff
@@ -374,12 +374,27 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
...
+        this._rpcBase.chain.getBlockHash(0),
+        this._rpcBase.chain.getRuntimeVersion()
+      ]);
+      const prebundlesKey = `${this._genesisHash}-${(this._runtimeVersion as RuntimeVersion).specVersion}`;
+      if (prebundlesKey in prebundles) {
+        try {
+          const rpcData = prebundles[prebundlesKey];
+          const metadata = new Metadata(rpcData);
+          this._runtimeMetadata = metadata;
+          this.runtimeMetadata.getUniqTypes(false);
```
amaurymartiny: Here it needs to be true, so that this function throws an error (in case there's a mismatch), get caught, and do the RPC call

packages/api/test/e2e/promise-metadata.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    const specVersion = '62';
+    const prebundles = {};
+    const key = `${genesisHash}-${specVersion}`;
+    prebundles[key] = rpcData;
+    const api = await ApiPromise.create({
+      prebundles
+    });
+
+    expect(api.genesisHash).toBeDefined();
+    expect(api.runtimeMetadata).toBeDefined();
```
amaurymartiny: `expect api.runtimeMetadata.toHex()` to be rpcData

packages/api/test/e2e/promise-metadata.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    expect(api.tx).toBeDefined();
+    expect(api.derive).toBeDefined();
+  });
+
+  it('Create API instance with incorect metadata map and makes the runtime, rpc, state & extrinsics available', async() => {
+    const rpcData = 'invalid data';
+    const genesisHash = '0x8579ded4438b26597e188cda8579fdd6abf09b74ab672e604b9584cd8b4c3d0b';
+    const specVersion = '62';
+    const prebundles = {};
+    const key = `${genesis
```
amaurymartiny: `expect api.runtimeMetadata.toHex()` to be rpcData (or: not to be 'invalid data')

packages/api/test/e2e/promise-metadata.spec.js
```diff
@@ -0,0 +1,68 @@
...
+    expect(api.runtimeVersion).toBeDefined();
+    expect(api.rpc).toBeDefined();
+    expect(api.query).toBeDefined();
+    expect(api.tx).toBeDefined();
+    expect(api.derive).toBeDefined();
+  });
+
+  it('Create API instance with incorect metadata map and makes the runtime, rpc, state & extrinsics available', async() => {
+    const rpcData = 'invalid data';
+    const genesisHash = '0x8579ded4438b26597e188cda8579fdd6abf09b74ab672e604b9584cd8b4c3d0b';
```
amaurymartiny: It's kind of annoying that each time I want to run this test with my local node, I need to change this line manually. Maybe we could create a promise-alex.spec.js or promise-ee.spec.js.

Edit: on top of an e2e test, we could actually add a unit test, with the mockProvider (inside rpc-provider) that will return these pre-defined values (hash, rpcData, specVersion)

packages/api/src/Base.ts
KarishmaBothara: Thanks for your comment.. May I know under what conditions we want to throw error.
I assume, if the rpcData(hex) is invalid then `new Metadata(rpcData)` will throw error and handle it. However if we have `this.runtimeMetadata.getUniqTypes(true);`, it will fail even with valid metadata with new types.  
Would like to understand the intention for this validation. Is there a case where in, genesis hash and spec version can produce two different metadata?

amaurymartiny: Basically, the question is: if there is a type mismatch between pre-bundled metadata and types in typeRegistry, what do we do?

Solution 1: `this.runtimeMetadata.getUniqTypes(true);`, this will throw an error in the `try/catch` block, so will disregard pre-bundled, and fetch metadata from RPC call.

Solution 2: `this.runtimeMetadata.getUniqTypes(false);` this will console.error a message, but will still use pre-bundled metadata. There might be a panic further down the road, when decoding a type that is no found in typeRegistry.

I prefer solution 1: it's safer.

> Is there a case where in, genesis hash and spec version can produce two different metadata

This should be fine: genesis+specVersion should uniquely identify metadata

ianhe8x: If there's such type mismatch, why fetch metadata from node can make any different? Unless the result is different than the prebundled one, then the question is how? If it's by mistyping, then i believe it will fail at new Metadata(), or it could be malicious？

amaurymartiny: > If there's such type mismatch, why fetch metadata from node can make any different?

That's the whole problem here: if we allow pre-bundled metadata, we need to trust that the runtime developer has put the correct metadata in `new Api({ prebundles })`. He can totally put a correct but outdated metadata (not malicious, but by mistake), and new Metadata() will not fail, nothing will fail on init; but some time later, something might fail (because of outdated metadata) and that will be super hard to debug.

So in my opinion there needs to be some sort of check that checks "pre-bundled metadata is compatible with node", or else it's hard-to-debug decoding errors guaranteed. I was thinking `getUniqTypes()` might be a good check, but after 2nd thought, it's not good enough. A good check would be to do `state_getMetadata` RPC call, and show warning if 2 are not same, as xlc proposed https://github.com/polkadot-js/api/issues/829#issue-430684062. But then, what's the point of having a pre-bundled metadata?

amaurymartiny: Another solution would be to keep @karishma's current implementation (i.e. without getUniqTypes check), but just add a `console.warn` clearly stating that this is an experimental feature, and that developers should be careful with this. wdyt @jacogr ?

packages/api/test/e2e/promise-metadata.spec.js
KarishmaBothara: Metadata.toHex() operation is not supported..
Screen Shot 2019-04-30 at 10 58 24 AM

packages/api/test/e2e/promise-metadata.spec.js
KarishmaBothara: Screen Shot 2019-04-30 at 10 58 24 AM
api.runtimeMetadata.toHex() is not supported, throws throw as shown in the screenshot attached.

packages/api/test/e2e/promise-metadata.spec.js
KarishmaBothara: Thanks..Changed as per suggestion.

packages/api/src/Base.ts
jacogr: Happy with the warning.

jacogr: Addressed, finally looking at the code now.

packages/api/src/Base.ts
```diff
@@ -374,12 +374,27 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
   }
 
   private async loadMeta (): Promise<boolean> {
+    const { prebundles = {} } = this._options;
```
jacogr: (nitpick) Blank line between var definitions and code blocks.

packages/api/src/Base.ts
```diff
@@ -374,12 +374,27 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
...
+      [this._genesisHash, this._runtimeVersion] = await Promise.all([
+        this._rpcBase.chain.getBlockHash(0),
+        this._rpcBase.chain.getRuntimeVersion()
+      ]);
+      const prebundlesKey = `${this._genesisHash}-${(this._runtimeVersion as RuntimeVersion).specVersion}`;
+      if (prebundlesKey in prebundles) {
+        try {
+          const rpcData = prebundles[prebundlesKey];
+          const metadata = new Metadata(rpcData);
+          this._runtimeMetadata = metadata;
```
jacogr: Why not just `this._runtimeMetadata = new Metadata(rpcData);` or even just `this._runtimeMetadata = new Metadata(prebundles[prebundlesKey]);`? Don't see metadata const used anywhere else.

packages/api/src/Base.ts
```diff
@@ -374,12 +374,27 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
...
+          l.warn('Experimental feature. Developers to be carefully while adding prebundled metadata');
+        } catch (e) {
+          this._runtimeMetadata = await this._rpcBase.state.getMetadata();
```
jacogr: Not 100% on this at all - we have 2 blocks following that does exactly the same. Honestly, I would not catch the error here - if it fails, it fails. We do expect valid things to be added, not just swallowed when invalid.

packages/api/src/promise/Api.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  it('Create API instance with metadata map and makes the runtime, rpc, state & extrinsics available', async () => {
+    const rpcData = await mock.send('state_getMetadata',[]);
+    const genesisHash = await mock.send('chain_getBlockHash',[]);
+    const specVersion = await mock.send('chain_getRuntimeVersion', []);
+    const prebundles: any = {};
+    const key = `${genesisHash}-${specVersion}`;
+    prebundles[key] = rpcData;
+    const api = await ApiPromise.create({ provider: mock, prebundles } as ApiOptions);
+    expect(api.genesisHash).toBeDefined();
```
jacogr: (nitpick) Blank line between vars and actual expects - doesn't const anything and reads cleaner.

packages/api/src/types.ts
```diff
@@ -132,6 +132,13 @@ export interface ApiOptions {
    * uses types not available in the base Substrate runtime.
    */
   types?: RegistryTypes;
+  /**
+   * @description prebundles is a map of 'genesis hash and runtime spec version' as key to metadata's hex string
+   * if genesis hash and runtime spec version matches, then use metadata, else fetch it from chain
+   */
+  prebundles?: {
```
jacogr: Alphabetical ordering of properties.

packages/api/src/Base.ts
```diff
@@ -374,12 +374,27 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
...
+        this._rpcBase.chain.getBlockHash(0),
+        this._rpcBase.chain.getRuntimeVersion()
+      ]);
+      const prebundlesKey = `${this._genesisHash}-${(this._runtimeVersion as RuntimeVersion).specVersion}`;
+      if (prebundlesKey in prebundles) {
+        try {
+          const rpcData = prebundles[prebundlesKey];
+          const metadata = new Metadata(rpcData);
+          this._runtimeMetadata = metadata;
+          l.warn('Experimental feature. Developers to be carefully while adding prebundled metadata');
```
jacogr: The issue with the warning here as it stands (since it is not tied to anything) is that it will pop up for all users - which is going to create more issues for mom-and-pop asking what is wrong ;) Not sure how to get around it at all, so I'd rather err on the side of not having it.

packages/api/src/types.ts
jacogr: Not sold on the naming of `prebundles` at all. Doesn't tell me what it is. `source.metadata[${genesisHash}-${specVersion}]` works as well

KarishmaBothara: Thanks for your comment.. I believe I can't simply put metadata(adding Partial to ApiBase<any, any>) in source as doing that, I would also need other things like _isReady, genesisHash, runtimeVersion. 
For now I renamed prebundles to metadata.

packages/api/src/promise/Api.spec.ts
KarishmaBothara: done

packages/api/src/Base.ts
KarishmaBothara: done

packages/api/src/Base.ts
KarishmaBothara: done

packages/api/src/Base.ts
KarishmaBothara: done

packages/api/src/types.ts
KarishmaBothara: done

packages/api/src/promise/Api.spec.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright 2017-2019 @polkadot/api authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import ApiPromise from '../promise';
+import { ApiOptions } from '../types';
+import Hash from '../../../types/src/type/Hash';
```
jacogr: This is on the border of a nitpick since it is a test, however - would rather import this from ‘@polkadot/types’ (Same 2 lines below) instead of relative.

KarishmaBothara: done

PR #857, opened 2019-04-23
Use custom hasher for Map

amaurymartiny:
Implement https://github.com/paritytech/substrate/pull/2268

Most notable changes:
- Add MetadataV4
- type-storage's `fromMetadata` now takes as argument a `Metadata` instead of a `MetadataV0`, and has different logic depending on meta version.

Cleaned some stuff:
- in v0 we had `default: Bytes` and `documentation: Vector<Text>`. Somehow in v1-v3 we now have `fallback: Bytes` and `docs: Vector<Text>`, no idea why this change. I harmonized to have `fallback: Bytes` (as initially agreed, b/c `default` is a reserved JS word) and `documentation: Vector<Text>`, because Rust also uses that.
  **Note**: stricto sensu this is a breaking change, because `api.query.some.method.meta.get('default')` won't work anymore, but I believe people use the `api.query.some.method.meta.default` getter, which still works
- renamed classes used in Metadata to follow exactly names in Rust code, e.g. `MetadataStorageType -> StorageFunctionType`
- finally added a v0 and v4 distinction in type-storage, the `.toV0()` workaround really is not maintainable.


packages/types/src/Metadata/v3/Storage.ts
```diff
@@ -13,13 +13,20 @@ import { MapType, PlainType } from '../v2/Storage';
 export class DoubleMapType extends Struct {
   constructor (value?: any) {
     super({
+      keyHasher: Text,
```
xlc: I don't think you can do this. The order have to match on substrate side

xlc:
In order to make metadata still contain all the necessary information when converted to v0 and not breaking backward compatibility, you need to do the same trick I did with `is_linked` with `hasher`

packages/types/src/Metadata/v0/Modules.ts
```diff
@@ -124,7 +124,7 @@ export class ModuleMetadata extends Struct {
 
 export class StorageFunctionModifier extends Enum {
   constructor (value?: any) {
-    super(['Optional', 'Default', 'Required'], value);
+    super(['Optional', 'fallback', 'Required'], value);
```
xlc: This is a backward compatibility breaking change, which doesn't matter anymore because no one should be using v0 now.

amaurymartiny: Ah, nice catch! This is a find/replace gone wrong, I'll fix it.

amaurymartiny: > which doesn't matter anymore because no one should be using v0 now

It actually matters, because v1-v4 all use v0's StorageFunctionModifier

amaurymartiny:
e2e tests work for storage!

TODO:
- [ ] Have a look if type-extrinsics needs some change (probably not)
- [ ] Fix TS type errors introduced with `StorageFunction<Meta>`

packages/type-storage/src/fromMetadata/v4/substrate.ts
```diff
@@ -0,0 +1,61 @@
...
+  type: 'u64'
+});
+
+export const authorityCount = createRuntimeFunction('authorityCount', ':auth:len', {
+  documentation: 'Number of authorities.',
+  type: 'u32'
+});
+
+export const authorityPrefix = createRuntimeFunction('authorityPrefix', ':auth:', {
+  documentation: 'Prefix under which authorities are storied.',
```
ltfschoen: typo `storied` should be `stored`, we'd also need to update paritytech/substrate/core/primitives/src/storage.rs where this came from

packages/type-storage/src/fromMetadata/v4/createFunction.ts
```diff
@@ -0,0 +1,81 @@
+// Copyright 2017-2019 @polkadot/storage authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { Compact, Text, createType, StorageKey, Bytes, U8a } from '@polkadot/types';
```
ltfschoen: change order `createType, Bytes, Compact, StorageKey, Text, U8a`?

packages/type-storage/src/fromMetadata/v4/createFunction.ts
```diff
@@ -0,0 +1,81 @@
...
+import { PlainType, StorageFunctionMetadata, StorageFunctionModifier, StorageFunctionType } from '@polkadot/types/Metadata/v4/Storage';
+import { StorageFunction } from '@polkadot/types/primitive/StorageKey';
+import { assert, isNull, isUndefined, stringLowerFirst, stringToU8a, u8aConcat } from '@polkadot/util';
+
+import getHasher from './getHasher';
+
+export interface CreateItemOptions {
+  isUnhashed?: boolean;
```
ltfschoen: alphabetical order

Stefie:
I tested your branch with `./apps`, it's working without problems with v3 and v4 metadata.

amaurymartiny:
Works on substrate-light-ui too

PR #861, opened 2019-04-29
Bump deps

jacogr:
Also updated the `combineLatest` as marked in https://github.com/polkadot-js/api/pull/861 (also pinned rxjs until 6.5.2 hits)


packages/api/package.json
```diff
@@ -29,13 +29,13 @@
...
-    "@babel/runtime": "^7.4.3",
+    "@babel/runtime": "^7.4.4",
     "@polkadot/api-derive": "^0.78.0-beta.0",
     "@polkadot/extrinsics": "^0.78.0-beta.0",
     "@polkadot/rpc-provider": "^0.78.0-beta.0",
     "@polkadot/rpc-rx": "^0.78.0-beta.0",
     "@polkadot/storage": "^0.78.0-beta.0",
     "@polkadot/types": "^0.78.0-beta.0",
-    "@polkadot/util-crypto": "^0.76.1"
+    "@polkadot/util-crypto": "^0.90.0-beta.1"
```
Stefie: As a follow up from Riot: 0.90.0 is also just random here?

jacogr: Since it is closer that API to 1.0 bumped common to 0.9 since it should hit 1.0 first - it should alleviate confusion as well, i.e. why are they the same.

PR #865, opened 2019-05-03
Update WithdrawReasons to match substrate

### LINKED ISSUE #864: balances.locks throws Mismatch decoding on a stash accounts
This happens for both Substrate and Alexander and only on stash accounts AFAICS.

`Mismatch decoding '15', computed as '7' with TransactionPayment,Transfer,Reserve`

To reproduce:
In extrinsics: balances > locks on a stash account e.g `Alice_stash` in substrate

or in the JS console:
```js
const [balancesLocksSubstrate, balancesLocksAlexander ] = await Promise.all([
  api.query.balances.locks('5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY'),
  api.query.balances.locks('5DWrJzdCLLEXFndG6qjCNaMFU74sMEdNwg8TuLLwRiGJcYsn')
]);

console.log('locks: ' + balancesLocksSubstrate);
console.log('locks: ' + balancesLocksAlexander);
```
image

xlc:
Should fixes #864 

https://github.com/paritytech/substrate/blob/3669cf7b7e2ab6bb130715a082f4db9b6014743b/srml/support/src/traits.rs#L471

### DIFF
--- a/packages/types/src/type/WithdrawReasons.ts
+++ b/packages/types/src/type/WithdrawReasons.ts
@@ -10,11 +10,12 @@ import Set from '../codec/Set';
  * The Substrate WithdrawReasons for staking
  */
 export default class WithdrawReasons extends Set {
-  constructor (value: any = ['header', 'body', 'justification']) {
+  constructor (value: any = []) {
     super({
       TransactionPayment: 0b00000001,
       Transfer: 0b00000010,
-      Reserve: 0b00000100
+      Reserve: 0b00000100,
+      Fee: 0b00001000
     }, value);
   }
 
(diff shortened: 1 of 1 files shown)


packages/types/src/type/WithdrawReasons.ts
```diff
@@ -10,11 +10,12 @@ import Set from '../codec/Set';
  * The Substrate WithdrawReasons for staking
  */
 export default class WithdrawReasons extends Set {
-  constructor (value: any = ['header', 'body', 'justification']) {
+  constructor (value: any = []) {
```
jacogr: O, my word. This is a serious copy-and-paste issue from me :(

PR #866, opened 2019-05-03
Basic same-type multi queries

jacogr:
- Replaces https://github.com/polkadot-js/api/pull/768
- Part of https://github.com/polkadot-js/apps/issues/800


jacogr:
In addition to the e2e test here, also tested the indices against apps

Issue #827, opened 2019-04-08
Check derive methods for broken implementations with substrate 1.0


Stefie:
confirmed broken:
~~subscribeNewHead~~ Doesn't seem to appear broken anymore
~~account.id lookups~~ Fixed by https://github.com/polkadot-js/api/pull/831/

jacogr:
Do we have something outstanding here?

jacogr:
Closing due to #871

PR #871, opened 2019-05-03
Fix derive methods

Stefie:
- v3 / v4 switch implementation for derive methods
- fix broken methods

Addresses https://github.com/polkadot-js/api/issues/827


Stefie:
@amaurymartiny  @jacogr  Could I get a quick feedback before I continue with the implementation on whether the way I'm doing the v3 / v4 switch implementation for derive methods is fine or not?

My thoughts were:

- Pass the metadata version to decorateDerive in api/Base.ts
- Always treat the current version as default (as of now v4)
- If the version is 1,2 or 3, load method overwrites from a folder called `v3` to make the derive package compatible with previous metadata versions.

That way we only need to delete the v3 folder and remove the conditional overwrite when we decide to drop v3 support.

PS: This is only a first draft and WIP, so I'm aware of console.logs/ comments etc.

amaurymartiny:
@Stefie tbh I don't think v3/v4 separation makes sense in api-derive.

The change `api.staking.validators() -> api.session.validators()` didn't happen during the v3->v4 transition, but well before. A change like this can happen on a block-basis (runtime upgrade), it's totally decoupled from the v3/v4 upgrades.

Catering for both staking.validators & session.validators is a difficult problem which I'm not sure how to solve easily. I'd propose to remove completely staking.validators, but I don't have a big picture of if any testnets/external teams still use staking.validators, @jacogr might know more if it's safe to remove

ltfschoen:
> @Stefie tbh I don't think v3/v4 separation makes sense in api-derive.
> 
> The change `api.staking.validators() -> api.session.validators()` didn't happen during the v3->v4 transition, but well before. A change like this can happen on a block-basis (runtime upgrade), it's totally decoupled from the v3/v4 upgrades.
> 
> Catering for both staking.validators & session.validators is a difficult problem which I'm not sure how to solve easily. I'd propose to remove completely staking.validators, but I don't have a big picture of if any testnets/external teams still use staking.validators, @jacogr might know more if it's safe to remove

we're still using it https://github.com/polkadot-js/apps/blob/master/packages/app-staking/src/Account/index.tsx#L478, which doesn't agree with https://github.com/paritytech/polkadot/blob/master/service/src/chain_spec.rs#L102

Stefie:
@amaurymartiny I went for this solution because that's what I took away as the prefered way to go from when I asked on Riot the other day. 

I tried different ways to implement backwards compatibility and having separate folders seemed to be the cleanest way to do it.
But I see that it's decoupled from the versions...

What would be the cleanest solution you can think of? Just dumping the support for older methods?

Stefie:
@ltfschoen
> we're still using it https://github.com/polkadot-js/apps/blob/master/packages/app-staking/src/Account/index.tsx#L478, which doesn't agree with https://github.com/paritytech/polkadot/blob/master/service/src/chain_spec.rs#L102

 That's fine (and I think on purpose) since Polkadot and I think also Embric Elm haven't been updated to the latest substrate, yet. That's why having backwards compatibility would be nice, to support both older versions of the runtime and the latest Substrate version

amaurymartiny:
If we keep backwards compat, then yes, folders would be the cleanest (would need to think of a good naming for these folders).

Then somewhere,
```
if (isFunction(api.staking.validator)) { `folder1` } else { `folder2` }
```

I think we can go like this for now.

Stefie:
Question is also if we only want to implement temporary backwards compatibility (as in 'until polkadot got updated, then remove'), then we could just do { `polkadot` } { `substrate` } as { `folder1` } else { `folder2` } or we just use substrate git commit hashes, which is not the prettiest, but we probably won't need that distinction anymore anyways  once substrate stabilizes

jacogr:
Both `staking.validators` and `session.validators` actually do exist. The one as a linked-map, eg. https://github.com/paritytech/substrate/blob/master/srml/staking/src/lib.rs#L451 (you can get the full list of an entry in the list),  the other as a vec, eg. https://github.com/paritytech/substrate/blob/master/srml/session/src/lib.rs#L201

It is not a v3/v4 split thing, would remove that completely.

jacogr:
> we're still using it https://github.com/polkadot-js/apps/blob/master/packages/app-staking/src/Account/index.tsx#L478, which doesn't agree with https://github.com/paritytech/polkadot/blob/master/service/src/chain_spec.rs#L102

Not sure why you are pointing to this as opposed to the actual srml. 

The 2 serve very different purposes, the UI code as linked is correct as to the purpose. It retrieves (since you have the stashId), the preferences - without a param it will (a very expensive query) perform the linked-map retrieval magic, basically starting at head and moving down until it has all entries. 

To re-iterate -

- session - current validators
- staking - all (previously called intentions)

Stefie:
@jacogr The folder structure was meant as a way to make the derive methods backwards compatible, the derive.staking.validators was just an example. Amaury already pointed out that it was independent of v3/v4. 

The question for me is whether we want `api-derive` to be backwards compatible with older versions of the runtime or not.
If yes, a folder structure is the cleanest solution. If not, I'm happy to just drop support for older versions, but that will likely break things for for applications using api-derive with older versions of the runtime.

I'm fine either way, I'd just strongly vote against adding some kind of conditionals inside the actual derive function calls because that will become hard to maintain.

jacogr:
Conditionals with a linked issue is easier imho, otherwise it almost becomes an all-or-nothing. Would def. no introduce yet more metadata folders since we cannot even properly cope with what we have.

jacogr:
There seems to be some confusion with versions here and what they mean and are intended for. There are 3 versions in substrate that are tracked (the latter 2 have open issues to track properly) -

- Metadata version
- Spec version
- Implementation version

Here we seem to equate the first to the last 2. So what is the metadata version? Is is an indication of the binary format used to exchange information around the modules. It does not care about the data, but rather the structure of the binary structures. So a history of these -

- v0 - The first version, rough and ready. It had a huge issue that we couldn't mark backwards-incompatible changes, i.e. when we added defaults, we just had to align all upgrades to happen at the same time - we couldn't support both. So enters v1...
- v1 - Added versioning and magic number. In addition it completely reworked the structures, so instead of us having the storage, calls and events separately, we know have modules with each of those sections.
- v2 - Linked-map support introduced
- v3 - Double-map support introduced (not used yet, so API only allows parsing, not usage)
- v4 - Configurable storage hasher support

The above is only focussed on the metadata formatting, not the modules or their versions. So when a new modules is introduced, the actual spec version is incremented (and impl. version normally set to the same) and subsequent patches or internal changed bump the impl. version.

Spec/impl has been used for signing in the past, i.e. when > 256 bytes and new version sign the hash, otherwise sign the data, and it is not a smooth process. For one each client implementation has their own versioning, i.e. Polkadot is independent of substrate, other clients as well (In most cases not, but it is the intent). So to do per-version feature toggles you need to -

- determine the client version
- determine the client type
- do magic, i.e. some old magic signing detection magic https://github.com/polkadot-js/api/blob/1ea1f6f9f56bdeb75fe095961a53ac301f6845a9/packages/types/src/type/SignaturePayload.ts#L98

Metadata versions are independent of modules, module names, module functions and actual implementation. A simple example, when fees was introduced, the spec in substrate changed and the polkadot spec changed as well (delayed), subsequently when it was removed the spec versions changed again. (And clients with different names were in various stages in-between), i.e. https://github.com/polkadot-js/api/blob/1ea1f6f9f56bdeb75fe095961a53ac301f6845a9/packages/api-derive/src/balances/fees.ts#L18 - there was no metadata bump anywhere close to this.

TL;DR - You certainly cannot do module function detection based on metadata, nor can are they dependent on each other (There may be changes in both at any point, but breaking changes in the one doesn't mean breaking changes in the other)

Stefie:
Split into smaller chunks, so I'm closing this.

Issue #835, opened 2019-04-10
Support Arrays & Tuples for contract ABI


jacogr:
As defined here https://github.com/Robbepop/pdsl/issues/40

`types/ContractAbi`

PR #873, opened 2019-05-07
feat: add vec & tuple contract abi return types

yjkimjunior:
#835

Question: currently the e2e tests for the contract abi promise-contract.spec.js is marked as `skip`, though the tests actually pass (except for the contract call test, which fails sometimes depending on timeout). 

What was/is the plan for this before it can be used? I'm planning on adding a function in incrementer that returns a (i32, bool) for example and using the generate wasm for testing.


packages/types/src/ContractAbi.ts
```diff
@@ -22,7 +25,7 @@ export type ContractABIMethod = ContractABIMethodBase & {
   mutates?: boolean,
   name: string,
   selector: number,
-  return_type: string | null
+  return_type: string | Vector<Bool | I8 | I16 | I32 | I64 | I128 | U8 | U16 | U32 | U64 | U128> | Tuple | null
```
amaurymartiny: We could create a type for this (it's in 2 places)

jacogr:
Are we actually supporting the parsing of the ABI as defined in https://github.com/paritytech/ink/issues/40 (Have not seen a real sample in the wild, but from the description there the arrays looks very different from anything we do elsewhere)

yjkimjunior:
> > Question: currently the e2e tests for the contract abi promise-contract.spec.js is marked as skip, though the tests actually pass (except for the contract call test, which fails sometimes depending on timeout).
> 
> The reason why we skipped e2e tests is that they require a running local node - which we don't have on travis. I think it'll be overkill to set up (even dev) nodes on travis, so the usual thing I do is remove the `.skip`, test locally on my machine, and if everything passes re-put the .skip.
> 
> The alex tests could be unskipped, but for now I'd argue to leave them skipped to prevent flaky tests, maybe when things get mature enough we can add back e2e tests with remote nodes

ah, that makes sense

packages/api/test/e2e/promise-contract.spec.js
```diff
@@ -12,7 +12,7 @@ import incrementer from '../data/incrementer.json';
 import erc20 from '../data/erc20.json';
 import Api from '../../src/promise';
 
-describe.skip('e2e contracts', () => {
+describe('e2e contracts', () => {
```
jacogr: .skip

packages/api/test/e2e/promise-contract.spec.js
```diff
@@ -89,6 +89,21 @@ describe.skip('e2e contracts', () => {
           }
         });
     });
+
+    it('decodes tuple return types', (done) => {
+      expct(address).toBeDefined();
+
+      api.tx.contract
+        .call(address, 12345, 500000, abi.messages.compare_and_get(1, 2))
```
jacogr: Don't see `compare_and_get` in the in-repo ABI?

yjkimjunior: I added this as a dumb function just to get a (u64, u64) return value to test on. Would it be better to have an entirely separate contract instead?

jacogr: Would really like to see a valid ABI, so we are sure -

- the ABI parsing is done correctly
- the e2e testing is done correctly

Adding an additional contract + ABI is fine for these

jacogr: To to expand on this - so basically take a peek at https://github.com/polkadot-js/api/pull/906 - it just adds the wasm for an updated test with the ABI.

So when we finally do `Option<T>` and `Result<T>` (and `Vec<T>` which has the same form), that is in place for tests -

- so the ABI parsing can be updated, i.e https://github.com/polkadot-js/api/blob/master/packages/types/src/ContractAbi.ts - that certainly does not cater for this type of parsing (also not sure of having it in "types", but could convince myself that it either belongs in types or api... well, needs to be _somewhere_)
- Once that is find and it can encode/decode based on the definitions, we can add a specific test that uses the ABI + WASM to test on-chain

I hope that explains my previous comments (slightly) better.

yjkimjunior:
Just upgraded my local substrate node to 2.0.0 and incrementer tests are failing

Stefie:
Which tests are failing for you? I just noticed that they are failing in my JS -> TS branch, too. That's probably for different reasons, but I'd generally suggest that they should be self-contained so that one test doesn't rely on the success/ result of another test.

At the moment, both the `allows contract create'` and `allows contract call` tests need a value from a previous passed test to run. If the previous test fails they never test what they're supposed to test.

jacogr:
@yjkimjunior Quite possible. The contracts are under active development, so these probably need an upgrade at this point. (Re-compile of the wasm). These were the first version that were actually working, but I saw in the ink! channel that even 2 week old contracts are having some issues atm due to the recent changes.

Got a fresh `flipper` from @kwingram25 this morning, I'll write the same tests around that one and see if they work - if so, can replace. (The test as they stand doesn't matter which contract, as long as it is _a_ contract and can be deployed and called with some events resulting from the call)

yjkimjunior:
@Stefie should've mentioned this also happens on master disregarding my changes.

@jacogr watched my computer burst into flames building and recompiling latest version of everything and it still seems to be an issue. I have an idea for how to actually parse the new ret_types though, but investigating this first before moving forward.

---
Done so far:
1. Created and compiled a dummy contract that simply returns the newly implemented ink! return types (fixed size vector and array)
2. Moved the ABI and wasm into `api/test/data` next to the `erc20` and `incrementer`
3. In `ContractAbi.validateMethods`, added logic to also allow an array return type (how Tuples are represented) as well as an object return type (how Vectors are represented) from each method
4. Added a `describe` block that creates that contract and then tries to call each of those new methods, and asserts the Extrinsic was a Success.

What I’ve understood so far:
1. I see that before these new return types were implemented, in the `ContractAbi` just expecting a `string | null` return type was enough for the primitive types. So I added a new return type for the Vector and Tuple representation jsons
2. I see the `ContractAbi` constructor goes through each of the methods specified in the abi and encodes the method as a ContractAbiFn to its messages property.
3. For each method getting encoded, you create a new constructor function that is basically a key value pair of each of the fields in the abi and their types.
    1. Right now, they are just being set with `base[name] = type` because type is just going to be a primitive.
    2. From my understanding, this is where I will need to put in the logic to parse the Vector and Tuple types and set the appropriate SCALE codec type
4. Then we define an encoder function that takes the method’s expected params and returns a bytes array representation of the method name mapping with its return type, padded with a length prefix.
5. Then that function gets added as a message in the `ContractAbi`’s `messages` field, which can later be queried and called with something like`contractAbi.messages[key]`

jacogr:
@kwingram25 Also had endless issues getting it going - it ended up being not compiling substrate with `--release`

Less worried about the returns, since, well, the contracts cannot return anything - everything via events.

The rest sounds sane.

PR #875, opened 2019-05-07
Add contract fees

jacogr:
Pull all contract fees, part of https://github.com/polkadot-js/apps/issues/983

### DIFF
--- a/packages/api-derive/src/balances/fees.ts
+++ b/packages/api-derive/src/balances/fees.ts
@@ -16,12 +16,8 @@ export function fees (api: ApiInterface$Rx) {
     return (combineLatest([
       api.query.balances.creationFee(),
       api.query.balances.existentialDeposit(),
-      api.query.fees
-        ? api.query.fees.transactionBaseFee()
-        : api.query.balances.transactionBaseFee(),
-      api.query.fees
-        ? api.query.fees.transactionByteFee()
-        : api.query.balances.transactionByteFee(),
+      api.query.balances.transactionBaseFee(),
+      api.query.balances.transactionByteFee(),
       api.query.balances.transferFee()
     ]) as any as Observable<[BN, BN, BN, BN, BN]>).pipe(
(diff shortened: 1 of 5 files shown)


packages/api-derive/src/balances/fees.ts
```diff
@@ -16,12 +16,8 @@ export function fees (api: ApiInterface$Rx) {
     return (combineLatest([
       api.query.balances.creationFee(),
       api.query.balances.existentialDeposit(),
-      api.query.fees
-        ? api.query.fees.transactionBaseFee()
-        : api.query.balances.transactionBaseFee(),
-      api.query.fees
-        ? api.query.fees.transactionByteFee()
-        : api.query.balances.transactionByteFee(),
+      api.query.balances.transactionBaseFee(),
```
jacogr: Fees modules has been removed for a long time now, safe to remove.

Stefie: Is that how you'd solve it for all deprecated methods?

jacogr: Depends on the actual case. So `transactionBaseFee` used to be in balances, then got moved to fees (and some networks supported it, some others not) and then got moved back from fees to balances.

In this case is was _exactly_ the same, just the module changed. (And quite easy to detect)

Same has gone in the past with the split into the `indices` modules (the detection there is already removed). These are easy cases - where the functionality ceases to exist, it is better just having a new file where things are done differently. (e.g. old intentions vs new controllers in staking)

Generally when this happens, it is viewed as a short-term override. As soon as the official Polkadot & Substrate test networks have been updated, it _should_ be removed. (Sadly these have not been logged each and every time - generally we do log these "support multiples" as issues, but I believe this specific one slipped through)

Stefie: Ok, so I'll add `support multiples`  to mine. Maybe with a date or commit hash of when it was changed in substrate master?

PR #878, opened 2019-05-07
Fix linked maps

amaurymartiny:
The `.toString()` of StorageFunctionType was completely wrong after #857, which caused `Linkage<T>`  types not being parsed correctly

### DIFF
--- a/packages/type-storage/src/fromMetadata/v4/createFunction.ts
+++ b/packages/type-storage/src/fromMetadata/v4/createFunction.ts
@@ -45,7 +45,7 @@ export default function createFunction (section: Text | string, method: Text | s
       assert(!isUndefined(arg) && !isNull(arg), `${meta.name} expects one argument`);
 
       const type = meta.type.asMap.key.toString();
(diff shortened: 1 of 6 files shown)


packages/type-storage/src/fromMetadata/v4/createFunction.ts
```diff
@@ -45,7 +45,7 @@ export default function createFunction (section: Text | string, method: Text | s
       assert(!isUndefined(arg) && !isNull(arg), `${meta.name} expects one argument`);
 
       const type = meta.type.asMap.key.toString();
-      const param = createType(type, arg).toU8a(false);
+      const param = createType(type, arg).toU8a();
```
jacogr: Are we sure about this? It will add the prefixes as it stands? (Or did we move from a non-`Bytes` or rather, does it require the length prefixes now?)

amaurymartiny: `isBare` get passed down into subtypes, so `toU8a(false)` should be strictly equivalent to `toU8a()`.

Issue #844, opened 2019-04-15
Refactor all the `asXxx` method to actually assert the underlying type is matching


xlc:
Currently all they does is just casting to the type without any checking so it is easy to bypass the type system and write bad code.

An assert will make it runtime error (still not as good as check time error) but it at least makes some error instead of silently enter inconsistent state.

Relates to #843

PR #879, opened 2019-05-07
assert type on all asXxx getters

jacogr:
Not the most fun exercise, but here it is. Closes https://github.com/polkadot-js/api/issues/844

### DIFF
--- a/packages/types/src/Metadata/MetadataVersioned.ts
+++ b/packages/types/src/Metadata/MetadataVersioned.ts
@@ -194,7 +194,7 @@ export default class MetadataVersioned extends Struct implements MetadataInterfa
     assert([2, 3, 4].includes(this.metadata.version), `Cannot convert metadata from v${this.metadata.version} to v1`);
 
     if (isUndefined(this._convertedV1)) {
-      this._convertedV1 = v2ToV1(this.metadata.asV2);
+      this._convertedV1 = v2ToV1(this.asV2);
     }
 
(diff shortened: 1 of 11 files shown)


packages/types/src/Metadata/MetadataVersioned.ts
```diff
@@ -20,7 +20,10 @@ import v2ToV1 from './v2/toV1';
 import v3ToV2 from './v3/toV2';
 import v4ToV3 from './v4/toV3';
 
-class MetadataEnum extends EnumType<Null | MetadataV1 | MetadataV2 | MetadataV3 | MetadataV4> {
+class MetadataDeprectated extends Null {
```
amaurymartiny: MetadataDeprecated

packages/types/src/Metadata/MetadataVersioned.ts
```diff
@@ -20,7 +20,10 @@ import v2ToV1 from './v2/toV1';
...
+}
+
+class MetadataEnum extends EnumType<MetadataDeprectated | MetadataV1 | MetadataV2 | MetadataV3 | MetadataV4> {
```
amaurymartiny: So here we're removing MetadataV0, but we can still convert asV0, which is weird

jacogr: Ummm... the change was _only_ supposed to be changing `Null` -> `MetadataDeprecated`

PR #883, opened 2019-05-08
Replace votingBalance -> derive.balances.all

kwingram25:
Remove defunct balances keys and update w new call

(Please check that calculations are accurate! @Tbaut @jacogr)

### DIFF
--- a/packages/api-derive/src/balances/all.ts
+++ b/packages/api-derive/src/balances/all.ts
@@ -59,7 +59,7 @@ export function all (api: ApiInterface$Rx) {
           ),
           availableBalance,
           lockedBalance,
-          vestedBalance
+          vestedBalance: vestedBalance.lt(new Balance(0)) ? new Balance(0) : vestedBalance
         };
       }),
       drr()
--- a/packages/api-derive/src/balances/all.ts
+++ b/packages/api-derive/src/balances/all.ts
@@ -1,0 +1,68 @@
+// Copyright 2017-2019 @polkadot/api-derive authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { combineLatest, of, Observable } from 'rxjs';
+import { map, switchMap } from 'rxjs/operators';
+import { ApiInterface$Rx } from '@polkadot/api/types';
+import { AccountId, AccountIndex, Address, Balance, BalanceLock, BlockNumber, Option, VestingSchedule } from '@polkadot/types';
+import { bnMax } from '@polkadot/util';
+
+import { idAndIndex } from '../accounts/idAndIndex';
+import { DerivedBalances } from '../types';
+import { drr } from '../util/drr';
+
+const EMPTY_ACCOUNT = new AccountId();
+
+export function all (api: ApiInterface$Rx) {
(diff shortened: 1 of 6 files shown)


packages/api-derive/src/balances/index.ts
```diff
@@ -2,6 +2,8 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
+export * from './all';
+export * from './allMulti';
```
jacogr: Not crazy about the `allMulti` naming here. :( However, I'm complaining without a better suggestion, so probably need to live with it.

packages/api-derive/src/balances/all.ts
```diff
@@ -0,0 +1,58 @@
...
+import { AccountId, AccountIndex, Address, Balance, BalanceLock, BlockNumber } from '@polkadot/types';
+import { bnMax } from '@polkadot/util';
+
+import { idAndIndex } from '../accounts/idAndIndex';
+import { DerivedBalances } from '../types';
+import { drr } from '../util/drr';
+
+const EMPTY_ACCOUNT = new AccountId(new Uint8Array(32));
+
+export function all (api: ApiInterface$Rx) {
```
jacogr: Would like an e2e test here so you can validate it actually does what it says on the tin - https://github.com/polkadot-js/api/blob/master/packages/api/test/e2e/promise-queries.spec.js

packages/api-derive/src/balances/all.ts
```diff
@@ -0,0 +1,58 @@
...
+import { map, switchMap } from 'rxjs/operators';
+import { ApiInterface$Rx } from '@polkadot/api/types';
+import { AccountId, AccountIndex, Address, Balance, BalanceLock, BlockNumber } from '@polkadot/types';
+import { bnMax } from '@polkadot/util';
+
+import { idAndIndex } from '../accounts/idAndIndex';
+import { DerivedBalances } from '../types';
+import { drr } from '../util/drr';
+
+const EMPTY_ACCOUNT = new AccountId(new Uint8Array(32));
```
jacogr: Just `new AccountId()` should do the trick.

packages/api-derive/src/balances/index.ts
jacogr: On second though, let's drop the multi. (We are not using it internally and hopefully the API will have native multi call support real-soon-now - it is my top API priority to do)

kwingram25:
@Tbaut It's "all" as in all derived balance figures, not "all" DOTs/units/etc :)

packages/api-derive/src/balances/all.ts
```diff
@@ -0,0 +1,68 @@
...
+          // get the maximum of the locks according to https://github.com/paritytech/substrate/blob/master/srml/balances/src/lib.rs#L699
+          lockedBalance = bnMax(...totals.map(({ amount }) => amount)) as Balance;
+        }
+
+        const { offset, perBlock } = vesting.unwrapOr({
+          offset: new Balance(0),
+          perBlock: new Balance(0)
+        } as VestingSchedule);
+
+        const vestedBalance = new Balance(perBlock.mul(bestNumber).add(freeBalance.sub(offset)));
```
Tbaut: This can end up being negative, so I think it would make sense to make it 0 in that case.

PR #885, opened 2019-05-08
Add traits replace rules. for invalid cases `::<type>` and `<type>::`.

satellitex:
Add traits replace rules. `::<type>` and `<type>::` to `<type>`.


satellitex:
@amaurymartiny
Yes. I see that `<type>::` or `::<type>` occurred situation. (surely alike `Hash::` was.)
Sorry, now I lost the code to reproduce it at hand. Please wait for a while if reproduction is necessary. I will prepare.

satellitex:
A `::<type>` appeared case is that used by `T::Balance` type in arguments.
`T::Balance` is regarded as `::Balance` on the client side.

jacogr:
So how do you replicate this `::Balance` from `T::Balance` - I'm specifically asking since exactly that happens a lot in the metadata, and it gets removed properly, i.e. this is the output - https://github.com/polkadot-js/api/blob/master/packages/types/src/Metadata/v4/latest.substrate.v4.json

So really just want to see exactly what cases that, which inputs being metadata or type injected related. 

(Have seen some questions around the type definitions lately, which does not allow traits lately only supporting the clear type, hence mentioning that)

amaurymartiny:
>  T::Balance is regarded as ::Balance on the client side.

Hmm I don't think so, unless there's a bug

https://github.com/polkadot-js/api/blob/bb2477876795bf96c007ded9eba77bd8c1cca081/packages/types/src/primitive/Type.ts#L194-L195

Note: your PR actually looks fine and will definitely solve the issue, but I'm trying to diagnose why we have `::Balance` in the first place

satellitex:
@amaurymartiny 
For example, `<T as balance::Trait>::Balance>` case is occurred invalid string `::Balance`. `<T as **::Trait>::**` is too. 
(ref: https://github.com/stakedtechnologies/Plasm/blob/66e82055ad9f62e6dfc2f7ea5d44a2528e387f7f/core/parent/src/mvp.rs#L216)

So, I rewrite `_removeTraits` rules. Please confirm.

PR #886, opened 2019-05-08
Linked map toString in v0

amaurymartiny:
**this is an ugly fix**

I'll clean metadata version once for all in a couple of days


jacogr:
CI passed, but not reflected here. (For some weird reason)

PR #887, opened 2019-05-08
multi calls for democracy.votes and staking.controllers

Stefie:


jacogr:
Nice! Thank you!

Stefie:
Yeah, not a big win for democracy.votes because we still need to map the accountId's. I changed that more because consistency

packages/api-derive/src/staking/controllers.ts
```diff
@@ -11,10 +11,7 @@ import { drr } from '../util/drr';
 
 function allBonds (api: ApiInterface$Rx, stashIds: Array<AccountId>) {
   return combineLatest(
```
amaurymartiny: can remove combineLatest

Stefie: Ok, I will add it to my next PR.

PR #888, opened 2019-05-09
Update extrinsics FromMetadata. Drop storages with doublemap type when converting v3 to v2

AlexWZ0129:
Hi, @amaurymartiny. This is the refactoring we mentioned in [Start supporting double map](https://github.com/polkadot-js/api/pull/843). I also updated the toV2, because in base.ts line 416 `Event.injectMetadata(this.runtimeMetadata.asV0);` is using asV0 as well.  So I changed the toV2 function to drop all the storages with doublemap type.


AlexWZ0129:
> > Event.injectMetadata(this.runtimeMetadata.asV0); is using asV0 as well.
> 
> Now we inject v0 or v4 in injectMetadata, so this function should also cater for both cases:
> https://github.com/polkadot-js/api/blob/d746815ad7cc81328d029881a28200247626d254/packages/types/src/type/Event.ts#L122
> 
> I'm also working on cleaning everything related to metadata right now (i'm basing myself on your branch), so if you add the `if (v0) {} else {}` in the function above, we can merge this PR asap as it'll will work, and I'll make a bigger refactor right after.

Yup, Okay, I will update the injectMetadata now and push to here asap. Thank you for the feedback amaurymartiny.

AlexWZ0129:
@amaurymartiny  btw, the code is working now after changing the tov2 function. Because in toV2 method, the storages contain double-map-type have all been removed, the `Event.injectMetadata(this.runtimeMetadata.asV0)` won't throw error.  Do we still want to change the injectMetadata here?

ianhe8x:
Now that asV0() in `Event.injectMetadata()` won't cause us error, I would like Alex to move on and focus on the DoubleMapStorage support. 
To refactor `Event.injectMetadata()` might involve a few code reorganization which are not decisions for us to make, for example I considered the possibility of creating a new package `types-events` mimicking the way we do for extrinsic and storage. Anyway that's why i would rather leave this to your guys. @amaurymartiny

AlexWZ0129:
Hi @amaurymartiny, I  updated the injectMetadata, please let me know if there is anything I need to change. Thank you!

amaurymartiny:
> for example I considered the possibility of creating a new package types-events mimicking the way we do for extrinsic and storage

What would type-events hold? Exactly this piece of code?

https://github.com/polkadot-js/api/blob/d746815ad7cc81328d029881a28200247626d254/packages/types/src/type/Event.ts#L122-L139

AlexWZ0129:
> lgtm, thanks! Your PR helped make things clearer, and I'll also refactor some other things on top of it. You can work on supporting double map finally :)
My pleasure. Thanks @amaurymartiny ! :D

AlexWZ0129:
Hi, @amaurymartiny. Shall we merge the [Start supporting double map](https://github.com/polkadot-js/api/pull/843) first? It will take me some time to finish the double map support. But  [Start supporting double map](https://github.com/polkadot-js/api/pull/843) now can prevent the api init error when the metadata contains doublemap type.

PR #896, opened 2019-05-13
Add exports from types/types

kwingram25:
IExtrinsic to start, are there any others that are like this?


jacogr:
The issue is that `IExtrinsic` is a type - all the other exports are classes. (i.e. _also_ available in pure JS). So would not do this export. (The `@polkadot/types` naming is unfortunate - since we expose all TS types in `<packagename>/types`)

Issue #900, opened 2019-05-14
EventRecord + DoubleMap conversion on metadata fails


xlc:
```
VM104:37 2019-05-14 14:11:18   API/DECORATOR: FATAL: Unable to initialize the API:  Cannot convert 'DoubleMapType' via asType
window.console.error @ VM104:37
m @ logger.js:83
error @ logger.js:129
(anonymous) @ Base.js:379
async function (async)
(anonymous) @ Base.js:364
u.emit @ index.js:201
emit @ Provider.js:343
onSocketOpen @ Provider.js:159
```

Relates to https://github.com/paritytech/substrate/pull/2491 as it introduce usage of double map in substrate
Should be fixed by https://github.com/polkadot-js/api/issues/776

jacogr:
EventRecord has a new `topics` field added - https://github.com/polkadot-js/api/blob/master/packages/types/src/type/EventRecord.ts#L81

https://github.com/paritytech/substrate/blob/master/srml/system/src/lib.rs#L344

jacogr:
Closed in #901 & #902

PR #901, opened 2019-05-14
Metadata update for latest master

jacogr:
- Part of https://github.com/polkadot-js/api/issues/900 that allows current master metadata to parse
  - Add `EventIndex`
  - Allow parsing (only via `Null`) for `DoubleMapType` constructors (NOTE: This is a step, not full support)

Outstanding for the above issue -

- new `EventRecord` structure (we need support for both old & new, not including here since there will be HACKs to get support in)
- proper `DoubleMap` support - not included here since I know there is ongoing work - so only the bare minimum to not fail on metadata


packages/types/src/Metadata/v4/latest.substrate.v4.json
```diff
@@ -142,6 +143,46 @@
...
+              "documentation": [
+                " The number of events in the `Events<T>` list."
+              ]
+            },
+            {
+              "name": "EventTopics",
+              "modifier": "Default",
+              "type": {
+                "DoubleMapType": {
+                  "hasher": "Blake2_256",
```
jacogr: Not convinced on the naming here of the hasher as opposed to line 167

amaurymartiny: There's an issue on substrate to convert key2Hasher from Text to StorageHasher, should be fine once that's done.

jacogr: Ahh, cool.

PR #902, opened 2019-05-14
new EventRecord with topics

jacogr:
- Final part of #900, support EventRecord for both old (v3 & v4) as well as new (v4)
- Some `DoubleMap` hacking since the existing code didn't work with v3 (and was reverted)
- Loads of additional e2e tests
- Cleanups and additional asserts for breakages during parsing


packages/types/src/codec/EnumType.ts
```diff
@@ -98,15 +98,19 @@ export default class EnumType<T> extends Base<Codec> implements Codec {
...
-    assert(index !== -1, `Cannot map input on JSON, unable to find '${key}' in ${keys.join(', ')}`);
+    assert(index !== -1, `Cannot map EnumType JSON, unable to find '${key}' in ${keys.join(', ')}`);
 
     return EnumType.createValue(def, index, value);
   }
 
   private static createValue (def: TypesDef, index: number = 0, value?: any): Decoded {
+    const Clazz = Object.values(def)[index];
+
+    assert(!isUndefined(Clazz), `Unable to create EnumType via index ${index}, in ${Object.keys(def).join(', ')}`);
```
jacogr: Previously we just failed with invalid errors, now we do an explicit check that there is an value - more info on parsing failures

packages/types/src/type/EventRecord.ts
```diff
@@ -72,6 +74,35 @@ export class Phase extends EnumType<ApplyExtrinsic | Finalization> {
   }
 }
 
+/**
+ * @name EventRecordV0
+ * @description
+ * A record for an [[Event]] (as specified by [[Metadata]]) with the specific [[Phase]] of
+ * application.
+ */
+export class EventRecordV0 extends Struct {
```
jacogr: Old. If we plan to support both (PR still WIP as to approach)

packages/api/src/nodeTypes.ts
```diff
@@ -0,0 +1,43 @@
...
+    }],
+    override: EventRecord77,
+    type: 'EventRecord'
+  }
+];
+
+export default function injectNodeTypes ({ specName, specVersion, implVersion }: RuntimeVersion): void {
+  types
+    .filter(({ nodes }) =>
+      nodes.filter((node) =>
```
amaurymartiny: nodes.some()

packages/types/src/type/Event.ts
```diff
@@ -119,23 +128,48 @@ export default class Event extends Struct {
...
-        const eventIndex = new Uint8Array([sectionIndex, methodIndex]);
-        const typeDef = meta.arguments.map((arg) => getTypeDef(arg));
-        const Types = typeDef.map(getTypeClass);
-
-        EventTypes[eventIndex.toString()] = class extends EventData {
-          constructor (value: Uint8Array) {
-            super(Types, value, typeDef, meta, sectionName, methodName);
-          }
-        };
+  static injectMetadata (metadata: Metadata): void {
```
amaurymartiny: so this is useless, could have kept the v0 version? I'll overwrite it with MetadataV4 anyways

packages/types/src/type/EventRecord.ts
```diff
@@ -72,6 +74,42 @@ export class Phase extends EnumType<ApplyExtrinsic | Finalization> {
   }
 }
 
+/**
+ * @name EventRecordV0
```
amaurymartiny: 77

packages/types/src/type/Event.ts
jacogr: I had conflicts since I changed it from the non-working version, so kept it. (Although I know it will be replaced)

jacogr:
Will merge in the morning so I can do the final pull-into-apps thing. (All should be ok, e2e tests for both local node and Alexander have been extensively tested - however never say never...)

PR #904, opened 2019-05-14
Fix for RPC bugs caused by uk spelling

Stefie:
Some of the RPC chain methods were still using UK spelling, specifically  `chain_getFinalizedHead `,
`chain_subscribeFinalizedHeads` and  `chain_unsubscribeFinalizedHeads ` .

That breaks these methods for newer versions of the metadata and which caused the the app-toolbox to crash on `chain.getFinalizedHead()`.

Just select `chain` > `getFinalizedHeader()` here for a live-demo: https://polkadot.js.org/apps/#/toolbox

This specific error was not mentioned here, but I came across it while I was trying to fix it. https://github.com/polkadot-js/apps/issues/1160


Stefie:
I only ran the tests on substrate because Alex is currently offline.

jacogr:
Ahh, ok, half job - didn't help that I converted apps at the same time without properly testing the aliasses.

PR #907, opened 2019-05-15
update extrinsics frommetadata

AlexWZ0129:
This fixes the issue using [#888](https://github.com/polkadot-js/api/pull/888) on v3. The reason for the issue is, I converted v3 to v4 when injecting events(because v1 to v4 are all using MetadataEvent), which is not allowed by `assert(this.isV4, 'Cannot convert '${this.type}' via asV4');`. (I started working on refactoring before this change. Sorry that I just run the unit test, didn’t do more check after rebase, will check more in the future) Now I use v0 as default, convert v3 to v0, which will keep the v3 as same as before. 

I ran the `promise-alex.spec.js` test, and the API init now is successful. (Sorry that I didn't run the e2e test before the #888, I thought the test cases are outdated because I saw they are skipped, will always run e2e in the future!) I still have an `Unable to decode Vector on index 0 Decoding 5EF7wdkP9XZq38fu2ioAvxoEhHUJUc1E2KbuSQSMGofRKhCL: Invalid decoded address checksum` error, but it also appears when running on the current master branch. 

Please feel free to contact me anytime if there is someplace I need to update. I will be online before 2 pm in Germany(12 am NZ).


packages/types/src/type/Event.ts
```diff
@@ -119,23 +120,48 @@ export default class Event extends Struct {
...
-        const typeDef = meta.arguments.map((arg) => getTypeDef(arg));
-        const Types = typeDef.map(getTypeClass);
-
-        EventTypes[eventIndex.toString()] = class extends EventData {
-          constructor (value: Uint8Array) {
-            super(Types, value, typeDef, meta, sectionName, methodName);
-          }
-        };
+  static injectMetadata (metadata: Metadata): void {
+    if (metadata.version === 4) {
```
AlexWZ0129: I updated here to keep the v3 as same as before

packages/types/src/Metadata/v3/toV2.ts
```diff
@@ -2,9 +2,37 @@
...
+
+  return new Option<Vector<StorageFunctionMetadata>>(
+    Vector.with(StorageFunctionMetadata),
+    new Vector<StorageFunctionMetadata>(StorageFunctionMetadata, storageV2)
+  );
+}
+
+function modulesV2 (v3: MetadataV3): Array<MetadataModule> {
+  return v3.modules.map((mod) =>
+    new MetadataModule({
```
jacogr: This is an additional issue I found and the one that I didn't want to debug completely, hence reverting - when constructing, you are losing the events & calls, i.e. https://github.com/polkadot-js/api/blob/master/packages/types/src/Metadata/v2/Metadata.ts#L22

packages/types/src/primitive/Method.ts
```diff
@@ -127,14 +128,21 @@ export default class Method extends Struct implements IMethod {
...
     };
   }
 
   // If the extrinsic function has an argument of type `Origin`, we ignore it
-  static filterOrigin (meta?: FunctionMetadata): Array<FunctionArgumentMetadata> {
+  static filterOrigin (meta?: MetaV0 | MetaV4): Array<FunctionArgumentMetadata> {
     // FIXME should be `arg.type !== Origin`, but doesn't work...
+    if (meta instanceof MetaV4) {
+      return meta
+      ? meta.args.filter(({ type }) =>
```
jacogr: Not for the PR, since it is the way it is - but we don't receive `Origin` anymore, so something to keep in mind to remove. (@amaurymartiny possibly when you rework?)

jacogr:
PS: @amaurymartiny did run the e2e tests, but I don't believe they were substantial enough to pick up the issue (or rather the Alex ones weren't complete). So only found both the issues when I added this - https://github.com/polkadot-js/api/blob/master/packages/api/test/e2e/promise-alex.spec.js#L82

packages/types/src/Metadata/v3/toV2.ts
AlexWZ0129: Thank you for the feedback! This is a tempory solution to prevent error when using v3 and double map at the same time. I can remove this if we won't use v3 and double map together.

AlexWZ0129: Otherwise, do we want to support double map in v3? The easiest way is converting v3 to v4.

jacogr: The issue is that we are not passing `events` though here, i.e. it should be something like -

```js
new MetadataModule({
      name: mod.name,
      prefix: mod.prefix,
      storage: storageV2(mod.storage),
      events: ... // missing currently
      calls: ... // missing currently
})
```

So basically when we convert, we don't have those populated

AlexWZ0129: Previously, there is an error if the metadata has a doublemap type and calling as v0. I change toV2 function to prevent that error. I see you have added [DoubleMapType to StorageFunctionType in v0](https://github.com/polkadot-js/api/blob/bd22da687e385983dca2b080b335204bcc4df60f/packages/types/src/Metadata/v0/Storage.ts#L82), which will prevent this error. So I think I can remove my changes about this toV2?

AlexWZ0129:
Hi @jacogr, I removed the changes of toV2(not needed because of the [change here](https://github.com/polkadot-js/api/blob/bd22da687e385983dca2b080b335204bcc4df60f/packages/types/src/Metadata/v0/Storage.ts#L82)) and injectEvent(already in master), only keep the updates of extrinsic fromMetadata here.

jacogr:
Perfect, thanks a million (as always, you guys rock!). Will just give a play-through with the updated e2e tests, but cannot spot massive things out of the ordinary here.

amaurymartiny:
@AlexWZ0129 I'm working on a refactor of metadata as said previously, and it will overwrite 90% of what you did here. I would propose to close this one, merge mine, and rework doublemap on top of it.

I'm aiming this afternoon for my PR. Does that sound fine to you?

AlexWZ0129:
> @AlexWZ0129 I'm working on a refactor of metadata as said previously, and it will overwrite 90% of what you did here. I would propose to close this one, merge mine, and rework doublemap on top of it.
> 
> I'm aiming this afternoon for my PR. Does that sound fine to you?

Yup. That sounds good for me. I have done some work about doublemap. I will move them to your branch.

AlexWZ0129:
> Perfect, thanks a million (as always, you guys rock!). Will just give a play-through with the updated e2e tests, but cannot spot massive things out of the ordinary here.

Thanks @jacogr !

Issue #767, opened 2019-03-10
Allow retrieval of multi keys in one storage subscription


jacogr:
(Split from https://github.com/polkadot-js/apps/issues/800)

The `state_storage` subs can support multiple keys per subscription. Currently we are only doing a single subscription for each query even when we know from the use side that we need to do multiples in a row. Doing multiples in one-go can reduce and optimise the load/number of subs greatly on both sides.

Even if at first we only implement this via the same type, maybe even `api.query.balances.freeBalance.multi(Alice, Bob, Dave, ...)` it will drop a lot of the use and can be used to make thing more optimal, for instances in the case of `api.derive.*`

From @amaurymartiny 

The `.multi` sounds good, it's clear on what it does (principle of least surprise), so I'm okay with this addition.

But as you said, it's only on one storage key, so I won't help in most `derive.*` where the storage keys are different, e.g. I can't do a multi-subscription to:

```js
api.query.balances.freeBalance(Alice)
api.query.system.accountNonce(Alice)
```

both in one subscription. I'd see something like:

```js
api.query.multi(
  api.query.balances.freeBalance(Alice),
  api.query.system.accountNonce(Alice)
).then/subscribe()
```

The freeBalance example above would be slightly more verbose:

```js
api.query.multi(
  api.query.balances.freeBalance(Alice),
  api.query.balances.freeBalance(Bob),
  api.query.balances.freeBalance(Dave),
)
```

so we could still add the `query.balances.freeBalance.multi` syntax as a shorthand. But I think the `query.multi()` one is more important.

PR #909, opened 2019-05-15
Expose queryMulti on api

jacogr:
Closes https://github.com/polkadot-js/api/issues/767


jacogr:
Not yet. Looking at the stuff, it will probably need some work to support e.g. the derive stuff which is nested with calls & derives. (However, first round will probably just use this and the re-evaluate where we stand - I balking at trying to be too clever without really being in it and understanding exactly what the different kinds of uses are.)

TL:DR - Sure this is not the last, building from the ground up, a little bite at a time.

Issue #910, opened 2019-05-15
Update derive to use api.queryMulti


jacogr:
As done here - https://github.com/polkadot-js/api/pull/909/files#diff-2c17a9848687821518cc295f7baa4d99R15

PR #911, opened 2019-05-15
Update api.derive to use api.queryMulti

jacogr:
- Closes https://github.com/polkadot-js/api/issues/910
- use api.queryMulti is session

### DIFF
--- a/packages/api-derive/src/balances/all.ts
+++ b/packages/api-derive/src/balances/all.ts
@@ -14,6 +14,22 @@ import { drr } from '../util/drr';
 
 const EMPTY_ACCOUNT = new AccountId();
 
+/**
+ * @name all
+ * @param {( ccountIndex | AccountId | Address | string )} address - An accounts Id in different formats.
+ * @returns An object containing the combined results of the storage queries for
+ * all relevant fees as declared in the substrate chain spec.
+ * @example
+ * <BR>
+ *
+ * ```javascript
+ * const ALICE = 'F7Hs';
+ *
+ * api.derive.balances.all(ALICE, ([accountId, lockedBalance]) => {
+ *   console.log(`The account ${accountId} has a locked balance ${lockedBalance} units.`);
+ * });
+ * ```
+ */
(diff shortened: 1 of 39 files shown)


packages/api-derive/src/balances/all.ts
```diff
@@ -12,8 +12,43 @@ import { idAndIndex } from '../accounts/idAndIndex';
...
+  }
+
+  const { offset, perBlock } = vesting.unwrapOr({
+    offset: new Balance(0),
+    perBlock: new Balance(0)
+  } as VestingSchedule);
+
+  const vestedBalance = new Balance(perBlock.mul(bestNumber).add(freeBalance.sub(offset)));
+
+  // const availableBalance = lockedBalance.eq(new BlockNumber('0xffffffffffffffffffffffffffffffff')) ? new Balance(0) : freeBalance && new Balance(freeBalance.sub(lockedBalance));
```
amaurymartiny: remove code comment?

packages/api-derive/src/session/info.ts
```diff
@@ -0,0 +1,69 @@
...
+import BN from 'bn.js';
+import { Observable, combineLatest } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { ApiInterface$Rx } from '@polkadot/api/types';
+import { BlockNumber, Option } from '@polkadot/types';
+
+import { drr } from '../util/drr';
+import { bestNumber } from '../chain';
+
+type Result = [BN, [BlockNumber?, Option<BlockNumber>?, BlockNumber?, BlockNumber?, BlockNumber?]];
```
amaurymartiny: Would it be possible to make that the previous function always returns something (never undefined), so that we can remove all the `?`

afaict all the observables return something, so there should ideally never be `?` in return types.

jacogr: Nope, looking at it is is actually worse than that - the original (before api-derive) and before we reworked the API and before defaults returned `undefined` - so this is ancient code. :) Anyway, reworked so it is modern and working.

PR #913, opened 2019-05-15
refactor: Metadata overhaul

### LINKED ISSUE #858: Clean types/Metadata, type-storage, type-extrinsics
While working on Metadata v4, I found out a lot of places which could do some cleaning. Some of them are already done in #857.

#### 1. Remove `toV1`, `toV0`... hacks

These are seriously really ugly. Afaik they are only used to create `api.{tx|query}.*` functions from type-storage and type-extrinsics.

**Proposal**: create `v4`, `v5`... folders inside type-storage and type-extrinsics. This is done in type-storage, but should be done in type-extrinsics too.

#### 2. (After 1) Factorize code between `v*`

We're already doing stuff like:

https://github.com/polkadot-js/api/blob/fafb5b46ddfd179aff0073247838afcbc4b342fd/packages/types/src/Metadata/v4/Storage.ts#L15-L24

But still a lot of stuff needs to be duplicated, and codeclimate is not happy. Same thins appears in type-storage.

#### 3. Figure out how to have different types by version

`Method.ts`, `Extrinsic.ts`, `StorageKey.ts` inside `types` will actually be different types depending on the metadata version.

**Proposal**: this one is hard. Some ideas:
- Use generic `StorageKey<Meta>`
- Subclass StorageKey
- completely different classes by version

1,2,3 can be done separately

amaurymartiny:
closes #858 

The main idea of this overhaul is the following: instead of creating v1ToV0 functions, we create v0ToV1 functions.

Why? In this case, the whole codebase (type-storage, type-extrinsics...) will only cater for the v4 (latest metadata) case (instead of having the whole codebase cater for v0). If one day there's a v5, we change the whole codebase to support v5, and create a v4ToV5 function.

This is based on the fact that each Metadata version bump _adds_ information: v1->v2 = linked map, v2->v3 = double map, v3->v4 = custom hasher. In this case, if `n<m`, we can convert from `MetadataVn` to `MetadataVm` without any loss of information (the other way round is false: we lose information when we convert e.g. v3->v2, which produced hacks all over the place).

Downside: if one day Metadata bump drops information, then it's either hacks all over again, or we need to find something else. I believe this case will only happen extremely rarely.

<details><summary>Original idea</summary>
<p>
My original idea of the overhaul was a bit different: it was to create `v0/`..`v4/` folders in `type-extrinsic`, `type-storage`, `Event.ts`, `Method.ts`... We then check the Metadata version, and call the correct function/class in the corresponding folder (see e.g. #907)

This solution is robust against metadata information drop. But it introduces a lot of duplication, so I think it's better to avoid it. This -3k loc is pretty satisfying.
</p>
</details>

Also:
- Rename all Metadata classes to match exactly Rust types (e.g. `MetadataCall` -> `FunctionMetadata`)

TODO:
- [x] update the script that generates Metadata docs (it uses v0)


packages/type-extrinsics/src/fromMetadata/createUnchecked.ts
```diff
@@ -16,15 +16,15 @@ import { assert } from '@polkadot/util';
...
+  indexCount: number,
+  meta: FunctionMetadata,
+  index: number
 ): MethodFunction {
-  const callIndex = new Uint8Array([index, meta.id.toNumber()]);
-  let extrinsicFn: any;
+  const callIndex = new Uint8Array([indexCount, index]);
 
-  const expectedArgs = Method.filterOrigin(meta);
+  const expectedArgs = meta.args;
```
jacogr: (nit) remove extra line before

packages/type-extrinsics/src/fromMetadata/index.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param metadata - The metadata to extend the storage object against.
+ */
+export default function fromMetadata (metadata: Metadata): ModulesWithMethods {
+  let indexCount = -1;
+
+  return metadata.asV4.modules.reduce((result, module: ModuleMetadata) => {
+    if (module.calls.isNone || module.calls.isEmpty) {
+      return result;
+    }
+    indexCount++;
```
jacogr: (nit) extra line before and after. 

(I would have preferred if we started at 0 and incremented after, more-or-less because of the -1 start, but it is fine & understandable - just mentioning, but doesn't really need to action. Doing this also aligns with https://github.com/polkadot-js/api/pull/913/files#diff-313a8826a9a3cd542eb82c29553391f1R131)

packages/type-storage/src/fromMetadata/createFunction.ts
```diff
@@ -10,7 +10,10 @@ import { assert, isNull, isUndefined, stringLowerFirst, stringToU8a, u8aConcat }
 import getHasher from './getHasher';
 
 export interface CreateItemOptions {
-  isUnhashed?: boolean;
+  /**
+   * We don't hash the keys defined in ./substrate.ts
+   */
+  skipHashing?: boolean;
```
jacogr: (nit) alphabetical (probably `//` comment as well, not really a doc)

packages/types/src/Metadata/v3/toV4.ts
```diff
@@ -0,0 +1,67 @@
...
+ * See https://github.com/paritytech/substrate/pull/2268 for details
+ */
+export default function toV4 (metadataV3: MetadataV3): MetadataV4 {
+  return new MetadataV4({
+    modules: metadataV3.modules.map((modul) => {
+      return {
+        name: modul.name,
+        prefix: modul.prefix,
+        storage: modul.storage.isSome
+          ? new Option(Vector.with(StorageFunctionMetadata), modul.storage
```
jacogr: Have some difficulty reading this one, would have suggested 

```js
new Option(
  Vector.with(StorageFunctionMetadata),
  modul.storage.unwrap().map(toV4StorageFunction)
)
```

packages/types/src/type/Event.ts
```diff
@@ -168,8 +146,10 @@ export default class Event extends Struct {
             }
           };
         });
-      });
-    }
+
+        sectionIndex += 1;
```
jacogr: (nit) sectionIndex++;

packages/types/src/type/Extrinsic.ts
```diff
@@ -58,16 +58,9 @@ export default class Extrinsic extends Struct implements IExtrinsic {
...
-        return new Uint8Array();
-      }
-
       const [offset, length] = Compact.decodeU8a(value);
-      const total = offset + length.toNumber();
-
-      assert(total <= value.length, `Extrinsic: required length less than remainder, expected at least ${total}, found ${value.length}`);
 
-      return value.subarray(offset, total);
+      return value.subarray(offset, offset + length.toNumber());
```
jacogr: I really don't like the fact that we dropped the assert here.

ianhe8x:
Will that still be too much of work when we need to bump Metadata version?

amaurymartiny:
> Will that still be too much of work when we need to bump Metadata version?

What we currently do:
1. Write v5ToV4 transition function
2. Write hacks in type-storage & type-extrinsic to support new features with old Metadata

What I propose to do:
1. Write v4TV5 transition function
2. Change type-storage & type-extrinsic & Event.ts to use by default newest Metadata

imo the amount of work will be +/- the same in both case, but the 2nd scenario is cleaner

packages/type-extrinsics/src/fromMetadata/createUnchecked.ts
```diff
@@ -16,15 +16,15 @@ import { assert } from '@polkadot/util';
 export default function createDescriptor (
   section: string,
   method: string,
-  index: number,
-  meta: FunctionMetadata
+  indexCount: number,
+  meta: FunctionMetadata,
+  index: number
```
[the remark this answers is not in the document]
amaurymartiny: Changed, but still didn't do it alphabetical, because it's function arguments. imo it doesn't make sense to have `methodIndex` before `sectionIndex`

packages/type-extrinsics/src/fromMetadata/index.ts
amaurymartiny: I had a big doubt while re-reading. Here we only increment `indexCount` (== `sectionIndex`) when there are calls in the module (which happens always in our case). But: we should increment `indexCount` _always_, correct?

If there's a module without calls, then the sectionIndexes of modules after this module will be wrong

packages/types/src/type/Extrinsic.ts
amaurymartiny: Sorry, copy-pasted and forgot to check

packages/type-extrinsics/src/fromMetadata/index.ts
jacogr: We need to check this against the original, cannot recall off-hand - I believe, it should only be incremented when there are none.

packages/type-extrinsics/src/fromMetadata/index.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Extend a storage object with the storage modules & module functions present
+ * in the metadata.
+ *
+ * @param extrinsics - An extrinsics object to be extended.
+ * @param metadata - The metadata to extend the storage object against.
+ */
+export default function fromMetadata (metadata: Metadata): ModulesWithMethods {
+  return metadata.asV4.modules
+    .filter((modul) => modul.calls.isSome)
+    .reduce((result, module: ModuleMetadata, sectionIndex) => {
```
jacogr: Maybe just `modul` here as well as in the previous line? (since module is Node)

amaurymartiny:
@ianhe8x I merged as-is. If you have a different idea on how to refactor metadata cleanly, I'm of course interested.

ianhe8x:
@amaurymartiny The approach i explored in my pr, use interface in places like Method Event, caster all version of metadata to that interface. I think you have read my code, though we haven't talked much about the downsides of it

PR #914, opened 2019-05-15
Extrinsic bug with undefined value

Stefie:
Addresses the `Method.js:104 Method: cannot decode value 'undefined' of type undefined` Bug in https://github.com/polkadot-js/apps/issues/1160 

Onload on https://polkadot.js.org/apps/#/toolbox
Screenshot 2019-05-15 at 14 41 31


amaurymartiny:
I'm okay to merge this, but tbh I think it won't change anything:

When value is undefined, we pass in `{}` to decodeExtrinsic

https://github.com/polkadot-js/api/blob/d9ecedba2545ac4ad0626c27e1e4cc8f86ae650c/packages/types/src/type/Extrinsic.ts#L41

so `value` never be undefined inside decodeExtrinsic's args afaict

Stefie:
> I'm okay to merge this, but tbh I think it won't change anything:

@amaurymartiny at least it makes this bug polkadot-js/apps#1160 in the toolbox app disappear  :)

packages/types/src/primitive/Method.ts
```diff
@@ -128,7 +128,7 @@ export default class Method extends Struct implements IMethod {
       args: new Uint8Array(),
       argsDef: {},
       meta: new FunctionMetadata(),
-      callIndex: new Uint8Array([255, 255])
+      callIndex: new Uint8Array()
```
jacogr: Throw on line 125, line 88 becomes -

`private static decodeMethod (value: DecodedMethod | Uint8Array | string = new Uint8Array(), _meta?: FunctionMetadata): DecodedMethod {`

jacogr:
Still think apps needs to be fixed to not pass crap info.

jacogr:
@amaurymartiny Yes, you are 100% right - so think it is looking at the wrong thing. (Don't understand how that makes the issue disappear)

Stefie:
> Don't understand how that makes the issue disappear)

Good to hear, because me neither! Good thing is that I fixed two other bugs while trying to investigate that...

packages/types/src/primitive/Method.ts
Stefie: I changed that (if I understood you correctly)

jacogr: 100%, line 125 should throw, everything below that should be removed. (No return - at that point, we give up)

Stefie:
The changes in `Method.ts` were not directly related to the bug, but I changed it to throwing an error.

Issue #366, opened 2018-10-30
codec/Set to extend Set


jacogr:
https://github.com/polkadot-js/api/pull/365/files#diff-31f4c64ff0ef4336c587e4512673ff11R16

PR #917, opened 2019-05-15
codec/Set extends JS Set

jacogr:
Closes https://github.com/polkadot-js/api/issues/366

### DIFF
--- a/packages/types/src/codec/Set.ts
+++ b/packages/types/src/codec/Set.ts
@@ -2,9 +2,8 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { isU8a, isNumber, isUndefined, u8aToHex } from '@polkadot/util';
+import { assert, isU8a, isNumber, isUndefined, u8aToHex } from '@polkadot/util';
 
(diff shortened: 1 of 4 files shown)


packages/types/src/codec/Set.ts
```diff
@@ -78,21 +78,34 @@ export default class Set extends Base<Array<string>> implements Codec {
...
-    return Set.encodeSet(this._setValues, this.raw);
+    return CodecSet.encodeSet(this._setValues, this.strings);
+  }
+
+  /**
+   * @description adds a value to the Set (extended to allow for validity checking)
+   */
+  add (key: string): this {
+    // we have the isUndefined(this._setValues) in here as well, add is used internally
+    // in the Set constructor (so it is undefined at this point, and should allow)
```
amaurymartiny: outdated comment?

jacogr: Maybe just unclear. So basically the assert does this `assert(isUndefined(this._setValues) || ...` - this is for the constructor (Set calls `add` internally)

amaurymartiny: Ah, understood

PR #919, opened 2019-05-16
Create submittable via hex

jacogr:
- Closes https://github.com/polkadot-js/api/issues/920
- Introduce `api.tx(encoded: string | Uint8Array)` that returns a Submittable
- This allows for offline signing and then submitting the tx via the api
- e2e usage test for this interface
- rework e2e promise tests with less duplication

cc @jnaviask

Just for use-clarity, as detailed in #920:

```js
// this is just the construction phase - can be any anywhere and in any form, i.e. via
// a mobile app, via a Rust app, via, well, anything - all we need is the encoded extrinsic hex
const nonce = await api.query.system.accountNonce(keyring.dave.address());
const hex = api.tx.balances
  .transfer(keyring.eve.address(), 12345)
  .sign(keyring.dave, { nonce })
  .toHex(); // hex UncheckedMortalExtrinsic

// this is the important bit added - with the encoded hex value, we can send
return api.tx(hex).send(({ events, status }) => { ... });
```

### DIFF
--- a/packages/api/test/e2e/promise-tx.spec.js
+++ b/packages/api/test/e2e/promise-tx.spec.js
@@ -8,9 +8,28 @@ import { randomAsHex } from '@polkadot/util-crypto';
 
 import Api from '../../src/promise';
 import WsProvider from '../../../rpc-provider/src/ws';
-import SingleAccountSigner from "../util/SingleAccountSigner";
+import SingleAccountSigner from '../util/SingleAccountSigner';
 
 describe.skip('e2e transactions', () => {
+  // log all events for the transfare, calling done() when finalized
+  const logEvents = (done) =>
+    ({ events, status }) => {
+      console.log('Transaction status:', status.type);
+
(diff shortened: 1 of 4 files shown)


packages/api/test/e2e/promise-tx.spec.js
```diff
@@ -59,43 +73,13 @@ describe.skip('e2e transactions', () => {
...
-          events.forEach(({ phase, event: { data, method, section } }) => {
-            console.log('\t', phase.toString(), `: ${section}.${method}`, data.toString());
-          });
-
-          if (events.length) {
-            done();
-          }
-        }
-      });
+      .send(loEvents(done));
```
jnaviask: Should this be `logEvents`?

jacogr: Ok, so how the hell did I manage to screw up that line between testing and committing. Thanks!

Issue #271, opened 2018-10-12
Remove EnumType, use only Enum


amaurymartiny:
Make Enum class as close as possible to Rust enum.

**Example 1:**
If Rust = 
```rust
| Something(123)
| Empty
```
then JS =
```javascript
new Enum({ Something: UInt, Empty: Null}, { Some: 123 }); 
```

**Example 2:** a Rust Option in JS is simply a:
```javascript
class Option extends Enum {
  constructor(Type, value) {
    super({ None: Null, Some: Type }, value)
  }
}
```
and U8a codec will be magically resolved: `[0]` for `None` and `[1, ...]` for `Some(...)`

**Example 3:** A simple Enum (i.e. without values)
Unfortunately we won't use TS enums at all (anyway they don't allow values), but a SimpleEnum is just an Enum using Null everywhere
```javascript
class SimpleEnum extends Enum {
  constructor(Type, value) {
    super({ VALUEA: Null, VALUEB: Null }, value)
  }
}
```

**Example 4:** MisbehaviorKind
```javascript
class MisbehaviorKind extends Enum {
  constructor(Type, value) {
    super({ 0x11: BftDoublePrepare, 0x12: BftDoubleCommit }, value)
  }
}
```
Just have a convention that instead of key strings, we can pass key numbers, and in this case the encoding is `[0x11, ...]` for BftDoublePrepare, and `[0x12, ...]` for BftDoubleCommit (i.e. 1st byte is not index position, but index value).

jacogr:
I suggest sticking to EnumType since it gets the most love. We can do -

- allow Array<string> as input
- as found, do { [name: string]: Null } (as suggested)
- delete Enum, rename EnumType -> Enum

PR #922, opened 2019-05-16
Combine Enum & EnumType

jacogr:
- Closes https://github.com/polkadot-js/api/issues/271
- Allow `EnumType` to construct with `['someString1' 'someString2', ...]`
- Remove `<T>` from `EnumType` - no real use there in the current
- Add old `Enum` tests (basic) into `EnumType`
- Covert all uses of `Enum` to the new `EnumType` 
- Rename `EnumType` (not yet, see this lil mishap - https://github.com/polkadot-js/api/pull/922#discussion_r284602223)

There is just one feature that `EnumType` does not currently cater for, this one - https://github.com/polkadot-js/api/blob/master/packages/types/src/codec/Enum.ts#L10 (It was also purposefully removed from EnumType in an earlier iteration.)


packages/types/src/codec/Enum.ts
```diff
@@ -2,94 +2,209 @@
...
+ * This implements an enum, that based on the value wraps a different type. It is effectively
+ * an extension to enum where the value type is determined by the actual index.
  */
 // TODO:
-//   - It would be great if this could actually wrap actual TS enums
-export default class Enum extends Base<number> implements Codec {
-  private _enum: EnumMap | Array<string>;
+//   - As per Enum, actually use TS enum
+//   - It should rather probably extend Enum instead of copying code
+export default class Enum extends Base<Codec> implements Codec {
```
jacogr: Fookit! The rename created issues here, now it looks like the actual file was adapted :(

jacogr: My first force-push to a PR to make this sane.

packages/types/src/codec/Enum.ts
```diff
@@ -2,129 +2,6 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { isString, isU8a, u8aToHex, assert } from '@polkadot/util';
+import Enum from './EnumType';
```
jacogr: For now. Should rename and not create chaos like the last time around - as it stands, keeping it like this until all-ok here.

packages/types/src/codec/EnumType.ts
```diff
@@ -25,62 +25,83 @@ type Aliasses = {
...
-export default class EnumType<T> extends Base<Codec> implements Codec {
+export default class Enum extends Base<Codec> implements Codec {
   private _def: TypesDef;
   private _index: number;
   private _indexes: Array<number>;
+  private _isBasic: boolean;
 
-  constructor (def: TypesDef, value?: any, index?: number | EnumType<T>, aliasses?: Aliasses) {
-    const decoded = EnumType.decodeEnumType(def, aliasses || {}, value, index);
+  constructor (def: TypesDef | Array<string>, value?: any, index?: number | Enum, aliasses?: Aliasses) {
```
amaurymartiny: Can we remove aliasses? It's used in one place (Finalised -> Finalized) and imo it's a bit useless now

packages/types/src/codec/EnumType.ts
```diff
@@ -25,62 +25,83 @@ type Aliasses = {
...
-    this._indexes = Object.keys(def).map((_, index) => index);
+    this._def = defInfo.def;
+    this._isBasic = defInfo.isBasic;
+    this._indexes = Object.keys(defInfo.def).map((_, index) => index);
     this._index = this._indexes.indexOf(decoded.index) || 0;
   }
 
-  private static decodeEnumType<T> (def: TypesDef, aliasses: Aliasses, value?: any, index?: number | EnumType<T>): Decoded {
+  private static extractDef (def: TypesDef | Array<string>): { def: TypesDef, isBasic: boolean } {
+    if (!Array.isArray(def)) {
```
amaurymartiny: What if I define `new Enum({a: Null, b: Null})`: that should be isBasic. i.e. `isBasic = Object.values(def).every(a=>a === Null)` (can even be a private method instead of a property in this case)

packages/types/src/codec/EnumType.spec.ts
```diff
@@ -2,132 +2,213 @@
...
-    expect(
-      new EnumType(
-        { 1: Text, 5: U32 },
-        new Uint8Array([1, 2 << 2, 49, 50])
-      ).toString()
-    ).toEqual('Text');
-  });
+    // We are currently not using this approach, none of the types in Substrate currently
+    // have any overrides. Instead of trying to support it (just-in-case), rather have it
+    // removed to simplify the code - it can be pulled-back if needed
```
amaurymartiny: Agree with this comment: we could remove _indexes and co in https://github.com/polkadot-js/api/pull/922/files#diff-438804f1fd82048ec30fdbe6c821d298R50

(not urgent)

packages/types/src/codec/EnumType.ts
jacogr: A bit sceptical there - since I'm not 100% what the Rust side does. So I know if you do this -

```
enum {
  'Zero',
  'One',
  'Two'
}
```

And it serializes to JSON, it is `1` for `One`. When using types, it actually serializes to the name `One` (or `One: value` when there is one). So it seems slippery - so really keep `isBasic` to just the case where there are strings.

packages/types/src/codec/EnumType.ts
jacogr: Good point, meant to look at those since I couldn't remember the use.

Issue #867, opened 2019-05-03
Remove HeaderExtended from typeRegistry and move to api-derive


Stefie:
`HeaderExtended`  is a type only used by the `api-derive´ module so it shouldn't be part of the  typeRegistry

PR #923, opened 2019-05-16
Move HeaderExtended type to api-derive

jacogr:
Closes https://github.com/polkadot-js/api/issues/867


jacogr:
Sheiße, I was sure I did that. Thank you. (Think I undid since I lost the copy and then never came back, rookie mistake.)

Issue #756, opened 2019-03-03
How to create and submit a mortal transaction?


xlc:
Could you provide some examples of how to create and submit a mortal transaction? e.g. I want this transaction expire after block 10000 or after 100 blocks from current height

All I can found is [ExtrinsicEra](https://github.com/polkadot-js/api/blob/master/packages/types/src/type/ExtrinsicEra.ts) which requires me to construct the details from bytes. It should provide helper methods to construct `ExtrinsicEra` from block number.

I did found this https://github.com/paritytech/wiki/blob/master/Extrinsic.md#transaction-era
I can try to understand the algorithm and implement the encode method but it will be good if it is done by someone already knows how to do it.

jacogr:
Both an enhancement and documentation -

- documentation: show how to submit the blockHash
- enhancement: provide the getters on ExtrinsicEra, encode fully (partly implemented)

jacogr:
Keep https://github.com/polkadot-js/api/issues/778 in mind - specify the era as part of `signAndSend` (if not already 100% clear based on the above discussions)

jacogr:
Mortals supported since https://github.com/polkadot-js/api/pull/927 - changing labels to reflect documentation and samples.

kushptlapigarage:
This will be useful to apps me and my team building.

jacogr:
We need to get the docs updated with samples. Eras have been supported since #927 (we wish to swap it to the default at some point, not the next stable, but most probably the one after)

kushptlapigarage:
Thanks for the response.

We used [this](https://github.com/polkadot-js/api/pull/927/files#diff-0ae13b274820f1e4ec131df1958a37f9R159) e2e test as base. The difficult part was calculating the fees as we show users as part Enzyme (Polkadot DApp Wallet) how much it would cost a transaction.

jacogr:
Closing, 0.82 now both support (already available in 0.81) and uses it by default.

PR #927, opened 2019-05-17
ERA changes

KarishmaBothara:


packages/api/test/e2e/promise-tx.spec.js
```diff
@@ -9,6 +9,7 @@ import { randomAsHex } from '@polkadot/util-crypto';
 import Api from '../../src/promise';
 import WsProvider from '../../../rpc-provider/src/ws';
 import SingleAccountSigner from '../util/SingleAccountSigner';
+import {ExtrinsicEra} from '@polkadot/types/type';
```
ianhe8x: import from external first

packages/api/test/e2e/promise-tx.spec.js
```diff
@@ -154,4 +155,44 @@ describe.skip('e2e transactions', () => {
...
+    // eraDeath - end of ERA validity (EXPIRY)
+    const eraBirth = exERA.asMortalEra.birth(currentHeight.toNumber());
+    const eraDeath = exERA.asMortalEra.death(currentHeight.toNumber());
+    console.log('STARTED AT :'+eraBirth+' EXPIRED AT :'+eraDeath);
+    const eraHash = await api.rpc.chain.getBlockHash(eraBirth);
+    const ex = api.tx.balances
+      .transfer(keyring.eve.address(), 12345);
+    const tx = await ex.signAndSend(keyring.dave, {blockHash: eraHash, era:exERA, nonce});
+
+    expect(tx.toHex()).toHaveLength(66);
```
ianhe8x: why length must be 66? not a good assertion

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,196 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
+import Enum from '../codec/Enum';
+import { isHex, isU8a, hexToU8a, isObject, assert, u8aToBn } from '@polkadot/util';
```
ianhe8x: be careful with import order

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,196 @@
...
+  get isMortalEra (): boolean {
+    return this.index >= 0;
+  }
+
+  /**
+   * @description Encodes the value as a Uint8Array as per the parity-codec specifications
+   * @param isBare true when the value has none of the type-specific prefixes (internal)
+   */
+  toU8a (isBare?: boolean): Uint8Array {
+    if (this.index === 0) {
```
xlc: `return this.value.toU8a()` is enough

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,196 @@
...
+   */
+  get isImmortalEra (): boolean {
+    return this.index === 0;
+  }
+
+  /**
+   * @description `true` if Mortal
+   */
+  get isMortalEra (): boolean {
+    return this.index >= 0;
```
ianhe8x: `> 0` or `!this.ImmortalEra`

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,196 @@
...
+   */
+  death (current: number) {
+    return this.birth(current) + this.period.toNumber();
+  }
+
+  /**
+   * @description convert the number to binary and get the trailing zero's.
+   */
+  private getTrailingZeros (period: number) {
+    const zeros: number[] = [];
```
xlc: Untested
```
const binary = period.toString(2);
let index = 0;
while (binary[binary.length - 1 - index] === '0') {
    index++;
}
return index;
```

packages/api/test/e2e/promise-tx.spec.js
```diff
@@ -1,7 +1,7 @@
 // Copyright 2017-2019 @polkadot/api authors & contributors
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
-
+import {ExtrinsicEra} from '@polkadot/types/type';
```
ianhe8x: spaces at start and end of {}

jacogr:
Brilliant! Will take a look ASAP.

packages/types/src/type/ExtrinsicEra.ts
ianhe8x: era has custom decode logic, the first byte as enum index is reused to construct the sub type

KarishmaBothara:
> Brilliant! Will take a look ASAP. (Part of #756)

We are aware of a unit test failure (Extrinsic.spec.ts). Will work and fix it soon.

packages/api/test/e2e/promise-tx.spec.js
jacogr: (nit) Empty line after `//` and before import

packages/types/src/type/ExtrinsicEra.spec.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import ExtrinsicEra from './ExtrinsicEra';
+
+describe('ExtrinsicEra', () => {
+
```
jacogr: (nit) Unneeded empty line

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
+import Enum from '../codec/Enum';
```
jacogr: Import order. First externals, <blank line>, files. (all alphabetical)

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
 
 /**
  * @name ExtrinsicEra
  * @description
  * The era for an extrinsic, indicating either a mortal or immortal extrinsic
  */
-export default class ExtrinsicEra extends U8a {
+export default class ExtrinsicEra extends Enum {
+  constructor (value?: any) {
+    super({ ImmortalEra, MortalEra }, ExtrinsicEra.decodeExtrinsicEra(value));
```
jacogr: This would read better (and align with what we have elsewhere), as

```js
super({ 
  ImmortalEra,
  MortalEra
}, ExtrinsicEra.decodeExtrinsicEra(value));
```

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+  }
+
+  private static decodeExtrinsicEra (value: MortalMethod | Uint8Array | string): Uint8Array | Object | undefined {
+    if (!value) {
```
jacogr: This can be replaced by making the signature `value: MortalMethod | Uint8Array | string = new Uint8Array([0])` (Or even `= new Uint8Array()` and just adjusting the check lower down to be 

```js
else if (isU8a(value)) {
      if (!value.length || value[0] === 0) {
```

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+    } else if (isU8a(value)) {
+      if (value[0] === 0) {
+        return new Uint8Array([0, 0]);
+      } else {
+        return new Uint8Array([1, value[0], value[1]]);
+      }
+    } else if (isObject(value)) {
+      return { MortalEra: value };
+    }
+    throw new Error('Invalid data passed to Era');
```
jacogr: (nit) Extra blank line before the throw

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+  }
+
+  /**
+   * @description Returns the item as a [[ImmortalEra]]
+   */
+  get asImmortalEra (): ImmortalEra {
+    assert(this.isImmortalEra, `Cannot convert '${this.type}' via asImmortalEra`);
+    return this.value as ImmortalEra;
```
jacogr: (nit) Here and for the remainder of the file. Whitespace to denote blocks are cheap and read better. So aligning with the rest of the codebase, this is preferred -

```js
assert(this.isImmortalEra, `Cannot convert '${this.type}' via asImmortalEra`);

return this.value as ImmortalEra;
```

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+  get isMortalEra (): boolean {
+    return this.index > 0;
+  }
+
+  /**
+   * @description Encodes the value as a Uint8Array as per the parity-codec specifications
+   * @param isBare true when the value has none of the type-specific prefixes (internal)
+   */
+  toU8a (isBare?: boolean): Uint8Array {
+    return this.value.toU8a();
```
jacogr: Pass `isBare` through into the `value`, i.e. `return this.value.toU8a(isBare)`

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+  }
+}
+
+const VALID_IMMORTAL = new U8a([0]);
+/**
```
jacogr: (nit) Blank line between the const and the comment. Even better would be to move the `const VALID_IMMORTAL` to the top of the file, just before the first class definition so all global constants are visible there.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+    super(value);
+
+    assert(this.eq(VALID_IMMORTAL), `IMMORTAL: expected ${VALID_IMMORTAL.toHex()}, found ${this.toHex()}`);
   }
+}
 
-  static decodeExtrinsicEra (value?: AnyU8a): Uint8Array {
-    if (value) {
-      const u8a = u8aToU8a(value);
+export type MortalEraValue = [U8, U8];
```
jacogr: Move all types to the top of the file -

- imports (with the correct ordering)
- all type definitions
- all constants
- all classes

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+
+/**
+ * @name MortalEra
+ * @description
+ * The MortalEra for an extrinsic, indicating period and phase
+ */
+export class MortalEra extends Tuple {
+  constructor (value?: any) {
+    super({
+      period : U8, phase : U8
```
jacogr: Adjust this to be -

```js
super({
  period: U8,
  phase: U8
}, MortalEra.decodeMortalEra(value));
```

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+  }
+
+  private static decodeMortalEra (value: MortalMethod | Uint8Array): MortalEraValue {
+    if (isU8a(value)) {
+      const first = u8aToBn(value.subarray(0, 1)).toNumber();
+      let second = u8aToBn(value.subarray(1, 2)).toNumber();
+      const encoded: number = first + (second << 8);
+      const period = 2 << (encoded % (1 << 4));
+      const quantizeFactor = Math.max(period >> 12, 1);
+      let phase = (encoded >> 4) * quantizeFactor;
```
jacogr: Why is this `let`, not assigned again?

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+      if (period >= 4 && phase < period) {
```
jacogr: (nit) Blank line between var definitions and code blocks.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+        return [new U8(period), new U8(phase)];
+      }
+      throw new Error('Invalid data passed to Mortal era');
```
jacogr: (nit) Blank line before throw

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+  constructor (value?: any) {
+    super({
+      period : U8, phase : U8
+    }, MortalEra.decodeMortalEra(value));
+  }
+
+  private static decodeMortalEra (value: MortalMethod | Uint8Array): MortalEraValue {
+    if (isU8a(value)) {
+      const first = u8aToBn(value.subarray(0, 1)).toNumber();
+      let second = u8aToBn(value.subarray(1, 2)).toNumber();
```
jacogr: Why is this not a `const`?

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+      let phase = (encoded >> 4) * quantizeFactor;
+      if (period >= 4 && phase < period) {
+        return [new U8(period), new U8(phase)];
+      }
+      throw new Error('Invalid data passed to Mortal era');
+    } else if (isObject(value)) {
+      const { current } = value;
+      const { period } = value;
+      let calPeriod = Math.pow(2, Math.ceil(Math.log2(period)));
+      calPeriod = Math.min(Math.max(calPeriod, 4), 1 << 16);
```
jacogr: Blank line before and after so the definitions and actual logic are separated and well denoted for future reading and maintenance.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+      const phase = current % calPeriod;
+      const quantizeFactor = Math.max(calPeriod >> 12, 1);
+      const quantizedPhase = phase / quantizeFactor * quantizeFactor;
+      return [new U8(calPeriod), new U8(quantizedPhase)];
```
jacogr: Blank line before return.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
     }
+    throw new Error('Invalid data passed to Mortal era');
```
jacogr: Blank line before throw.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+  }
+  /**
```
jacogr: (nit) Blank line between the end of the functions and the comments blocks (also applies to the rest of the file). It really helps understanding and reading.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+   */
+  toU8a (isBare?: boolean): Uint8Array {
+    const period = this.period.toNumber();
+    const phase = this.phase.toNumber();
+    const quantizeFactor = Math.max(period >> 12, 1);
+    const trailingZeros = this.getTrailingZeros(period);
+    const encoded = Math.min(15, Math.max(1, trailingZeros - 1)) + (((phase / quantizeFactor) << 4));
+    const first = encoded >> 8;
+    const second = encoded & 0xff;
+    return new Uint8Array([second, first]);
```
jacogr: (nit) Blank line before.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+  /**
+   * @description Encodes the value as a Uint8Array as per the parity-codec specifications
+   * @param isBare true when the value has none of the type-specific prefixes (internal)
+   */
+  toU8a (isBare?: boolean): Uint8Array {
+    const period = this.period.toNumber();
+    const phase = this.phase.toNumber();
+    const quantizeFactor = Math.max(period >> 12, 1);
+    const trailingZeros = this.getTrailingZeros(period);
+    const encoded = Math.min(15, Math.max(1, trailingZeros - 1)) + (((phase / quantizeFactor) << 4));
```
jacogr: In this case, it would make sense to have a short comment of how the structure operates here - it makes maintenance and checking the logic easier without having to know the specs off-by-hand. (Additionally, this can then be checked against any future spec changes as well)

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,189 @@
...
+    return this.birth(current) + this.period.toNumber();
+  }
+
+  /**
+   * @description convert the number to binary and get the trailing zero's.
+   */
+  private getTrailingZeros (period: number) {
+    const binary = period.toString(2);
+    let index = 0;
+    while (binary[binary.length - 1 - index] === '0') {
```
jacogr: (nit) Spacing before and after the loop.

packages/types/src/types.ts
```diff
@@ -124,6 +124,7 @@ export interface IMethod extends Codec {
 
 export interface IExtrinsicSignature extends Codec {
   readonly isSigned: boolean;
+  era: Codec;
```
jacogr: Not quite correct here - this means it can be any Codec, so have the interface defined (which would then be implemented by the Era classes)

KarishmaBothara: Thanks.. Can you please suggest some interface. We used Codec cause we didn't find anything in common in Mortal and Immortal Era which we could place in interface and moreover its only used in ExtrinsicSignature to encode U8a

packages/api/test/e2e/promise-tx.spec.js
KarishmaBothara: done

packages/types/src/type/ExtrinsicEra.spec.ts
KarishmaBothara: done

packages/types/src/type/ExtrinsicEra.ts
KarishmaBothara: added in comments section

packages/types/src/type/ExtrinsicEra.ts
KarishmaBothara: done

packages/types/src/type/ExtrinsicEra.ts
KarishmaBothara: done

packages/types/src/types.ts
jacogr: Well, `era` is an `ExtrinsicEra` - possibly `IExtrinsicEra` (with `asMortal` & `asImmortal`) - that is if it is actually used where the interface is passed.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,224 @@
...
 
+import { assert, hexToU8a, isHex, isU8a, isObject, u8aToBn } from '@polkadot/util';
 import { AnyU8a } from '../types';
+import Enum from '../codec/Enum';
+import Tuple from '../codec/Tuple';
+import U8 from '../primitive/U8';
+import U8a from '../codec/U8a';
 
-import { u8aToU8a } from '@polkadot/util';
+export type MortalEraValue = [U8, U8];
```
jacogr: Does this need to be exported? Cannot see it used externally from a quick search.

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,224 @@
...
-    super(
-      ExtrinsicEra.decodeExtrinsicEra(value)
-    );
+    super(value);
+
+    assert(this.eq(VALID_IMMORTAL), `IMMORTAL: expected ${VALID_IMMORTAL.toHex()}, found ${this.toHex()}`);
+  }
+}
+
+interface MortalMethod {
```
jacogr: Move `interface` (also a type), up to the other type defs in this file.

packages/types/src/type/ExtrinsicEra.ts
KarishmaBothara: no, shouldn't be exported..changed

packages/types/src/types.ts
ianhe8x: Sure we will make this change as you said, actually we've thought about that, only problem is the return type of asMortal and asImmortal would be a real implementation class, which seems against the reason we use interface here.

jacogr: Yes, that would be the quick one - which doesn't require adding yet more.

packages/api/test/e2e/promise-tx.spec.js
```diff
@@ -2,6 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
+import { ExtrinsicEra } from '@polkadot/types/type';
```
amaurymartiny: nit: alphabetical ordering

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,224 @@
...
+/**
+ * @name MortalEra
+ * @description
+ * The MortalEra for an extrinsic, indicating period and phase
+ */
+export class MortalEra extends Tuple {
+  constructor (value?: any) {
+    super({
+      period: U8,
+      phase: U8
```
amaurymartiny: both of these should extend U64: https://github.com/paritytech/substrate/blob/0e6a407a13fbc5a5ad200645aed72bb8a8e528d7/core/sr-primitives/src/generic/era.rs#L25

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,224 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { AnyU8a } from '../types';
+import { assert, hexToU8a, isHex, isU8a, isObject, u8aToBn } from '@polkadot/util';
+import { AnyU8a, IExtrinsicEra } from '../types';
```
amaurymartiny: nit: blank line above this line

packages/types/src/type/ExtrinsicEra.ts
```diff
@@ -2,33 +2,224 @@
...
+    }
+  }
+}
+
+/**
+ * @name ImmortalEra
+ * @description
+ * The ImmortalEra for an extrinsic
+ */
+export class ImmortalEra extends U8a {
```
amaurymartiny: Should extend Null: https://github.com/paritytech/substrate/blob/0e6a407a13fbc5a5ad200645aed72bb8a8e528d7/core/sr-primitives/src/generic/era.rs#L32

ianhe8x: ImmortalEra has encodedLength of 1, while Null is 0

packages/types/src/type/ExtrinsicEra.ts
ianhe8x: period and phase are u64 for arithmetic operation and encoded as u8

packages/types/src/type/ExtrinsicEra.ts
amaurymartiny: So there are 2 ways to encode/decode this:

1. `Enum<ImmortalEra | MortalEra>`. Normally Enum encoding starts with 0 or 1, which denotes which enumIndex. So if it's 0, then we have ImmortalEra, and ImmortalEra should be of encodedLength=0; If it's not 0, then we don't care about the 0/1, and decode MortalEra.

2. `Enum<ImmortalEra | MortalEra>`, where we manually remove the starting 0/1 enumIndex, in this case we set encodedLength(ImmortalEra)=1 and and `ImmortalEra.toU8a() = [0]`;

Both work, but imo 1 makes a little bit more sense: we don't need the `assert(this.eq(VALID_IMMORTAL))` anymore.

packages/types/src/type/ExtrinsicEra.ts
amaurymartiny: Yes. So decodeMortalEra should take [U8, U8], do the magic decoding, and return [U64, U64] to be passed into the constructor.

Looking at the test now: `expect(extrinsicEra.asMortalEra.phase.toNumber()).toEqual(20000);`, where phase is a U8...

amaurymartiny: > I think we don't have to follow that because in js side they are anyway a BN.

Ah, just read your edit. OK this is debatable, though I still prefer U64 for correctness. If I'm not mistaken, there are only these 2 lines to change, if won't affect anywhere else.

packages/types/src/type/ExtrinsicEra.ts
amaurymartiny: Actually, come to think of it, my proposal 1/ is not necessarily cleaner, we can keep like this. Maybe U8aFixed makes more sense though?

packages/types/src/type/ExtrinsicEra.ts
ianhe8x: my concern is if we make them u64, will we potentially have overflow issues?

packages/types/src/type/ExtrinsicEra.ts
ianhe8x: For ImmortalEra, it can go both ways. but for MortalEra, the first byte is not fixed to [1], which makes it logically conflict with your proposal 1.

I think enum still make sense in this case, and in rust side it's an enum.

packages/types/src/type/ExtrinsicEra.ts
ianhe8x: i think you are right

packages/types/src/type/ExtrinsicEra.ts
amaurymartiny: Sounds good 👍

PR #928, opened 2019-05-17
Show section.method on storage read error

jacogr:
Allows for better debugging when things go south, it allows the logging of the exact key that is throwing an exception on storage decoding.

cc @jnaviask

### DIFF
--- a/packages/types/src/primitive/StorageKey.ts
+++ b/packages/types/src/primitive/StorageKey.ts
@@ -48,17 +48,16 @@ export default class StorageKey extends Bytes {
   }
 
   static decodeStorageKey (value?: AnyU8a | StorageKey | StorageFunction | [StorageFunction, any]): Decoded {
-    if (!value || isString(value)) {
+    if (value instanceof StorageKey) {
+      return {
+        key: value,
+        method: value.method,
+        section: value.section
+      };
+    } else if (!value || isString(value) || isU8a(value)) {
       // let Bytes handle these inputs
       return {
         key: value
-      };
-    } else if (isU8a(value)) {
-      // let Bytes handle these inputs
-      return {
-        key: value,
-        method: (value as StorageKey).method,
-        section: (value as StorageKey).section
       };
     } else if (isFunction(value)) {
       return {
--- a/packages/types/src/primitive/StorageKey.ts
+++ b/packages/types/src/primitive/StorageKey.ts
@@ -11,12 +11,18 @@ import { StorageFunctionMetadata as MetaV4 } from '../Metadata/v4/Storage';
 
 export interface StorageFunction {
   (arg?: any): Uint8Array;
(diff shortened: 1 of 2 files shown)


packages/types/src/primitive/StorageKey.ts
```diff
@@ -81,16 +110,30 @@ export default class StorageKey extends Bytes {
...
   /**
-   * @description The metadata or `null` when not available
+   * @description The metadata or `undefined` when not available
    */
   get meta (): MetaV4 | undefined {
     return this._meta;
   }
 
+  /**
+   * @description The key mehod or `undefined` when not specified
```
amaurymartiny:
```suggestion
   * @description The key method or `undefined` when not specified
```

packages/types/src/primitive/StorageKey.ts
```diff
@@ -26,27 +32,50 @@ export interface StorageFunction {
...
-  static decodeStorageKey (value?: AnyU8a | StorageKey | StorageFunction | [StorageFunction, any]): Uint8Array | string | undefined {
-    if (!value || isU8a(value) || isString(value)) {
+  static decodeStorageKey (value?: AnyU8a | StorageKey | StorageFunction | [StorageFunction, any]): Decoded {
+    if (!value || isString(value)) {
       // let Bytes handle these inputs
-      return value;
+      return {
+        key: value
+      };
+    } else if (isU8a(value)) {
```
amaurymartiny: Totally equivalent, but I would put `instanceof StorageKey` here, and isU8a in the above case (which must then go below in the if/else block).

packages/types/src/primitive/StorageKey.ts
```diff
@@ -26,27 +32,50 @@ export interface StorageFunction {
...
     this._meta = StorageKey.getMeta(value as StorageKey);
+    this._method = method;
     this._outputType = StorageKey.getType(value as StorageKey);
+    this._section = section;
   }
 
-  static decodeStorageKey (value?: AnyU8a | StorageKey | StorageFunction | [StorageFunction, any]): Uint8Array | string | undefined {
-    if (!value || isU8a(value) || isString(value)) {
+  static decodeStorageKey (value?: AnyU8a | StorageKey | StorageFunction | [StorageFunction, any]): Decoded {
+    if (!value || isString(value)) {
```
amaurymartiny: We should probably throw on undefined, just to be sure. Also, isString should be isHex?

jacogr: Since we don't have a hex type and the input is a string, kept to as `isString` - I'd rather let Bytes sort it out for all those cases.

jacogr: `undefined` is the empty case, i.e. `new StorageKey()` - which should just return the default value.

Issue #863, opened 2019-05-02
optimize double map first key hash


thiolliere:
https://github.com/paritytech/substrate/pull/2451

thiolliere:
OK I'm finally not sure about the way to go we can also postponed this as it doesn't warm.

Also I don't know is this suppose to follow v1.0 or master ?

jacogr:
API needs to support - 1.x as used by devs, testnets & master. (Which takes some juggling in some cases)

PR #934, opened 2019-05-20
support double map storage function

AlexWZ0129:
#776 Hi, here is an implement to support double map with minimal changes, which uses an array to represent keys of the double map, I am working on changing signature to use two arguments.

### DIFF
--- a/packages/api/test/e2e/promise-queries.spec.js
+++ b/packages/api/test/e2e/promise-queries.spec.js
@@ -168,4 +168,19 @@ describe.skip('e2e queries', () => {
       done();
     });
   });
+
+  it('queries state using double map key', async () => {
+    // TODO Update ['any', '0x1234'] to the key of a known event topic and update '[]' to the expected value
+    const eventTopics = await api.query.system.eventTopics(['any', '0x1234']);
+
+    expect(eventTopics.toString()).toEqual('[]');
+  });
+
+  it('subscribes to queries using double map key', async (done) => {
+    // TODO Update ['any', '0x1234'] to the key of a known event topic and update '[]' to the expected value
+    api.query.system.eventTopics(['any', '0x1234'], (eventTopics) => {
+      expect(eventTopics.toString()).toEqual('[]');
+      done();
+    });
+  });
 });
--- a/packages/type-storage/src/fromMetadata/createFunction.spec.ts
+++ b/packages/type-storage/src/fromMetadata/createFunction.spec.ts
@@ -2,7 +2,9 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { stringToU8a, u8aConcat } from '@polkadot/util';
+import { StorageHasher, Text } from '@polkadot/types';
+import { StorageFunction } from '@polkadot/types/primitive/StorageKey';
+import { stringToU8a, u8aConcat, u8aToHex } from '@polkadot/util';
 
 import createFunction from './createFunction';
 
@@ -52,5 +54,61 @@ describe('createFunction', () => {
         stringToU8a(':auth:len')
       )
     );
+  });
+
+  describe('the created double map function', () => {
+    let storageFn: StorageFunction;
+    beforeAll(() => {
+      storageFn = createFunction(
+        'GenericAsset',
+        'FreeBalance', {
+          name: 'metaName',
+          type: {
+            isDoubleMap: true,
+            asDoubleMap: {
+              hasher: new StorageHasher('Blake2_256'),
+              key1: new Text('AccountId'),
+              key2: new Text('AccountId'),
+              value: new Text('Balance'),
+              key2Hasher: new Text('twox_128')
+            }
+          }
+        } as any
+      );
+    });
+
+    it('should return correct key', () => {
+      const result = storageFn(['5DXUeE5N5LtkW97F2PzqYPyqNkxqSWESdGSPTX6AvkUAhwKP', '5DXUeE5N5LtkW97F2PzqYPyqNkxqSWESdGSPTX6AvkUAhwKP']);
+      expect(u8aToHex(result)).toEqual('0x00fa40e72d7173e69ee54b980345ea01cb81e64258502e0247af4303dee91ec0aec2ecd3a60ab080cff7b52a8f6d543b');
+    });
+
+    it('needs two arguments', () => {
+      expect(
+        () => storageFn(['5DXUeE5N5LtkW97F2PzqYPyqNkxqSWESdGSPTX6AvkUAhwKP'])
+      ).toThrow(/metaName expects two arguments/);
+    });
+  });
+
+  it('allows creates double map function with a Null type key', () => {
+    const storageFn = createFunction(
+        'System',
+        'EventTopics',
+        {
+          type: {
+            isDoubleMap: true,
+            asDoubleMap: {
+              hasher: new StorageHasher('Blake2_256'),
+              key1: new Text('Null'),
+              key2: new Text('Hash'),
+              value: new Text('Vec<(BlockNumber,EventIndex)>'),
+              key2Hasher: new Text('blake2_256')
+            }
+          }
+        } as any
+      );
+    // the value of the Null type key does not effect the result
+    expect(u8aToHex(storageFn(['any', [1, 2, 3]]))).toEqual(u8aToHex(storageFn([[1, 2, 3], [1, 2, 3]])));
+    // the value of the not Null type key does effect the result
+    expect(u8aToHex(storageFn(['any', [1, 2, 3]]))).not.toEqual(u8aToHex(storageFn(['any', [1, 2, 3, 4]])));
   });
 });
--- a/packages/type-storage/src/fromMetadata/createFunction.ts
+++ b/packages/type-storage/src/fromMetadata/createFunction.ts
@@ -31,7 +31,6 @@ export default function createFunction (section: Text | string, method: Text | s
   const rawKey = stringToU8a(stringKey);
 
   // Get the hashing function
-  // FIXME Hash correctly for double map too
   let hasher: HasherFunction;
   let key2Hasher: HasherFunction;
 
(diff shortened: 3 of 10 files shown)


ianhe8x:
also solves #863

packages/type-storage/src/fromMetadata/createFunction.ts
```diff
@@ -32,15 +32,36 @@ export default function createFunction (section: Text | string, method: Text | s
...
   // - storage.timestamp.blockPeriod()
   const storageFn = (arg?: any): Uint8Array => {
     let key = rawKey;
+
+    if (meta.type.isDoubleMap) {
+      assert(!isUndefined(arg) && !isNull(arg) && !isUndefined(arg[0]) && !isNull(arg[0]) && !isUndefined(arg[1]) && !isNull(arg[1]), `${meta.name} expects two arguments`);
+      const type1 = meta.type.asDoubleMap.key1.toString();
+      const type2 = meta.type.asDoubleMap.key2.toString();
+
+      const param1Encoded = u8aConcat(key, createType(type1, arg[0]).toU8a(true));
```
amaurymartiny: `true` here means we don't length-prefix the U8a. Intuitively I would say we need to add the length prefix, but I may be wrong.

Would like to see an e2e test (with event_topics)

packages/type-storage/src/fromMetadata/getHasher.ts
```diff
@@ -2,19 +2,39 @@
...
+  switch (text.toString()) {
+    case 'blake2_128':
+      return new StorageHasher('Blake2_128');
+    case 'blake2_256':
+      return new StorageHasher('Blake2_256');
+    case 'twox_128':
+      return new StorageHasher('Twox128');
+    case 'twox_256':
+      return new StorageHasher('Twox256');
+    default:
```
amaurymartiny: Could you add a `twox_64_concat` case here too?

packages/type-storage/src/fromMetadata/getHasher.ts
```diff
@@ -2,19 +2,39 @@
...
+      return new StorageHasher(text.toString());
```
amaurymartiny: I think we can throw here instead, not sure what `new StorageHasher(random stuff)` will do

packages/types/src/codec/createType.ts
```diff
@@ -148,7 +149,8 @@ export function getTypeDef (_type: Text | string, name?: string): TypeDef {
     value.sub = getTypeDef(subType);
   } else if (startingWith(type, 'DoubleMap<', '>')) {
     // FIXME As a first step, only allow parsing of DoubleMap, don't actually init
```
amaurymartiny: Can remove this comment

packages/type-storage/src/fromMetadata/createFunction.ts
```diff
@@ -32,15 +32,36 @@ export default function createFunction (section: Text | string, method: Text | s
 
   // Get the hashing function
   // FIXME Hash correctly for double map too
```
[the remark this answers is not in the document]
AlexWZ0129: > This looks very good, thanks @AlexWZ0129 👍
> 
> Could you add one e2e test inside `/api/tests/e2e/promise-queries.js`? With event_topics, to make sure everything works.

Thanks for the feedback @amaurymartiny . Will add an e2e test here.

packages/type-storage/src/fromMetadata/createFunction.ts
AlexWZ0129: yep. will double check in the e2e test

packages/type-storage/src/fromMetadata/getHasher.ts
AlexWZ0129: okay

packages/type-storage/src/fromMetadata/getHasher.ts
AlexWZ0129: yep.

packages/type-storage/src/fromMetadata/getHasher.ts
AlexWZ0129: Yep. Btw, in StorageHasher, I only see 'Twox128Concat', does 'twox_64_concat' match the 'Twox128Concat'?

amaurymartiny: Actually I think I was mistaken, iirc at some point there was Twox128Concat in the substrate code, but now everything is Twox64Concat.

So in api, we should only have Twox64Concat and twox_64_concat

AlexWZ0129: Cool. I see. Thanks @amaurymartiny !

packages/types/src/primitive/StorageKey.ts
```diff
@@ -70,6 +70,14 @@ export default class StorageKey extends Bytes {
 
       assert(isFunction(fn), 'Expected function input for key construction');
 
+      if (Array.isArray(arg[0])) {
```
jacogr: Can we just add a comment here to indicate why we are doing this? (2 months from now this will look like black magic without background/info)

packages/type-storage/src/fromMetadata/createFunction.ts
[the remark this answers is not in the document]
AlexWZ0129: > Can remove this comment now

Hi @amaurymartiny , please excuse me if it's a silly question, do you know if there is any event topic I can query for testing on substrate? Or anyway I can create one?

packages/types/src/primitive/StorageKey.ts
AlexWZ0129: Thanks for the review. Yeah, I will add some commands here.

packages/type-storage/src/fromMetadata/createFunction.ts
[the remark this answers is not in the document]
AlexWZ0129: > Can remove this comment now
I didn't find any extrinsic to create an event topic on the substrate. But I tested on our chain(using double map for the balance), it's working.

AlexWZ0129:
> This looks very good, thanks @AlexWZ0129 👍
> 
> Could you add one e2e test inside `/api/tests/e2e/promise-queries.js`? With event_topics, to make sure everything works.

I didn't find any extrinsic to create an event topic on the substrate. But I tested on our chain(using double map for the balance), it's working.

packages/type-storage/src/fromMetadata/createFunction.ts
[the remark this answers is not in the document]
AlexWZ0129: > > Can remove this comment now
> 
> Hi @amaurymartiny , please excuse me if it's a silly question, do you know if there is any event topic in the state I can query for testing on substrate, or anyway I can create one?

I didn't find any extrinsic to create an event topic on the substrate. But I tested on our chain(using double map for the balance), it's working.

packages/type-storage/src/fromMetadata/createFunction.spec.ts
```diff
@@ -53,4 +55,60 @@ describe('createFunction', () => {
...
+            }
+          }
+        } as any
+      );
+    });
+
+    it('should return correct key', () => {
+      const result = storageFn(['5DXUeE5N5LtkW97F2PzqYPyqNkxqSWESdGSPTX6AvkUAhwKP', '5DXUeE5N5LtkW97F2PzqYPyqNkxqSWESdGSPTX6AvkUAhwKP']);
+      expect(u8aToHex(result)).toEqual('0x00fa40e72d7173e69ee54b980345ea01cb81e64258502e0247af4303dee91ec0aec2ecd3a60ab080cff7b52a8f6d543b');
+    });
```
AlexWZ0129: This expected value '0x00fa40e72d7173e69ee54b980345ea01cb81e64258502e0247af4303dee91ec0aec2ecd3a60ab080cff7b52a8f6d543b' here is from the intergration test I did on our chain. This value is the key on the chain. And the storagefunction created here returns the same key.

packages/type-storage/src/fromMetadata/createFunction.ts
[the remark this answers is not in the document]
jacogr: Happy to know it works. Would like something more repeatable as part of the e2e tests, so for the srml system module, this is available - https://github.com/paritytech/substrate/blob/master/srml/system/src/lib.rs#L344

packages/types/src/primitive/StorageKey.ts
```diff
@@ -70,6 +70,14 @@ export default class StorageKey extends Bytes {
 
       assert(isFunction(fn), 'Expected function input for key construction');
 
+      if (fn.meta.type.isDoubleMap) {
```
jacogr: For safety, rather use `if (fn.meta && fn.meta.type. isDoubleMap)` - it is possible that it is just a function passed in since it can come from anywhere. (The `fn.method` and `fn.section` below doesn't have those issues since they will just be `undefined`)

AlexWZ0129: Ah. cool! Thank you @jacogr ! I will change it now.

packages/type-storage/src/fromMetadata/createFunction.ts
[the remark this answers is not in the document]
AlexWZ0129: > https://github.com/paritytech/substrate/blob/master/srml/system/src/lib.rs#L344

Okay. I will add an e2e test here. But as I know, there aren't any EventTopics in the state of the chain? I will just in any two random parameters and get `[]` as a result, is that fine?

packages/api/test/e2e/promise-queries.spec.js
```diff
@@ -168,4 +168,19 @@ describe.skip('e2e queries', () => {
...
+    expect(eventTopics.toString()).toEqual('[]');
+  });
+
+  it('subscribes to queries using double map key', async (done) => {
+    // TODO Update ['any', '0x1234'] to the key of a known event topic and update '[]' to the expected value
+    api.query.system.eventTopics(['any', '0x1234'], (eventTopics) => {
+      expect(eventTopics.toString()).toEqual('[]');
+      done();
+    });
+  });
```
AlexWZ0129: I added an e2e test here, now it actually doesn't test anything here... But if we update ['any', '0x1234'] to the key of a known event topic, it will test the double map.

jacogr: Yes, in but not fully used. Cannot win them all. It is fine, not perfect, but fine. (Since you have verified against a real supported chain, will accept that - and the above gives the gap to properly extend once substrate moves along a bit)

AlexWZ0129: Thanks. Yeah, we could extend this later.

AlexWZ0129:
> lgtm!

Thanks @amaurymartiny !

PR #935, opened 2019-05-20
ExtrinsicSignature init with address/signature strings

jacogr:
- Address can take ss-58 inputs
- Signature can take hex inputs

### DIFF
--- a/packages/types/src/type/ExtrinsicSignature.ts
+++ b/packages/types/src/type/ExtrinsicSignature.ts
@@ -128,7 +128,7 @@ export default class ExtrinsicSignature extends Struct implements IExtrinsicSign
   /**
    * @description Adds a raw signature
    */
-  addSignature (_signer: Address | Uint8Array, _signature: Uint8Array, _nonce: AnyNumber, _era: Uint8Array = IMMORTAL_ERA): ExtrinsicSignature {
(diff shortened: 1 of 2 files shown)


packages/types/src/type/ExtrinsicSignature.ts
```diff
@@ -128,7 +128,7 @@ export default class ExtrinsicSignature extends Struct implements IExtrinsicSign
   /**
    * @description Adds a raw signature
    */
-  addSignature (_signer: Address | Uint8Array, _signature: Uint8Array, _nonce: AnyNumber, _era: Uint8Array = IMMORTAL_ERA): ExtrinsicSignature {
+  addSignature (_signer: Address | Uint8Array | string, _signature: Uint8Array | string, _nonce: AnyNumber, _era: Uint8Array = IMMORTAL_ERA): ExtrinsicSignature {
```
amaurymartiny: It's a bigger refactor, but we could add a `Hex` type in api, which basically is `type Hex = string;`, but at least will make it clearer where we want strings and where Hex

jacogr: Yes, each and everytime I do this type of thing, I think of "we should have a Hex type".

Issue #908, opened 2019-05-15
Re-work test setup for e2e tests


Stefie:
While working on the api-derive tests I noticed that the test setup in API could urgently need some attention. 
Now that we're more people working on it who don't necessarily know the whole code by heart (including external contributors) we should put more effort in making the tests more bulletproof and the setup more intuitive.

- [ ] Reorder and better categorize existing tests
- [ ] Have a central file where all local tests are imported to be able to toggle all tests that require a local node (`.skip`)
- [ ] Have a setup that runs all tests on the **ApiPromise** and **ApiRx** for both **Polkadot** (wss://poc3-rpc.polkadot.io/) and the latest **Substrate** master on a local machine
- [ ] Replace console.logs in tests with actual [Jest methods](https://jestjs.io/docs/en/expect.html)
- [ ] Try to bring up test coverage
- [ ] Fix broken / outdated tests
- [ ] Add some documentation for external contributors

Related:
#329 
#868

Stefie:
I'd pick up the first three (setup/ reordering) next, I just want to finish 2 other things today.

amaurymartiny:
> Have a central file where all local tests are imported to be able to toggle all tests that require a local dev node

Related to this: I would propose to put **all** e2e tests into `api/`, because when testing e2e, we should test what users use, namely `const api = new Api()`. This includes:
- `rpc-core`: replace those tests with `api.rpc.*.*()` tests, so that's it's really e2e
- `api-derive`: they test the decorated `api.derive.*.*`, so they are already e2e tests on api itself
- `type-extrinsics`: there are some tests hanging there, move them to `api.tx.*.*` tests

Stefie:
Maybe we should split this ticket up into smaller chunks, it's a monster anyways....

jacogr:
Can just link the PRs to it and tick off.

amaurymartiny:
Or instead of having a central file which imports all other files to have a central `.skip`, we could do what YJ proposed here: https://github.com/polkadot-js/api/pull/1059/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R23

Stefie:
Yeah, I've seen it and that looks really handy!

amaurymartiny:
So a possible folder architecture:
```
e2e/
  |- alex/
      |- backwardsCompat/
          |- tests to test backwards compat go here
      |- promise.spec.ts
      |- promise-archive.spec.ts
  |- dev/
[...]

PR #936, opened 2019-05-21
Convert tests to ts and fix existing

Stefie:
Fixes https://github.com/polkadot-js/api/issues/329

as preparation for https://github.com/polkadot-js/api/issues/908


jacogr:
Brilliant! Was taking a peek this weekend to see how much was left, the provider ones (without looking at details), had me down into a cold sweat. Really happy to see this.

Stefie:
Yes! There's still quite a bit of work to do, but I figured that I'd rather clean that out of the way before starting to refactor. (As this also gives me quite a good overview of which tests are there already) :)

packages/rpc-provider/src/http/polyfill.spec.ts
```diff
@@ -2,12 +2,15 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
+import { Global } from '@polkadot/rpc-provider/mock/types';
+
+declare const global: Global;
+
 describe('http/polyfill', () => {
-  let origFetch;
+  let origFetch: GlobalFetch;
```
Stefie: I could need a second pair of eyes on this file / tests

packages/rpc-provider/src/ws/polyfill.spec.ts
```diff
@@ -2,25 +2,18 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-describe('ws/polyfill', () => {
-  let origWs;
+import { Global } from '@polkadot/rpc-provider/mock/types';
```
Stefie: I also could need a second pair of eyes on this file / tests.

packages/rpc-provider/src/ws/send.spec.ts
```diff
@@ -31,7 +34,6 @@ describe('send', () => {
 
     if (mock) {
       mock.done();
-      mock = null;
     }
   });
```
Stefie: Here I could also need a second pair of eyes.
I'm getting this error with my `Global` solution:

```
packages/rpc-provider/src/ws/send.spec.ts:45:5 - error TS2740: Type 'typeof WebSocket' is missing the following properties from type 'WebSocket': binaryType, bufferedAmount, extensions, onclose, and 15 more.

45     global.WebSocket = class {
```

and 

```

packages/rpc-provider/src/ws/send.spec.ts:56:5 - error TS2532: Object is possibly 'undefined'.

56     websocket.onopen();
```

Stefie:
I'm down to three 3 tslint errors and one broken test  🎉 (from 150+), but now I could need a second pair of eyes on the last, really resistant ones:

The errors can be found here:
https://travis-ci.com/polkadot-js/api/builds/112971448

packages/api/test/e2e/promise-tx.spec.ts
```diff
@@ -138,61 +145,62 @@ describe.skip('e2e transactions', () => {
...
-    const eraDeath = exERA.asMortalEra.death(currentHeight.toNumber());
-    console.log('STARTED AT :'+eraBirth+' EXPIRED AT :'+eraDeath);
-    const eraHash = await api.rpc.chain.getBlockHash(eraBirth);
-    const ex = api.tx.balances
-      .transfer(keyring.eve.address(), 12345);
-    const hash = await ex.signAndSend(keyring.dave, {blockHash: eraHash, era:exERA, nonce});
-
-    expect(hash.toHex()).toHaveLength(66);
-    done();
+    // const nonce = await api.query.system.accountNonce(keyring.dave.address());
```
Stefie: I also commented out the ERA tests because they're broken on different  levels and I haven't looked into the ERA implementation yet. I'd be thankful if someone else would take over here

packages/api/test/e2e/promise-alex-archive.spec.ts
```diff
@@ -2,10 +2,13 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import Api from '../../src/promise';
+import { ApiPromiseInterface } from '@polkadot/api/promise/types';
+import { Extrinsic } from '@polkadot/types';
+
+import Api from '@polkadot/api/promise';
```
amaurymartiny: it's my personal opinion, but I prefer relative imports in tests, so that we know in a glance what we are testing

packages/api/test/e2e/promise-alex-archive.spec.ts
```diff
@@ -2,10 +2,13 @@
...
 
 describe.skip('alex archive queries (local)', () => {
-  let api;
+  let api: ApiPromiseInterface;
```
amaurymartiny: `Api` is more conceptually correct

packages/api/test/e2e/promise-alex-archive.spec.ts
```diff
@@ -34,25 +37,23 @@ describe.skip('alex archive queries (local)', () => {
...
-
-        return true;
-      }
-    });
-
-    expect(failed).toHaveLength(0);
-
-    done();
+    api.rpc.chain.getBlock('0x85c62b581f38cb81c3e443d34392672beb1fb877017fd7237cc87704113259dc', result => {
+      const failed: boolean = result.block.extrinsics.filter((extrinsic: Extrinsic) => {
```
amaurymartiny: failed is not a boolean here, it's an array of extrinsics. If we manually cast something here, it's better to do it on `result` here

packages/api/test/e2e/promise-alex.spec.ts
```diff
@@ -32,30 +34,31 @@ describe.skip('alex queries', () => {
...
+        done();
+      }).then().catch();
     });
 
     it('Gets the hash of the last finalized header', async (done) => {
       api.rpc.chain.getFinalizedHead((head) => {
         expect(head instanceof Hash).toBe(true);
         done();
-      });
+      }).then().catch();
```
amaurymartiny: .then is useless. Also, we could console.error the .catch just in case

packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -41,57 +42,55 @@ describe.skip('e2e queries', () => {
   it('queries state for a balance', async () => {
     const balance = await api.query.balances.freeBalance(keyring.alice.address());
 
-    expect(
-      balance.isZero()
-    ).toBe(false);
```
amaurymartiny: This test is stronger, we should keep it

packages/api/test/e2e/rx-queries.spec.ts
```diff
@@ -29,10 +32,8 @@ describe.skip('e2e queries', () => {
 
   it('queries state for a balance', (done) => {
     api.query.balances.freeBalance(keyring.alice.address()).subscribe((balance) => {
-      expect(
-        balance.isZero()
-      ).toBe(false);
 
+      expect(balance).toBeInstanceOf(BN);
```
amaurymartiny: Same: we should test isZero false here

packages/rpc-provider/src/http/index.spec.ts
```diff
@@ -4,10 +4,12 @@
...
 import { TEST_HTTP_URL } from '../../test/mockHttp';
 
-import Http from '.';
+import { ProviderInterface } from '@polkadot/rpc-provider/types';
+
+import Http from '@polkadot/rpc-provider/http';
 
 describe('Http', () => {
-  let http;
+  let http: ProviderInterface;
```
amaurymartiny: Same here: `Http` is more correct

packages/rpc-provider/src/http/polyfill.spec.ts
```diff
@@ -19,7 +22,7 @@ describe('http/polyfill', () => {
   });
 
   it('polyfills with no exceptions (with fetch)', () => {
-    global.fetch = () => true;
+    expect(global.fetch).toBeTruthy();
```
amaurymartiny: No: we should keep `global.fetch = () => true;` here, or else this test is the same as the one before

packages/rpc-provider/src/http/polyfill.spec.ts
```diff
@@ -19,7 +22,7 @@ describe('http/polyfill', () => {
...
 
     expect(require('./polyfill')).toBeDefined();
   });
```
amaurymartiny: Overall these tests were actually wrong before: `require(...)` will always be defined.

Here's the corrected version:
```typescript
// Copyright 2017-2019 @polkadot/rpc-provider authors & contributors

import { Global } from '@polkadot/rpc-provider/mock/types';

declare const global: Global;

describe('http/polyfill', () => {
  let origFetch: GlobalFetch;

  beforeEach(() => {
    origFetch = global.fetch;
  });

  afterEach(() => {
    global.fetch = origFetch;
  });

  it('polyfills with no exceptions (without fetch)', () => {
    (global as any).fetch = undefined;
    require('./polyfill');

    expect(global.fetch).toBeDefined();
  });

  it('polyfills with no exceptions (with fetch)', () => {
    (global as any).fetch = () => true;
    require('./polyfill');

    expect(global.fetch).toBeDefined();
  });
});

```

packages/rpc-provider/src/http/send.spec.ts
```diff
@@ -32,7 +34,7 @@ describe('send', () => {
     return http
       .send('test_body', ['param'])
       .then((result) => {
-        expect(mock.body['test_body']).toEqual({
+        expect(Object.values(mock.body)[0]).toEqual({
```
amaurymartiny: I believe test_body is easier to read

packages/rpc-provider/src/mock/index.ts
```diff
@@ -38,6 +38,7 @@ const l = logger('api-mock');
  * @return {ProviderInterface} The mock provider
  */
 export default class Mock implements ProviderInterface {
+  public body: any;
```
amaurymartiny: I really don't think there should be a body here. Body is used to mock the http response

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -69,7 +69,7 @@ export default class WsProvider implements WSProviderInterface {
   private autoConnect: boolean;
   private coder: Coder;
   private endpoint: string;
-  private handlers: {
+  public handlers: {
```
amaurymartiny: I still think these should be private, end-users touch WsProvider, and they shouldn't care what handlers are.

To silent the TS accessing private methods, just put @ts-ignore and a small comment

packages/rpc-provider/src/ws/index.spec.ts
```diff
@@ -4,29 +4,29 @@
...
 
-function createWs (requests, autoConnect) {
-  mock = mockWs(requests);
-  ws = new Ws(TEST_WS_URL, autoConnect);
+let ws: WsProvider;
+let mock: Mock;
 
+function createWs (requests: Array<any>, autoConnect: Boolean) {
+  mock = mockWs(requests);
+  ws = new WsProvider(TEST_WS_URL, autoConnect = true);
```
amaurymartiny: Afaict we're assigning `autoConnect = true` here, which is not what we want

packages/rpc-provider/src/ws/onOpen.spec.ts
```diff
@@ -4,14 +4,16 @@
...
+import WsProvider from '@polkadot/rpc-provider/ws';
 
-function createWs (requests, autoConnect) {
+let ws: WsProvider;
+let mock: Mock;
+
+function createWs (requests: Array<any>, autoConnect: Boolean) {
   mock = mockWs(requests);
-  ws = new Ws(TEST_WS_URL, autoConnect);
+  ws = new WsProvider(TEST_WS_URL, true);
```
amaurymartiny: Can keep autoConnect

packages/rpc-provider/src/ws/polyfill.spec.ts
amaurymartiny: Agreed, they are not testing what they should be. Have a look at the HTTP one above

packages/rpc-provider/src/mock/types.ts
```diff
@@ -2,6 +2,20 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
+import { Server } from 'mock-socket';
+
+export interface Global extends NodeJS.Global {
+  WebSocket: WebSocket;
```
amaurymartiny: here should be `Constructor<WebSocket>`

packages/rpc-provider/src/ws/send.spec.ts
amaurymartiny: global.WebSocket = class extends WebSocket

packages/type-extrinsics/test/e2e/transfer.spec.ts
```diff
@@ -0,0 +1,62 @@
+// Copyright 2017-2019 @polkadot/extrinsics authors & contributors
```
amaurymartiny: ooh I forgot we had e2e tests in type-extrinsics. @jacogr I propose to remove this one, and only put them in api

packages/api/test/e2e/promise-alex-archive.spec.ts
Stefie: I tried to make it a bit more consistent. We're mostly // > 50% using  the `@polkadot/...` imports in the other files, that's why I changed it in the test files, too.

packages/rpc-provider/src/ws/polyfill.spec.ts
Stefie: Yeah, that also doesn't really test what it's supposed to. Both tests are basically testing the same thing

packages/api/test/e2e/promise-alex.spec.ts
Stefie: Otherwise I'm getting the `Promises must be handled appropriately` error on all promises without `.then().catch()`.

Do you know a better solution?

packages/api/test/e2e/promise-alex-archive.spec.ts
Stefie: Ok, I changed the type to `Array<Extrinsic>`.
Not sure what you mean with the second sentence, there's no `result` in the test.

Stefie: Well, actually, the test fails when I change the test to `Array<Extrinsic>` , it passes with `boolean`
Screenshot 2019-05-24 at 12 18 38

packages/api/test/e2e/promise-queries.spec.ts
Stefie: Tslint: `Property 'isZero' does not exist on type 'Codec'.`

packages/api/test/e2e/promise-alex-archive.spec.ts
amaurymartiny: I meant: `(result: Block)`

packages/api/test/e2e/promise-alex-archive.spec.ts
amaurymartiny: I believe the rule of thumb is:
- all imports from the same package should be relative
- all imports from other packages should be absolute

packages/api/test/e2e/promise-alex.spec.ts
amaurymartiny: - .catch should be there
- .then can be removed

packages/api/test/e2e/promise-queries.spec.ts
Stefie: Ah, ok, I can just write it as a callback like I did in quite a few other places:

```
api.query.balances.freeBalance(keyring.alice.address(), balance => {
      expect(
        balance.isZero()
      ).toBe(false);
    });
```

packages/api/test/e2e/rx-queries.spec.ts
Stefie: Doesn't work for rx, though, here I'm still getting the `Property 'isZero' does not exist on type 'Codec'.` tslint error....

packages/rpc-provider/src/http/polyfill.spec.ts
Stefie: Ah, I'm just seeing this now! Some comments were hidden and I was busy addressing the ones on the top.
Thanks for that!

packages/api/test/e2e/rx-queries.spec.ts
amaurymartiny: This is because we right now don't have exact types in the `api.query` query functions, so they only return Codec. We need to manually cast.

`(balance as Balance).isZero()` should work

packages/rpc-provider/src/http/send.spec.ts
Stefie: `Element implicitly has an 'any' type because type 'Object' has no index signature.`
`mock.body['test_body']`

packages/rpc-provider/src/mock/index.ts
Stefie: Ah, that was a leftover from an attempt to fix another linting error. Consider it removed.

packages/rpc-provider/src/http/index.spec.ts
Stefie: Changed.

packages/api/test/e2e/promise-alex-archive.spec.ts
Stefie: Ah, ok, the test still fails, though.

packages/api/test/e2e/rx-queries.spec.ts
Stefie: thanks, solved!

packages/rpc-provider/src/http/send.spec.ts
Stefie: `(mock.body as any)['test_body']` is working.

packages/rpc-provider/src/ws/onOpen.spec.ts
Stefie: ok, done.

packages/api/test/e2e/promise-tx.spec.ts
```diff
@@ -138,38 +144,37 @@ describe.skip('e2e transactions', () => {
...
     const eraBirth = exERA.asMortalEra.birth(currentHeight.toNumber());
     const eraDeath = exERA.asMortalEra.death(currentHeight.toNumber());
-    console.log('STARTED AT :'+eraBirth+' EXPIRED AT :'+eraDeath);
+    console.log('STARTED AT :' + eraBirth + ' EXPIRED AT :' + eraDeath);
     const eraHash = await api.rpc.chain.getBlockHash(eraBirth);
-    const ex = api.tx.balances
-      .transfer(keyring.eve.address(), 12345);
-    const hash = await ex.signAndSend(keyring.dave, {blockHash: eraHash, era:exERA, nonce});
+    const ex = api.tx.balances.transfer(keyring.eve.address(), 12345);
+    const hash = await ex.signAndSend(keyring.dave, { blockHash: eraHash, era: exERA, nonce });
```
Stefie: @amaurymartiny any Idea here? I'm getting:

```
Argument of type '{ blockHash: Codec; era: ExtrinsicEra; nonce: Codec; }' is not assignable to parameter of type 'StatusCb'.
  Object literal may only specify known properties, and 'blockHash' does not exist in type 'StatusCb'.
```

packages/rpc-provider/src/ws/Provider.ts
Stefie: Ok, I did that for all the tests the need to access the `handlers`, `websocket` properties and the `onSocketMessage` method

packages/rpc-provider/src/ws/send.spec.ts
Stefie: Closed because it's a duplicate of https://github.com/polkadot-js/api/pull/936#discussion_r287268555

packages/api/test/e2e/promise-alex-archive.spec.ts
amaurymartiny: TS typings have no impact no tests, so the test is failing for some other reason.

In this file, I think you need to run an Alex archive node locally. We can `.skip` for now, if one day there's a public URL pointing to an archive node we put it back.

packages/api/test/e2e/promise-tx.spec.ts
```diff
@@ -93,27 +93,33 @@ describe.skip('e2e transactions', () => {
...
+  });
 
-    //with callback
+  it('makes a transfer (signAndSend via Signer)  with the wrong keyring pair with a callback', async () => {
+    // with callback
     await expect(api.tx.balances
       .transfer(keyring.eve.address(), 12345)
-      .signAndSend(keyring.alice.address(), ({ events, status }) => {
+      .signAndSend(keyring.alice.address(), (cb: any) => {
+        expect(cb).toBeInstanceOf(Object);
```
amaurymartiny: I think we should not expect anything here: if there's something, the test should fail

packages/api/test/e2e/promise-alex-archive.spec.ts
Stefie: Yes, I'm an idiot.  Of course we don't have access to a random old block on alex... Sorry things are starting to get blurry, I've spent too much time with tslint :)

packages/api/test/e2e/promise-tx.spec.ts
amaurymartiny: Not sure here, you can just `{ blockHash: eraHash, era: exERA, nonce } as any`

packages/api/test/e2e/promise-tx.spec.ts
```diff
@@ -178,21 +183,22 @@ describe.skip('e2e transactions', () => {
...
-        const hash = await ex.signAndSend(keyring.alice, {blockHash: eraHash, era:exERA, nonce});
+
+    const unsubscribe = await api.rpc.chain.subscribeNewHead(async (header) => {
+      if (header.blockNumber.toNumber() === eraDeath - 1) {
+        const hash = await ex.signAndSend(keyring.alice, { blockHash: eraHash, era: exERA, nonce });
 
         expect(hash).toBeUndefined();
         done();
       }
+      unsubscribe();
```
amaurymartiny: Should probably unsubscribe before the done

packages/rpc-provider/src/ws/polyfill.spec.ts
```diff
@@ -2,25 +2,32 @@
...
   });
 
   afterEach(() => {
     global.WebSocket = origWs;
   });
 
-  it('polyfills with no exceptions (without WebSocket)', () => {
-    expect(require('./polyfill')).toBeDefined();
+  it('polyfills with no exceptions (with WebSocket)', () => {
+    (global as any).fetch = undefined;
```
amaurymartiny:
```suggestion
    (global as any).WebSocket = undefined;
```

packages/api/test/e2e/promise-alex-archive.spec.ts
Stefie: I added a comment and skipped the test for now.

packages/api/test/e2e/promise-tx.spec.ts
Stefie: ok, that is working. thanks!

packages/api/test/e2e/promise-alex-archive.spec.ts
```diff
@@ -34,25 +37,23 @@ describe.skip('alex archive queries (local)', () => {
...
-        console.error(error);
-
-        return true;
-      }
-    });
-
-    expect(failed).toHaveLength(0);
-
-    done();
+    api.rpc.chain.getBlock('0x6f6f9bba0eed8e3ae9446c37eee763f93118b52a315a7b46090453ba6288da1f', (result: SignedBlock) => {
```
amaurymartiny: I'm not sure we should change the block hash

packages/api/test/e2e/promise-tx.spec.ts
Stefie: I just removed it completely.

packages/api/test/e2e/promise-tx.spec.ts
Stefie: The test is explicitly `makes a transfer (...) with a callback`.
Tslint complains when the block is empty.

I could do `(cb: any) => { return; }` instead?

amaurymartiny: What I sometimes do is `(cb: any) => { /*do nothing */ }`, and tslint is happy. Your solution works too!

packages/api/test/e2e/promise-alex-archive.spec.ts
Stefie: reverted.

packages/api/test/e2e/promise-tx.spec.ts
Stefie: Ok, I'll stick to `/*do nothing */` for consistency, then :)

packages/api/test/e2e/promise-alex-archive.spec.ts
Stefie: Changed it to match that pattern.

packages/rpc-provider/src/mock/types.ts
Stefie: hmm... this in combination with `packages/rpc-provider/src/ws/send.spec.ts` is the last one. I can't quite wrap my head around it.

Stefie: Tried your solutions with different implementations but I only get more linting errors. If you have a few minutes I'd really appreciate a PR with what you're thinking is working best here!

packages/type-extrinsics/test/e2e/transfer.spec.ts
jacogr: Agreed. Nobody knows about them here, so they never get used.

packages/api/src/checkTypes.manual.ts
```diff
@@ -39,7 +39,7 @@ export default async function test () {
 
   const unsub = await api.tx.balances
     .transfer(keyring.bob.address(), 12345)
-    .signAndSend(keyring.alice, ({ status }) => {
+    .signAndSend(keyring.alice, ({ status }: any) => {
```
jacogr: No any, This is a SumbittableResult

packages/api/test/e2e/promise-alex.spec.ts
```diff
@@ -32,30 +32,31 @@ describe.skip('alex queries', () => {
...
+        done();
+      }).catch();
     });
 
     it('Gets the hash of the last finalized header', async (done) => {
       api.rpc.chain.getFinalizedHead((head) => {
         expect(head instanceof Hash).toBe(true);
         done();
-      });
+      }).catch();
```
jacogr: Not on-board with these `.catch()`-es - rather return the promise from the test, ie. `it('...', async () => { return api...`

packages/api/test/e2e/promise-alex.spec.ts
```diff
@@ -64,29 +65,18 @@ describe.skip('alex queries', () => {
...
 
-
   it('makes a query at a latest block (specified)', async () => {
-    const header = await api.rpc.chain.getHeader();
-    const events = await api.query.system.events.at(header.hash);
-
-    events.forEach(({ event: { data, method, section }, phase, topics }, index) => {
-      console.error(index, phase.toString(), `: ${section}.${method}`, data.toString(), topics.toString());
-    });
+    const header: any = await api.rpc.chain.getHeader();
```
jacogr: `chain.getHeader() as Header`

packages/api/test/e2e/promise-alex.spec.ts
```diff
@@ -64,29 +65,18 @@ describe.skip('alex queries', () => {
...
+    const events: any = await api.query.system.events.at(header.hash);
```
jacogr: cast to type

packages/api/test/e2e/promise-contract.spec.ts
```diff
@@ -6,17 +6,21 @@ import fs from 'fs';
...
-import Api from '../../src/promise';
+import { SubmittableResult } from './../../src';
 
 describe.skip('e2e contracts', () => {
-  let address;
-  let codeHash;
-  let keyring;
-  let api;
+  let address: any;
+  let codeHash: any;
```
jacogr: Hash?

packages/api/test/e2e/promise-contract.spec.ts
```diff
@@ -6,17 +6,21 @@ import fs from 'fs';
...
 import erc20 from '../data/erc20.json';
-import Api from '../../src/promise';
+import { SubmittableResult } from './../../src';
 
 describe.skip('e2e contracts', () => {
-  let address;
-  let codeHash;
-  let keyring;
-  let api;
+  let address: any;
```
jacogr: AccountId?

packages/api/test/e2e/promise-contract.spec.ts
```diff
@@ -41,27 +45,25 @@ describe.skip('e2e contracts', () => {
...
 
             if (record) {
               codeHash = record.event.data[0];
-
-              done();
             }
+            done();
           }
-        });
+        }).catch();
```
jacogr: As above `.catch()`

jacogr:
Ok, yes, I have to echo's Amaury here - this is a MASSIVE, MASSIVE help.

packages/type-extrinsics/test/e2e/transfer.spec.ts
Stefie: Ah, dammit, they're failing anyways...

packages/api/test/e2e/promise-alex.spec.ts
Stefie: Yeah, I had that, but I stumbled upon quite a few problems with `Codec` which was the reason for ending up choosing `any` just because of the sheer number of tslint errors I was getting.

Stefie:
@jacogr  Ok, I'm not gonna address your comments today anymore, that wouldn't make sense. My brain is just producing garbage. Monday morning it is, then. 
I'm aware of the `any`'s . The goal of this PR is mainly to convert all existing files & tests to valid Typescript without loosing any.

packages/api/test/e2e/promise-alex.spec.ts
jacogr: The tricky bits are the derives - the actual derive values we have (see eg. DeriveBalances) are not Codec. So for those, they need to be casted (with a relevant FIXME). We even atm, do this internal to derive, drives me up the wall. If I had $0.02 for each time I've done that, well...

The `api.rpc` and `api.query` are actually fine - they return valid `Codec` types. So a simple `as <Type>` works in those cases.

jacogr:
I'm not asking for the types to make your life a living missery, there is method behind the madness. For instance, I actually added this file inside api - https://github.com/polkadot-js/api/blob/master/packages/api/src/checkTypes.manual.ts since despite the fact that we test everywhere, we still had issues that came back reported from external parties since we messed up the typings.

So that one is useless, doesn't run anywhere - but it complains on types so we can get it right and have the libs not only functionally working, but also working for TS users. Which is just absolutely crazy...

So take your time, happy to wait a bit, but this not not just admin work, it really adds a MASSIVE amount of value. And we can after this, even get rid of the type of, to put it lightly, crap I did in that file above - since now, we can actually check properly for linting/use in real-used files as well.

packages/api/test/e2e/promise-alex.spec.ts
Stefie: I ran into quite a few of these cases, so as a general question addressing all of them:

Would you declare `const header = ...` like this:
`const header: Header = await api.rpc.chain.getHeader() as Header;`
or like this:
`const header = await api.rpc.chain.getHeader() as Header;`
because it's already explicitly declared ` as Header` anyways?

packages/type-extrinsics/test/e2e/transfer.spec.ts
Stefie: The tests are broken in the current master, too, so they didn't break as part of this PR and are unrelated.
Therefore I added two `@TODO`s to code:

```
// @TODO 1: All three tests below are broken, fix or find better way to test transfers.
// @TODO 2: Move tests out of ./type-extrinsics to ./api folder
```

packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+    expect(api.runtimeVersion).toBeDefined();
+    expect(api.rpc).toBeDefined();
+    expect(api.query).toBeDefined();
+    expect(api.tx).toBeDefined();
+    expect(api.derive).toBeDefined();
+  });
+
+  it('queries state for a balance', async () => {
+    return(
+      api.query.balances.freeBalance(keyring.alice.address(), balance => {
```
jacogr: Generally you could just do `return api.query.balances.freeBalance(keyring.alice.address(), balance => {` (all tests, but it is neither here, nor there. (So nothing to do)

Stefie:
I just noticed that my Comment regarding the Extrinsics tests is hidden in the collapsed part:

https://github.com/polkadot-js/api/pull/936#discussion_r287815922

packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -0,0 +1,196 @@
...
+        expect(balance).toBeInstanceOf(BN);
+        expect(balance.isZero()).toBe(false);
+      })
+    );
+  });
+
+  it('subscribes to rpc', (done) => {
+    return(
```
jacogr: This file seems like it has a lot of `return(` as opposed to `return (` - just a simple global search/replace could fix.

packages/api/test/e2e/promise-queries.spec.ts
jacogr: (nit) prefer `(balance) =>` as opposed to the bare version

packages/api/test/e2e/promise-tx.spec.ts
```diff
@@ -4,35 +4,37 @@
...
-        console.log('Completed at block hash', status.value.toHex());
-        console.log('Events:');
+    if (status.isFinalized) {
+      console.log('Completed at block hash', status.value.toHex());
+      console.log('Events:');
 
-        events.forEach(({ phase, event: { data, method, section } }) => {
-          console.log('\t', phase.toString(), `: ${section}.${method}`, data.toString());
-        });
+      events.forEach(({ phase, event: { data, method, section } }: any) => {
```
jacogr: `: EventRecord`

packages/api/test/e2e/rx-queries.spec.ts
```diff
@@ -41,11 +42,11 @@ describe.skip('e2e queries', () => {
     api.rpc.chain
       .getHeader()
       .pipe(
-        switchMap((header) =>
-          api.query.system.events.at(header.hash)
+        switchMap(({ hash }: any) =>
```
jacogr: `: Header`

packages/api/test/e2e/rx-queries.spec.ts
```diff
@@ -41,11 +42,11 @@ describe.skip('e2e queries', () => {
...
+          api.query.system.events.at(hash)
         )
       )
-      .subscribe((events) => {
+      .subscribe((events: any) => {
```
jacogr: `Vector<EventRecord>`

packages/api/test/e2e/rx-tx.spec.ts
```diff
@@ -26,14 +28,14 @@ describe.skip('e2e transactions', () => {
       .accountNonce(keyring.alice.address())
       .pipe(
         first(),
-        switchMap((nonce) =>
+        switchMap((nonce: any) =>
```
jacogr: `: Index`

packages/api/test/e2e/rx-tx.spec.ts
```diff
@@ -45,14 +47,14 @@ describe.skip('e2e transactions', () => {
...
```
jacogr: `: Index`

packages/api/test/e2e/promise-queries.spec.ts
Stefie: Whoops, merged before reading this. Github did show me an 'approved' from you so there must have been an overlap.
I'll address your comments in another PR

packages/api/test/e2e/promise-queries.spec.ts
Stefie: true, I relied too  much on the linting here

packages/api/test/e2e/rx-queries.spec.ts
Stefie: With:
```
switchMap(({ hash }: Header) =>
  api.query.system.events.at(hash)
)
```
I'm getting 
`Argument of type 'OperatorFunction<Header, Codec>' is not assignable to parameter of type 'OperatorFunction<Codec, Codec>'.`
` Type 'Codec' is missing the following properties from type 'Header': blockNumber, digest, extrinsicsRoot, hash, and 20 more.`

What am I missing here?

packages/api/test/e2e/rx-tx.spec.ts
Stefie: Same `Codec`  error ashere https://github.com/polkadot-js/api/pull/936#discussion_r287831252 (probably with the same solution)

packages/api/test/e2e/promise-queries.spec.ts
Stefie: https://github.com/polkadot-js/api/pull/955

packages/api/test/e2e/promise-queries.spec.ts
Stefie: addressed in https://github.com/polkadot-js/api/pull/955

packages/api/test/e2e/promise-tx.spec.ts
Stefie: https://github.com/polkadot-js/api/pull/955

Stefie:
Moved further discussion to https://github.com/polkadot-js/api/pull/955

Issue #942, opened 2019-05-22
Update dependencies in documentation examples


ltfschoen:
When I go to https://polkadot.js.org/api/examples/promise/ and try and run one of the examples,
the dependencies in the package.json files are outdated (i.e. the @polkadot/api version) so for example if I run https://polkadot.js.org/api/examples/promise/10_upgrade_chain/ then I'll get an error like:
```
2019-05-22 10:37:13        RPC-CORE: getMetadata (block: Hash): Metadata:: Object.values(...)[index] is not a constructor
2019-05-22 10:37:13   API/DECORATOR: FATAL: Unable to initialize the API:  getMetadata (block: Hash): Metadata:: Object.values(...)[index] is not a constructor
```

Suggest we at least update the documentation so the user knows how to do it themselves.

And possibly also add Dependabot like we did for https://github.com/paritytech/js-libs

jacogr:
this only gets updated as part of a release. (Which is today anyway - for the most part we don't track these on the minor bumps)

Greenkeeper doesn't track examples/* neither does depndabot.

ltfschoen:
Noted that this commit makes our lives easier https://github.com/polkadot-js/api/pull/944

PR #943, opened 2019-05-22
docs: Relates to #942. Update instructions so user knows how to update outdated dependencies

ltfschoen:
Possibly also add Dependabot to polkadot-js/api in addition to this update

Note that I've used `npm install -g yarn-upgrade-all;` since installing it with `yarn global add yarn-upgrade-all` on Debian 9 doesn't appear to work.

### DIFF
--- a/docs/examples/promise/README.md
+++ b/docs/examples/promise/README.md
@@ -12,7 +12,14 @@ substrate --dev
 
 ## Running the examples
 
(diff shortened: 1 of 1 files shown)


jacogr:
Oki, so the dependabot thing - 

- we have greenkeeper
- the issue is that these packages are not in the root nor packages/*
- so even the "auto-update" scripts does not cater for these

Closing - manual updates as per major releases.

PR #948, opened 2019-05-23
Use name instead of prefix

### LINKED ISSUE #947: Metadata uses prefix, not name, for calls
https://github.com/polkadot-js/api/blob/master/packages/type-extrinsics/src/fromMetadata/index.ts#L24

Basically this means if you have a module with no storage, i.e. metadata returns an empty prefix, you cannot call extrinsics.

cc @amaurymartiny

amaurymartiny:
fixes #947

### DIFF
--- a/packages/type-extrinsics/src/fromMetadata/index.ts
+++ b/packages/type-extrinsics/src/fromMetadata/index.ts
@@ -21,12 +21,12 @@ export default function fromMetadata (metadata: Metadata): ModulesWithMethods {
   return metadata.asV4.modules
     .filter((modul) => modul.calls.isSome)
     .reduce((result, modul: ModuleMetadata, sectionIndex) => {
-      const prefix = stringCamelCase(modul.prefix.toString());
+      const section = stringCamelCase(modul.name.toString());
 
-      result[prefix] = modul.calls.unwrap().reduce((newModule, callMetadata, methodIndex) => {
+      result[section] = modul.calls.unwrap().reduce((newModule, callMetadata, methodIndex) => {
         const funcName = stringCamelCase(callMetadata.name.toString());
 
-        newModule[funcName] = createUnchecked(prefix, sectionIndex, methodIndex, callMetadata);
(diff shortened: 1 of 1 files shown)


packages/type-extrinsics/src/fromMetadata/index.ts
```diff
@@ -21,12 +21,12 @@ export default function fromMetadata (metadata: Metadata): ModulesWithMethods {
   return metadata.asV4.modules
     .filter((modul) => modul.calls.isSome)
```
Stefie: Unfortunately delayed nitpick: We have a quite german name for `module` here . Don't know if it's worth another PR, but I'd rename `modul` to the english `module`

amaurymartiny: module is a reserved name in js, I renamed to modul here (and in multiple other places) on purpose

Stefie: Ah, ok, makes sense. it looked liked an oversight to me.

jacogr: Couple of these where we deviate from Rust -

- `arguments` -> `args`
- `call` -> `method`
- `module` -> `modul` (only for loops on decorating, it not exposed anywhere by this name)

amaurymartiny: Come to think about it: `type` is a reserved TS word. Though it seems to work okay `const { type } = a;` everywhere, so we don't need to change it.

jacogr: I would hate `something.typ` as a getter :)

amaurymartiny: I think rust uses `.ty`, but not better

Stefie: Rust also has `typ` instead of `type`? There really were some germans involved in developing that language 😄 

These two make sense
```
arguments -> args
call -> method
```

PR #952, opened 2019-05-24
Fix some more tests in TS

amaurymartiny:


Stefie:
Sooo good to see the tests finally running in the CI!

PR #956, opened 2019-05-27
rpc-core e2e tests in typescript

Stefie:
Follow-up of #936 (convert missed folder)


packages/rpc-core/test/e2e/krummelanke.test.ts
```diff
@@ -2,43 +2,48 @@
...
+          [storage.balances.freeBalance, ALICE]
         ],
-        (data) => {
+        (data: StorageChangeSet) => {
           expect(data).toHaveLength(2);
-          expect(data[0].toNumber()).toEqual(0);
-          expect(data[1].toNumber()).not.toEqual(0);
+          expect([data as any][0].toNumber()).toEqual(0);
+          expect([data as any][1]).toBeInstanceOf(Balance);
+          expect([data as any][1].toNumber()).not.toEqual(0);
```
Stefie: Any better ideas here?

jacogr: I would drop this one completely - we don't have Krumme Lanke anymore. (Or at least rename to be current and point to an Alexander RPC)

As for the `StorageChangeSet`, this shows how old there are. So https://github.com/polkadot-js/api/blob/master/packages/types/src/rpc/StorageChangeSet.ts has the `.changes` part which this needs to query. 

So basically, something like (typing out of the top of my head), since changes in a option -

```js
// could also unwrap and test the result
expect(data.changes[0].value.isSome).toBe(true);
```

Stefie: Hmm, but the tests are actually passing and subscribing to these two
```
[storage.system.accountNonce, randomAccount],
[storage.balances.freeBalance, ALICE]
```
I'm getting back a subscription that is returning an array of type `[Nonce, Balance]` with the correct results?!

jacogr: Ahh, yes, you are correct, I am wrong - rpc does return an Array. (I should know _very_ well - not sure how I "forgot" that rpc-core mangles the storage return)

Since it returns an `Array<Codec>` it is not each, not great to cast it to a Tuple.

PR #957, opened 2019-05-28
update signature of doublemap storage function, add e2e test cases

AlexWZ0129:
#776 Hi, I updated the signature of double map storage function to use two arguments to represent keys. Also, I added some e2e test cases to ensure my change won't break anything we currently have.

Here are the covered situations of those test cases:
1. Promise query

  | Double Map | Map | Plain
-- | -- | -- | --
query | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
subscribe query | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
At | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
hash | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
key | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
multi | passed, tested on our chain | passed, tested on substrate | plain type doesn’t need multi
subscribe multi | passed, tested on our chain | passed, tested on substrate | plain type doesn’t need multi
size | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate

2. Rx query

  | Double Map | Map | Plain
-- | -- | -- | --
query | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
At | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
hash | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
key | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate
multi | passed, tested on our chain | passed, tested on substrate | plain type doesn’t need multi
size | passed, tested on our chain | passed, tested on substrate | passed, tested on substrate

Please let me know if there is something I need to change or we don't want those test cases.


packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -177,20 +177,197 @@ describe.skip('Promise e2e queries', () => {
...
-      api.query.system.eventTopics(['any', '0x1234'], (eventTopics) => {
-        expect(eventTopics.toString()).toEqual('[]');
+  describe('with map type', async () => {
+    it('queries correct value', async () => {
+      const balance = await api.query.balances.freeBalance(keyring.alice.address()) as Balance;
+
+      expect(balance.isZero()).toBe(false);
+    });
+
+    it('queries correct value at a specified block', async () => {
```
AlexWZ0129: These test cases assume the account Alice is only used in tests(the balance of Alice won't change between `api.query.balances.freeBalance(keyring.alice.address());` and `api.query.balances.freeBalance.at(header.hash, keyring.alice.address())` ). We can change this to just check the queried balance is greater than 0.

packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -177,20 +177,197 @@ describe.skip('Promise e2e queries', () => {
...
+  describe('with double map type', async () => {
+    const KEY1 = 'any';
+    const KEY2 = '0x1234';
+    it('queries correct value', async () => {
+      const eventTopics = await api.query.system.eventTopics(KEY1, KEY2);
+
+      expect(eventTopics.toJSON()).toEqual([]);
+    });
+
+    it('queries correct value at a specified block', async () => {
```
AlexWZ0129: Here is a special case: `await api.query.system.eventTopics.at(header.hash, KEY1, KEY2);` will throw an `Encoding for input doesn't match output, created 0x00 from 0x` error. This error is from [here](https://github.com/polkadot-js/api/blob/761682373def42dd64587832c1ebd554340ed1a8/packages/types/src/codec/createType.ts#L244). This `at` function only throw this error when the returned value from the chain is `null`. If the returned value is not null, this function will work. I added a unit test below for this issue

packages/types/src/codec/createType.spec.ts
```diff
@@ -232,4 +232,18 @@ describe('createType', () => {
       createType('{"_enum": {"A": null, "B": "u32", "C": null} }', 1).toJSON()
     ).toEqual({ B: 0 });
   });
+
```
AlexWZ0129: This is the unit test I added for the `Encoding for input doesn't match output, created 0x00 from 0x` issue above. Looks like this error appear even if the first argument of `createType` is not a double map type (`Vec<(BlockNumber,EventIndex)>`).

packages/api/test/e2e/rx-queries.spec.ts
```diff
@@ -51,4 +51,181 @@ describe.skip('Rx e2e queries', () => {
         done();
       });
   });
+
+  describe('with plain type', async () => {
```
jacogr: These are great, however could have stopped at promise - since underlying promise uses the rxjs interfaces, so if that works, we are (mostly) golden. There are obviously cases where the one may break and other other not, but at long as there is basic other coverage, it will be caught.

TL;DR - These are great, but probably technically not needed

packages/api/test/e2e/rx-queries.spec.ts
```diff
@@ -51,4 +51,181 @@ describe.skip('Rx e2e queries', () => {
...
+    const TOTAL_ISSUANCE = 12582912;
+    it('queries correct value', async (done) => {
```
jacogr: We are using `async` for all the rxjs tests - should not be needed since we are not using `await`?

packages/types/src/codec/createType.spec.ts
jacogr: Yes, `null` would be a funny one here. Not sure it would happen from storage, if that path does get taken we have some other issues there. (Which is the only place where `isPedantic` is currently used - if the decoding goes wrong, it goes horribly wrong there...)

packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -177,20 +177,197 @@ describe.skip('Promise e2e queries', () => {
     );
   });
 
-  it('queries state using double map key', async () => {
-    // TODO Update ['any', '0x1234'] to the key of a known event topic and update '[]' to the expected value
-    const eventTopics = await api.query.system.eventTopics(['any', '0x1234']);
+  describe('with plain type', async () => {
+    const TOTAL_ISSUANCE = 12582912;
```
jacogr: Bit sceptical about using this for dev chains, but it certainly matches with a fresh DB. Maybe we should rather use something simpler for dev, i.e. `balances.existentialDeposit`, i.e. `500`? (This should be a relative easy replace for all tests)

packages/api/src/Base.ts
```diff
@@ -589,37 +592,50 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
...
+            ? [creator, args]
+            : [creator, ...args]),
+        [arg1, arg2]
       ) as unknown as HashResult<C, S>;
 
-    decorated.key = (arg?: CodecArg): string =>
-      u8aToHex(compactStripLength(creator(arg))[1]);
+    decorated.key = (arg1?: CodecArg, arg2?: CodecArg): string =>
+      creator.meta.type.isDoubleMap
+      ? u8aToHex(creator([arg1, arg2]))
```
jacogr: This really is weird the way it is done here - one type has the compact removed, the other not. The reason the underlying storage has the length is that so it can go into a StorageKey, i.e. https://github.com/polkadot-js/api/blob/master/packages/types/src/primitive/StorageKey.ts and be treated correctly.

packages/api/src/Base.ts
```diff
@@ -589,37 +592,50 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
...
 
-    decorated.size = (arg?: CodecArg): U64Result<C, S> =>
+    decorated.size = (arg1?: CodecArg, arg2?: CodecArg): U64Result<C, S> =>
       onCall(
-        (arg: CodecArg) => this._rpcRx.state.getStorageSize([creator, arg]),
-        [arg]
+        (...args: CodecArg[]) => this._rpcRx.state.getStorageSize(
+          creator.meta.type.isDoubleMap
+            ? [creator, args]
+            : [creator, ...args]),
```
jacogr: My only concern here is that it effectively ends up being `[creator, <value of arg1>, undefined]` when called with 1 arg, instead of `[creator, <value of arg1>]`. Atm, no difference since the normal keys only accept one arg and would be split there, but just worried it could have some consequences down the line?

packages/api/test/e2e/rx-queries.spec.ts
AlexWZ0129: yeah, right, let me remove them

packages/api/src/Base.ts
AlexWZ0129: Okay. let me check it

AlexWZ0129: I updated the implementation of doublemap, all types use compact now.

packages/api/test/e2e/rx-queries.spec.ts
AlexWZ0129: My mistaken.

packages/api/src/Base.ts
AlexWZ0129: what about this?
``` 
decorated.size = (arg1?: CodecArg, arg2?: CodecArg): U64Result<C, S> =>
      onCall(
        (arg1?: CodecArg, arg2?: CodecArg) => this._rpcRx.state.getStorageSize(
          creator.meta.type.isDoubleMap
            ? [creator, [arg1, arg2]]
            : [creator, arg1]),
        [arg1, arg2]
      ) as unknown as U64Result<C, S>;
```
Now when called with 1 arg, it will be `[creator, <value of arg1>]`

jacogr: It think that is fine since the second arg is anyway (currently) useless in any other non-DoubleMap operations.

AlexWZ0129: Cool. I updated `.at()`, `.hash()` and `.size()`. Also I rerun the test, and all tests passed.

packages/api/test/e2e/promise-queries.spec.ts
AlexWZ0129: Sorry, forget to fix this

AlexWZ0129: just updated test cases to use `balances.existentialDeposit` for plain type

packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -178,43 +178,43 @@ describe.skip('Promise e2e queries', () => {
   });
 
   describe('with plain type', () => {
-    const TOTAL_ISSUANCE = 12582912;
+    const EXISTENTIAL_DEPOSIT = 500;
     it('queries correct value', async () => {
-      const totalIssuance = await api.query.balances.totalIssuance() as Balance;
+      const totalIssuance = await api.query.balances.existentialDeposit() as Balance;
```
jacogr: `const existentialDeposit = await api.query.balances.existentialDeposit()` :)

AlexWZ0129: ah.. my bad...

PR #958, opened 2019-05-28
Update for latest substrate master metadata

jacogr:


jacogr:
NOT happy with the `exec::StorageKey` alias here. But no other option.

packages/types/src/Metadata/v4/latest.substrate.v4.json
```diff
@@ -3147,7 +3147,7 @@
               "type": {
                 "PlainType": "Schedule"
               },
-              "fallback": "0x0000000001000000000000000100000000000000010000000000000001000000000000000100000000000000010000000000000001000000000000000100000000000000000001001000000000",
+              "fallback": "0x000000000100000000000000010000000000000001000000000000000100000000000000010000000000000001000000000000000100000000000000010000000000000001000000000000000400000000000100100000000020000000",
```
amaurymartiny: The fallback length has changed: Schedule.ts needs to be updated too

jacogr: We should use isPedantic there in the creation

PR #959, opened 2019-05-28
Redeemable and Unlocking in staking.info

Tbaut:
- adds redeemable `BN`
- adds unlocking `{value:BN, remainingBlocks: BN}[]`

Meant to be used (and displayed) for stash accounts.

- Needs e2e testing


packages/api-derive/src/staking/info.ts
```diff
@@ -2,30 +2,39 @@
...
+        [api.query.staking.stakers, stashId],
+        [api.query.staking.validators, stashId]
+      ])
+    ]) as any as Observable<[BN, BlockNumber, [Option<AccountId>, Option<StakingLedger>, [Array<AccountId>], Exposure, [ValidatorPrefs]]]>
+  )
+  .pipe(
+    map(([eraLength, bestNumber, [nextKeyFor, stakingLedger, [nominators], stakers, [validatorPrefs]]]) => ({
       accountId,
       controllerId,
+      unlocking: calculateUnlocking(stakingLedger.unwrap().unlocking, eraLength, bestNumber),
```
jacogr: For safety, would do `unwrapOr(null)` and pass the `StakingLedger | null` into the calc

packages/api-derive/src/staking/info.ts
```diff
@@ -34,13 +43,66 @@ function withStashController (api: ApiInterface$Rx, accountId: AccountId, contro
       stakingLedger: stakingLedger.isSome
         ? stakingLedger.unwrap()
         : undefined,
+      redeemable: redeemableSum(stakingLedger.unwrap().unlocking, eraLength, bestNumber),
```
jacogr: We are unwrapping it twice, or rather 2 times - so I propose moving stakingLedger out, i.e.

```js
const stakingLedger = stakingLedger.isSome ? stakingLedger.unwrap() | undefined
```

And passing that value around (not the null suggested earlier). Have a check, in the case of `isUndefined(stakingLedger)`, return `undefined`

packages/api-derive/src/staking/info.ts
```diff
@@ -34,13 +43,66 @@ function withStashController (api: ApiInterface$Rx, accountId: AccountId, contro
...
+      map[key] = map[key].add(value);
+    }
+
+    return map;
+  }, {} as { [index: string]: BN });
+}
+
+function redeemableSum (unlockings: Vector<UnlockChunk>, eraLength = new BN(0), bestNumber= new BlockNumber(0)) {
+  return unlockings
+  .filter((chunk) => remainingBlocks(chunk.era, eraLength, bestNumber).eqn(0))
```
jacogr: Weird indentation here, 2 spaces in

packages/api-derive/src/staking/info.ts
```diff
@@ -34,13 +43,66 @@ function withStashController (api: ApiInterface$Rx, accountId: AccountId, contro
...
+  .reduce((curr, prev) => {
+    return curr.add(prev.value);
+  }, new BN(0));
+}
+
+function remainingBlocks (era: BN, eraLength: BN, bestNumber: BlockNumber) {
+
+  if (!bestNumber || !eraLength || era.lten(0)) {
```
jacogr: (nit) Remove extra blank line

packages/api-derive/src/staking/info.ts
```diff
@@ -34,13 +43,66 @@ function withStashController (api: ApiInterface$Rx, accountId: AccountId, contro
...
         : undefined,
+      redeemable: redeemableSum(stakingLedger.unwrap().unlocking, eraLength, bestNumber),
       stashId,
       validatorPrefs
     })),
     drr()
   );
 }
 
+function calculateUnlocking (unlockings: Vector<UnlockChunk>, eraLength = new BN(0), bestNumber= new BlockNumber(0)) {
```
jacogr: Would suggest moving the util functions (all these following), to the top of the file

packages/api-derive/src/types.ts
```diff
@@ -60,8 +60,10 @@ export type DerivedSessionInfo = {
 export type DerivedStaking = {
   accountId: AccountId,
   controllerId?: AccountId,
+  unlocking?: {value: BN, remainingBlocks: BN}[],
```
jacogr: Alphabetical ordering. Also prefer `Array<{ remainingBlocks: BN, value: BN }>` (No shortcut & alphabetical)

packages/api-derive/src/staking/info.ts
```diff
@@ -34,13 +43,66 @@ function withStashController (api: ApiInterface$Rx, accountId: AccountId, contro
...
+  if (unlockingChunks.length) {
+    // group the Unlockchunks that have the same era and sum their values
+    const groupedResult = groupByEra(unlockingChunks);
+    const results: {value: BN, remainingBlocks: BN}[] = [];
+
+    Object.keys(groupedResult).map((eraString) => (
+      results.push({ value: groupedResult[eraString], remainingBlocks: remainingBlocks(new BlockNumber(eraString), eraLength, bestNumber) })
+    ));
+
+    return results.length ? results : undefined;
```
jacogr: This is great, one of the issues currently is that it is either non-existent, or empty - so returning nothing for both these helps a lot.

packages/api-derive/src/staking/info.ts
```diff
@@ -34,13 +43,66 @@ function withStashController (api: ApiInterface$Rx, accountId: AccountId, contro
...
 }
 
+function calculateUnlocking (unlockings: Vector<UnlockChunk>, eraLength = new BN(0), bestNumber= new BlockNumber(0)) {
+  // select the Unlockchunks that can't be redeemed yet.
+  const unlockingChunks = unlockings.filter((chunk) => remainingBlocks(chunk.era, eraLength, bestNumber).gtn(0));
+
+  if (unlockingChunks.length) {
+    // group the Unlockchunks that have the same era and sum their values
+    const groupedResult = groupByEra(unlockingChunks);
+    const results: {value: BN, remainingBlocks: BN}[] = [];
```
jacogr: This form `{value: BN, remainingBlocks: BN}` gets used a lot, define it and export from `./types.ts`, import in this file

jacogr: ... Additionally, follow the alphabetical suggestions for this type & `Array<MyCoolType>` as opposed to the `[]` shorthand. (Just consistency)

packages/api-derive/src/staking/info.ts
jacogr: Both these should not be undefined, so no need for the defaults

packages/api-derive/src/staking/info.ts
```diff
@@ -2,42 +2,110 @@
...
+      return new StructAny({
         accountId,
         controllerId,
         nextSessionId: nextKeyFor.isSome
           ? nextKeyFor.unwrap()
           : undefined,
         nominators,
+        redeemable: redeemableSum(_stakingLedger, eraLength, bestNumber),
         stakers,
         stakingLedger: stakingLedger.isSome
```
jacogr: Can use `_stakingLedger` from above, i.e. `stakingLedger: _stakingLedger`

packages/api-derive/src/staking/info.ts
```diff
@@ -2,42 +2,110 @@
...
 import { map, switchMap } from 'rxjs/operators';
-import { ApiInterface$Rx } from '@polkadot/api/types';
-import { AccountId, Exposure, Option, StakingLedger, StructAny, ValidatorPrefs } from '@polkadot/types';
 
+import { bestNumber } from '../chain/bestNumber';
 import { drr } from '../util/drr';
+import { eraLength } from '../session/eraLength';
+import { isUndefined } from '@polkadot/util';
+
+function calculateUnlocking (stakingLedger: StakingLedger | undefined, eraLength: BN, bestNumber: BlockNumber) {
```
jacogr: Prefer the return types specifies on functions, especially here 

```js
function calculateUnlocking (..): Array<{ remainingBlock: BlockNumber, value: BN > | undefined {
```

packages/api-derive/src/staking/info.ts
```diff
@@ -2,42 +2,110 @@
...
+  if (isUndefined(stakingLedger)) return undefined;
```
jacogr:
```js
if (isUndefined(stakingLedger)) {
  return undefined;
}
```

packages/api-derive/src/staking/info.ts
```diff
@@ -2,42 +2,110 @@
...
+  // select the Unlockchunks that can't be redeemed yet.
+  const unlockingChunks = stakingLedger.unlocking.filter((chunk) => remainingBlocks(chunk.era, eraLength, bestNumber).gtn(0));
+
+  if (unlockingChunks.length) {
+    // group the Unlockchunks that have the same era and sum their values
+    const groupedResult = groupByEra(unlockingChunks);
+    const results: DerivedUnlockingMap = [];
+
+    Object.keys(groupedResult).map((eraString) => (
+      results.push({ value: groupedResult[eraString], remainingBlocks: remainingBlocks(new BlockNumber(eraString), eraLength, bestNumber) })
```
jacogr: One of the reasons for the return type on the function, we are not actually using `eraLength` or `bestNumber`

Should just be, with readability enhancements -

```js
results.push({
  remainingBlocks: remainingBlocks(new BlockNumber(eraString)),
  value: groupedResult[eraString]
})
```

packages/api-derive/src/staking/info.ts
```diff
@@ -2,42 +2,110 @@
...
+    } else {
+      map[key] = map[key].add(value);
+    }
+
+    return map;
+  }, {} as { [index: string]: BN });
+}
+
+function redeemableSum (stakingLedger: StakingLedger | undefined, eraLength: BN, bestNumber: BlockNumber) {
+  if (isUndefined(stakingLedger)) return new BN(0);
```
jacogr: Prefer

```js
if (isUndefined(stakingLedger)) {
  return new BN(0);
}
```

packages/api-derive/src/staking/info.ts
```diff
@@ -2,42 +2,110 @@
...
+    .filter((chunk) => remainingBlocks(chunk.era, eraLength, bestNumber).eqn(0))
+    .reduce((curr, prev) => {
+      return curr.add(prev.value);
+    }, new BN(0));
+}
+
+function remainingBlocks (era: BN, eraLength: BN, bestNumber: BlockNumber) {
+  if (!bestNumber || !eraLength || era.lten(0)) {
+    return new BN(0);
+  } else {
```
jacogr: We don't need the else, so rather just -

```js
if (!bestNumber || !eraLength || era.lten(0)) {
  return new BN(0);
}

const remaining = eraLength.mul(era).sub(bestNumber);

return remaining.lten(0) ? new BN(0) : remaining;
```

packages/api-derive/src/staking/info.ts
```diff
@@ -2,42 +2,110 @@
...
+
+  return stakingLedger.unlocking
+    .filter((chunk) => remainingBlocks(chunk.era, eraLength, bestNumber).eqn(0))
+    .reduce((curr, prev) => {
+      return curr.add(prev.value);
+    }, new BN(0));
+}
+
+function remainingBlocks (era: BN, eraLength: BN, bestNumber: BlockNumber) {
+  if (!bestNumber || !eraLength || era.lten(0)) {
```
jacogr: bestNumber & eraLength should be here - since they come from other derives?

jacogr:
Also, just run it though the existing e2e tests - uncomment `.skip`, add `it.only` - you can inspect the logged console value. (Would obviously suggest unlocking some beforehand... :))

Tbaut:
Thanks, I've been testing it successfully with the cumbersome copy/pasting in `node_modules`. I might touch a bit the tests to take the new fields into account.

packages/api-derive/src/staking/info.ts
```diff
@@ -2,42 +2,111 @@
...
+
+  if (!unlockingChunks.length) {
+    return undefined;
+  }
+
+  // group the Unlockchunks that have the same era and sum their values
+  const groupedResult = groupByEra(unlockingChunks);
+  const results: DerivedUnlocking = [];
+
+  Object.keys(groupedResult).map((eraString) => (
```
jacogr: As discussed, this would probably be a better form -

```js
const results = Object.entries(groupedResult).map(([eraString, value]) => ({
    value,
    remainingBlocks: remainingBlocks(new BlockNumber(eraString), eraLength, bestNumber)
}));
```

Tbaut: Thanks. I haven't pushed the modifications yet as I make my way on an e2e test.

Tbaut:
The test isn't ideal:
- it generically checks for unlocking not undefined (although it could be not undefined for other reasons).
  - Why not?: I should check for the current era, do the calculation of the `amount_of_blocks_left`, and verify that the array unlocking has an object with `unlocking.remainingBlocks`== `amount_of_blocks_left` with a `value`>=`UNBONDED_VALUE`. I found this simply a little too complex..
- it doesn't check for redeemable.
  - Why not?: I should make sure the chain has an `staking.unbondingDuration` of x, a `session.eraLength` of y, unbond, and wait for x*y to see if `unlocking.redeemable` >= `UNBONDED_VALUE`.  I found this simply a little too complex.. and would take ages.

packages/api-derive/test/e2e/promise.spec.ts
```diff
@@ -100,4 +103,35 @@ describe.skip('derive e2e', () => {
       done();
     });
   });
+
+  describe.only('verifies derive.staking.unlocking',() => {
```
jacogr: Remove the `.only`

packages/api-derive/test/e2e/promise.spec.ts
```diff
@@ -100,4 +103,35 @@ describe.skip('derive e2e', () => {
...
+    it('bondsExtra funds for Alice Stash', () => {
+      return api.tx.staking.bondExtra(BOND_VALUE)
+        .signAndSend(aliceStashPair, ({ events = [], status }) => {
+          expect(status.type).toMatch(/Ready/i);
+        });
+    });
+
+    it('unbonds dots for Alice (from Alice Stash)', () => {
+      return api.tx.staking.unbond(UNBOND_VALUE)
+      .signAndSend(alicePair, ({ events = [], status }) => {
```
jacogr: (nit) spacing mishap, indent by extra 2 spaces here

packages/api-derive/test/e2e/promise.spec.ts
```diff
@@ -100,4 +103,35 @@ describe.skip('derive e2e', () => {
...
+    const ALICE_STASH = '5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY';
+    const ALICE = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
+    const keyring = testKeyring();
+    const aliceStashPair = keyring.getPair(ALICE_STASH);
+    const alicePair = keyring.getPair(ALICE);
+
+    it('bondsExtra funds for Alice Stash', () => {
+      return api.tx.staking.bondExtra(BOND_VALUE)
+        .signAndSend(aliceStashPair, ({ events = [], status }) => {
+          expect(status.type).toMatch(/Ready/i);
```
jacogr: You actually want to wait for finalised before it is "ok"

packages/api-derive/test/e2e/promise.spec.ts
```diff
@@ -100,4 +103,35 @@ describe.skip('derive e2e', () => {
...
+        });
+    });
+
+    it('unbonds dots for Alice (from Alice Stash)', () => {
+      return api.tx.staking.unbond(UNBOND_VALUE)
+      .signAndSend(alicePair, ({ events = [], status }) => {
+        expect(status.type).toMatch(/Ready/i);
```
jacogr: Flinalized

packages/api-derive/test/e2e/promise.spec.ts
```diff
@@ -100,4 +103,35 @@ describe.skip('derive e2e', () => {
...
+      });
+    });
+
+    it('verifies that derive.staking.unlocking isn\'t empty/undefined', () => {
+      return api.derive.session.info(ALICE_STASH, (info: DerivedStaking) => {
+        expect(info.unlocking).toBeDefined();
```
jacogr: Some tweaks, adjust as needed -

```js
expect(info.unlocking).toHaveLength(1);
expect(info.unlocking[0].value.eqn(UNBOND_VALUE)).toBe(true);
```

Tbaut: The thing here is that `unlocking` could have a length bigger than `1` ( e.g if you've unbonded before, like you've run the test before, or if you have redeemable funds).
You can't be sure about any `unlocking[x].value.eqn` because the value is a sum (in case you've unbonded several times in the same session, you might have a value>`UNBOND_VALUE`). Also you can't be sure about what x is because things aren't ordered anymore. Just did a test with a change of bonding duration and although the extrinsic keeps the order of when the events came in, things get shuffled because of the derive calculation.

I'll use `expect(info.unlocking).toBeGreaterThan(0)` which should be the same as `.toBeDefined()` considering how we built the derive.

jacogr: That is fine, just comment it with "this needs a clean chain"

packages/api-derive/test/e2e/promise.spec.ts
Tbaut: Thanks, totally missed that, not familiar with the `status` yet

packages/api-derive/test/e2e/promise.spec.ts
```diff
@@ -100,4 +107,35 @@ describe.skip('derive e2e', () => {
...
+    const ALICE_STASH = '5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY';
+    const ALICE = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';
+    const keyring = testKeyring();
+    const aliceStashPair = keyring.getPair(ALICE_STASH);
+    const alicePair = keyring.getPair(ALICE);
+
+    it('bondsExtra funds for Alice Stash', () => {
+      return api.tx.staking.bondExtra(BOND_VALUE)
+        .signAndSend(aliceStashPair, ({ events = [], status }) => {
+          expect(status.type).toMatch(/Flinalized/i);
```
jacogr: Finalized doesn't come in in the first round, so you need some hoops, such as - https://github.com/polkadot-js/api/pull/960/files#diff-a2134f09bc132a8d3d7e01270da62a86R53

PR #967, opened 2019-06-04
Add SignaturePayloadRaw

jacogr:
- Extracted from https://github.com/polkadot-js/extension/blob/master/packages/extension/src/background/RawPayload.ts
- The need is not only for the above, but also comes up with other offline signers (e.g. Polkawallet), hence making it available at the base to ease working with payloads


jacogr:
It is only used in the signing - so SubmittableExtrinsic & Payloads. Hppy with the interface since, yes, apart from tests, it doesn't actually get used as-is, it really is the types only.

PR #974, opened 2019-06-06
Support more consensus engines with digest info

jacogr:


packages/types/src/rpc/Digest.ts
```diff
@@ -178,26 +222,37 @@ export class PreRuntime extends Tuple {
...
+  Other, // 0
+  AuthoritiesChange, // 1
+  ChangesTrieRoot, // 2
+  SealV0, // 3
+  Consensus, // 4
+  Seal, // 5
+  PreRuntime // 6
+};
+
+type DigestItemTypes = keyof typeof DigestItemEnumMap;
```
jacogr: The above map extracted to allow type-checking on the logs filtering

packages/types/src/rpc/Digest.ts
```diff
@@ -37,6 +44,57 @@ export class ChangesTrieRoot extends Hash {
  * A 4-byte identifier (actually a [u8; 4]) identifying the engine, e.g. for Aura it would be [b'a', b'u', b'r', b'a']
  */
 export class ConsensusEngineId extends U32 {
+  static idToString (input: number | BN): string {
+    return bnToBn(input).toArray('le').reduce((result, value) =>
```
amaurymartiny: really small nit, but .map().join() is imo easier to read

PR #977, opened 2019-06-07
Add rewardDestination to derive.staking.info

Tbaut:
- add rewardDestination to `derive.staking.info` + tests
- remove the part of `derive.staking.unlocking` tests that's failing with a brand new node (Alice_stash has no available funds).
- fix test where `derive.staking.unlocking` was actually not verified [here](https://github.com/polkadot-js/api/compare/tbaut-rewardDestination?expand=1#diff-c23a90c8db5d294132f8695fba4792f5L152)

### DIFF
--- a/packages/api-derive/src/staking/info.ts
+++ b/packages/api-derive/src/staking/info.ts
@@ -3,7 +3,7 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { ApiInterface$Rx } from '@polkadot/api/types';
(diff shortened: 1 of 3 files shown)


Tbaut:
CI says:

> Function withControllerLedger has 27 lines of code (exceeds 25 allowed). Consider refactoring.

Not sure I should do that that. Any hint?

jacogr:
We have we too many of those from CodeClimate - some useful, some where we have throw hands in the air. (Yes CodeClimate, you are right, but I really no clue how to actually make it better without making is messier)

I this case, as it stands it is spread out, but readable - can squash the `nextSessionId` into a single line (or even `nextSessionId: nextKeyfor.unwrapOr(null) || undefined`). That removes the warning and at least the second doesn't make it worse.

Issue #980, opened 2019-06-10
 Base.disconnect() does not work


AlexWZ0129:
Hi, I found the provider will keep re-connecting even if the [disconnect()](https://github.com/polkadot-js/api/blob/0b66f27a8c1ad28f15d6549d4a751f246398f91c/packages/api/src/Base.ts#L289) is called

AlexWZ0129:
cannot reproduce

AlexWZ0129:
I know what is happening here, this issue will happen only if `provider` in `ProviderInterface` is undefined.

jacogr:
Closed in #983

PR #983, opened 2019-06-10
fix issue that creates duplicate providers

AlexWZ0129:
Hi, this is a quick fix for #980. Seems like when creating an API instance, if the `provider` in `ApiOptions` is undefined, there will be two Rpc instances created at [`this._rpcBase = new RpcBase(thisProvider);`](https://github.com/polkadot-js/api/blob/0b66f27a8c1ad28f15d6549d4a751f246398f91c/packages/api/src/Base.ts#L112) and [`this._rpcRx = new RpcRx(thisProvider);`](https://github.com/polkadot-js/api/blob/0b66f27a8c1ad28f15d6549d4a751f246398f91c/packages/api/src/Base.ts#L116) with two provider instances. More exactly, at [here](https://github.com/polkadot-js/api/blob/0b66f27a8c1ad28f15d6549d4a751f246398f91c/packages/rpc-core/src/index.ts#L61) the constuctor of Rpc will create a new WsProvider if provider if undefined.

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -113,7 +113,7 @@ export default abstract class ApiBase<CodecResult, SubscriptionResult> implement
 
     assert(this.hasSubscriptions, 'Api can only be used with a provider supporting subscriptions');
 
-    this._rpcRx = new RpcRx(thisProvider);
+    this._rpcRx = new RpcRx(this._rpcBase._provider);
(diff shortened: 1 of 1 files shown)


jacogr:
Yes, good catch. Thank you.

AlexWZ0129:
> Yes, good catch. Thank you.

my pleasure 😸

PR #986, opened 2019-06-11
Storage Size u32

kwingram25:


packages/types/src/type/ContractInfo.ts
```diff
@@ -20,10 +21,11 @@ export class AliveContractInfo extends Struct {
   constructor (value?: any) {
     super({
       trieId: TrieId,
-      storageSize: U64,
+      storageSize: U32,
```
jacogr: As per https://github.com/paritytech/substrate/pull/2602/files#diff-59240eb8b4386f31565195c1211430cfR190

PR #991, opened 2019-06-12
Update to Metadata v5

Stefie:
Addresses #988

PR #993, opened 2019-06-12
Add metadata v5 changes to release notes

Stefie:
moved over from https://github.com/polkadot-js/api/pull/991

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Fix for disconnection when default providers (non-specified) are used
 - Update `ContractInfo` to match substrate master
 - Additional examples for subscriptions using multi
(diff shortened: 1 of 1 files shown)


jacogr:
Don't want to merge it into mine as of yet, since it is not in merged into master as of yet. Simple enough to add as part of follow-up or even final bump (when these are done anyway), so not a major issue atm.

Stefie:
Ah, I thought this was a prepared PR for once the metadata changes are merged.

jacogr:
No issues. That actually is a bit problematic with the bump PRs, since unless you stop CI, you end up with merge conflicts after the facts. So once metadata is in, apps updated and just run-though tested probably close to having API for a stable release. 

(There are a couple of smaller nigglies left though for current node-template and substrate 1.x)

Issue #989, opened 2019-06-11
Add corresponding "off" method for API event listeners.


jnaviask:
The ApiBase currently provides an `on` function for listening to connection-related events, but there's no way to unregister these event handlers. It would be nice to have an `off` function or similar that allows us to remove them. Thanks!

PR #999, opened 2019-06-13
Add .off method to ApiBase

axelchalon:
Close https://github.com/polkadot-js/api/issues/989


jacogr:
Could you just add this to the CHANGELOG as well, well worth mentioning in there

PR #1003, opened 2019-06-14
fix: set isPedantic false if fallback exists

### LINKED ISSUE #1002: isPedantic is conflict with fallback field in metadata
when `await api.query.staking.lastEraLengthChange.at(blockHash)` i have this error :
```
Unable to decode storage staking.lastEraLengthChange: Encoding for input doesn't match output, created 0x0000000000000000 from 0x
```
it's fine without `at()` which go through the subscription path.

the return of state query is null, but since in metadata there's a fallback for LastEraLengthChange.
```
{
              "name": "LastEraLengthChange",
              "modifier": "Default",
              "type": {
                "PlainType": "BlockNumber"
              },
              "fallback": "0x0000000000000000",
              "documentation": [
                " The session index at which the era length last changed."
              ]
            },
```
should not throw but use the fallback. Turn out in `packages/rpc-core/src/index.ts#241` isPedantic is passed as true.

I guess the solution would be to check `meta.fallback` and pass isPedantic false if fallback exists.

ianhe8x:
Close #1002

### DIFF
--- a/packages/rpc-core/src/index.ts
+++ b/packages/rpc-core/src/index.ts
@@ -232,17 +232,18 @@ export default class Rpc implements RpcInterface {
     // outputType that we have specified. Fallback to Data on nothing
     const type = key.outputType || 'Data';
     const meta = key.meta || EMPTY_META;
+    const isPedantic = meta.fallback ? false : true;
 
     if (meta.type.isMap && meta.type.asMap.isLinked) {
-      return createType(type, base, true);
+      return createType(type, base, isPedantic);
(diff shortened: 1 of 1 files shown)


jacogr:
Not the correct fix, as per https://github.com/polkadot-js/api/issues/1002#issuecomment-501960884

The actual issue here is that https://github.com/polkadot-js/api/blob/master/packages/rpc-core/src/index.ts#L230 (unlike the set subscription) does not apply the fallback like it should.

ianhe8x:
cool

Issue #1008, opened 2019-06-14
Create an ergonomic contracts interface


jacogr:
This needs to go after https://github.com/polkadot-js/api/pull/1007

Basically, currently we only allow people to operate directly on the raw contract calls, i.e. https://github.com/polkadot-js/api/blob/master/packages/api/test/e2e/promise-contract.spec.ts#L92

In the above, we would really just like something along the following lines -

```js
import { Contract } from '@polkadot/api-contract';

await new Contract(<api instance>, <abi>, <addr>)
  .flip(<value to go along>, <max gas>, <...params>)
  .signAndSend(<account>, (submittableResult) => {...});
```

That is the end-result, earlier in the chain, we want to be able to deploy or attach code -

```js
import { Blueprint } from '@polkadot/api-contract';

// here it becomes a bit fuzzy... we need to 
//  - make the deploy call
//  - as a result, we want a Contract instance (previous code block)
//  - basically, it would take the address and abi to construct one
await new Blueprint(<api instance>, <abi>, <hash>)
  .deploy(<value>, <gas>, <...params>)
  .signAndSend(<account>, (submittableResult, contract) => {...});

// even better
const contract = await new Blueprint(...).deploy(<account>, ...);
```

Deployment of code (putCode) needs something similar -

```js
import { Code } from '@polkadot/api-contract';

// same as above with a bit of fuzziness
//  - as a result, we would like to retrieve a Blueprint instance
//  - basically it would take the retrieved hash and abi, to construct one
await new Code(<api instance>, <abi>, <wasm bytes>)
  .deploy(<endowment>, <max gas>)
  .signAndSend(<account>, (submittableResult, blueprint) => {...});

// even better
const blueprint = await new Code(...).deploy(<account>, ...);
```

So as a crazy code returns blueprint, returns contract, it ends up -

```js
const blueprint = await new Code(...).deploy(<account>, ...);
const contract = await blueprint.deploy(<account>, ...);

await contract
  .flip(<value to go along>, <max gas>, <...params>)
  .signAndSend(<account>, (submittableResult) => {...});
```

... or something like that

All-in-all, needs some playing to see what works and what doesn't, how to integrate and how to return values. Basically we want to move away from people having to manually inspect events, we can just handle all that transparently and have explicit interfaces.

Additionally, need to see how we tie this to all the type of api interfaces we are supporting, current RxJs & Promise (Similar approach as in Submittables?)

jacogr:
[...]

PR #1015, opened 2019-06-14
More ergonomic contracts interfaces (RxJs only initial)

jacogr:


jacogr:
@amaurymartiny As discussed briefly. This is just some idea, `Code.ts` is the actual simple interface, RxJs only.

amaurymartiny:
Looks good for now

packages/api-contract/src/RxBase.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import Abi from './Abi';
+
+// NOTE Experimental, POC, bound to change
+export default abstract class RxBase implements IContractBase<'rxjs'> {
+  readonly abi: Abi;
+  readonly api: ApiObject<'rxjs'>;
+  readonly apiContracts: SubmittableModuleExtrinsics<'rxjs'>;
+
+  constructor (api: ApiObject<'rxjs'>, abi: ContractABI | Abi) {
```
amaurymartiny: I think we can remove ApiObject, and always pass in ApiRx (like we always pass in ApiRx into Derive and into SubmittableExtrinsic).

Then, from ApiRx, create 2 interfaces.

---

Alternatively (but more in the future, once we're sure the 2 interfaces works), pass in ApiRx and decorateMethod, and there shouldn't be any duplicates anymore.

jacogr: Yes, I purposefully kept this generic at the base object, although it is called RxBase. (And only called such because the `apiContracts` give issues when the Base is made fully generic.) I didn't want to lose track of the `ApiType` thing too much. Although, it may be fully dropped with Promise.

jacogr: I would prefer the `ApiRx` & `decorateMethod` approach. But let's see how it can be used with more than one - always think better when things are actually on-screen.

jacogr:
Lays some groundwork for https://github.com/polkadot-js/api/issues/1008

PR #1016, opened 2019-06-14
Additional types for latest substrate master

jacogr:
Would not trust the new derive.session.info - as a first pass, do the best with the info available, but need this in, looked at on a running system and expanded. (The calcs here are mostly empty, since very limited information is now available - even worse, some hard-coded due to the runtime)

~~Metadata generated using https://github.com/paritytech/substrate/pull/2876~~ (now merged)


packages/types/src/Metadata/v5/latest.substrate.v5.json
```diff
@@ -2,2575 +2,3503 @@
...
-          "name": "BlockHash",
-          "modifier": "Default",
-          "type": {
-            "MapType": {
-              "hasher": "Blake2_256",
-              "key": "BlockNumber",
-              "value": "Hash",
-              "isLinked": false
+          ],
+          "calls": null,
```
jacogr: Something is SERIOUSLY wrong here - https://github.com/paritytech/substrate/blob/master/node/runtime/src/lib.rs#L250

Calls are exposed and do not appear here (e.g. new setCode)

jacogr: Logged a substrate issue for this - as soon as that is available, need to update 

(The bottom-line is that the changes in Method/extrinsics here would need to be re-adjusted when system does return calls, i.e. that will become the first index and everything else will move down, i.e. this would be bumped by 1 https://github.com/polkadot-js/api/pull/1016/files#diff-fdb8bb0e6f1aabc8679a048b348e91adR20)

PR #1030, opened 2019-06-18
Removed auth storage keys

Stefie:
`:auth:` and `:auth:len` have been removed from the `well_known_keys ` as of https://github.com/paritytech/substrate/pull/2802/files#diff-c9dc9bd631e45d41bbbcf740d562a4c7

### DIFF
--- a/docs/METHODS_STORAGE.md
+++ b/docs/METHODS_STORAGE.md
@@ -494,12 +494,6 @@ ___
 
 _These are keys that are always available to the runtime implementation_
 
-▸ **authorityCount**(): `u32`
-- **summary**: Number of authorities.
-
-▸ **authorityPrefix**(): `u32`
-- **summary**: Prefix under which authorities are stored.
-
 ▸ **changesTrieConfig**(): `u32`
 - **summary**: Changes trie configuration is stored under this key.
 
--- a/docs/METHODS_STORAGE.md
+++ b/docs/METHODS_STORAGE.md
@@ -494,6 +494,12 @@ ___
 
 _These are keys that are always available to the runtime implementation_
 
+▸ **authorityCount**(): `u32`
+- **summary**: Number of authorities.
+
+▸ **authorityPrefix**(): `u32`
+- **summary**: Prefix under which authorities are stored.
+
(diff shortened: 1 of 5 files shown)


docs/METHODS_STORAGE.md
```diff
@@ -494,12 +494,6 @@ ___
 
 _These are keys that are always available to the runtime implementation_
 
-▸ **authorityCount**(): `u32`
```
jacogr: Revert. This us auto-generated.

docs/METHODS_STORAGE.md
```diff
@@ -494,12 +494,6 @@ ___
...
-- **summary**: Number of authorities.
-
-▸ **authorityPrefix**(): `u32`
```
jacogr: As above, revert changes to auto-generated files, CI does it.

packages/types/src/scripts/METHODS_STORAGE_SUBSTRATE.md
```diff
@@ -5,12 +5,6 @@
 
 _These are keys that are always available to the runtime implementation_
 
-▸ **authorityCount**(): `u32`
```
jacogr: As above.

amaurymartiny:
Ah yeah, good point.

packages/types/src/scripts/METHODS_STORAGE_SUBSTRATE.md
jacogr: Sorry, this removal is valid - so it solves the problem of not being in docs.

Stefie:
What do you mean by "make it deprecated" ? Just marking it as deprecated in the docs?

packages/type-storage/src/fromMetadata/substrate.ts
```diff
@@ -50,6 +42,16 @@ export const authorityPrefix = createRuntimeFunction('authorityPrefix', ':auth:'
   type: 'u32'
 });
 
+export const code = createRuntimeFunction('code', ':code', {
```
jacogr: Ordering here are find, deprecated first, then the rest as per substrate. +1

PR #1037, opened 2019-06-19
refactor: Remove rpc-rx

### LINKED ISSUE #642: Merge rpc-rx and rpc-core packages
These two packages could be merged into one.

Right now, rpc-core is a thin layer over the rpc-provider, which basically turns WS requests/responses from JSON to a `@polkadot/types` type, and returns a Promise or a callback. Then rpc-rx takes this rpc-core, and transforms this Promise/callback into an Observable. (And worse: further down the road, the OnCall in api transforms this Observable back again into a Promise/callback!)

One simple solution is to just:
- remove rpc-rx completely, only keep rpc-core
- let rpc-core return Observables instead of Promise/callback. In case of a Promise it's just an Observable that completes after the 1st value.

amaurymartiny:
closes #642


packages/api/test/e2e/promise-tx.spec.ts
```diff
@@ -192,10 +192,12 @@ describe.skip('Promise e2e transactions', () => {
...
         if (header.blockNumber.toNumber() === eraDeath - 1) {
-          const hash = await ex.signAndSend(keyring.alice, { blockHash: eraHash, era: exERA, nonce } as any);
-
-          expect(hash).toBeUndefined();
-          done();
+          try {
+            await ex.signAndSend(keyring.alice, { blockHash: eraHash, era: exERA, nonce } as any);
+          } catch (error) {
+            expect(error.message).toMatch(/1010: Invalid Transaction \(0\)/);
+            done();
```
amaurymartiny: So this is a breaking change, or more like a breaking fix.

Before `ex.signAndSend` at an invalid block would resolve with undefined. Now `ex.signAndSend` will throw, which I think should be the desired behavior. cc @KarishmaBothara i think you wrote this test

jacogr: `throw` is correct.

packages/rpc-core/src/index.ts
```diff
@@ -117,23 +123,40 @@ export default class Rpc implements RpcInterface {
...
-      }
+      // Here, logically, it should be `of(this.formatInputs(method, values))`.
+      // However, formatInputs can throw, and when it does, the above way
+      // doesn't throw in the "Observable loop" (which is internally wrapped in
+      // a try/catch block). So we:
+      // - first do `of(1)` - won't throw
+      // - then do `map(()=>this.formatInputs)` - might throw, but inside Observable.
+      return of(1)
+        .pipe(
+          map(() => this.formatInputs(method, values)),
```
axelchalon: just for info `defer(() => of(this.formatInputs(method, values))` also does the trick

PR #1047, opened 2019-06-21
GenericCall & Event: prepare getting rid of injection/globals

axelchalon:
Part of https://github.com/polkadot-js/api/issues/580

The second parameter to the Method type constructor (method metadata) is now required. The method metadata must now be passed as second argument to the Extrinsic constructor (had to make the first argument required for this).

This lets us get rid of `injectMethods` and of storing the metadata globally in `Method.ts`. (Event will follow).

Ready for review but I have to check if this is still usable in apps; might make more updates here as I'm updating apps.


jacogr:
Edgeware, Apps and the extension will be affected, at least from a findMethod perspective. Basically, since it decodes e.g. proposals, it needs to find the method. 

With `Method()` does not having a default anymore (which breaks away from what is in parity-scale-codec as well), this could be problematic.

amaurymartiny:
The way we construct `api.tx.*.*` today, we need to pass metadata to Method. The PR as-is will break this package, someone needs to inject metadata to Method.

A good test is to see if all e2e tests pass, they are failing on this PR.

jacogr:
Let me just clarify - I'm not against the idea at all this opens up the door to some nice things, things. The `findMeta` is a good idea that solves the most pressing issues. I am however against this idea in the next 1 month. 

(That is until we are stable and have feature parity between 2.x and 1.x at least, there are just a lot of unknowns as it stands and I don't particularly want to add some more)

axelchalon:
:+1: I'm adding 'on ice' to the title but feel free to close the PR if you prefer

> The way we construct api.tx.*.* today, we need to pass metadata to Method. The PR as-is will break api, someone needs to inject metadata to Method.

The metadata is already passed to Method when constructing `api.tx.*.*`: https://github.com/polkadot-js/api/blob/7e6e609f2a63b22a648f0ab31e4210632ced8ec8/packages/type-extrinsics/src/fromMetadata/createUnchecked.ts#L34-L37

jacogr:
I think we can actually start pulling at least parts in - like the findMeta, ie. That is not breaking and we can already start pointing to using that. (So eg. Apps, Extension, Edgeware and I believe, SingleSource can already start using this interface)

jacogr:
Just to clarify - the findMeta interface without getting rid of the older version. (And then start pointing to using that in examples and changelog). Same I guess with the Event equivalent.

amaurymartiny:
If end-developers are using `.findMeta`, I would propose something like:

```javascript
Method.findMeta(callIndex: Uint8Array, metadata: Metadata);
Method.findMeta(sectionName: string, methodName: string, metadata: Metadata);

// Then users can use like:
import {Method} from '...'
const api = new Api();
Method.findMeta([3,0], api.runtimeMetadata);
Method.findMeta('balances', 'balanceOf', api.runtimeMetadata);
```

The last argument is the whole metadata, so that users don't need to do `Method.findMeta('balances', 'balanceOf', extrinsicsFromMetadata(api.runtimeMetadata));`

axelchalon:
Starting to update apps now and seeing what issues come up. Might update this PR along the way.

axelchalon:
Ready for review/merge

packages/types/src/primitive/Method.ts
```diff
@@ -65,7 +67,25 @@ export class MethodIndex extends U8aFixed {
...
 
-  constructor (value: any, meta?: FunctionMetadataV5) {
+  /**
+   * Method constructor
+   *
+   * @param value
+   * @param meta - Metadata to use, so that the `injectMethods` lookup is not
+   * necessary. This argument will soon be required so as to get rid of globals.
+   */
+  constructor (value: any, metaOrMethod?: FunctionMetadataV5 | MethodFunction) {
```
amaurymartiny: In extension, you changed to https://github.com/polkadot-js/extension/blob/8d9463878602d59e6bca313ae8b49f01ee9365e0/packages/extension-ui/src/Popup/Signing/Details.tsx#L33

```javascript
   const method = new Method(data, Method.findMetaByValue(data, meta));
```

Instead of `FunctionMetadataV5 | MethodFunction`, we could do `FunctionMetadataV5 | Metadata`? So that you would just

```javascript
   const method = new Method(data, meta); // reads better
```

Edit: or for better future-proofness, `const method = new Method(data, {meta});`? Not sure...

---

There's another reason I prefer to pass the whole Metadata to Method/Event/Extrinsic. It's that the api/the end-user doesn't need to know what to extract from Metadata to pass to Method. Just pass the whole thing, and Method itself should extract the relevant information it needs.

packages/types/src/primitive/Method.ts
```diff
@@ -222,13 +310,17 @@ export default class Method extends Struct implements IMethod {
 
   /**
    * @description Returns the name of the method
+   *
+   * @deprecated Will be removed soon.
```
amaurymartiny: tbh I think there's no need to deprecate this.

Phase 1 (now): we warn users that global inject will be gone soon, so that they do `new Method(value, meta)` in their code, and we extract methodName/sectionName from meta
Phase 2: we remove global inject, and make meta in constructor obligatory. We can then still extract methodName, sectionName.

packages/types/src/type/Extrinsic.ts
```diff
@@ -33,10 +33,10 @@ type ExtrinsicValue = {
  * - left as is, to create an inherent
  */
 export default class Extrinsic extends Struct implements IExtrinsic {
-  constructor (value?: ExtrinsicValue | AnyU8a | Method) {
+  constructor (value?: ExtrinsicValue | AnyU8a | Method, meta?: FunctionMetadata) {
```
amaurymartiny: Would put `Metadata` or `{ meta: Metadata }` here, as above.

packages/types/src/primitive/Method.ts
axelchalon: If we pass a FunctionMetadataV5 to the constructor, then we don't have the section name

amaurymartiny: Agreed. This is why I would advocate to pass in whole metadata always.

axelchalon: in my mind it somehow made sense to have the section and method name stored in MethodFunction, a wrapper around FunctionMetadata; and if we need to know/use the section/method name, then we use a MethodFunction

packages/types/src/primitive/Method.ts
axelchalon: future-proofness is a good idea! :+1: passing the runtime metadata to the Method constructor feels to me way more practical but somehow less clean/functional (i.e. now `types` is somehow tied to the structure of the runtime metadata). but let's go with ease of use

packages/types/src/primitive/Method.ts
axelchalon: however, considering the section/method name will have been passed to the constructor via meta if the developer wants to access it, I think it's safe to keep the getters :+1:

packages/types/src/primitive/Method.ts
amaurymartiny: > (i.e. now types is somehow tied to the structure of the runtime metadata).

I see what you mean. But I would argue that even with a `MethodFunction` passed into the constructor, it's still tied to the structure of the runtime Metadata, because `MethodFunction.meta` is a `FunctionMetadataV5`, which can change with different metadata version.

packages/types/src/primitive/Method.ts
```diff
@@ -19,8 +25,12 @@ interface DecodeMethodInput {
 interface DecodedMethod extends DecodeMethodInput {
   argsDef: ArgsDef;
   meta: FunctionMetadataV5;
+  method?: string;
+  section?: string;
 }
 
+export type MetaLike = FunctionMetadataV5 | MethodFunction | Metadata;
```
amaurymartiny: I'm still not convinced about MethodFunction.

And as you proposed in Riot, i would maybe go for removing MethodFunction in the whole codebase.

packages/types/src/type/SignaturePayload.ts
```diff
@@ -44,10 +49,10 @@ function sign (signerPair: IKeyringPair, u8a: Uint8Array): Uint8Array {
 export default class SignaturePayload extends Struct {
   protected _signature?: Uint8Array;
 
-  constructor (value?: SignaturePayloadValue | Uint8Array) {
+  constructor (value?: SignaturePayloadValue | Uint8Array, options: ConstructorOptions = {}) {
     super({
       nonce: Nonce,
-      method: Method,
+      method: (isU8a(value) && options.meta) ? Method.withMeta(options.meta) : Method,
```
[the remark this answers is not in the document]
axelchalon: right, I agree it's a bit confusing
I guess I was trying to make explicit that if we pass in a `value: SignaturePayloadValue`, then `options.meta` will not be taken into account, because the `Method` is already instanciated inside `value` (as `value.method`)

packages/types/src/primitive/Method.ts
```diff
@@ -64,16 +74,35 @@ export class MethodIndex extends U8aFixed {
  */
 export default class Method extends Struct implements IMethod {
   protected _meta: FunctionMetadataV5;
+  protected _method?: string;
```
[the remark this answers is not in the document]
axelchalon: right, I didn't think about it. I think the character case might be different in MethodFunction.section and in meta.section as well. I'm keeping it like this (i.e. like the way it was before this PR) for now, and adding a comment

packages/types/src/primitive/Method.ts
```diff
@@ -86,21 +119,45 @@ export default class Method extends Struct implements IMethod {
...
-  private static decodeMethod (value: DecodedMethod | Uint8Array | string = new Uint8Array(), _meta?: FunctionMetadataV5): DecodedMethod {
+  private static decodeMethod (value: DecodedMethod | Uint8Array | string = new Uint8Array(), _meta?: MetaLike): DecodedMethod {
+
+    const unwrapMeta = (callIndex: Uint8Array, meta?: MetaLike): {
+      meta: FunctionMetadataV5;
+      method?: string;
+      section?: string;
+    } => {
+      const isFunctionMetadata = (x: any): x is FunctionMetadataV5 =>
+        !!(x && 'args' in x);
```
amaurymartiny: `x instanceOf FunctionMetadataV5`?

packages/types/src/type/SignaturePayload.ts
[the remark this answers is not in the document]
amaurymartiny: Ah I get it now. So maybe
```
value instanceOf SignaturePayloadValue ? Method : Method.withMeta(options.meta)
```

the problem is that `AnyU8a = Uint8Array | Array<number> | string`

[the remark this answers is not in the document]
axelchalon: `SignaturePayloadValue` is a `type`, so can't use `instanceof`: ` 'SignaturePayloadValue' only refers to a type, but is being used as a value here.ts(2693)`

> the problem is that AnyU8a = Uint8Array | Array<number> | string

what do you mean?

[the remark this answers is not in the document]
amaurymartiny: Ah yeah, true.

> what do you mean?

Ah sorry, I was actually looking at the constructor of Extrinsic. There the isU8a seems a bit weirder, we can (and we do) construct extrinsics by hex. `new Extrinsic('0x1234', { meta })` will never use meta

axelchalon:
@amaurymartiny regarding your suggestion: the function part of `MethodFunction` is used in `decorateExtrinsicEntry`, so I'll keep it in the interface for now

packages/types/src/type/SignaturePayload.ts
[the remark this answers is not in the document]
amaurymartiny: What do you think of removing the isU8a part here...

(to be continued, see below)

packages/types/src/primitive/Method.ts
```diff
@@ -86,21 +119,42 @@ export default class Method extends Struct implements IMethod {
...
       return {
         args: value.subarray(2),
         argsDef: Method.getArgsDef(meta),
         callIndex,
-        meta
+        meta,
+        method,
+        section
       };
     } else if (isObject(value) && value.callIndex && value.args) {
```
amaurymartiny: ... and add another case here:

```javascript
else if (value instanceof Method) {
  return Method.{args,argsDef,callIndex,meta,method,section}
}
```

So that the logic "ignore options.meta" is done here, not in constructors of all other types that need Method

packages/types/src/primitive/Event.ts
```diff
@@ -126,8 +150,46 @@ export default class Event extends Struct {
     };
   }
 
+  static getDataType (soughtEventIndex: Uint8Array, metadata: Metadata): Constructor<EventData> | undefined {
+    const methods = ([] as {meta: EventMetadataV5, eventIndex: Uint8Array, sectionName: string}[]).concat(
```
amaurymartiny: 2 things:
- the casting is wrong imo: `metadata.asV5.modules` is `ModuleMetadata[]`, so should be the empty array
- small formatting nit: .concat could go on a new line

packages/types/src/primitive/Event.ts
```diff
@@ -16,6 +16,11 @@ import Null from './Null';
 import U32 from './U32';
 import Unconstructable from './Unconstructable';
 
+interface ConstructorOptions {
+  DataType?: Constructor<EventData>;
```
amaurymartiny: Same question like last time: do you think it's useful to keep this `DataType` field in the options?

axelchalon: I think it makes sense to be able to instanciate an Event/EventRecord and provide it with its DataType, rather than only be able to pass in the whole metadata from which the DataType is extracted

e.g. useful in tests as well

packages/types/src/primitive/Event.ts
axelchalon: > the casting is wrong imo: metadata.asV5.modules is ModuleMetadata[], so should be the empty array

I'm mapping `metadata.asV5.modules` so that it becomes an array of arrays of `{meta, eventIndex, sectionName}` ; and then I'm using `[].concat(...arraysOfArrays)` to flatten. The end array is an array of `{meta, eventIndex, sectionName}`, so I don't think the casting is wrong?

amaurymartiny: Ah my bad. I was reading `([] as ...).concat(...metadata.asV5.modules).filter(...).map(...)`.

In this case the .concat formatting is better like this.

jacogr:
Tests need to pass. 

There is a broader discussion here which I cannot quite have on a phone. However, making the meta required at this time, is not correct. 

Effectively when we get reflected metadata (apart from primitives, no type classes), we need to effectively (akin to what do do with the derives) have the types wrappped in a metadata closure. 

Anything trying to do too much at this point is premature and will probably be reworked (have to be, types as it stands can’t live on). 

Less is more.

amaurymartiny:
> making the meta required at this time

meta will never be required.

> have the types wrappped in a metadata closure

So as I understand:

```typescript
class Event {
  constructor() {
    const a = metadata; // won't be undefined, if the whole class is in a closure
  }
}
```

which is fine. I suppose it'll throw an Error if we instantiate outside of a closure, e.g.
```typescript
const Event = api.fromTypeRegistry('Event'); // Once we have 1328, get Event from metadata
const e = new Event(value); // will throw, because it doesn't have metadata in a closure
```

What this PR is doing is somewhat similar. The only difference: the injection of metadata is not done via closure, but via constructor.
Same behavior:
```typescript
const Event = api.fromTypeRegistry('Event'); // Once we have 1328, get Event from metadata
const e = new Event(value); // will throw, because it doesn't have metadata in its constructor
```
but if it's used inside the api, more specifically inside the `createType` function, then the `createType` will inject metadata to every type, so that those who need it can use it. See https://github.com/polkadot-js/api/issues/580#issuecomment-506778954 where I tried to explain this.

---

This is indeed a big change for api, so I'm happy to wait a bit before merging this, we can discuss this once you come back.

jacogr:
Yes, agreed similar. However having it in a closure applies to all. Would prefer to keep types as new Type - although when we have reflected types, the new will never be exposed.

jacogr:
And yes, anything like this is massive - but the reflected types will break down the way we do things completely. There are no more classes to new.

axelchalon:
Haven't followed up on the recent changes in api, but I think the reflected metadata types might make this PR obsolete and as such this is safe to close? otherwise the closure solution also sounds good to me, provided we'll always instantiate the types using createType or such

amaurymartiny:
> I think the reflected metadata types might make this PR obsolete

Getting rid of injection/globals still needs to be done, so I believe this PR is still relevant.

> otherwise the closure solution also sounds good to me

What will e.g. Method.ts look like in the closure solution?

currently:

```
```

packages/types/src/primitive/Event.ts
amaurymartiny: I'm still half-convinced about this. For me, from a user's perspective, the user should not handle parts of the metadata, but only the metadata as a whole, and tests should use that too

Not a big deal though, we can still keep it.

packages/types/src/primitive/Method.ts
```diff
@@ -142,12 +202,32 @@ export default class Method extends Struct implements IMethod {
...
 
     return injected[callIndex.toString()] || FN_UNKNOWN;
   }
 
+  /**
+   * Retrieves a function from the runtime metadata
+   * @param callIndex Call index of the function, e.g. new Uint8Array([3, 0])
+   * @param metadata Runtime metadata, e.g. api.runtimeMetadata
+   */
+  static findByCallIndex (callIndex: Uint8Array, metadata: Metadata): MethodFunction | undefined {
```
amaurymartiny: I think we can remove this. Types should be as simple as possible. If requested by a user, we could add this function as a helper function somewhere else.

packages/types/src/primitive/Event.ts
```diff
@@ -96,24 +102,43 @@ export class EventId extends U8aFixed {
...
-  static decodeEvent (value: Uint8Array = new Uint8Array()) {
+  static withDataType (DataType: Constructor<EventData>): Constructor<Event> {
+    return class extends Event {
+      constructor (value?: any) {
+        super(value, { DataType });
+      }
+    };
+  }
+
+  static withMeta (meta: Metadata): Constructor<Event> {
```
amaurymartiny: I would remove the two with*s here, not sure they are useful

packages/types/src/primitive/Method.ts
```diff
@@ -2,15 +2,21 @@
...
+import Metadata from '@polkadot/types/Metadata';
 import { assert, isHex, isObject, isU8a, hexToU8a } from '@polkadot/util';
 
 import { getTypeDef, getTypeClass } from '../codec/createType';
 import Struct from '../codec/Struct';
 import U8aFixed from '../codec/U8aFixed';
 import { FunctionMetadata as FunctionMetadataV6, FunctionArgumentMetadata } from '../Metadata/v6/Calls';
 
+interface ConstructorOptions {
+  meta?: MetaLike;
```
amaurymartiny: Same here, better to enforce meta always being a Metadata

jacogr:
@amaurymartiny Happy to revive without the extras, i.e. just making `new GenericCall(callData, { metadata })` work

Be aware that currently in most cases it will always be constructed via `createType('Call', callData)` so in general operation (e.g. as part of an extrinsic), this the data will never be available as a first step. However what we save on is this dance - 

https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/Popup/Signing/Details.tsx#L39-L41

Most other uses, should more-or-less be catered for with the `api.findCall`, e.g https://github.com/polkadot-js/apps/blob/master/packages/app-extrinsics/src/Selection.tsx#L139

packages/types/src/primitive/Method.ts
axelchalon: findByCallIndex is used internally in decodeMethod, to extract the MethodFunction from the runtime metadata + call index

amaurymartiny: Ah yeah, I see. Maybe private static then? Or just a function

axelchalon:
The way I understood the closure idea was that the types constructors would be able to access some predefined globals, ultimately always guaranteed to be set, since the type constructors would only be instantiated from within api.createType, which would set the globals before calling the constructor, and unset them after (iirc the way React works internally?).

```ts
export default class Method extends Struct implements IMethod {
 constructor(value) {
   console.log(globalMetadata, globalTypeRegistry, /* ... */)
   // ...
 }
}
```

To ensure that the globals are always the same to an instance of the type constructor, we would need to enforce that the globals not be stored in the type class and not be used after the constructor is called (e.g. in a public method), and there's no way to enforce this afaik.

This being said, an advantage of this approach is that the type constructor doesn't need to pass the metadata etc. forward to other types it constructs (e.g. Extrinsic using Method and having to pass the received metadata forward to Method). The type constructor options are then type-specific options.

axelchalon:
(Addresses https://github.com/polkadot-js/api/issues/1224 in part)

- Not sure about the naming in the options, do we want `meta` (as in this PR) or `metadata` as key in the options?
- Will have to create an issue to rewrite the tests in Method.spec.ts using the runtime metadata, since Call doesn't accept the function metadata anymore
- A bit unsure if EventRecord still passes the metadata forward, haven't gotten my head around the latest changes (EventRecord being now an interface)

I'm getting some weird test errors that seem to me unrelated to this PR, will have to investigate.

jacogr:
@axelchalon Would still with `meta` for the method-specific version (like we have everywhere), and `metadata` for the full thing, i.e. everything an the kitchen-sink.

axelchalon:
Getting weird errors that seem to me unrelated to this PR, more related to metadata v6/v7.. a bit at a loss what to make of them or how to fix them

jacogr:
So, this looks deceptively simple... between the registry, not having classes and trying to make things non-global, well, a bit over-the-top. (Just saying - underatand wtf you are dealing with here). Onto the stuff -

- `TypeError: Right-hand side of 'instanceof' is not an object` never seen this before - and it is the bulk of them. Will need to dig in specifically

- `Unable to determine type from {"info":5,"type":"GenericAccountId"}` - this mean the `injector` can not been called, i.e. the type is not there and not available registered. However... since `Generic/Event.spec.ts` should have called this (I'm not checking, just assumed it has) as a loss as to why...

- `Unable to determine type from {"info":5,"name":"foo","type":"u32"}` - weirdest of them all. It has a class of these, it should have been injected with the injector.

axelchalon:
Well, this took me a while to track down... TypeRegistry is in urgent need of a rework (i.e. getting rid of global), there's a lot of weird shit going on under the hood (circular deps & nested calls) because of the way things are set up and I feel like this is a ticking timebomb of very hard-to-debug problems. For example I had the dynamic `require` statement only return part of the module's exports... and other weird shit.

I still don't fully understand why things didn't work, but one part was something along the lines of this: the test uses a `new TypeRegistry()` and then registers a type using a string (e.g. `registry.register({U32Renamed: 'u32'});`). `register` then internally tries to fetch `u32` from the global type registry. The global type registry doesn't exist yet, so we start the process to create it: first step is we dynamically require the `index.types` that will be later injected to the global type registry. `index.types` re-exports `Call` which imports `@polkadot/api-metadata/extrinsics/fromMetadata` which imports `createUnchecked` which imports `import { createType, Metadata } from '@polkadot/types'`. The index file of `@polkadot/types` runs the injector, which tries to get the global type registry, which doesn't exist yet, so it tries to create it, and re-requires `index.types`, etc.

In other words, when `getDefaultRegistry` wants to create the global registry, the dependencies make it call `getDefaultRegistry` again (and the global registry still doesn't exist, so it's probably created twice, once at each nesting level). Reading this I don't even understand why everything isn't already falling apart. Feels like we're walking a very fine line.

jacogr:
The registry vs createType is an interesting thing - effectively there is duplication and it is very messy. (The fact that we support multiple registries, but only use one which is a global anyway, is even worse)

axelchalon:
Isn't the unification of constructor signatures (as done in this PR for Call, Event & Extrinsic) (https://github.com/polkadot-js/api/issues/1224) still relevant for when we will want to provide a single access point to initialize types ([api.type](https://github.com/polkadot-js/api/issues/1213)) ?

jacogr:
Yes, but we are going the wrong way around here. We are unifying without even understanding how things are used everywhere, even worse is that the types will come from the metadata as well, putting even more pressure on the register/create funnel.

It is the base of everything, that only played a very small roll up till very recently and is now used everywhere. And the base now does things it is not supposed to.

axelchalon:
What would you recommend as a next step / next issue to tackle, to fix the base ?

jacogr:
So if you are asking where to start, I would start here - https://github.com/polkadot-js/api/issues/1263

What we should have, in terms of dependencies -

- 1. codec types, they only depend on each other
- 2. primitive types, they only depend on codec and each other
- 3. registry and creators, they only depend downwards
- 4. inherited types
- 5. apps

Effectively at this point the first 3 layers contain circular deps among them, which, as you've noticed, creates some really ugly issues. The above list is a pyramid, higher can depend down, but lower cannot depend up - which is currently the case.

After this, we need to sort out the registry/createType mess. It cross-depends on each other, it should be one clean lil bundle. (Not convinced we need 2, i.e. `createType` is what we tell people to use, no `getRegistry.get`)

jacogr:
Closing, replaced by https://github.com/polkadot-js/api/pull/1592

Issue #1048, opened 2019-06-22
Support new Polkadot metadata types


jacogr:
Current polkadot master: 

Unknown types found, no types for 

AuctionIndex,Bidder,IncomingParachain,LeasePeriod,LeasePeriodOf,NewBidder,ParaIdOf,SlotRange,SubId,UpwardMessage,WinningData,grandpa::AuthorityId

polkadot metadata here https://gist.github.com/jacogr/7864197585e12ee751225f25a534129c

So some new, some needs to be aliassed (`grandpa::AuthorityId` -> `AuthorityId`) some are just names for existing, e.g. `ParaIdOf` (Same way `Balance` -> `BalanceOf`) gets handled

PR #1050, opened 2019-06-22
Latest polkadot metadata

### LINKED ISSUE #1049: Pass validateTypes with flag, not throwing on init
It should -

- log the missing types
- not throw
- emit event as per types - this can be used to display info to users (assuming the API is used in an UI, on the console the logging displays, as well as the error event raised)

It will throw when trying to work with any missing types, which is events is certainly almost immediately fatal. However, this allows developers to at least do basic operations.

This is the functionality we had initially before closing down completely - since we needed to action. (Can still action with logs)

It actually does part of this already, i.e.

image

jacogr:
Closes https://github.com/polkadot-js/api/issues/1048, Closes #1049

As a bonus, we now can handle `[<Type>; <length>]` definitions (not only `[u8; <length]`)


jacogr:
Merging, need it for parachains apps as of right now. Tests pass, no substrate types touched, no existing polkadot types either.

Issue #1043, opened 2019-06-20
Pull parameter_types! from metadata


jacogr:
... when available there, i.e. via https://github.com/paritytech/substrate/issues/2921

- A number of parameters, previously in storage, have already moved, e.g. https://github.com/paritytech/substrate/blob/master/node/runtime/src/lib.rs#L171
- More are on the way/in-progress, e.g. all fees

Once available on metadata, add to the api (like we have `genesisHash`, `runtimeVersion`) so it is available to all that want to query it. Additionally, update at least the derive interfaces to pull known/interesting values from here, if available.

amaurymartiny:
This issue can be solved in 2 steps:

- step 1: create MetadataV6 to cater for https://github.com/paritytech/substrate/commit/4c6a54e60631fd4ec72ebc38b4e13849b8008483

- step 2: think about an interface to expose these constants.

For step 2, the natural interface I can think about is:

```javascript
api.constants.council.candidacyBond; // = new U64(3)
```

amaurymartiny:
Just a small note about step 2. The natural way of doing this (if we follow the current pattern) is to create a new package called `type-constants` and put a fromMetadata function, like in `type-{storage,extrinsics}`.

However, I feel this package `type-constants` is too small, i.e. not worth to put in a separate package.

I would propose the following change: merge `type-{extrinsic,storage,constants}` into one package `type-metadata`, and create `{extrinsic,storage,constants}/` folders inside.

---

Realistically, I guess for now, for step 2, let's first create a `type-constants` package, and consider refactoring later on.

jacogr:
Yes, happy with merging - it has been in the back of my mind. As it stands the existing ones is not very big either, and rather a hold-over from the initial hard-coded poc-2 versions.

Stefie:
Moving over here from @axelchalon 's PR: 

> > Stefie: Is that going to be the final format?
> 
> Amaury: Still changeable, but that's my favorite format.


In Javascript world I like `api.constants` better, but since it's an implementation of the substrate `parameter_types`, your second suggestion `api.parameters` also makes a lot of sense. 

On the other hand we're using `query` for `storage` and `tx` for `extrinsics` , so `constants` for `parameters` wouldn't break any logic naming-wise.

So in short: +1 for `api.constants`

PR #1059, opened 2019-06-27
feat: Metadata V6 with module constants

yjkimjunior:
addressing part 1 of #1043

~work in progress while waiting for https://github.com/paritytech/substrate/commit/4c6a54e60631fd4ec72ebc38b4e13849b8008483
to get merged to test and see what breaks.~


packages/types/src/Metadata/v6/Constants.ts
```diff
@@ -0,0 +1,48 @@
...
+import Text from '../../primitive/Text';
+import Type from '../../primitive/Type';
+import Struct from '../../codec/Struct';
+import Vector from '../../codec/Vector';
+
+export class ModuleConstantMetadata extends Struct {
+  constructor (value?: any) {
+    super({
+      name: Text,
+      ty: Text,
```
amaurymartiny: Here, it can be `ty: Type`, which is a subclass of Text which deals with rust types.

amaurymartiny: And actually we use `type: Type` in JS, it's not the same as Rust, but ty is just weird

packages/types/src/Metadata/v6/Constants.ts
```diff
@@ -0,0 +1,48 @@
...
+      value: Bytes,
```
amaurymartiny: yeah, we can leave Bytes now. To be perfectly correct, it should be `value: ty` (i.e. value should be the type of what was returned inside ty). But we do the same for `fallback: Bytes` right now, so let's keep it like this.

amaurymartiny:
Ready to be tested! I'll build Basti's branch too

packages/types/src/Metadata/v6/Constants.ts
yjkimjunior: true, i was matching directly with whatever is in Rust, plus type is a reserved keyword so to me that was more confusing 😄

packages/types/src/Metadata/v5/toV6.ts
```diff
@@ -0,0 +1,25 @@
...
+export default function toV6 (metadataV5: MetadataV5): MetadataV6 {
+  return new MetadataV6({
+    modules: metadataV5.modules.map((modul) => {
+      return {
+        name: modul.name,
+        prefix: modul.prefix,
+        storage: modul.storage,
+        calls: modul.calls,
+        events: modul.events,
+        constants: []
```
amaurymartiny: Here it should be None, i.e. `new Option.with(...)(undefined)`

yjkimjunior: how about `new Vector(ModuleConstantMetadata, [])` ? since constants is not an optional

amaurymartiny: Ah! My bad, i thought is was optional like the 3 above. Then let's keep it like this.

packages/api-derive/test/e2e/rx.spec.ts
```diff
@@ -179,7 +179,7 @@ describe.skip('derive e2e', () => {
     });
 
     // FIXME https://github.com/polkadot-js/api/issues/868
-    describe.skip('getHeader', () => {
+    describe('getHeader', () => {
```
amaurymartiny: add skip back

packages/api/src/promise/Api.spec.ts
```diff
@@ -8,7 +8,7 @@ import Mock from '@polkadot/rpc-provider/mock/index';
 import { ApiPromise } from './..';
 import { ApiOptions } from './../types';
 
-describe.skip('Metadata queries', () => {
+describe('Metadata queries', () => {
```
amaurymartiny: same

packages/api/test/e2e/rx-tx.spec.ts
```diff
@@ -43,13 +43,13 @@ describe.skip('Rx e2e transactions', () => {
       });
   });
 
-  it('makes a proposal', (done) => {
+  it.only('makes a proposal', (done) => {
```
amaurymartiny: remove only

amaurymartiny:
One more thing: Could you add a line in the CHANGELOG to say we're supporting MetadataV6?

packages/types/src/primitive/StorageKey.spec.ts
```diff
@@ -123,4 +124,32 @@ describe('StorageKey', () => {
       ).toEqual(Uint8Array.from([1, 1, 48, 7, 201, 255, 112, 39, 246, 89, 0, 171, 205, 252, 164, 253, 177, 7, 234, 212, 126, 42, 158, 53, 88, 224, 27, 105, 27, 15, 74, 95, 133, 24, 212, 135, 50, 102, 20, 240, 102, 65, 99, 8, 191, 106, 164, 229, 4, 29, 25, 73, 146, 142, 75, 38, 237, 233, 142, 60, 235, 179, 106, 59, 23, 38]));
     });
   });
+
+  describe('with MetadataV6', () => {
```
Stefie: These tests should reflect what changed in this metadata version. In my v5 PR i just copied the  two double map tests because `EventTopics` is the only entry in the metadata v5 (and v4) that uses `DoubleMapType` and the changes in v5 were related to `DoubleMapType`.

Here we should test if the values of some of the modules that are using `parameter_types` now are added correctly.

Stefie: But feel free to merge and put it in a second PR! I know  I'm a little late to the party with my comment here

yjkimjunior: AFAICT, storageKey hasn't changed between v5->v6 so i just copied the same test, i.e. should expect the same result with v6 as with v5.

Stefie: Ah, ok, makes sense!

jacogr:
Why the meta label? That applies to a bundle of issues - unless WIP or breaking, let’s forgo labels on PRs

Issue #1034, opened 2019-06-19
Acomodate 2.x (alongside 1.x) Vote interfaces


jacogr:
Intially from https://github.com/polkadot-js/apps/issues/1315

The 2.x `democracy.vote` doe not take a `i8` anymore, but rather a `Vote` (alongside a conviction). https://github.com/paritytech/substrate/blob/master/srml/democracy/src/lib.rs#L145

This has a specific encoding and decoding override for a compact encoding.

jacogr:
This one could be problematic, i.e. the Vote is encoded differently - unlike decoding, it is not clear how to handle different params on creation. Options -

- Just treat it as a byte, i.e. pass already encoding
- For the above, could expose encoding features on the class, but caller takes care
- Pass in additional version into constructor (which means this is not default anymore, however could be done optionally)

amaurymartiny:
Closed in #1061

PR #1061, opened 2019-06-28
feat: Vote interface as U8a

yjkimjunior:
https://github.com/polkadot-js/api/issues/1034

todo:
* [x] - decode vote from ~struct input~ bits
* [x] - test it is compatible with v1


amaurymartiny:
> test it is compatible with v1

Isn't it simpler to just add it to node_compat?

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,98 @@
...
 
 /**
  * @name Vote
  * @description
  * A number of lock periods, plus a vote, one way or the other.
  */
-export default class Vote extends I8 {
+export default class Vote extends U8a {
+  private _aye: Boolean | I8;
+  private _conviction?: Conviction;
```
amaurymartiny: I think these 2 are not necessary.

I would add 2 getters in this class:
- `get aye (): Boolean`
- `get conviction (): Conviction`

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,98 @@
...
+
   constructor (value?: any) {
-    super(Vote.decodeVote(value));
+    const decoded = Vote.decodeVote(value);
```
amaurymartiny: in `decodeVote`, it will be much simpler to always return a Uint8Array. then below, you don't need to distinguish the cases between the 2 decoded, just `super(decoded)` always.

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,98 @@
...
   }
 
   private static decodeVote (value?: any): any {
     if (isBoolean(value)) {
-      return value ? -1 : 0;
+      if (value === true) {
+        return { aye: new I8(-1) };
+      } else {
+        return { aye: new I8(0) };
+      }
```
amaurymartiny: so as as above:
```typescript
if (isBoolean(value)) {
  return [0b10000000] or [0b0] // we assume Conviction is None
} else if (isNumber(value)) {
  // use https://github.com/polkadot-js/common/blob/master/packages/util/src/bn/toU8a.ts#L31
  // with isNegative
} else if (isObject(value)) {
  // encode value to U8a, following rust
} else if (isU8a(value)) {
  return value; // return as-is
} else {
  throw
}
```

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,98 @@
...
   /**
-   * @description true is the wrapped value is a negative vote
+   * @description true if the wrapped value is a negative vote
    */
   get isNay (): boolean {
     return !this.isAye;
   }
+
+  toNumber (): number {
+    if (this._aye instanceof I8) {
```
amaurymartiny: would just need to do `u8aToBn(this, { isNegative: true })` here

packages/types/src/type/Vote.spec.ts
```diff
@@ -17,15 +19,76 @@ describe('Vote', () => {
...
+
+  it('constructs with Vote aye is true, conviction is Locked4x', () => {
+    expect(
+      new Vote({
+        aye: new Boolean(true),
+        conviction: new Conviction('Locked4x')
+      })
+      .toU8a()
+    )
+      .toEqual(new Uint8Array([1, 0, 0, 0, 0, 1, 0, 0]));
```
amaurymartiny: this is incorrect!

it should be `new Uint8Array([0b10000100])`, i.e. vote.toU8a() is *always* one byte

yjkimjunior:
unit tests pass, fixing CI...

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,79 @@
...
   constructor (value?: any) {
-    super(Vote.decodeVote(value));
+    // decoded is just 1 byte
+    // Aye: Most Significant Bit
+    // Conviction: 0000 - 0101
+    const decoded = Vote.decodeVote(value);
+
+    super(decoded);
+
+    const msb = decoded[0] << 7;
```
amaurymartiny: shouldn't it be >> here?

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,79 @@
...
+
+    this._aye = new Boolean(msb);
+    this._conviction = new Conviction(conviction);
   }
 
-  private static decodeVote (value?: any): any {
+  private static decodeVote (value?: any): Uint8Array {
     if (isBoolean(value)) {
-      return value ? -1 : 0;
+      return value ? new Uint8Array([-1]) : new Uint8Array([0]);
```
amaurymartiny: -1 is incorrect, it should be `new Uint8Array([0b10000000])`

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,79 @@
...
     } else if (value instanceof Boolean) {
       return Vote.decodeVote(value.valueOf());
+    } else if (isNumber(value)) {
+      return value < 0 ? bnToU8a(1, { bitLength: -1, isLe: true, isNegative: true }) : bnToU8a(0);
```
amaurymartiny: If value is positive, we shouldn't return 0, we should also take the conviction into account

here, bnToU8a is enough

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,79 @@
...
+    } else if (isObject(value)) {
+      const aye = value.aye;
+      const index = value.conviction.toNumber();
```
amaurymartiny: you should check that value.conviction is 1/ not undefined and 2/ value.conviction can be a normal JS number (in which case toNumber doesn'ent exist)

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,79 @@
...
+      const conviction = index >>> 0;
```
amaurymartiny: I've never used `>>>` before, so not 100% sure, but I would say this line does nothing?

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,79 @@
...
+      const result = conviction | (aye.eq(true) ? 0b1 << 7 : 0b0 >>> 7);
```
amaurymartiny: Again, `0b0 >>> 7` seems wrong, because `0b0 >>> 7 === 0b0`

advice: `0b0 === 0b00000000`, you don't need to shift right to "fill up" 0s.

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,79 @@
...
-    super(Vote.decodeVote(value));
+    // decoded is just 1 byte
+    // Aye: Most Significant Bit
+    // Conviction: 0000 - 0101
+    const decoded = Vote.decodeVote(value);
+
+    super(decoded);
+
+    const msb = decoded[0] << 7;
+    const conviction = decoded[0] & ((1 << 2) - 1);
```
amaurymartiny: this might work, but I need too much time to understand it.

trick: you can use a bit mask: `const convition = decoded[0] & 0b01111111`

packages/types/src/type/Vote.spec.ts
```diff
@@ -17,15 +19,76 @@ describe('Vote', () => {
...
   });
 
-  it('has isYay for negative', () => {
+  it('has isNay for negative', () => {
     expect(new Vote(new Boolean(false)).isNay).toBe(true);
   });
 
-  it('is positive for negative numbers', () => {
+  it('is Aye for negative numbers', () => {
     expect(new Vote(-999).isAye).toBe(true);
```
amaurymartiny: I would change 999 to something else: `new Vote(a)` -> `a` needs to be between -128 and 127, so 999 won't work.

Also, I would add another test: `new Vote(a).toNumber` should be equal to `a`

yjkimjunior: > Also, I would add another test: new Vote(a).toNumber should be equal to a

should it? my understanding was that a positive vote is a positive vote, and the weight to it is only determined by the conviction. so toNumber should be -1 for `new Vote(a)`, `a > 0` and 0 for `new Vote(a)`, `a < 0` and `a ≠0`

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,78 @@
...
 
 /**
  * @name Vote
  * @description
  * A number of lock periods, plus a vote, one way or the other.
  */
-export default class Vote extends I8 {
+export default class Vote extends U8a {
+  private _aye: Boolean;
+  private _conviction?: Conviction;
```
amaurymartiny: No need for ?

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,78 @@
...
-  private static decodeVote (value?: any): any {
+  private static decodeVote (value?: any): Uint8Array {
     if (isBoolean(value)) {
-      return value ? -1 : 0;
+      return value ? new Uint8Array([0b10000000]) : new Uint8Array([0b0]);
     } else if (value instanceof Boolean) {
       return Vote.decodeVote(value.valueOf());
+    } else if (isNumber(value) && value !== 0) {
+      return value < 0 ? new Uint8Array([0b10000000]) : new Uint8Array([0b0]);
+    } else if (isObject(value) && value.aye && value.conviction) {
```
amaurymartiny: value.aye can be false, and value.conviction can be 0. we should use isUndefined here

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,78 @@
...
+      const aye = value.aye;
+      const convictionIndex = isNumber(value.conviction) ? value.conviction : value.conviction.toNumber();
```
amaurymartiny: If I pass
```
new Vote({
  aye: true,
  conviction: 'Locked2x'
})
```

it should work too.

here you could do `const aye = new Boolean(...)`, `const conviction = new Conviction(...)` (maybe with a try/catch if those two can fail)

packages/types/src/type/Vote.ts
```diff
@@ -2,41 +2,78 @@
...
 
-  private static decodeVote (value?: any): any {
+  private static decodeVote (value?: any): Uint8Array {
     if (isBoolean(value)) {
-      return value ? -1 : 0;
+      return value ? new Uint8Array([0b10000000]) : new Uint8Array([0b0]);
     } else if (value instanceof Boolean) {
       return Vote.decodeVote(value.valueOf());
+    } else if (isNumber(value) && value !== 0) {
+      return value < 0 ? new Uint8Array([0b10000000]) : new Uint8Array([0b0]);
```
amaurymartiny: cf Riot chat, to keep backwards compatibility, this should be fixed

packages/types/src/index.spec.ts
```diff
@@ -8,7 +8,7 @@ import { Constructor } from './types';
 import * as Classes from './index.types';
 
 const Types = Classes as { [index: string]: Constructor };
-const UNCONSTRUCTABLE = ['origin', 'usize'];
+const UNCONSTRUCTABLE = ['origin', 'usize', 'vote'];
```
amaurymartiny: why is this unconstructable?

yjkimjunior: it seems nonsensical to me to construct a Vote with a default value

amaurymartiny: yeah, you're right

packages/types/src/type/Vote.ts
```diff
@@ -2,39 +2,84 @@
...
       return Vote.decodeVote(value.valueOf());
+    } else if (isNumber(value) && value !== 0) {
+      return value < 0
+        ? new Uint8Array([0b10000000])
+        : new Uint8Array([0b0]);
+    } else if (isObject(value) && !isUndefined(value.aye) && !isUndefined(value.conviction)) {
+      const aye =
+        isBoolean(value.aye)
+          ? new Boolean(value.aye)
+          : value.aye;
```
amaurymartiny: `const aye = new Boolean(value.aye)` is enough, it already encompasses all possible cases internally

Edit: new Bool

packages/types/src/type/Vote.spec.ts
```diff
@@ -2,30 +2,97 @@
...
+      .toEqual(true);
+  });
+
+  it('isNay getter works', () => {
+    expect(
+      new Vote({
+        aye: new Boolean(true),
+        conviction: new Conviction('None')
+      }).isNay)
+      .toEqual(false);
```
amaurymartiny: Would like to see an additional test case:

```
  it('Conviction getter works', () => {
    expect(
      new Vote({
        aye: true,
        conviction: 'Locked2x'
      }).conviction.toString()
    )
      .toEqual('Locked2x');
  });
```

packages/types/src/type/Vote.ts
```diff
@@ -2,39 +2,84 @@
...
+
+      const convictionIndex =
```
amaurymartiny: For the above test case to work, I would put here `const convictionIndex = new Conviction(value.conviction)`

packages/types/src/type/Vote.ts
```diff
@@ -2,39 +2,84 @@
...
+import Boolean from '../primitive/Bool';
 
 /**
  * @name Vote
  * @description
  * A number of lock periods, plus a vote, one way or the other.
  */
-export default class Vote extends I8 {
+export default class Vote extends U8a {
+  private _aye: Boolean;
```
amaurymartiny: AH my bad, i just saw this. It's better to return the `Codec` type, i.e. Bool

packages/types/src/type/Vote.ts
```diff
@@ -2,39 +2,84 @@
...
+    // Aye: Most Significant Bit
+    // Conviction: 0000 - 0101
+    const decoded = Vote.decodeVote(value);
+
+    super(decoded);
+
+    const msb = decoded[0] >> 7;
+    const conviction = decoded[0] & 0b01111111;
+
+    this._aye = new Boolean(msb);
```
amaurymartiny: new Bool

packages/types/src/primitive/StorageKey.spec.ts
```diff
@@ -56,7 +56,7 @@ describe('StorageKey', () => {
       ).toBe('0x7f864e18e3dd8b58386310d2fe0919eef27c6e558564b7f67f22d99d20f587bb');
     });
 
-    it(`should correctly get Alice's freeBalance storage key (u8a)`, () => {
+    it.only(`should correctly get Alice's freeBalance storage key (u8a)`, () => {
```
amaurymartiny: rwmove only

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - **Breaking change** The api now uses the module name instead of the prefix to generate the storage methods. The methods of the grandpa module changed from `api.query.grandpaFinality` to `api.query.grandpa`. 
 - Update with latest substrate 2.x types
 - **Breaking Change** StorageFunction has been renamed to StorageEntry.
+- **Breaking Change** Vote interface extends U8a instead of I8.
```
jacogr: Why is this marked breaking? The users of the interface shouldn’t care what happens under the hood. (And if indeed breaking from an API user POV, the comment doesn’t say what to do to fix it)

I would assume the entry would be non-breaking, just supports substrate 2.x interfaces for voting

yjkimjunior:
```suggestion
- **Breaking Change** Vote interface extends U8a instead of I8 and no longer supports the `toNumber()` method. Vote properties can be accessed via the `isAye`, `isNay`, and `conviction` getters. 
Votes can still be constructed as before with a raw JS boolean, a SCALE encoded Boolean, an i8 number, or a JS object with properties `aye` and `conviction` defined. 
```

amaurymartiny: Since I8 is a BN, all BN methods that users might be using are gone

amaurymartiny:
```suggestion
- **Breaking Change** Vote interface extends U8a instead of I8. Vote properties can be accessed via the `isAye`, `isNay`, and `conviction` getters. 
Votes can still be constructed as before with a raw JS boolean, a SCALE encoded Boolean, an i8 number, or a JS object with properties `aye` and `conviction` defined. 
```

yjkimjunior: I the only thing that would have been breaking would be if people were using `toNumber()`, which it doesn't look like apps or any or the other interfaces were. I will document it better anywhow

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 - **Breaking change** The api now uses the module name instead of the prefix to generate the storage methods. The methods of the grandpa module changed from `api.query.grandpaFinality` to `api.query.grandpa`. 
 - Update with latest substrate 2.x types
 - **Breaking Change** StorageFunction has been renamed to StorageEntry.
+- **Breaking Change** Vote interface extends U8a instead of I8 and no longer supports the `toNumber()` method. Vote properties can be accessed via the `isAye`, `isNay`, and `conviction` getters.
```
yjkimjunior:
```suggestion
- **Breaking Change** Vote interface extends U8a instead of I8. Vote properties can be accessed via the `isAye`, `isNay`, and `conviction` getters. 
```

amaurymartiny:
Thanks YJ!

yjkimjunior:
thanks for the help!

Issue #1064, opened 2019-07-02
Change api.derive (fees, staking, session) to use constants


jacogr:

jacogr:
Detection of availability and fallback (So it continues to support old and new). Mostly balances.fees, contracts.fees and the session info (as I remember)

jacogr:
Relevant (now merged) PR - https://github.com/paritytech/substrate/pull/2883

PR #1066, opened 2019-07-02
Pull parameter_types! from metadata

axelchalon:
* [x] `@polkadot/{extrinsics,storage}` moved to `@polkadot/metadata` (as per https://github.com/polkadot-js/api/issues/1043#issuecomment-506449622)
* [x] Expose the constants in the API (using ~~`@polkadot/metadata/constants`~~`@polkadot/metadata/consts`), e.g. ~~`api.constants.council.candidacyBond; // = new U64(3)`~~`api.consts.council.candidacyBond; // = new U64(3)`

Close #1043

Issue #868, opened 2019-05-03
Add tests for api-derive methods


Stefie:
None of the methods have any tests at the moment. they should gradually be added.

jacogr:
Closing this. 

Like any urgent vs important discussion (which is how the next-up it prioritized), this doesn't mean it it won't add value, but rather that it won't get any focus. So the options are to have it sit gathering more dust or just be honest that based on capacity and focus-wise it will not get addressed.

If there is appetite for PRs to move things in the direction intended here, they will always be welcome.

Issue #603, opened 2019-01-18
Extend api.derive with JSDOC - it it very useful to API users


jacogr:
Currently these are not included in the generated documentation, but should be.

(Related to https://github.com/polkadot-js/api/issues/589 where we add some actual sample of use)

PR #1068, opened 2019-07-03
Temporary solution for missing parameter_types

Stefie:
Resolves https://github.com/polkadot-js/api/issues/1064.

Has some follow up tasks once https://github.com/polkadot-js/api/pull/1066 is merged. Marked with
`// @TODO replace this with calls to `api.consts` once implemented`
in the code.

I also added some tests as part of  https://github.com/polkadot-js/api/issues/868 and added some code comments for JSDOC https://github.com/polkadot-js/api/issues/603


amaurymartiny:
#1066 is 95% finished, so if you want to start rebasing on top of that i think it's safe now.

Stefie:
Great, will give it a shot!

Stefie:
Ah, no, @axelchalon  actually pinged me on Riot yesterday evening saying that the PR doesn't work with the latest substrate yet, only with the slightly outdated static version of the metadata. Since there have been no commits since then I'd guess that that's still the case.

axelchalon:
turns out it was my node which had a problem (served outdated metadata), so I think in the end my PR as it stands should still work with latest substrate. wrapping it up this afternoon anyway.

Stefie:
:) Ok, I'll just give it a try and roll back if things break!

PR #1071, opened 2019-07-04
comment out debug logs

Stefie:
The debug logs that spit out the complete hex-encoded metadata multiple times make it quite hard to follow the tests and bump up the time the tests are running significantly without adding any value, unless someone is actually updating the metadata.

If needed, the logs can be easily reactivated.

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -437,7 +437,7 @@ export default abstract class ApiBase<ApiType> {
       } catch (_error) {
         const error = new Error(`FATAL: Unable to initialize the API: ${_error.message}`);
 
-        l.error(error);
+       l.error(error);
 
         this.emit('error', error);
       }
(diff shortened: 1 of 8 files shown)


Stefie:
I changed it only comment out those logs that produce a lot of noise

Stefie:
> right now, NODE_ENV=abc yarn test ... runs tests without any noise. We could always change this logic to only show noise on let's say DEBUG=debug yarn test, but I don't think adding/removing debug lines manually on each test is the way to go.

I didn't know that! I'm really for setting the default to "no noise'" since you really only need 'some'(!) of the metadata log when you're working with the metadata directly and never all of them.

As opposed to that, running tests is the default behaviour when developing and testing all branches and PR's and the logs really slow that down.

amaurymartiny:
So we can close this? It's a polkadot/common issue

jacogr:
Yes, doesn’t make sense at all as is. 

Not sure about the “log slowdown” either - if you want to inspect specific, run those tests only or switch of logging completely with the NODEENV option - “debug” always logs and that will not change on common. 

(Yes, they can be very noisy and provide too much info - but as somebody who has probably run yarn test as much as anybody, not seeing what this means to address)

jacogr:
Closing.

Issue #1070, opened 2019-07-04
Add storage test to promise-queries.spec.ts


axelchalon:
I removed the following test from `@polkadot/api/test/e2e/promise-queries.spec.ts` in https://github.com/polkadot-js/api/pull/1066, because the test used `existentialDeposit` to test the storage with a well-known value. However, `existentialDeposit` has since been moved to `consts`. The test needs to be rewritten to use another storage.

```js
describe('with plain type', () => {
    const EXISTENTIAL_DEPOSIT = 500;
    it('queries correct value', async () => {
      const existentialDeposit = await api.query.balances.existentialDeposit() as Balance;

      expect(existentialDeposit.toNumber()).toEqual(EXISTENTIAL_DEPOSIT);
    });

    it('queries correct value at a specified block', async () => {
      const header = await api.rpc.chain.getHeader() as Header;
      const existentialDepositAt = await api.query.balances.existentialDeposit.at(header.hash) as Balance;

      expect(existentialDepositAt.toNumber()).toEqual(EXISTENTIAL_DEPOSIT);
    });

    it('subscribes to query and get correct result', (done) => {
      return api.query.balances.existentialDeposit((existentialDeposit: Balance) => {
        expect(existentialDeposit.toNumber()).toEqual(EXISTENTIAL_DEPOSIT);
        done();
      });
    });

    it('queries correct hash', async () => {
      const hash = await api.query.balances.existentialDeposit.hash();

      expect(hash).toBeDefined();
    });

    it('gets correct key', async () => {
      const key = api.query.balances.existentialDeposit.key();
      const existentialDepositData = await api.rpc.state.getStorage(key) as Option<any>;
      const existentialDepositRPC = new Balance(existentialDepositData.unwrapOr(undefined));

      expect(existentialDepositRPC.toNumber()).toEqual(EXISTENTIAL_DEPOSIT);
    });

    it('queries correct size', async () => {
      const size = await api.query.balances.existentialDeposit.size();

      expect(size.toNumber()).not.toEqual(0);
    });
  });
```

amaurymartiny:
Putting P1, as these tests help a lot in whether the PRs we make are correct or not. I would love to see them back in master asap, it just doesn't feel right to remove tests

axelchalon:
I added the test back in my branch as I think Stefie's backward-compatibility PR (previously https://github.com/polkadot-js/api/pull/1068) will make the test continue to work
cc @Stefie

Stefie:
[...]

PR #1072, opened 2019-07-05
Switch from existentialDeposit to SessionIndex

Stefie:
Closes https://github.com/polkadot-js/api/issues/1070

### DIFF
--- a/packages/api/test/e2e/promise-queries.spec.ts
+++ b/packages/api/test/e2e/promise-queries.spec.ts
@@ -9,7 +9,7 @@ import { DerivedBalances } from '@polkadot/api-derive/types';
 import WsProvider from '@polkadot/rpc-provider/ws';
 import testingPairs from '@polkadot/keyring/testingPairs';
 import { LinkageResult } from '@polkadot/types/codec/Linkage';
-import { Balance, EventRecord, Header, Index, Option, ValidatorPrefs, Vector } from '@polkadot/types';
+import { Balance, EventRecord, Hash, Header, Index, Option, SessionIndex, ValidatorPrefs, Vector } from '@polkadot/types';
 
 import Api from './../../src/promise';
 
@@ -17,7 +17,7 @@ const ZERO = new BN(0);
 const WS_URL = 'ws://127.0.0.1:9944';
 // const WS_URL = 'wss://poc3-rpc.polkadot.io/';
 
-describe.skip('Promise e2e queries', () => {
+describe('Promise e2e queries', () => {
   const keyring = testingPairs({ type: 'ed25519' });
   let api: Api;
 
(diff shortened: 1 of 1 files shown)


packages/api/test/e2e/promise-queries.spec.ts
```diff
@@ -200,43 +200,42 @@ describe.skip('Promise e2e queries', () => {
...
+      const sessionIndexData = await api.rpc.state.getStorage(key) as Option<any>;
+      const sessionIndexRPC = new SessionIndex(sessionIndexData.unwrapOr(undefined));
 
-      expect(existentialDepositRPC.toNumber()).toEqual(EXISTENTIAL_DEPOSIT);
+      expect(sessionIndexRPC.toNumber()).toBeGreaterThanOrEqual(0);
     });
 
     it('queries correct size', async () => {
-      const size = await api.query.balances.existentialDeposit.size();
+      const size = await api.query.session.currentIndex.size();
```
amaurymartiny: hmm this test fails for me

```
Promise e2e queries › with plain type › queries correct size

    expect(received).not.toEqual(expected) // deep equality

    Expected: not 0

      238 |       const size = await api.query.session.currentIndex.size();
      239 | 
    > 240 |       expect(size.toNumber()).not.toEqual(0);
          |                                   ^
      241 |     });
      242 |   });
      243 | 

      at Object.toEqual (packages/api/test/e2e/promise-queries.spec.ts:240:35)
```

Stefie: which chain? And at which block?

I tested with 1.0 and substrate master and they were passing. I'll try with a fresh node.

amaurymartiny: latest master from 3h ago, fresh chain

amaurymartiny:
Also, this failing test has been hanging around since a while. I propose to just put `[]` to make it pass, as per the commented code.

```
Promise e2e queries › with double map type › queries correct value at a specified block

    expect(received).toEqual(expected) // deep equality

    Expected: undefined
    Received: []

      341 |       // TODO check: this will throw the error: Encoding for input doesn't match output, created 0x00 from 0x
      342 |       const eventTopicsAt = await api.query.system.eventTopics.at(header.hash, KEY1, KEY2);
    > 343 |       expect(eventTopicsAt).toEqual(undefined);
          |                             ^
      344 | 
      345 |       // const eventTopicsAt = await api.query.system.eventTopics.at(header.hash, KEY1, KEY2);
      346 |       // expect(eventTopicsAt.toJSON()).toEqual([]);

      at Object.toEqual (packages/api/test/e2e/promise-queries.spec.ts:343:29)

```

packages/api/test/e2e/promise-queries.spec.ts
Stefie: Ah, ok, ja, a fresh node during the first session it returns `0`

Stefie:
The double map type suite is failing on 1.0, but that's expected...

amaurymartiny:
> The double map type suite is failing on 1.0, but that's expected since `eventTopics` does not exist...

yeah... until we have something that says "if spec_version >= 123, then run double-map tests", I think there's nothing much to do.

PR #1075, opened 2019-07-05
Rewrite SubmittableExtrinsic.ts (close #1045)

### LINKED ISSUE #1045: Rewrite SubmittableExtrinsic.ts
This file: https://github.com/polkadot-js/api/blob/master/packages/api/src/SubmittableExtrinsic.ts

It's hard to maintain, everytime I read this file, I need a 10 good minutes to get what it's doing.

Some ideas:
- Only use TS classes (loaded at runtime, but still possible), no `Object.defineProperties`
- See if `decorateMethod` can be less hacky inside
- https://github.com/polkadot-js/api/blob/5e5bb9984b1f42f0dbc3b4d42a41fdb17aa60898/packages/api/src/SubmittableExtrinsic.ts#L69-L71

How do I pass in options and a callback?
- https://github.com/polkadot-js/api/blob/5e5bb9984b1f42f0dbc3b4d42a41fdb17aa60898/packages/api/src/SubmittableExtrinsic.ts#L79
This seems to be used nowhere...

axelchalon:
close https://github.com/polkadot-js/api/issues/1045


axelchalon:
cf failing travis build / tsc : no idea why `typeof _options extends Callback<ISubmittableResult>` isn't matched for `signAndSend (account: IKeyringPair | string | AccountId | Address, _options?: Partial<SignatureOptions> | Callback<ISubmittableResult>, statusCb?: Callback<ISubmittableResult>)` when calling it with `.signAndSend(keyring.alice, ({ status }: SubmittableResult) => { ... }`

ianhe8x:
SubmittableExtrinsic used to extends Extrinsic like the change you made, i changed it to use Object.defineProperties because in our project we have custom extrinsic implementation. you can check `https://github.com/polkadot-js/api/pull/739` for more info.

jacogr:
Yes, Ian is right - we should not extend it, specifically for the reasons mentioned, ie. Custom extrinsic implementations. (We are Not just catering for Polkadot and simple substrate chains here)

ianhe8x:
I did another version(#1083 ), feedback welcomes.

ianhe8x:
this DynamicExtrinsicSubmittable ideally  only need to be created once, maybe we can store it in base as a protected member and reuse

ianhe8x:
i just think of some cases it will be handy if i have access to it, so i can do some thing with its prototype

axelchalon:
actually, what prevents us from registering DynamicExtrinsicSubmittable inside the TypeRegistry? i.e. having `SubmittableExtrinsic` in `@polkadot/types` just like `Extrinsic`; only downside I see is that if we register a custom extrinsic, we also have to register a custom submittableextrinsic. is this it?

ianhe8x:
Only problem is there would be two SubmittableExtrisic, one for PromiseApi the other for Rx

axelchalon:
> Yes, Ian is right - we should not extend it, specifically for the reasons mentioned, ie. Custom extrinsic implementations. (We are Not just catering for Polkadot and simple substrate chains here)

gotcha, but extending the dynamically fetched Extrinsic class from the type registry (like now this in PR) should be ok, or?

axelchalon:
cc @jacogr does the code of this PR look good and do we want to merge it? if so I'll merge from master and fix the conflicts. else we can close the PR

jacogr:
So, things have changed quite a bit... which actually means what you set out to do initially is now actually possible.

Look at `api-derive/type/HeaderExtended`, it extends via `ClassOf('Header')` which pulls the Header from the registry.

PR #1077, opened 2019-07-07
Auto-generate TS definitions from types

jacogr:
With metadata v8 (plus or minus a generation or two), the reflection will be available in the actual metadata, basically removing the need to define any types explicitly (apart from: primitives, since these have a defined meaning, any types where there are special user displays, e.g. AccountId, and any types where custom encoding/decoding is used, e.g. Address/Vote)

This starts dipping the toe into making some types (only for parachains here), non-constructable via `new ClassName`, while creatable via `createType` - all while having definitions available for these types. This could apply to most/all the rpc/type definitions.

What it does -

- it uses the JSON format (single file), to define the types. Here we do lose some typing, however since the generated TS interface are checkable (and quickly show issues), we are actually not losing typing

- A generator (under no illusions that this is a POC, and could be re-done - it was done for readability, not optimality - and only covers the bases as used here) parses the definitions via the codec and generates a definition output

- The API injects the definitions upon startup

- The generated types are actually richer than what we normally have in the case of `Tuple`, since it has the actual index types as well. In the case of `Enum`, it adds the `isXyz/asXyz` getters as well, which previously was not available like we have in `Struct` getters.

This is already breaking, but non-impacting. i.e. if people used `MyCoolType: AttestedCandidate` in their definitions (TS), this won't work. Similarly, the imports for the types are now done on a different path. 

(Overall net this looks like it is adding code, however this is due to the generator and once-off enhancements to enums, i.e. generates `is*` and `as*` getters. It should be more compact overall in terms of number of lines, however much less in terms of available files to check, manage and import from)

Shortcuts -

- Generator - https://github.com/polkadot-js/api/pull/1077/files#diff-469a8ce707a5a8d33ea83df325ecb559R1
- Definitions - https://github.com/polkadot-js/api/pull/1077/files#diff-fa47946df5505559639b134bc41e08b6R1
- Generated output - https://github.com/polkadot-js/api/pull/1077/files#diff-c54a766d99474535df2f5c626b6b92c8R1


packages/types/src/Metadata/util/testUtil.ts
```diff
@@ -20,6 +28,8 @@ export function decodeLatestSubstrate<Modules extends Codec> (
   latestSubstrate: object
 ) {
   it('decodes latest substrate properly', () => {
+    injectDefinitions();
```
jacogr: Not convinced this is the proper way to do this before the type checking tests. However mimics API, without doing magic in createType to just "pull these in".

packages/types/src/codec/EnumType.ts
```diff
@@ -127,6 +127,30 @@ export default class Enum extends Base<Codec> implements Codec {
     return class extends Enum {
       constructor (value?: any, index?: number) {
         super(Types, value, index);
+
+        Object.keys(this._def).forEach((_key) => {
```
jacogr: `is*` and `as*` getters - basically in the same form as we have for the struct getters

packages/types/src/codec/createType.ts
```diff
@@ -262,6 +262,11 @@ export function createClass (type: Text | string): Constructor {
   );
 }
 
+// alias for createClass
+export function ClassOf (name: string): Constructor {
```
jacogr: Unused atm. However just an alias so you could do `MyCoolType: ClassOf('AttenstedCandidate')` - this is actually sorely needed, i.e. in the case of `Balance` we are currently stuck at `u128` with no real way to override this from an API-user POV.

packages/types/src/scripts/srmlTS.ts
```diff
@@ -0,0 +1,164 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import fs from 'fs';
```
jacogr: Ugly as sin, but follow-able.

packages/types/src/srml/parachains/definitions.ts
```diff
@@ -0,0 +1,94 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+export default {
```
jacogr: This is where we actually define the parachain types in a JSON-like format - and it uses the JSON injection path.

packages/types/src/srml/parachains/types.ts
```diff
@@ -0,0 +1,167 @@
+// Auto-generated, do not edit
```
jacogr: At least this gets sorted.

packages/types/src/srml/parachains/types.ts
```diff
@@ -0,0 +1,167 @@
...
+
+import { Enum, Struct, Tuple, Vector } from '../../codec';
+import { AccountId, Bytes, Hash, Signature, U32, u64 } from '../../primitive';
```
jacogr: Only allowed imports are from codec and primitives. (There needs to be a way to link across modules as well)

packages/types/src/srml/parachains/types.ts
```diff
@@ -0,0 +1,167 @@
...
+  readonly headData: HeadData;
+  readonly balanceUploads: Vector<BalanceUpload>;
+  readonly egressQueueRoots: Vector<EgressQueueRoot>;
+  readonly fees: u64;
+  readonly blockDataHash: Hash;
+}
+
+export interface CollatorSignature extends Signature {}
+
+export interface EgressQueueRoot extends Tuple {}
```
jacogr: These should really convey more information, i.e. `EgressQueueRoot = (ParaId, Hash)` (as defined), here none of the actual part info is available.

This is not valid (`An interface can only extend an identifier/qualified-name with optional type arguments.ts(2499)`) -

```js
export interface EgressQueueRoot extends [ParaId, Hash] implements Codec {}
```

However this work-around does do the trick -

```js
export type EgressQueueRootTuple = [ParaId, Hash];

export interface EgressQueueRoot extends EgressQueueRootTuple implements Codec {}
```

i.e. in the above you can do `[paraId, hash] = foo` (where foo is `EgressQueueRoot`) and `a` & `b` would have the `ParaId` and `Hash` types respectively.

packages/types/src/srml/parachains/types.ts
```diff
@@ -0,0 +1,167 @@
+// Auto-generated, do not edit
+
+import { Enum, Struct, Tuple, Vector } from '../../codec';
```
jacogr: Only allowed imports are from codec and primitives. (There needs to be a way to link across modules as well, not available as of yet, i.e. `AuthorityId` would be imported from e.g. `grandpa`)

packages/types/src/codec/EnumType.ts
```diff
@@ -127,6 +127,30 @@ export default class Enum extends Base<Codec> implements Codec {
...
+          const askey = `as${_key}`;
+          const iskey = `is${_key}`;
+
+          // do not clobber existing properties on the object
+          if (isUndefined((this as any)[iskey])) {
+            Object.defineProperty(this, iskey, {
+              enumerable: true,
+              get: () => this.type === _key
```
jacogr: Need to add an explicit test for this - I'm not convinced it is working as intended for basic enums. (Generally the `toNumber() === x` approach is used there)

packages/types/src/srml/parachains/definitions.ts
jacogr: The alternative would just be that the `getTypeDef` does not throw (if set as not-throwing), allowing at least the generation to work. However, this may create problems around the injection, i.e. the above will solve the ordering imports for the type extraction, but may not for actual use creation - to be looked at.

packages/types/src/codec/EnumType.ts
jacogr: And has tests added for these as well.

packages/types/src/primitive/index.ts
```diff
@@ -14,6 +14,9 @@ export { default as AccountId, AccountIdOf } from './AccountId';
 export { default as AccountIndex } from './AccountIndex';
 export { default as AccountInfo } from './AccountInfo';
 export { default as Address } from './Address';
+// Balance is technically not a primitive, however since it is used in a number of interfaces
```
jacogr: This is where the `ClassOf` definition above can come in handy - but that is a phase II. Only to be done once all simple types are moved to "non-constructable" and the primitives can be trimmed to be "only real primitives, some of these in here are questionable as well)

packages/types/src/srml/parachains/definitions.ts
jacogr: Wanted to log this explicitly, but not worth the effort when we are moving to reflected - depends on the timing, it will hit the contracts first.

jacogr:
Ok, so removing WIP, it is good enough as a first stab. Most probably won't cater for everything. (i.e. Compact will fail, but for the initial use was not needed). 

Some areas are actually vastly improved from current, i.e. with this Tuples are actually usable without magic casts. The  "JSON" definition overall is not too bad, since it is more compact and any omissions are tracked through the output (i.e. at compile-time).

Would really want to start boiling the frog slowly, i.e. getting devs to start importing "types" as opposed to "classes" - when swapping to reflected metadata, it should be as simple as removing the definitions and pointing the generator at the metadata from the node.

Next steps after this, initially, would be just to remove as much of the simple types as possible and organising into `srml/*` (anything that is a straight extend, no specific encoding/decoding, and hopefully not widely used, i.e. would _not_ move something like `Balance` in a rush but for the lesser-used stuff, cough, contracts, cough and any very recent stuff can be moved first)

packages/types/src/codec/createType.ts
amaurymartiny: Agreed, this is cool

packages/types/src/srml/parachains/types.ts
amaurymartiny: Why is this file committed?

The workflow I was thinking about was:
- the user runs smrlTypes.ts script, it outputs this file somehwere
- the user then needs to manually add this file to tsconfig's typeRoot

(and have a look if 0x and typechain does something more intelligent too)

jacogr: Agreed with your workflow.

However, would at least like to provide the default types for polkadot longer term. (Actually happy to add the other chain defaults as well, as long as I don't need to generate them)

As of now, don't want to boil the water too much - so there are still imports, just not classes. (At least for some)

amaurymartiny:
Another question:
```javascript
const b = new ClassOf('Balance'); // So b will be a BN
// Right now ClassOf() returns a Constructor<Codec>

b.addn(); // <- Will this have autocomplete?
```

The question I have is that, when I define new ClassOf(...), how will it know to use the typings defined in the generated file?

jacogr:
Ahhh, good one on ClassOf - we will need something like `<T>` there, so in the case of `Balance`, which I always get back to.

So should be able to do something like -

```js
import { Struct, ClassOf } from '@polkadot/types';
import { Balance } from '@polkadot/types/types'; // interface

class MyCoolThing extends Struct {
  constructor (value?: any) {
    super({
      MyValue: ClassOf<Balance>('Balance')
    }, value);
  }
}
```

Just thoughts, not quite there as of yet. For getter values it is ok, since `Balance` has all the info.

PR #1079, opened 2019-07-08
signAndSend with options & callback

jacogr:
@amaurymartiny Does this align with your new types - had the add the missing options + callback variant.

### DIFF
--- a/packages/api/src/SubmittableExtrinsic.ts
+++ b/packages/api/src/SubmittableExtrinsic.ts
@@ -76,11 +76,11 @@ export interface SubmittableExtrinsic<ApiType> extends IExtrinsic {
 
   sign (account: IKeyringPair, _options: Partial<SignatureOptions>): this;
 
-  signAndSend (account: IKeyringPair | string | AccountId | Address, options?: Partial<Partial<SignatureOptions>>): SumbitableResultResult<ApiType>;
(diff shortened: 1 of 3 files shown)


packages/api/src/SubmittableExtrinsic.ts
```diff
@@ -181,13 +184,13 @@ export default function createSubmittableExtrinsic<ApiType> (
         }
       },
       signAndSend: {
-        value: function (account: IKeyringPair | string | AccountId | Address, _options?: Partial<Partial<SignatureOptions>> | Callback<ISubmittableResult>, statusCb?: Callback<ISubmittableResult>): SumbitableResultResult<ApiType> | SumbitableResultSubscription<ApiType> {
+        value: function (account: IKeyringPair | string | AccountId | Address, optionsOrStatus?: Partial<Partial<SignatureOptions>> | Callback<ISubmittableResult>, statusCb?: Callback<ISubmittableResult>): SumbitableResultResult<ApiType> | SumbitableResultSubscription<ApiType> {
```
amaurymartiny: unless I'm missing something, `Partial<Partial<T>> === Partial<T>`

jacogr: I dunno why it is like that, copied from the line above. We were both blind letting that slip in. Me now ... twice...

PR #1080, opened 2019-07-08
Remove edgeware node from compat

jacogr:


jacogr:
Will be harmless now even for old, when moving to fallback it is not logged.

PR #1081, opened 2019-07-08
Swap to Mortal transactions by default

jacogr:
- signAndSend will retrieve the block (unless era or eraLength is specified)
- eraLength is new - setting this to -1 indicated immortal, otherwise it is the length in blocks (defaults to 10)
- I don't like the code that returns the nonce and signedBlock, so suggestions welcome


packages/api/src/SubmittableExtrinsic.ts
```diff
@@ -197,19 +197,37 @@ export default function createSubmittableExtrinsic<ApiType> (
...
           return decorateMethod(
             () => ((
-              isUndefined(options.nonce)
-                ? api.query.system.accountNonce<Index>(address)
-                : of(new Index(options.nonce))
+              combineLatest([
+                isUndefined(options.nonce)
+                  ? api.query.system.accountNonce<Index>(address)
+                  : of(new Index(options.nonce)),
+                isUndefined(options.era) && options.eraLength !== -1
```
amaurymartiny: ||

packages/api/src/SubmittableExtrinsic.ts
```diff
@@ -197,19 +197,37 @@ export default function createSubmittableExtrinsic<ApiType> (
...
+                  : of(null)
+              ])
             ).pipe(
               first(),
-              mergeMap(async (nonce) => {
+              mergeMap(async ([nonce, signedBlock]) => {
+                let blockHash: Hash | undefined;
+                let era: ExtrinsicEra | undefined;
+
+                if (signedBlock) {
```
amaurymartiny: In which case would signedBlock be falsy?

jacogr: when either -

- era is set, or
- eraLength is set to something that is not -1

jacogr: Added an explicit comment around the of as to why

amaurymartiny: So what I dislike about this, it's that
1/ this -1 has hidden meaning, i.e. you need to read docs to know about it
2/ the api user won't know what happens at a glance when I do
`signAndSend(..., { eraLength: -1, era: new Era('something mortal')}) // which one takes precedence?`
or even 
`signAndSend(..., { eraLength: 10, era: new Era({ ..., period: 20 })}) // which one takes precedence?`

I would remove eraLength, and change type of era to e.g.
```
type era = 'Immortal' | Era | AnyNumber;
```

So that 
```
signAndSend(..., { era: 'Immortal' });
signAndSend(..., { era: new Era('the immortal enum') });
signAndSend(..., { era: new Era('some mortal era') });
signAndSend(..., { era: 10 }); // should be clear that here the 10 means era length, esp if the user passes blockHash too
```

jacogr: Cool, yes, that is sane - I didn't like the `eraLength` as well as `era` definitions. (And at the same time didn't want to remove `era` completely)

Adjusted.

jacogr:
Blocked on https://github.com/polkadot-js/api/pull/1082 - sadly cannot actually do e2e tests here at all since all the indexes are off and it may or may not be operational.

packages/api/src/SubmittableExtrinsic.ts
```diff
@@ -76,11 +85,11 @@ export interface SubmittableExtrinsic<ApiType> extends IExtrinsic {
 
   sign (account: IKeyringPair, _options: Partial<SignatureOptions>): this;
 
-  signAndSend (account: IKeyringPair | string | AccountId | Address, options?: Partial<SignatureOptions>): SumbitableResultResult<ApiType>;
+  signAndSend (account: IKeyringPair | string | AccountId | Address, options?: Partial<SignerOptions>): SumbitableResultResult<ApiType>;
```
amaurymartiny: Two lines above it's still `SignatureOptions`, I guess SignerOptions is more flexible there too

docs/examples/promise/07_make_transfer_with_allowed_block_permissions_only/index.js
```diff
@@ -22,22 +22,21 @@ async function main () {
...
+  // construct an mortal era
+  const era = new ExtrinsicEra({ current: currentHeight, period: 10 });
 
-  // get the beggining of the current era
-  const eraBirth = exERA.asMortalEra.birth(currentHeight.toNumber());
-
-  // get the hash of the beginning of the current era
-  const eraHash = await api.rpc.chain.getBlockHash(eraBirth);
+  era.asMortalEra.birth(currentHeight);
+  era.asMortalEra.death(currentHeight);
```
amaurymartiny: Should we console.log these two? They do nothing right now

packages/api/src/SubmittableExtrinsic.ts
jacogr: Yes, cannot quite change the sign signature, which is a muck-up - since that sits in `IExtrinsic`. Had it changed and then it complains bitterly.

amaurymartiny: Ah i see what you mean, ok

PR #1083, opened 2019-07-09
refactor: use typeRegistry.extends to create SubmittableExtrinsic

ianhe8x:
for #1075 
In this PR, typeRegistry.extends() is introduced to extends type in the typeRegistry.

It works except some places I have to use `as any` to silence tsc.

This is a poc, not ready for merge.

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -4,7 +4,7 @@
 
 import { ProviderInterface } from '@polkadot/rpc-provider/types';
 import { Storage } from '@polkadot/storage/types';
-import { AnyFunction, Codec, CodecArg, RegistryTypes } from '@polkadot/types/types';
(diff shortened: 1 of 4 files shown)


ianhe8x:
actually the `extends()` isn't really necessary.

jacogr:
I like the idea, and this type of approach is what I had in the back of my mind, not only for here, but elsewhere as well.

However, I would prefer not not starting using `getRegistry().extends(...)` everywhere, which in reality is more-or-less where we will end up with the reflected metadata.

This is an alternative approach to the same, https://github.com/polkadot-js/api/pull/1077/files#diff-a86daab9b4c7e3dcd4eb7d91f8feb064R267 - the only difference is that the `getRegistry()` itself sits inside. (the extends as we have is nicer in terms of returning the generic type, so I will certainly "steal" that for the above)

PR #1084, opened 2019-07-09
Include api.consts in api.derive methods

Stefie:
Follow-up of 
https://github.com/polkadot-js/api/pull/1066 
and
https://github.com/polkadot-js/api/pull/1068


packages/api-derive/src/balances/fees.ts
```diff
@@ -22,25 +20,30 @@ const ZERO = new BN(0);
  *
  * ```javascript
  * api.derive.balances.fees(([creationFee, transferFee]) => {
- *   console.log(`The fee for creating a new account on this chain is ${creationFee} units. The fee required for making a transfer is ${transferFee} units.`);
+ *   // console.log(`The fee for creating a new account on this chain is ${creationFee} units. The fee required for making a transfer is ${transferFee} units.`);
```
jacogr: Why is this commented?

packages/api-derive/src/contracts/fees.ts
```diff
@@ -6,43 +6,114 @@ import { ApiInterface$Rx } from '@polkadot/api/types';
...
- * Retrieves all the contract fees
+ * @name fees
+ * @returns An object containing the combined results of the queries for
+ * all relevant contract fees as declared in the substrate chain spec.
+ * @example
+ * <BR>
+ *
+ * ```javascript
+ * api.derive.contracts.fees(([creationFee, transferFee]) => {
+ *   // console.log(`The fee for creating a new contract on this chain is ${creationFee} units. The fee required to call this contract is ${transferFee} units.`);
```
jacogr: As above, the same commented question.

packages/api-derive/src/contracts/fees.ts
```diff
@@ -6,43 +6,114 @@ import { ApiInterface$Rx } from '@polkadot/api/types';
...
-        rentDepositOffset,
-        transactionBaseFee,
-        transactionByteFee,
-        transferFee,
-        tombstoneDeposit
-      } as DerivedContractFees)),
+    if (api.query.contract && !api.query.contract.rentByteFee) {
+      // Only for 1.0 support. rentByteFee, rentDepositOffset, tombstoneDeposit are not available in substrate 1.0.
+      // @TODO remove this once 1.0 support is dropped
+      const ZERO = new BN(0);
```
jacogr: Move this const to file-level rather.

packages/api-derive/src/balances/fees.ts
```diff
@@ -22,25 +20,30 @@ const ZERO = new BN(0);
...
-        ? api.queryMulti([
-          api.query.balances.creationFee,
-          api.query.balances.existentialDeposit,
-          api.query.balances.transactionBaseFee,
-          api.query.balances.transactionByteFee,
-          api.query.balances.transferFee
-        ]) as any as Observable<[BN, BN, BN, BN, BN]>
-      : of([ZERO, ZERO, ZERO, ZERO, ZERO])
-    ).pipe(
+    return (api.consts.balances
```
jacogr: (nit-y) comment - this would read better as

```js
return (
  api.consts.balances
    ? of([ ... ])
    : api.queryMulti([...])
)
```

packages/api-derive/src/contracts/fees.ts
```diff
@@ -6,43 +6,114 @@ import { ApiInterface$Rx } from '@polkadot/api/types';
...
+          queryBase.callBaseFee,
+          queryBase.contractFee,
+          queryBase.createBaseFee,
+          queryBase.creationFee,
+          queryBase.transactionBaseFee,
+          queryBase.transactionByteFee,
+          queryBase.transferFee
+        ])
+      ]) as any as Observable<[[BN, BN, BN], [BN, BN, BN, BN, BN, BN, BN]]>).pipe(
+        map(([[
```
jacogr: This is a bit difficult to follow, would suggest (if you really want to spread the params into multiple lines), as

```js
map([
  [
    ...
  ],
  [
    ...
  ]
```

packages/api-derive/src/contracts/fees.ts
```diff
@@ -6,43 +6,114 @@ import { ApiInterface$Rx } from '@polkadot/api/types';
...
+            rentByteFee,
+            rentDepositOffset,
+            tombstoneDeposit
+          ],[
```
jacogr: If spread over multi lines, don't `], [`, rather `],\n[`

packages/api-derive/src/contracts/fees.ts
```diff
@@ -6,43 +6,114 @@ import { ApiInterface$Rx } from '@polkadot/api/types';
...
+            tombstoneDeposit,
+            transactionBaseFee,
+            transactionByteFee,
+            transferFee
+          } as DerivedContractFees)),
+        drr()
+      );
+    }
+
+    return (api.consts.contracts
```
jacogr: Same as the readability ni comment above, i.e.

```js
return (
  api.consts.contracts
    ? of([..])
    : api.queryMulti([...])
)
```

packages/api-derive/src/session/info.ts
```diff
@@ -81,13 +82,16 @@ export function info (api: ApiInterface$Rx) {
...
+        map(createDerived_pre_94),
         drr()
       )
     : (combineLatest([
-      // sessionsPerEra, hardcoded, due to https://github.com/paritytech/substrate/pull/2802/files#diff-5e5e1c3aec9ddfde0a9054d062ab3db9R156
-      of(new BN(6)),
+      // sessionsPerEra:
+      // substrate spec_version >= 94 : get from parameter_types exposed as api.consts
+      // https://github.com/paritytech/substrate/pull/2802/files#diff-5e5e1c3aec9ddfde0a9054d062ab3db9R156
+      // @TODO Switch to `api.constants` after https://github.com/polkadot-js/api/pull/1066 is merged.
```
jacogr: TODO can be removed now?

packages/api-derive/src/balances/fees.ts
Stefie: Oh, a leftover from development. I commented all console.logs to do less scrolling when comparing the results of different nodes. So an oversight.

packages/api-derive/src/contracts/fees.ts
Stefie: I changed it to use the same formatting for the type declarations and the parameters

packages/api-derive/src/contracts/fees.ts
Stefie: changed

packages/api-derive/src/contracts/fees.ts
Stefie: Actually, what about:
```
map(([
  [...],
  [...]
]) => ({
```

packages/api-derive/src/balances/all.ts
```diff
@@ -32,8 +33,8 @@ function calcBalances ([accountId = EMPTY_ACCOUNT, bestNumber = ZERO, [freeBalan
 
   // offset = balance locked at genesis, perBlock is the unlock amount
   const { offset, perBlock } = vesting.unwrapOr(new VestingSchedule());
-  const vestedNow = perBlock.mul(bestNumber);
-  const vestedBalance = vestedNow.gt(offset)
+  const vestedNow: BN = perBlock.mul(bestNumber);
```
amaurymartiny: Can be for another PR, but I'm wondering why in api-derive are we putting BNs everywhere? surely BlockNumber and Balance are more semantically correct

packages/api-derive/src/balances/fees.ts
jacogr: Phew, was really wondering about that one :)

packages/api-derive/src/contracts/fees.ts
jacogr: I prefer your second version.

Stefie: That's commited already :)

packages/api-derive/src/balances/all.ts
Stefie: Good point! But I'd also leave that for another PR since it means touching all files & tests

Stefie:
Anything holding back here? @jacogr @amaurymartiny ?

packages/api-derive/src/balances/fees.ts
```diff
@@ -29,19 +27,25 @@ const ZERO = new BN(0);
 export function fees (api: ApiInterface$Rx) {
   return (): Observable<DerivedFees> => {
     return (
-      // HACK This is a very quick fix to actually get the types in to continue with
-      // other work - so as of now, the fees are completely broken
-      api.query.balances.creationFee
-        ? api.queryMulti([
+      api.consts.balances
```
amaurymartiny: I think this will throw on older nodes `api.consts is undefined`

packages/api-derive/src/session/info.ts
```diff
@@ -13,13 +13,13 @@ import { BlockNumber, Option } from '@polkadot/types';
...
 
-type Result0_94 = [BN, [BN, Option<BlockNumber>, BN, BN, BN]];
+type Result_pre_94 = [BN, [BN, Option<BlockNumber>, BN, BN, BN]];
 type Result = [BN, [BN, BN]];
 
 const ZERO = new BN(0);
 
 // internal helper to just split the logic - take all inputs, do the calculations and combine
-function createDerived0_94 ([bestNumber, [currentIndex, _lastLengthChange, sessionLength, lastEraLengthChange, sessionsPerEra]]: Result0_94): DerivedSessionInfo {
+function createDerived_pre_94 ([bestNumber, [currentIndex, _lastLengthChange, sessionLength, lastEraLengthChange, sessionsPerEra]]: Result_pre_94): DerivedSessionInfo {
```
amaurymartiny: Let's have the convention everywhere `someThingWhatever_{start_spec}_{end_spec}`, e.g. we already have EventRecord_0_76, here it'd be `createDerived_0_94`

packages/api-derive/src/contracts/fees.ts
```diff
@@ -6,43 +6,107 @@ import { ApiInterface$Rx } from '@polkadot/api/types';
...
+          transactionBaseFee,
+          transactionByteFee,
+          transferFee
+        } as DerivedContractFees)),
+        drr()
+      );
+    }
+
+    return (
+      api.consts.contracts
```
amaurymartiny: same, possibily undefined

amaurymartiny:
ahah, added some comment 3s late

packages/api-derive/src/balances/fees.ts
amaurymartiny: or not, actually

packages/api-derive/src/session/info.ts
jacogr: Sorry :(

Stefie: Ah, that was the `0`! 💡

packages/api-derive/src/balances/fees.ts
Stefie: I tested them with matching nodes for the conditions, so I hope they don't. :)
But you're right.... they should. Even looking at this gives me a "Cannot get property x of undefined"-error.

packages/api-derive/src/session/info.ts
Stefie: https://github.com/polkadot-js/api/pull/1104

PR #1089, opened 2019-07-09
Enforce import oder with tslint

axelchalon:
We have some style guidelines we use, namely: modules imported alphabetically, exports of a module imported alphabetically, absolute imports before relative imports. However:
- Those rules aren't documented anywhere
- They are not integrated at CI level, so people doing PRs cannot know that they didn't respect the style guide
- Because of this, it makes the PR process unnecessary longer (reviewer asks for style changes)
- Makes contributors to the codebase in general lose time sorting imports according to our guidelines. This can be particularly time-consuming when renaming e.g. a package and then searching & replace where it's used: if we want to respect our styleguide, we need to sort the imports again in every file.

As such, this PR proposes:
- Translating our imports to a set of rules in `tslint.json`, such that the linting check will fail if imports aren't sorted properly.
- Enforcing them automatically (autofix) when running `yarn lint` (unless in CI), so that people don't have to bother sorting manually anymore
- Fix the import order of all files

Questions:
- I'm not sure if we want the tslint rules in this repo, or in [`@polkadot/dev`](https://github.com/polkadot-js/dev/blob/master/packages/dev/config/tslint.js) ?
- On another note, is anyone using the coverage report in `yarn test`? It makes me have to scroll up each time (even when running just a single test). But if somebody uses it we can keep it.


jacogr:
Yes, coverage report gets used... quite a bit by me actually. Happy with a a specific coverage option, provided -

- it runs on CI so the information is submitted
- doesn't break the dev scripts

All rules are in dev, never the specific repos - it applies to all. However, since we will be moving to eslint, would not put any effort into tslint at all (Eslint is already in dev, just not gotten around to start enabling, common first, then up the tree, to see what pops out of the woodwork).

packages/api-contract/src/method.ts
```diff
@@ -2,12 +2,12 @@
...
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { CodecArg, Constructor } from '@polkadot/types/types';
-import { ContractABIFn, ContractABIFn$Arg, ContractABIMethod, ContractABIMethodBase, ContractABITypes } from './types';
-
 import { Compact, createClass } from '@polkadot/types';
+import { CodecArg, Constructor } from '@polkadot/types/types';
 import { assert, isString, isUndefined, stringCamelCase } from '@polkadot/util';
 
+import { ContractABIFn, ContractABIFn$Arg, ContractABIMethod, ContractABIMethodBase, ContractABITypes } from './types';
```
jacogr: This is actually wrong, which is why we have not enabled any rules.

- type imports first
- external imports, i.e. bn.js
- @polkadot imports
- local imports

packages/api-derive/src/accounts/idAndIndex.ts
```diff
@@ -2,16 +2,16 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { Observable, of } from 'rxjs';
-import { map, startWith } from 'rxjs/operators';
 import { ApiInterface$Rx } from '@polkadot/api/types';
 import { AccountId, AccountIndex, Address } from '@polkadot/types';
 import { isU8a } from '@polkadot/util';
 import { decodeAddress } from '@polkadot/util-crypto';
+import { Observable, of } from 'rxjs';
```
jacogr: As per the above comment, we always have the non-polkadot externals first... which is why it hasn't been added

jacogr:
The other issue is, it is great for this repo - but all the polkadot-js repos follow exactly the same rules, so there are no surprises. So one repo cannot decide to go in a different way unless there is actually a concerted effort to apply to all.

As such, here it is just more effort than worth to be honest - we can certainly document better.

axelchalon:
Closing as we're switching to eslint soon.

jacogr:
Soon is relative :) It is a lot of work accross a lot of repos. (Should actually do a ui-related one, i.e. extension is nice and small, before the api itself as well - to test everything works. Hence doing common first, it is lowest in the tree of dependencies)

jacogr:
Addition to the above - the reason for swapping back to eslint (originally was eslint, then tslint), is because that is what the TS team themselves are swapping to as well. (And the tslint maintainers are actually on-board with that move as well, so tslint is going to be deprecated)

Issue #1063, opened 2019-06-29
Support optional extrinsic “tip” enum


jacogr:
As implemented in https://github.com/paritytech/substrate/pull/2930

PR #1090, opened 2019-07-09
Extrinsics with version decoding handler

jacogr:
Supporting infrastructure for https://github.com/polkadot-js/api/issues/1063

Previously, we didn't pay any attention to the actual version identifier in `Extrinsic`. This paves the way for supporting multiple versions, as required in the linked issue -

- it decodes the version byte, extracting the info
- based on the version (signing flag removed), it constructs an Extrinsic (fields may be incompatible)

Effectively it uses a simmilar approach to `Address`, where the first byte indicates the type of decoding used and then just wraps a `raw` container with the actual data. (And the raw container decodes by it's own rules)

It is a bit more involved than Address in that the `IExtrinsic` interface exposes a lot more getters, so the base extrinsic does a number of proxy calls to the top.

### DIFF
--- a/packages/types/src/type/Extrinsic.ts
+++ b/packages/types/src/type/Extrinsic.ts
@@ -4,21 +4,20 @@
 
 import { AnyNumber, AnyU8a, ArgsDef, Codec, IExtrinsic, IKeyringPair, SignatureOptions } from '../types';
 
-import { assert, isHex, isU8a, u8aToHex, u8aToU8a } from '@polkadot/util';
-import { blake2AsU8a } from '@polkadot/util-crypto';
+import { assert, isHex, isU8a, u8aToU8a } from '@polkadot/util';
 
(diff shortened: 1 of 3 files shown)


packages/types/src/type/Extrinsic.ts
```diff
@@ -4,21 +4,20 @@
...
 import ExtrinsicSignature from './ExtrinsicSignature';
+import ExtrinsicV1, { ExtrinsicValueV1 } from './ExtrinsicV1';
 
-type ExtrinsicValue = {
-  method?: Method
-  signature?: ExtrinsicSignature
-};
+type ExtrinsicValue = ExtrinsicValueV1;
+
+const UNMASK_VERSION = 0b01111111;
```
amaurymartiny: Where is this defined in the rust code?

PR #1091, opened 2019-07-09
Fix staking.info

### LINKED ISSUE #1060: props for derive.staking.info not updated
An issue appeared in `apps` and is related to `derive.staking.info`. Updates are not pushed through the props, so that the info don't update unless the page is reloaded.

I've tracked down the issue and it happens since v0.82 https://github.com/polkadot-js/api/pull/987

How to reproduce:
- navigate to staking>actions
- change the reward destination
- once the tx succeeds, the reward destination isn't updated
- refresh the page or change tab and get back -> the destination is updated.

### LINKED ISSUE #1092: api.queryMulti should cache values
```javascript
api.queryMulti([
  [api.query.session.nextKeyFor, controllerId],
  [api.query.staking.ledger, controllerId],
], callback)
```

First result: [something1, something2],

**Expected:** 

second result: [something1, something3]

**Actual:**

second result: [null, something3]

amaurymartiny:
fixes #1060 
fixes #1092

### DIFF
--- a/packages/api-derive/src/staking/info.ts
+++ b/packages/api-derive/src/staking/info.ts
@@ -99,7 +99,7 @@ function withStashController (api: ApiInterface$Rx, accountId: AccountId, contro
     map(([eraLength, bestNumber, [nextKeyFor, _stakingLedger, _nominators, rewardDestination, stakers, _validatorPrefs]]) => {
       const nominators = _nominators ? _nominators[0] : undefined;
       const validatorPrefs = _validatorPrefs ? _validatorPrefs[0] : undefined;
-      const stakingLedger = _stakingLedger.unwrapOr(undefined) || undefined;
+      const stakingLedger = _stakingLedger ? _stakingLedger.unwrapOr(undefined) : undefined;
 
       return {
         accountId,
--- a/packages/api-derive/src/staking/info.ts
+++ b/packages/api-derive/src/staking/info.ts
@@ -7,7 +7,7 @@ import { DerivedStaking, DerivedUnlocking } from '../types';
 
 import BN from 'bn.js';
 import { combineLatest, Observable, of } from 'rxjs';
-import { map, switchMap } from 'rxjs/operators';
+import { map, switchMap, tap } from 'rxjs/operators';
 import { AccountId, BlockNumber, Exposure, Keys, Option, RewardDestination, SessionKey, SessionKeys, StakingLedger, ValidatorPrefs, UnlockChunk } from '@polkadot/types';
 
 import { isUndefined } from '@polkadot/util';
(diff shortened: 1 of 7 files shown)


packages/api-derive/src/staking/info.ts
```diff
@@ -83,21 +83,23 @@ function withStashController (api: ApiInterface$Rx, accountId: AccountId, contro
...
+        [api.query.staking.validators, stashId]
+      ])
+    ]) as any as Observable<[BN, BlockNumber, [Option<Keys | SessionKey>, Option<StakingLedger>, [Array<AccountId>], RewardDestination, Exposure, [ValidatorPrefs]]]>
   ).pipe(
-    map(([eraLength, bestNumber, [nextKeyFor, _stakingLedger, [nominators], rewardDestination, stakers, [validatorPrefs]]]) => {
-      const stakingLedger = _stakingLedger.unwrapOr(null) || undefined;
+    map(([eraLength, bestNumber, [nextKeyFor, _stakingLedger, _nominators, rewardDestination, stakers, _validatorPrefs]]) => {
+      const nominators = _nominators ? _nominators[0] : undefined;
+      const validatorPrefs = _validatorPrefs ? _validatorPrefs[0] : undefined;
+      const stakingLedger = _stakingLedger ? _stakingLedger.unwrapOr(undefined) : undefined;
```
amaurymartiny: So the errors came from here, undefined & null...

packages/api-derive/src/util/drr.ts
```diff
@@ -12,6 +15,11 @@ import { Observable } from 'rxjs';
  */
 export const drr = () => <T>(source$: Observable<T>): Observable<T> =>
   source$.pipe(
+    catchError(err => {
+      l.error(err);
```
amaurymartiny: ... and were never logged inside the Observable loops. I added an explicit log here

packages/api-derive/test/e2e/promise.spec.ts
```diff
@@ -171,5 +167,45 @@ describe.skip('derive e2e', () => {
         done();
       });
     });
+
+    // FIXME This test doesn't pass
+    // https://github.com/polkadot-js/api/issues/1060
```
Stefie: Can we please not add it, then? Or just put it in an PR for fixing the issue mentioned in the comment?

I'm just working on making all e2e tests pass again at least in 1.0 and master.
I fixed all the derive ones in https://github.com/polkadot-js/api/pull/1084 and now I'm on the contract tests, the child storage tests and some other ones that were failing.

If we add tests, they should pass or the methods should be fixed if they're broken.

jacogr: Making that test pass is the whole point of this PR - so the comment will be removed as soon as it does.

Stefie: Ah, that wasn’t clear from the title, I thought it was primarily about logging/ error handling. Looking forward to it, then!

jacogr: The initial fix part was reverted when we realised we need more work and it is not quite the correct solution. (The test actually did pass on it, however it had some effects on API users which we are not keen on)

amaurymartiny:
OK frankenstein now passes

Stefie:
One of the tests actually fails on 1.0:
Screenshot 2019-07-10 at 18 45 27

jacogr:
@Stefie Ok, it really has been a long day for me, but I cannot quite seem how the hell that happens - so the failing test is Frankenstein line 171 onwards, but is shows a failure on the previous test line. Something is very weird.

PR #1099, opened 2019-07-10
eslint rules

jacogr:
- Same conversion as in https://github.com/polkadot-js/common/pull/435
- Checking each one individually, there has been some horrors with the auto-fix on these.
- there are some overrides that should be fixed - https://github.com/polkadot-js/api/pull/1099/files#diff-e4403a877d80de653400d88d85e4801aR7


jacogr:
Ok, cutting for now and leaving it exactly where I am to get back to at some point - some new issues introduced, a number of complaints fixed. Need to stop mucking about for now and get some real work done...

jacogr:
Removing WIP. This is around half-way, but a nightmare to keep current with merges. So did the following -

- made sure tslint still passes
- kept tslint as the default
- added `lint:new` as an eslint option

Checking each one individually, there has been some horrors with the auto-fix on these.

PR #1100, opened 2019-07-10
Add a docker compose file for e2e test environment

Stefie:
Preparation for https://github.com/polkadot-js/api/issues/908

But the containers can already be used for running the tests agains the different Polkadot/ Substrate versions.

It fires up 5 Docker containers with these ws endpoints:

- Substrate Master: ws://127.0.0.1:9945/
- Substrate 1.0: ws://127.0.0.1:9946/
- Substrate 2.0: ws://127.0.0.1:9947/
- Polkadot Master: ws://127.0.0.1:9948/
- Polkadot Alexander: ws://127.0.0.1:9949/


docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+  js-api-tests:
+    image: node:11
+    volumes: 
+      - ./:/polkadot-js/api
+    working_dir: /polkadot-js/api
+    network_mode: host
+    tty: true
+    environment: 
+      - TERM=xterm-256color`
+    command: bash -c "NODE_ENV=abc /usr/local/bin/yarn && /usr/local/bin/yarn test packages/"
```
amaurymartiny: test:all here

Stefie: `test:all` doesn't exist. Do you just mean "remove `packages/`" ?

amaurymartiny: https://github.com/polkadot-js/api/blob/master/package.json#L24

Stefie:
> We then should organize tests on which ones run on which nodes

Yeah, totally. We can talk about that tomorrow! This is just the setup to actually be able to run them once we reorganized them.

docker-compose.yml
Stefie: Hmm...
```
StefanieMBP:api stefaniedoll$ yarn run test:all
yarn run v1.16.0
error Command "test:all" not found.
```

Stefie: Ah, I see, Jaco just added it an hour ago. https://github.com/polkadot-js/api/commit/1a80bdef4be4777610a12945b7cb9fca5abb6ab6

docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+# Find more Docker images of Substrate https://hub.docker.com/r/parity/substrate/tags
+# Find more Docker images of Polkadot https://hub.docker.com/r/parity/polkadot/tags
+
+
+version: "3.7"
+services:
+  node-master:
+    image: parity/substrate:latest
+    ports:
+      - "9944:9944"
```
jacogr: Don't believe this should conflict with local at all, would rather adjust by 1 upwards - the test command can pass in the URL to connect to so that by default it connects to the local node for e2e and with a flag, it adjusts and points to the others exposed here.

Stefie: That was the other possible setup. I went for this so you don't need to run a local node at all anymore, but thinking about it again, this probably makes much more sense! People developing against a node that is not substrate master for development would still want access to a local node on `:9944` and also, substrate master is sometimes broken

Stefie: changed to `9945` and up

Stefie:
Can I merge this? I'd like to use it for my test branch

Issue #1039, opened 2019-06-19
One-shot storage queries should use `getStorage` instead of `subscribeStorage`


amaurymartiny:
In decorateStorageEntry:

https://github.com/polkadot-js/api/blob/aed73e2bb3d395d4210e193f996b19bcfab0ee91/packages/api/src/Base.ts#L585

so right now:
```javascript
const api = new ApiPromise();
await api.query.system.accountNonce(addr1); // uses subscribeStorage, with immediate unsubscribe after resolve
```

Expected: uses `state_getStorage`.

Edit: same for multi

jacogr:
Agreed with the `getStorage` for single-shot (we had it that way, I think with the Rx/Promise combination is was dropped for the sake of simplicity). However that doesn't cater for multiple keys, so in that case either -

- subscribe/unsubscribe (current)
- log a feature request with substrate

PR #1103, opened 2019-07-10
Refactor decorateMethod

axelchalon:
Internal refactor ; no api changes.

Previously:
- We passed a `decorateMethod` function to each `decorate*` function. Those `decorate*` functions were called twice, once with `rxDecorateMethod` (identity function) for decorating internally (`this._rx`) and once with `this.decorateMethod` for exposing (`this.{rpc,queryMulti,extrinsics,query}`). Those functions had to pass down `decorateMethod` to the other functions they would call.

Now:
- We don't pass `decorateMethod` anymore. Internally in Api we work with observables, so it makes sense that the `decorate*` functions work with observables (equivalent to being passed `rxDecorateMethod` before). We expose the api properties in a second stage, building them from `this._rx` using the concrete Api class's `expose{Rpc,Extrinsics,Multi,Storage}` functions, which take e.g. `this._rx.rpc` and return what should be exposed as `this.rpc`. (This would also allow overrides if needed, e.g. for https://github.com/polkadot-js/api/issues/1039.)

This avoids polluting the codebase with `decorateMethod` being passed down everywhere. Now we know that all the Api (internal) functions are using RxJs, and we only do the conversion / care about the ApiType/exposed properties in the `expose*` methods. I find it makes it easier to reason about the code.


jacogr:
Like the concept, going to leave the heavy-lifting to @amaurymartiny here, will only take a cursory look myself.

packages/api/src/promise/Api.ts
```diff
@@ -249,4 +249,73 @@ export default class ApiPromise extends ApiBase<'promise'> {
...
+  }
+
+  protected exposeExtrinsics (_extrinsics: SubmittableExtrinsics<'rxjs'>): SubmittableExtrinsics<'promise'> {
+    const creator = this.exposeMethod(_extrinsics);
+
+    return Object.keys(_extrinsics).reduce((result, sectionName) => {
+      const section = _extrinsics[sectionName];
+
+      result[sectionName] = Object.keys(section).reduce((result, methodName) => {
+        result[methodName] = this.exposeMethod(section[methodName]) as any;
```
jacogr: Not crazy at all about this list of stuff to do. This is where we are going to drop the ball. Same with storage, where it is even worse.

packages/api/src/promise/Api.ts
```diff
@@ -249,4 +249,73 @@ export default class ApiPromise extends ApiBase<'promise'> {
...
+        result[methodName].at = this.exposeMethod(section[methodName].at);
+        result[methodName].hash = this.exposeMethod(section[methodName].hash);
+        result[methodName].key = section[methodName].key;
+        result[methodName].multi = this.exposeMethod(section[methodName].multi);
+        result[methodName].size = this.exposeMethod(section[methodName].size);
+        result[methodName].creator = section[methodName].creator;
+        (result[methodName] as any).meta = (section[methodName] as any).meta;
```
jacogr: We `as any` a lot in here and above, it is a danger sign that we don't actually know what we have and are working with.

axelchalon:
I think with the latest commit it's reasonably clean (i.e. no more casting to `any` and less repetition; also noticed that decorateMethod used to return StorageEntryPromiseOverloads which is not true e.g. for decorating extrinsics afaik).

> the whole reason of doing decorateMethod was to avoid the duplication of nested double Object.keys().reduce() in all expose* functions, which we need to once in Base, once in ApiPromise.

I hear you. However, another argument is that currently in `master` for ApiRx, Base calls `this.decorate{Rpc,Multi,Extrinsics,Storage}` once for internal `._rx`, and once for the `ApiType`. If the `ApiType` is `rxjs`, then we're iterating over all the sections and creating the entries twice. With `expose*` we do 2x less computing (since it just returns `this._rx.extrinsics` rather than re-compute it) for `ApiRx`.

amaurymartiny:
> Now: We don't pass decorateMethod anymore. Internally in Api we work with observables, so it makes sense that the decorate* functions work with observables (equivalent to being passed rxDecorateMethod before). We expose the api properties in a second stage, building them from this._rx using the concrete Api class's expose{Rpc,Extrinsics,Multi,Storage} functions, which take e.g. this._rx.rpc and return what should be exposed as this.rpc. (This would also allow overrides if needed, e.g. for #1039.)

I still don't understand how this is simpler. With `decorateMethod`, there's only one place where you do the conversion `Observable->{Promise,rxjs,whatever}`, and you're done. Now it seems that there are 4 `expose*` functions that you need to handle.

So overall I think this version is harder to read than only 1 `decorateMethod`. I 100% agree with the computing though. If e2e tests pass, then i'm okay to merge this.

@jacogr I'll let you decide on this one. My opinion: it adds reading complexity, 60 lines for same func, and would prefer to focus on other more pressing issues.

jacogr:
Will take a proper look tonight.

axelchalon:
Obviously don't want to waste your time reviewing perceived low-added-value PRs. I don't mind closing it, and will refrain from spontaneous refactors for a while :stuck_out_tongue:

jacogr:
I've looked at this in-depth now, it is certainly cleaned up from the previous versions. The main benefits here seem to be only the 2x processing - which is reality is not that much of an impact timing wise, but certainly good to have.

This does seem more complex to me overall since we have circled back to doing multiple specific things for each type of API. In areas like in Promise where we need to explicitly track what is happening on the base (looking at storage = switch), has me concerned. Instead of just extending the storage, now there are multiple places to tweak. (In nothing else, the storage decoration here is something that I would not merge in as-is)

If the only actual benefit this provides is not double-decorating extrinsics/storage twice - in the case of the RxJs API, there must be another solution.

axelchalon:
:+1: Let's leave things as they are for now

PR #1106, opened 2019-07-11
Export storage prefix on StorageEntry

jacogr:
- Part of https://github.com/polkadot-js/apps/issues/1375 (needs appropriate UI changes - well, probably not, section will now just refer to the correct section, so once updated should just work)
- Basically, since we swapped from prefix -> section, the prefix was still sent as the section name. Now it has the -
  - explicit prefix as used (more as info than anything else, but never say never)
  - the section (as part of the map, existing - swapped from prefix to real name as exposed)
  - the method (as part of the map, existing, as it was)

### DIFF
--- a/packages/api-metadata/src/storage/fromMetadata/createFunction.ts
+++ b/packages/api-metadata/src/storage/fromMetadata/createFunction.ts
@@ -2,7 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { createType, Bytes, Compact, StorageKey, Text, U8a } from '@polkadot/types';
+import { createType, Bytes, Compact, StorageKey, U8a } from '@polkadot/types';
 import { PlainType, StorageEntryMetadata, StorageEntryModifier, StorageEntryType } from '@polkadot/types/Metadata/v6/Storage';
 import { StorageEntry } from '@polkadot/types/primitive/StorageKey';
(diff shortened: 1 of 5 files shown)


jacogr:
@amaurymartiny Would like your eyes on this as well.

packages/api-metadata/src/storage/fromMetadata/createFunction.ts
```diff
@@ -24,10 +31,10 @@ export interface CreateItemOptions {
  * are not known at runtime (from state_getMetadata), they need to be supplied
  * by us manually at compile time.
  */
-export default function createFunction (section: Text | string, method: Text | string, meta: StorageEntryMetadata, options: CreateItemOptions = {}): StorageEntry {
+export default function createFunction ({ meta, method, prefix, section }: CreateItemFn, options: CreateItemOptions = {}): StorageEntry {
```
jacogr: Changed the interface since 5 params is going stoopid - however, it should not be used externally, so not adding it to changelog, the prefix is already there anyway.

packages/api-metadata/src/storage/fromMetadata/createFunction.ts
```diff
@@ -24,10 +31,10 @@ export interface CreateItemOptions {
...
   const stringKey = options.key
     ? options.key
-    : `${section.toString()} ${method.toString()}`;
+    : `${prefix} ${method}`;
```
amaurymartiny: Ah yeah, this is even a bug fix

PR #1108, opened 2019-07-12
Bump metadata

jacogr:
- Current metadata with the `aura` module
- `::Proposal` -> `Proposal`
- `Phantom` -> `Null`
- Update call indexes


package.json
```diff
@@ -22,6 +22,7 @@
     "postinstall": "polkadot-dev-yarn-only",
     "test": "jest --coverage --testPathIgnorePatterns e2e",
     "test:all": "jest",
+    "test:one": "jest",
```
jacogr: `jest test singlefile.spec.ts` does not work anymore, well, this is probably better anyway since it doesn't add the `--coverage` stuff, which, when running single, you don't care about.

amaurymartiny: There's some trick by using something like
```
"test": "jest --coverage --testPathIgnorePatterns e2e --"
```

then single files still work

PR #1114, opened 2019-07-12
Expose blockNumber to external signers (useful in eras)

jacogr:
Signers can now display the birth/death


packages/api/src/SubmittableExtrinsic.ts
```diff
@@ -254,6 +255,7 @@ export default function createSubmittableExtrinsic<ApiType> (
                 } else if (api.signer) {
                   updateId = await api.signer.sign(_extrinsic, address, {
                     ...expandOptions({ ...options, ...eraOptions, nonce }),
+                    blockNumber: header ? header.blockNumber : new BN(0),
```
amaurymartiny: Why BN here? I'd put BlockNumber

jacogr: I actually want this whole interface to become JSON-ish, strings, primitives. So it is just a pure data transfer and there is a break between "out app" and the "external app" - external can do whatever he wants with the data.

So basically - 

- blockHash/genesisHash - hex
- method - hex
- era - hex (JSON enum)

Hence I opted for the "more primitive version", so we don't start relying on Codec.

Not sure if it is a great idea atm.

PR #1117, opened 2019-07-12
Add describeE2E

amaurymartiny:
Part of #908 

Quickly discussed with Stefie about it, here's a proposal to have better control on which tests should run where.
```diff
- describe('this test', () => {});
+ describeE2E(options)('this test', () => {});
```

where options describe on which nodes we want to run each describe().

Note: also added a `WITH_DOCKER` flag:
- `WITH_DOCKER yarn test:all` runs the tests everywhere according to the `describeE2E` options, expecting that docker-compose is running
- `yarn test:all` only the tests that work on `'local'`

Notes:
- no nested `describeE2E` allowed. Possible doable, but too much code and not worth it, just don't nest them
- no more `.skip`, and let's avoid them altogether

### DIFF
--- a/packages/api/test/util/describeE2E.ts
+++ b/packages/api/test/util/describeE2E.ts
@@ -1,0 +1,72 @@
+// Copyright 2017-2019 @polkadot/api authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { WsProvider } from '@polkadot/rpc-provider';
+
+import { ApiPromise, ApiRx } from '../../src';
+
+const WS_ENDPOINTS = {
+  'local': 'ws://127.0.0.1:9944/',
+  'substrate-master': 'ws://127.0.0.1:9945/',
+  'substrate-1.0': 'ws://127.0.0.1:9946/',
+  'substrate-2.0': 'ws://127.0.0.1:9947/',
+  'polkadot-master': 'ws://127.0.0.1:9948/',
+  'polkadot-alexander': 'ws://127.0.0.1:9949/',
+  'remote-polkadot-alexander': 'wss://poc3-rpc.polkadot.io/',
+  'remote-substrate-1.0': 'wss://substrate-rpc.parity.io/'
+};
+
+type WsName = keyof typeof WS_ENDPOINTS;
+
+interface Options {
+  except?: WsName[]; // Only one of the two keys `either` `only` should be set
+  only?: WsName[];
+  apiType?: 'promise' | 'rxjs';
+}
+
+export default function describeE2E (options?: Options) {
+  const apiType = options && options.apiType === 'rxjs' ? 'rxjs' : 'promise';
+  const Api = options && options.apiType === 'rxjs' ? ApiRx : ApiPromise;
+
+  return function (
+    message: string,
+    inner: (api: typeof apiType extends 'rxjs' ? ApiRx : ApiPromise) => any
+  ) {
+    let wsEndpoints: WsName[] = []; // The ws endpoints to test
+    if (options && options.only) {
+      wsEndpoints = options.only;
+    } else {
+      wsEndpoints = (Object.keys(WS_ENDPOINTS) as WsName[])
+        .filter((wsName) => !options || !options.except || !options.except.includes(wsName));
+    }
+
+    // If there's no WITH_DOCKER flag, we only run local node
+    if (!process.env.WITH_DOCKER) {
+      wsEndpoints = wsEndpoints.filter((wsName) => wsName === 'local');
+    }
+
+    wsEndpoints
+      .map((wsName) => WS_ENDPOINTS[wsName])
+      .forEach((wsUrl) => {
+        describe(`${message} with ${wsUrl}`, () => {
+          let api: any;
+
+          beforeAll(() => {
+            jest.setTimeout(30000);
+          });
+
+          beforeEach(async (done) => {
+            api = await Api.create(new WsProvider(wsUrl));
+            done();
+          });
+
+          afterAll(() => {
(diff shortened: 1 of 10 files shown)


packages/api/test/util/describeE2E.ts
```diff
@@ -0,0 +1,72 @@
...
+const WS_ENDPOINTS = {
+  'local': 'ws://127.0.0.1:9944/',
+  'substrate-master': 'ws://127.0.0.1:9945/',
+  'substrate-1.0': 'ws://127.0.0.1:9946/',
+  'substrate-2.0': 'ws://127.0.0.1:9947/',
+  'polkadot-master': 'ws://127.0.0.1:9948/',
+  'polkadot-alexander': 'ws://127.0.0.1:9949/',
+  'remote-polkadot-alexander': 'wss://poc3-rpc.polkadot.io/',
+  'remote-substrate-1.0': 'wss://substrate-rpc.parity.io/'
+};
```
Stefie: I already added them to the jest globals, so they should be available with `global.ws_local`, `global.ws_polkadot_master`etc https://github.com/polkadot-js/api/blob/master/jest.config.js#L23

If we only need them here, I should remove them from the config.

amaurymartiny: I took the same names as in the docker-compose files, (with `remote-*` for remote ones), but I can take your naming in jest config if you want

Stefie: Nah, dash is fine. I have no strong feelings about the naming

Stefie:
I have to leave, but I'm really looking forward to play around with it!

packages/api/test/util/describeE2E.ts
```diff
@@ -0,0 +1,63 @@
...
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+const WS_ENDPOINTS = {
+  'local': 'ws://127.0.0.1:9944/',
+  'substrate-master': 'ws://127.0.0.1:9945/',
+  'substrate-1.0': 'ws://127.0.0.1:9946/',
+  'substrate-2.0': 'ws://127.0.0.1:9947/',
+  'polkadot-master': 'ws://127.0.0.1:9948/',
+  'polkadot-alexander': 'ws://127.0.0.1:9949/',
+  'remote-polkadot-alexander': 'wss://poc3-rpc.polkadot.io/',
```
jacogr: Just be aware that unless we -

- setup a specific one, or
- use a service such as Polkasource or Unfrastructure

that this will not be possible anymore with Kusama for the Parity-hosted URLs. We get too much abuse which affects service to real users. 

Actual UI dev (hopefully) won't be affected, but I'll chat to the Parity folks around how that works, once it is setup and verified working. e2e,m unless tied to a specific IP though, will not be supportable.

packages/api/test/util/describeE2E.ts
```diff
@@ -0,0 +1,63 @@
...
+  'remote-substrate-1.0': 'wss://substrate-rpc.parity.io/'
+};
+
+type WsName = keyof typeof WS_ENDPOINTS;
+
+interface Options {
+  except?: WsName[]; // Only one of the two keys `either` `only` should be set
```
jacogr: +1 for the `[]` :)

packages/api/test/util/describeE2E.ts
```diff
@@ -0,0 +1,63 @@
...
+  only?: WsName[];
+  apiType?: 'promise' | 'rxjs';
+}
+
+// From the options and the WITH_DOCKER flag, calculate on which endpoints we
```
jacogr: Possible to have a `"test:docker": "WITH_DOCKER yarn test packages/api/test/e2e"` command?

- no coverage
- only runs the e2e stuff (assuming all in the same location)

packages/api/test/util/describeE2E.ts
```diff
@@ -0,0 +1,63 @@
...
+    wsEndpoints = options.only;
+  } else {
+    wsEndpoints = (Object.keys(WS_ENDPOINTS) as WsName[])
+      .filter((wsName) => !options || !options.except || !options.except.includes(wsName));
+  }
+
+  // If there's no WITH_DOCKER flag, we only run local node
+  if (!process.env.WITH_DOCKER) {
+    wsEndpoints = wsEndpoints.filter((wsName) => wsName === 'local');
+  }
```
Stefie: And can we maybe add the reverse condition as well? I noticed that since I added the docker compose I’m not running a local node anymore because I have the 5 endpoints incl. the latest masters  at my fingertip anyways. So my usecase would be to only test against against the docker images and only if I’d want a specific version I’d actually compile a local node

Stefie: But no blocker, I can also add that later

jacogr:
I broke this in https://github.com/polkadot-js/api/pull/1099, sorry.

packages/api/test/util/describeE2E.ts
amaurymartiny: I added test:docker as per Jaco's suggestion, is that alright for you?

packages/api/test/util/describeE2E.ts
```diff
@@ -0,0 +1,62 @@
...
+  // If there's no WITH_DOCKER flag, we only run local node
+  if (!process.env.WITH_DOCKER) {
+    wsEndpoints = wsEndpoints.filter((wsName): boolean => wsName === 'local');
+  }
+
+  return wsEndpoints;
+}
+
+export default function describeE2E (options?: Options): (message: string, inner: (wsUrl: string) => void) => void {
+  return function (message: string, inner: (wsUrl: string) => void): void {
```
amaurymartiny: Unfortunately need to write twice the funciton signature here

PR #1121, opened 2019-07-13
Added a folder on validating Polkadot addresses 

JesseAbram:


jacogr:
Brilliant! Can I ask to move this to the common repo? (Since it is utility usage)

JesseAbram:
Of course makes sense, does it fit in the util-crypto section or the keyring section since it does use both

PR #1125, opened 2019-07-14
add isSubstrateV2

Tbaut:
- as per https://github.com/polkadot-js/apps/pull/1397#discussion_r303215263 this gives an easy way to check on the UI side if we are connected to a Substrate v2
- test is passing for Alex (and failing with a localhost v2 as expected)

### DIFF
--- a/packages/api/src/Base.ts
+++ b/packages/api/src/Base.ts
@@ -546,7 +546,7 @@ export default abstract class ApiBase<ApiType> {
     return true;
   }
 
-    /**
+  /**
    * @description Returns true if the api is connected to a substrate node v2.
    *
    *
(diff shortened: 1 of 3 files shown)


packages/api/src/Base.ts
```diff
@@ -474,6 +476,10 @@ export default abstract class ApiBase<ApiType> {
           this.emit('ready', this);
         }
 
+        if (!Object.keys(this.consts).length) {
+          this._isSubstrateV2 = false;
```
jacogr: I believe this is the other way around?

Tbaut: It's maybe a little hard to read, but if the length is 0 it's a v1, the `!` will make it go into the `if` to set it to `false`.

Tbaut:
I re-read your comment too late in https://github.com/polkadot-js/apps/pull/1397#discussion_r303215263, you wrote `ui-api` from the start. I'll put it there

packages/api/src/Base.ts
jacogr: You are right :)

Issue #1113, opened 2019-07-12
arg[0] is undefined in decorateMulti from /api/src/Base.ts


Tbaut:
On a recent v2.0 Substrate node

The test "derive e2e › retrieves all staking info (for controller)" fails with the following:
```bash
    TypeError: Cannot read property 'creator' of undefined

      560 |           // the input is a QueryableStorageEntry, convert to StorageEntry
      561 |           Array.isArray(arg)
    > 562 |             ? [arg[0].creator, ...arg.slice(1)]
          |                       ^
      563 |             : [arg.creator] as any
      564 |         );
      565 | 

      at creator (packages/api/src/Base.ts:562:23)
          at Array.map (<anonymous>)
      at Object.map [as queryMulti] (packages/api/src/Base.ts:559:30)
      at queryMulti (packages/api-derive/src/staking/info.ts:125:9)
      at SwitchMapSubscriber.withControllerLedger [as project] (packages/api-derive/src/staking/info.ts:167:17)
      at SwitchMapSubscriber.Object.<anonymous>.SwitchMapSubscriber._next (node_modules/rxjs/src/internal/operators/switchMap.ts:123:21)
      at SwitchMapSubscriber.Object.<anonymous>.Subscriber.next (node_modules/rxjs/src/internal/Subscriber.ts:99:12)
      at RefCountSubscriber.Object.<anonymous>.Subscriber._next (node_modules/rxjs/src/internal/Subscriber.ts:139:22)
      at RefCountSubscriber.Object.<anonymous>.Subscriber.next (node_modules/rxjs/src/internal/Subscriber.ts:99:12)
      at ReplaySubject.Object.<anonymous>.Subject.next (node_modules/rxjs/src/internal/Subject.ts:70:17)
      at ReplaySubject.Object.<anonymous>.ReplaySubject.nextInfiniteTimeWindow (node_modules/rxjs/src/internal/ReplaySubject.ts:46:15)
      at ConnectableSubscriber.Object.<anonymous>.Subscriber._next (node_modules/rxjs/src/internal/Subscriber.ts:139:22)
      at ConnectableSubscriber.Object.<anonymous>.Subscriber.next (node_modules/rxjs/src/internal/Subscriber.ts:99:12)
      at Object.next [as callback] (packages/rpc-core/src/index.ts:183:22)
      at WsProvider.callback [as onSocketMessageSubscribe] (packages/rpc-provider/src/ws/Provider.ts:352:15)
[...]

PR #1127, opened 2019-07-15
SessionId extraction for 2.x

jacogr:
Closes https://github.com/polkadot-js/api/issues/1113

### DIFF
--- a/packages/types/src/type/SessionKeys.ts
+++ b/packages/types/src/type/SessionKeys.ts
@@ -13,22 +13,14 @@ import SessionKey from './SessionKey';
 export default class SessionKeys extends Struct {
   public constructor (value?: any) {
     super({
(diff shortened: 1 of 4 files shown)


packages/types/src/type/SessionKeys.ts
```diff
@@ -13,22 +13,14 @@ import SessionKey from './SessionKey';
 export default class SessionKeys extends Struct {
   public constructor (value?: any) {
     super({
-      grandpaKey: SessionKey,
-      auraKey: SessionKey
+      ed25519: SessionKey
```
amaurymartiny: Doesn't this break backward-compat?

jacogr: Yes and No. Keys were added in 2.x, so in this case it is ok to break the compat.

PR #1132, opened 2019-07-15
Reorganize test setup

Stefie:
Addresses the following points of https://github.com/polkadot-js/api/issues/908:
- Reorder and better categorize existing tests
- Move all e2e tests to ./api

I'm working through this part bit by by now for all supported versions 
- Fix broken / outdated tests

Also in this PR:
I split off the e2e test flags to be able to run the tests from the 3 different sources independently.
`yarn test:e2e`: Will run the e2e tests with the local node on `9944`
`yarn test:e2e-docker`: Will run the e2e tests with the 5 Docker images
`yarn test:e2e-remote`: Will run the e2e tests with the 2 remote endpoints

The goal is that in the end everyone developing on the API can spot failing tests on all supported substrate and polkadot versions easily.
Screenshot 2019-07-15 at 18 18 08


packages/api/test/e2e/api-derive/promise.spec.ts
```diff
@@ -7,13 +7,12 @@ import testingPairs from '@polkadot/keyring/testingPairs';
 import { WsProvider } from '@polkadot/rpc-provider';
 import { RewardDestination } from '@polkadot/types';
 
-import { HeaderExtended } from '../../src/type';
-import { DerivedBalances, DerivedFees, DerivedSessionInfo, DerivedStaking } from '../../src/types';
-import { SubmittableResult } from '../../../api/src';
+import { HeaderExtended } from '../../../../api-derive/src/type';
+import { DerivedBalances, DerivedFees, DerivedSessionInfo, DerivedStaking } from '../../../../api-derive/src/types';
```
amaurymartiny: I think here and above we can just use `@polkadot/api-derive`;

packages/api/test/e2e/api-derive/promise.spec.ts
```diff
@@ -7,13 +7,12 @@ import testingPairs from '@polkadot/keyring/testingPairs';
...
+import { SubmittableResult } from '../../../../api/src';
```
amaurymartiny: This one can be shortened too

packages/api/test/e2e/api/promise-queries.spec.ts
```diff
@@ -313,12 +329,28 @@ describeE2E()('Promise e2e queries', (wsUrl): void => {
...
 
       expect(size.toNumber()).not.toEqual(0);
     });
   });
+});
+
+// The following tests only run on nodes that support doubleMapType introduced by metadata v5
+describeE2E({
+  only: [
+    'docker-polkadot-master',
```
amaurymartiny: we can add local here

packages/api/test/util/describeE2E.ts
```diff
@@ -4,24 +4,35 @@
...
 
-// From the options and the WITH_DOCKER flag, calculate on which endpoints we
+function filterProcessEnv (wsEndpoints: WsName[]): WsName[] {
+  // If there's a TEST_DOCKER flag, we  only run tests to Docker endpoints
+  // If there's a TEST_REMOTE flag, we only run tests to remote endpoints
+  // If none of the two is present, we only run tests on local node
+  return process.env.TEST_REMOTE || process.env.TEST_DOCKER
+    ? process.env.TEST_REMOTE
+      ? wsEndpoints.filter((wsName): boolean => String(wsName).startsWith('remote-'))
+      : wsEndpoints.filter((wsName): boolean => String(wsName).startsWith('docker-'))
```
amaurymartiny: String() is useless here

packages/api/test/util/describeE2E.ts
```diff
@@ -4,24 +4,35 @@
...
   'remote-substrate-1.0': 'wss://substrate-rpc.parity.io/'
 };
 
 type WsName = keyof typeof WS_ENDPOINTS;
 
 interface Options {
-  except?: WsName[]; // Only one of the two keys `either` `only` should be set
+  except?: WsName[]; // Only one of the two keys `except` `only` should be set
   only?: WsName[];
   apiType?: 'promise' | 'rxjs';
```
amaurymartiny: You can actually remove apiType in all tests, I added it but it's actually useless

packages/api/test/e2e/api/promise-contract.spec.ts
```diff
@@ -6,16 +6,16 @@ import fs from 'fs';
...
-import { SubmittableResult } from '../../src';
-import ApiPromise from '../../src/promise';
-import describeE2E from '../util/describeE2E';
+import { SubmittableResult } from '../../../src';
+import ApiPromise from '../../../src/promise';
+import describeE2E from '../../util/describeE2E';
 
-const flipperCode = fs.readFileSync(path.join(__dirname, '../../../api-contract/test/contracts/flipper-pruned.wasm')).toString('hex');
+import flipperAbi from '../../util/contracts/flipper.json';
+const flipperCode = fs.readFileSync(path.join(__dirname, '../../util/contracts/flipper-pruned.wasm')).toString('hex');
```
jacogr: Not sure I agree having contract code in util.

packages/api/test/e2e/api/promise-queries.spec.ts
```diff
@@ -234,29 +144,135 @@ describeE2E()('Promise e2e queries', (wsUrl): void => {
       expect(size.toNumber()).toBeGreaterThanOrEqual(0);
     });
   });
+});
+
+// The following tests require a development node with the default funded testing accounts
+describeE2E({
```
jacogr: I would probably suggest just having one `describeE2E` block per file. That way we categorize things slightly better?

amaurymartiny:
Would be cool to take this into account too: https://github.com/polkadot-js/api/blob/master/packages/api/test/e2e/promise-tx.spec.ts#L122

packages/api/test/e2e/api/promise-queries.spec.ts
amaurymartiny: +1

packages/api/test/e2e/api-derive/promise.spec.ts
Stefie: I got eslint errors for that, but that might be a confused VSCode after the the switch from tslint. I'll look into it.

packages/api/test/e2e/api/promise-contract.spec.ts
Stefie: Ok. I just moved it here because it's only used  here in the e2e tests (apart from one ABI test in api-contract).
Do you generally agree moving the files to the e2e test folder in `api` or would you rather leave them in `api-contract`, even though they are not being used there, but logically belong to that api module?

Stefie:
> Should all tests pass?
No, they shouldn't. 
As I said in the comments, I'm going through them bit by bit. I'd just like to merge this PR because I have another PR open for which I need the changes in here and I don't want to pull this one in and then keep both op to date with master and each other, because the file structure for the tests changed in this one.
And there will be more e2e tests passing after this PR got merged than before.

packages/api/test/e2e/api/promise-queries.spec.ts
Stefie: Yes, true, but naming the files could get quite hard then, but I'll try to think of something.

Stefie:
> Would be cool to take this into account too: https://github.com/polkadot-js/api/blob/master/packages/api/test/e2e/promise-tx.spec.ts#L122

My current approach is: 
1. I'll go through all the e2e tests files and will make all tests pass or report bugs for all 5 supported versions
2. refactor tests / add tests/ tackle TODO's

I'd take care of the promise-tx.spec file in one of my next PR's

packages/api/test/e2e/api/promise-queries.spec.ts
Stefie: I split this file into three separate ones.

packages/api/test/util/describeE2E.ts
Stefie: removed.

packages/api/test/util/describeE2E.ts
Stefie: removed

packages/api/test/e2e/api/promise-queries.spec.ts
Stefie: Yes, absolutely right, I kind of assumed that it includes `local`  by default even though I touched the code that defines it

packages/api/test/e2e/api-derive/promise.spec.ts
Stefie: changed to `@polkadot` import

packages/api/test/e2e/api-derive/promise.spec.ts
Stefie: changed to `@polkadot` import

Stefie:
> Should all tests pass?
>  - test:e2e (on local node): 6 failed

I generally won't care about the local tests passing btw because that's unpredictable. I could never know which version / node a person is running on their machine.
So for local tests I'll only make sure that the tests are running, but I don't care about the results.

packages/api/test/e2e/api/promise-contract.spec.ts
jacogr: In this case it is a bit of a toss-up -

- it really is contract data
- it actually does get used in the contracts tests (well, the ABIs do - and where not all, they should - just waiting on the new interfaces)

So happy with the e2e tests in api, but the data since it is also used for "normals" (not the wasm, but should be kept with the ABIs)

Stefie: I moved it back now

PR #1141, opened 2019-07-16
Adding derives for election sets and approval flags.

### LINKED ISSUE #1118: Add derives for SetIndex in Elections.
Elections now uses chunking for approvals and voters (see: https://github.com/paritytech/substrate/blob/master/srml/elections/src/lib.rs#L243). It would be useful to have a set of derives that let the caller "just get everything" without worrying about the chunking. There is already an internal function that does this (https://github.com/paritytech/substrate/blob/master/srml/elections/src/lib.rs#L1047) but no way to access it via RPC.

The derive would also be useful if it converted ApprovalFlag to booleans, as seen here: https://github.com/paritytech/substrate/blob/master/srml/elections/src/lib.rs#L1030

Thanks!

jnaviask:
Fixes #1118 

WIP


jacogr:
I will add e2e tests here.

PR #1153, opened 2019-07-21
Move contracts classes to srml/contracts (now as types-only)

jacogr:
- Some more types after https://github.com/polkadot-js/api/pull/1077 (touching contracts as indicated in the first)
- Slightly extend the srmlTs types generation, here to cater for Option and Compact
- Make Compact generic (this actually does help a bundle in the `unwrap()`)


packages/types/src/codec/Compact.ts
```diff
@@ -20,13 +20,13 @@ import UInt from './UInt';
  * used by other types to add length-prefixed encoding, or in the case of wrapped types, taking
  * a number and making the compact representation thereof
  */
-export default class Compact extends Base<UInt | Moment> implements Codec {
-  public constructor (Type: Constructor<UInt | Moment>, value: Compact | AnyNumber = 0) {
-    super(Compact.decodeCompact(Type, value));
+export default class Compact<T> extends Base<UInt | Moment> implements Codec {
```
jacogr: Need some help here - this does not look 100% -

- Base needs to use `T` as well
- All `UInt | Moment` to be `T`

... however, cannot quite get all the checkers happy, so the current attempt is better, but far from technically correct. (Please pull down, play here and give suggestions - or just push a proper fix)

.codeclimate.yml
```diff
@@ -4,3 +4,4 @@ exclude_patterns:
 - "**/*.test.js"
 - "**/*.test.ts"
 - "docs/**/*.js"
+- "packages/types/src/srml/"
```
jacogr: Exclude these specifically, since it is generated.

packages/types/src/scripts/srmlTS.ts
```diff
@@ -22,6 +22,7 @@ interface TypeImports {
 
 const HEADER = '/* eslint-disable @typescript-eslint/no-empty-interface */\n// Auto-generated, do not edit\n\n';
 const FOOTER = '\n';
+const OUTPUT_FILE = 'index.ts';
```
jacogr: See initial PR comment - on the fence around the naming, `types.ts` probably fits better and it clearer in intent (although slightly more effort to use)

packages/types/src/codec/Compact.ts
jacogr: Thanks to @amaurymartiny a fix is in :)

jacogr:
Yes, I actually ended up at the same conclusions, but didn't want to revert that - since I already asked the question. Worth having a thought about anyway. Will swap it back.

PR #1156, opened 2019-07-22
Fix compact type

amaurymartiny:
something like that?

### DIFF
--- a/packages/types/src/codec/Compact.ts
+++ b/packages/types/src/codec/Compact.ts
@@ -20,12 +20,12 @@ import UInt from './UInt';
  * used by other types to add length-prefixed encoding, or in the case of wrapped types, taking
  * a number and making the compact representation thereof
  */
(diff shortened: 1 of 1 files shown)


packages/types/src/codec/Compact.ts
```diff
@@ -20,12 +20,12 @@ import UInt from './UInt';
  * used by other types to add length-prefixed encoding, or in the case of wrapped types, taking
  * a number and making the compact representation thereof
  */
-export default class Compact<T> extends Base<UInt | Moment> implements Codec {
-  public constructor (Type: Constructor<UInt | Moment>, value: Compact<T> | AnyNumber = 0) {
+export default class Compact<T extends UInt | Moment> extends Base<T> implements Codec {
```
jacogr: I swear, I did the `<T extends UInt | Moment>` and it complained bitterly for me.

Issue #1160, opened 2019-07-22
Polkadot master: api.query.staking.validators() broken


Stefie:
Calling `api.query.staking.validators()` on the latest Polkadot master with Metadata v6 is throwing the following error: 
`Unable to decode storage undefined.undefined: createType(AccountId):: Input doesn't match output, received 0x, created 0x0000000000000000000000000000000000000000000000000000000000000000`

It's fine on all Substrate versions (1.0, 2.0 and Master) and on Polkadot Alexander.

<img width="804" alt="Screenshot 2019-07-22 at 17 11 04" src="https://user-images.githubusercontent.com/125398/61644936-f0bcd800-aca5-11e9-94e1-a44f6d0675bb.png">

jacogr:
So technically `isPedantic` is too pedantic here since `0x` and `0x000...000` is equivalent. So this is one of those special cases actually. 

Should adjust the check to do `input === '0x' && decoded.isEmpty()`

(Not sure about `undefined.undefined`, however it is a linked map, so we potentially do something weird in setting the module/function for those in `StorageKey`

jacogr:
The weird one here is that I don't get it with a freshly-built polkadot -

![image](https://user-images.githubusercontent.com/1424473/61666063-e404a800-acd6-11e9-842d-499e26237815.png)

jacogr:
Ok, well, the above is wrong, so the problem is twofold -

- headKey (where this fails) does not have section/meta (as of now the linked PR fixes that, tested)
- linked maps does not apply defaults in rpc-core (they are ignored on isNull)

jacogr:
So the PR has the fixes. 

However, there is something up with the Polkadot config, I don't believe it has been fully adjusted for all the new staking/session logic. The validator set should not return empty based on genesis config for dev. 

(It is good that it did, since it uncovered the above issues, however as it stands the above node is not really ready for operation.)

Stefie:
> However, there is something up with the Polkadot config, I don't believe it has been fully adjusted for all the new staking/session logic. The validator set should not return empty based on genesis config for dev.

Yeah, it seems off in a few places. I'll keep that in mind when working on the other tests. 
Thanks for the quick fix anyways!

PR #1162, opened 2019-07-22
Fix linked maps with defaults

jacogr:
- Closes https://github.com/polkadot-js/api/issues/1160
- headKey now has section/method for display
- add a non-empty fallback value for linked map head
- adjust fallback value for linked map entries (also needs to cater for the `Linkage`)


jacogr:
Will add the issue as soon as merged to point to the FIXME

PR #1166, opened 2019-07-23
feat: Implement twox64concat

### LINKED ISSUE #1122: Add Twox64Concat hasher
https://github.com/polkadot-js/api/blob/master/packages/api-metadata/src/storage/fromMetadata/getHasher.ts#L34

amaurymartiny:
reference: https://github.com/paritytech/substrate/blob/ff7d929ba8eea4a06156584480e8a946dd154f00/srml/support/src/storage/hashed/generator.rs#L34

fixes #1122

### DIFF
--- a/packages/api-metadata/src/storage/fromMetadata/getHasher.ts
+++ b/packages/api-metadata/src/storage/fromMetadata/getHasher.ts
@@ -4,6 +4,7 @@
 
 import { StorageHasher } from '@polkadot/types/primitive';
 import { blake2AsU8a, xxhashAsU8a } from '@polkadot/util-crypto';
+import { bufferToU8a, isString, stringToU8a, u8aConcat } from '@polkadot/util';
 
 type HasherInput = string | Buffer | Uint8Array;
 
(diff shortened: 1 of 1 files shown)


packages/api-metadata/src/storage/fromMetadata/getHasher.ts
```diff
@@ -31,7 +32,20 @@ export default function getHasher (hasher?: StorageHasher): HasherFunction {
     return (data: HasherInput): Uint8Array => xxhashAsU8a(data, 256);
   }
 
-  // FIXME Add Twox128Concat
+  if (hasher.isTwox64Concat) {
+    return (data: HasherInput): Uint8Array => {
+      let dataAsU8a: Uint8Array;
```
jacogr: You could just use `const dataAsU8a = u8aTou8a(data)` and skip the ifs. (It does the u8a/string/buffer dance on your behalf) Apart form that, the logic matches.

amaurymartiny: ah, I knew there was something like that

PR #1168, opened 2019-07-23
Small rpc-core storage cleanups

jacogr:


packages/rpc-core/src/index.ts
```diff
@@ -305,9 +305,7 @@ export default class Rpc implements RpcInterface {
     const type = key.outputType || 'Data';
     const meta = key.meta || EMPTY_META;
 
-    if (meta.type.isMap && meta.type.asMap.isLinked) {
-      return createType(type, isNull ? meta.fallback : base, true);
-    } else if (meta.modifier.isOptional) {
+    if (meta.modifier.isOptional) {
```
jacogr: The asMap if was the same as the default return in both cases, removed.

PR #1179, opened 2019-07-24
DoubleMap extended tests

### LINKED ISSUE #1174: Add e2e tests for DoubleMap
Can use this - https://github.com/polkadot-js/api/pull/1171/files#diff-a60854aee74a1ede6dc8edf6544d4240R1404

- set key
- query to see if it has changed in the map

jacogr:
Closes #1174

### DIFF
--- a/packages/types/src/primitive/Type.ts
+++ b/packages/types/src/primitive/Type.ts
@@ -8,6 +8,8 @@ type Mapper = (value: string) => string;
 
 const ALLOWED_BOXES = ['Compact', 'Option', 'Vec'];
 
+// f4adb4c4f708c4b753657373696f6e204e6578744b657973343a73657373696f
+// 6e3a6b65797354b75224d766c852ac60eb44e1329aec5058574ae8daf703d43bc2fbd9f33d6c
 /**
  * @name Type
  * @description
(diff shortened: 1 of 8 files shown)


packages/types/src/primitive/Type.ts
```diff
@@ -57,8 +57,8 @@ export default class Type extends Text {
       Type._alias('Compact<Index>', 'IndexCompact'),
       // alias Vec<u8> -> Bytes
       Type._alias('Vec<u8>', 'Bytes'),
-      // alias &[u8] -> Bytes
-      Type._alias('&\\[u8\\]', 'Bytes'),
+      // alias &[u8] -> Data (this has no length-prefix)
+      Type._alias('&\\[u8\\]', 'Data'),
```
amaurymartiny: Why not just remove Data type and use U8a everywhere? Is it a rust type?

jacogr: U8a is not exported, I tried that first. And since we already use Data to indicate exactly those, just kept it at `Data` as to not have more sprawl.

amaurymartiny: Ok

PR #1181, opened 2019-07-24
Improve voter derive calls

kwingram25:
- Use voterSetSize, return more useful index object
- Add nextVoterSet to info


packages/api-derive/src/elections/voters.ts
```diff
@@ -25,14 +25,24 @@ import { voterSets } from './voterSets';
...
+        (voterPositions: DerivedVoterPositions): Vector<AccountId> =>
           new Vector(
             AccountId,
-            Object.keys(voterSets).map(
-              (address: string): AccountId => new AccountId(address)
-            )
+            Object.entries(voterPositions)
+              .sort(
+                (a, b): number => {
+                  if (a[1].globalIndex.lt(b[1].globalIndex)) {
```
jacogr: Isn't there a `a[1].globalIndex.cmp(b[1].globalIndex)` available?

packages/api-derive/src/elections/voterPositions.ts
```diff
@@ -0,0 +1,57 @@
...
+        const [setSize, voters] = result;
+        return voters.reduce((result: DerivedVoterPositions, vec, setIndex): DerivedVoterPositions => {
+          vec.forEach((e, index): void => {
+            // re-create the index based on position 0 is [0][0] and likewise
+            // 64 (0..63 in first) is [1][0] (the first index value in set 2)
+            const accountId: AccountId | null = e.unwrapOr(null);
+
+            if (accountId) {
+              result[accountId.toString()] = {
+                globalIndex: new BN((setIndex * setSize.toNumber()) + index),
```
jacogr: `setSize.muln(setIndex).addn(index)` - just in case it overflows at some point? (No new BN needed)

PR #1185, opened 2019-07-25
Unskip additional tests

jacogr:


packages/api/src/promise/Api.spec.ts
```diff
@@ -26,7 +26,7 @@ describe.skip('Metadata queries', (): void => {
     const api = await ApiPromise.create({ provider: mock, metadata } as unknown as ApiOptions);
 
     expect(api.genesisHash).toBeDefined();
-    expect(api.runtimeMetadata.toJSON()).toEqual(new Metadata(rpcData).toJSON());
+    expect(api.runtimeMetadata).toBeDefined();
```
jacogr: This is with the toJSON() atm is this change - https://github.com/polkadot-js/api/blob/master/packages/api-metadata/src/storage/fromMetadata/createFunction.ts#L120

So the fallback for these (3 of them), are different. I believe it is fine as-is, the parsing of the data is tested extensively.

jacogr:
Yes, I saw. However, really need to get the types PR in and release the extension this morning, so my time is limited. (This was my early morning wake-up PR)

My comments on the above 3 when I looked -

- The enum test can be removed completely. It is functionality we don't have (it was removed to simplify since it was unused) and when re-added, will look different anyway
- Both the Tuple and Vector tests are the same thing - it decodes a proposal, however the proposal maps to a method where the metadata indexes changed. (No idea what the original was). Here the hex-strings should be updated (and probably commented like we have elsewhere where we split out the hex to indicate the section/method with human-readable descriptions, so breakages can just update the indexes)

PR #1190, opened 2019-07-25
ClassOf does not require <T>

### LINKED ISSUE #1177: `ClassOf<T>('SomeType')` should not require the `<T>`
**Expected:**

```typescript
const g = createType('Gas', 1234);
g.ad... // should autocomplete because it's a BN
```

**Actual:**

Doesn't autocomplete. We need to do the following:

```typescript
import { Gas } from '@polkadot/types/smrl/types';

const g = createType<Gas>('Gas', 1234);
g.ad... // autocompletes correctly
```

which is okay for 1 or 2 types, but gets really redundant for more, since we're really typing the same word twice

**Solution:**

The idea is to create an `InterfaceRegistry`, empty at first, and the script `srmlTs.ts` populates `InterfaceRegistry` with the types it finds, via module augmentation.

- Step 1. Create `InterfaceRegistry`. The keys are strings with represent the srml type (like 'Gas'), and the values are the actual interface (like `Gas` in `types/smrl/contracts/types`).

```typescript
// in `@polkadot/types/smrl/types.ts`

export interface InterfaceRegistry {
  // empty at first
}
```

- Step 2. Modify `srmlTs.ts` to populate `InterfaceRegistry` via module augmentation. For example, for gas, the script would output in `types/srml/contract/types.ts`:

```typescript
declare module '@polkadot/types/srml/types' {
  interface InterfaceRegistry {
    Gas: Gas // maps the key "Gas" to the interface `Gas`
  }
}
```

- Step 3. In `ClassOf`:
    - if we're calling `const a = createType<SomeOtherType>('Gas', 1234)`, then `a` should be of type `SomeOtherType` (overridden by user)
    - if we're calling `const a = createType('Gas', 1234)`, then `ClassOf` should check if `InterfaceRegistry['Gas']` is present, if yes, take it, if no, return `Codec`

I'm not 100% this will actually work, so needs some exploration first.

amaurymartiny:
fixes #1177 

The central part is the registry `'@polkadot/types/interfaceRegistry'`, which is populated by srmlTs script. I don't think this `InterfaceRegistry` should be user-facing though, so for now it's only there, not re-exported from e.g. inside `'@polkadot/types/types'`


packages/api/src/checkTypes.manual.ts
```diff
@@ -90,4 +90,10 @@ export default async function test (): Promise<void> {
 
       unsub2();
     });
+
+  // check correct types with `createType`
+  const balance = createType('Balance', 2);
+  const overridden = createType<Nonce>('Gas', 2);
+  const gas = createType('Gas', 2);
+  console.log(balance, overridden, gas);
```
amaurymartiny: pull down and play :)

packages/types/src/codec/createType.ts
```diff
@@ -315,11 +320,13 @@ function initType<T extends Codec = Codec> (Type: Constructor<T>, value?: any, i
   }
 }
 
-export default function createType<T extends Codec = Codec> (type: Text | string, value?: any, isPedantic?: boolean): T {
-  // l.debug(() => ['createType', { type, value }]);
-
+export default function createType<T extends Codec = Codec, K extends Text | string = Text | string> (
```
amaurymartiny: I didn't remove Text yet, could do it now, could do it later when we cater for https://github.com/polkadot-js/api/issues/1177#issuecomment-514588305

packages/types/src/interfaceRegistry.ts
```diff
@@ -0,0 +1,24 @@
...
+// Holds a registry of all generated interfaces
+// https://github.com/polkadot-js/api/issues/1177
+export interface InterfaceRegistry {
+  AccountId: AccountId;
+  AccountIdOf: AccountIdOf;
+  AccountIndex: AccountIndex;
+  AccountInfo: AccountInfo;
+  Address: Address;
+  Balance: Balance;
+  // TODO Finish this
```
amaurymartiny: Unfortunately for primitives and jsonrpc types, we still need to populate InterfaceRegistry manually.

will remove WIP when I finish this

jacogr: We could actually add those as part of the generation as well?

i.e. pull in `primitives/index.ts`, apply the same logic and generate an output file?

packages/types/src/codec/createType.ts
amaurymartiny: I'll do this in a subsequent PR, this one is big enough

PR #1192, opened 2019-07-25
Convert RPC classes to primitive classes or /srml types

jacogr:
- Move Vector -> Vec
- Remove rpc/
- More primitive types to definitions


packages/type-jsonrpc/src/types.ts
```diff
@@ -2,14 +2,14 @@
...
-import { CodecTypes } from '@polkadot/types/classes';
+// import { CodecTypes } from '@polkadot/types/classes';
 
 type PubSub = [string, string, string];
 
 export interface RpcParam {
   isOptional: boolean;
   name: string;
-  type: CodecTypes;
+  type: string; // CodecTypes;
```
jacogr: Yes. Basically at this point need to add more `@ts-igore` lines that actual code. However... believe this is solvable now in the same was as `ClassOf` (some work required, but solvable)

packages/types/src/codec/Compact.ts
```diff
@@ -2,18 +2,20 @@
...
-import { AnyNumber, Constructor } from '../types';
-import AbstractInt, { UIntBitLength } from './AbstractInt';
+import { UIntBitLength } from './AbstractInt';
+import UInt from './UInt';
 import Base from './Base';
 
 // List of codec types that are compact-encodable
-export const COMPACT_ENCODABLE = [AbstractInt, Moment];
-export type CompactEncodable = AbstractInt | Moment; // FIXME is there a way to do it not-manually from the array?
+export const COMPACT_ENCODABLE = [UInt, Moment];
```
jacogr: @amaurymartiny This works now. In hindsight, it should not be applied to signed actually? (The checks are only for unsigned, shifting will lose bits...)

packages/types/src/codec/Linkage.ts
```diff
@@ -35,8 +35,8 @@ export default class Linkage<T extends Codec> extends Struct {
 export class LinkageResult extends Tuple {
   public constructor ([TypeKey, keys]: TypeWithValues, [TypeValue, values]: TypeWithValues) {
     super({
-      Keys: Vector.with(TypeKey),
-      Values: Vector.with(TypeValue)
+      Keys: Vec.with(TypeKey),
```
jacogr: This could be moved out as well.

jacogr: Not really, since it needs the Types on construction. (So really it fits at the base)

jacogr:
Merging as-is.

PR #1193, opened 2019-07-26
 Add hash: IHash on Codec base

jacogr:
Builds on #1192


jacogr:
Build is fixed on #1192, so will recover. (Once that is merged will re-base to master)

packages/types/src/primitive/Null.ts
```diff
@@ -18,6 +22,13 @@ export default class Null implements Codec {
     return 0;
   }
 
+  /**
+   * @description returns a hash of the contents
+   */
+  public get hash (): IHash {
+    return createType('Hash', blake2AsU8a(this.toU8a(), 256));
```
amaurymartiny: hmm, it's sometimes `createType('Hash',)`, sometimes `new U8a(...)`. For consistency I would just use createType everywhere

jacogr: Can't use `createType` in codec without running into the issues of `extending from null`

i.e. in U8a -

- U8aFixed extends U8a
- U8a imports createType
- createType imports U8aFixed
- circular dep, all hell breaks loose :)

amaurymartiny: ah, f, I see

PR #1199, opened 2019-07-26
CreateType is typed, add createTypeUnsafe just in case

amaurymartiny:
```
const gas = createType('Gas', 123); // works
const random = createType('Random', 123); // TS error
```

Also: removes the generic `<T>` from createType, so this is not possible: 
```
createType<Gas>('Random', 123); // TS error, this is unsafe!
```
---

if you know what you're doing, use `createTypeUnsafe`

```
const gas = createTypeUnsafe('Gas', 123); // works
const random = createTypeUnsafe<Gas>('Random', 123); // works, will throw!
const random2 = createTypeUnsafe('(u32,u32)', [123,123]); // works
```
Notes:
- createType, ClassOf etc don't accept Text as input anymore, only string

TODO:
- [x] ClassOfUnsafe, so that types are consistent
- [x] clean up some stuff


packages/types/src/types.ts
```diff
@@ -105,7 +105,7 @@ export interface IHash extends U8a { }
 export type CodecTo = 'toHex' | 'toJSON' | 'toString' | 'toU8a';
 
 export interface Constructor<T = Codec> {
-  Fallback?: Constructor<T>;
+  Fallback?: Constructor<Codec>;
```
amaurymartiny: needed to change here. This is more correct, as e.g. EventRecord.Fallback == EventRecord0To76, which is not T

PR #1205, opened 2019-07-26
Move Extrinsic/Method to generic

jacogr:
Also includes #1203 (so merging with master is only done in one)


packages/types/src/primitive/index.ts
```diff
@@ -25,10 +25,8 @@ export { default as i64 } from './I64';
 export { default as Fixed64 } from './I64Fixed';
 export { default as i128 } from './I128';
 export { default as i256 } from './I256';
-export { default as Method } from './Method';
 export { default as Moment } from './Moment';
 export { default as Null } from './Null';
-export { default as Origin } from './Origin';
 export { default as StorageHasher } from './StorageHasher';
 export { default as StorageKey } from './StorageKey';
```
jacogr: These 2 are left - StorageHash & StorageKey - I think they are relatively innocent since only user internally (but should be moved), I just don't have energy anymore...

PR #1207, opened 2019-07-27
createType allows multiple params

jacogr:
No effects on the exposed public `createType` interface, internal (i.e. unsafe) unaffected.


packages/api-metadata/src/extrinsics/fromMetadata/createUnchecked.ts
```diff
@@ -33,7 +33,7 @@ export default function createDescriptor (
       `Extrinsic ${section}.${funcName} expects ${expectedArgs.length.valueOf()} arguments, got ${args.length}.`
     );
 
-    return new (ClassOf('Method'))({
+    return createType('Method', {
```
jacogr: Because of these. There are also a places where this affects apps (and extension), so it is really worth-while having this work properly and without surprises.

PR #1211, opened 2019-07-27
Remove TypeDef duplication

jacogr:


packages/types/src/types.ts
```diff
@@ -115,8 +115,6 @@ export interface Constructor<T = Codec> {
 
 export type ConstructorDef<T = Codec> = Record<string, Constructor<T>>;
 
-export type TypeDef = Record<string, Codec>;
```
jacogr: Exported from codec/types (unused as an actual export)

PR #1217, opened 2019-07-27
Linked queries accept params again

### LINKED ISSUE #1170: Linked maps don't respect parameter inputs
Doing the query on the UI for a single value, I would expect the default to come back, however, everything is returned -

image

Need to confirm with e2e test -

- when no params are passed - return the full list
- when a param is passed - return a single entry

jacogr:
Closes #1170

On Alex, this work now -

```js
// return all
console.log(await api.query.staking.validators())

// return known (with value)
console.log(await api.query.staking.validators('DDB2HMziKeVb4HuQhmVg94zvDkPgmkSjLfhTGJUbyYPGsfy'))

// return known (with default)
console.log(await api.query.staking.validators('DMkn1GtZRcaPAJcHn8XyEgSrTNbo58FPJEYq9KcgZJTg3GK'))

// return non-existing (fallback)
console.log(await api.query.staking.validators('EG2FrW7bE4GnXP8Di3ewigfiNSdzzsQLXSPqs2kDzhGpmK8'))
```

### DIFF
--- a/packages/types/src/codec/createType.spec.ts
+++ b/packages/types/src/codec/createType.spec.ts
@@ -273,22 +273,6 @@ describe('createType', (): void => {
     ).toEqual(new Uint8Array([0x00, 0x12, 0x00, 0x23, 0x00, 0x45, 0x00, 0x67]));
   });
 
-  it('throw error when create base is a StorageData with null value and isPedantic is true', (): void => {
-    const base = createType('StorageData', null);
-
-    expect(
-      (): Codec => createTypeUnsafe('DoubleMap<Vec<(BlockNumber,EventIndex)>>', [base], true)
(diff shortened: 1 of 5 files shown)


packages/types/src/codec/createType.spec.ts
```diff
@@ -273,22 +273,6 @@ describe('createType', (): void => {
     ).toEqual(new Uint8Array([0x00, 0x12, 0x00, 0x23, 0x00, 0x45, 0x00, 0x67]));
   });
 
-  it('throw error when create base is a StorageData with null value and isPedantic is true', (): void => {
```
jacogr: Invalidated since we now check the stripped values

PR #1225, opened 2019-07-29
Update examples for 0.90

jacogr:


riusricardo:
I'm trying to execute the 05_read_storage example and I get a function error. Should this issue needs to be handled separately or this PR could include a fix to that example?

```
TypeError: blockPeriod.toNumber is not a function
```
I'm testing it on the Substrate v1.0 branch.

jacogr:
Those are going difficult, actually. Between 1.x and 2.x things have moved around a lot. So in things using the API with both 1.x & 2.x support, there are a lot of feature detection going on.

For these sample though, they should work on both 1.x and 2.x - so whatever is picked as samples should just work. I'll take a peek when I have a gap to try and find something else suitable :(

jacogr:
On the above, I actually think that may be a new issue (introduced in 0.9, should be adapted here), will take a peek.

riusricardo:
Ok, thanks! I can leave that example alone and test it later :)

PR #1229, opened 2019-07-30
Generic Asset module type definitions

### LINKED ISSUE #1228: Types for GenericAsset module
Add runtime type for [srml-generic-asset module](https://github.com/paritytech/substrate/tree/master/srml/generic-asset)

xlc:
Add type definitions for Generic Asset module. Also fixed a missing import issue for code generation logic.
Closes #1228

I spent some time debugging issue caused by out or order type definitions. It will be good if the type definitions are order independent so we don't need to worry about using a type orders. But I guess with type-metadata, the definition file will be automatically generated which shouldn't have this issue anymore.


xlc:
Is possible to make codeclimate ignore generated files?

jacogr:
Yes and yes.

- I was looking at the order and then decide it is wasted effort (but it is a bit of a nightmare to track as well).
- The generated files should be ignored, believe I had that in, but seems to have snuck back

PR #1240, opened 2019-07-30
Re-add StorageData from 0.82.0-beta.96 (not an alias)

jacogr:


Stefie:
Calling the state methods with a `StorageKey` is working again with this fix!

However, the other odd behavior with the undefined `meta, method, outputType, section` of the StorageKeys remains, but that seems to be unrelated.

Screenshot 2019-07-31 at 08 45 21

Stefie:
Do you know where the `4` came from? That drove me crazy!

jacogr:
The undefined will be in there since the input is a key from storage itself, i.e. it doesn't have the stuff associated with it populated. The 4 was due to the subarray part inside Bytes being followed.

PR #1243, opened 2019-07-31
Update DoubleMap e2e test for new setKeys

jacogr:
... it doesn't pass anymore...


jacogr:
Non-functionality change, going to merge so can actually use this with further investigation

PR #1248, opened 2019-08-01
Support new v3 extrinsic (payload with genesisHash)

jacogr:
- Required for https://github.com/paritytech/substrate/pull/3286


jacogr:
My concerns here is that this stuff is turning into a mini-Metadata (so really needs to be more generic here). However, since we will drop 1 & 2 at the same time as metadata < 7 (when we drop 1.x support), it is ok. For now.

PR #1249, opened 2019-08-01
Removing approvalFlagsToBool and fixing vote marshalling.

jnaviask:
We somehow had a duplicate functionality to `approvalFlagToBool` written in `approvalFlagsToBools` that performed the logic incorrectly. This patch fixes that by removing the duplicate function and switching the logic to work correctly.


jacogr:
We should have tests for this so we don't end up with `[true, false]` turning into `[false, true]`

@jnaviask Would appreciate a sample output (JSON or values) and can do the dirty on that.

Issue #1252, opened 2019-08-02
Inconvenience using new signer interface


ianhe8x:
I'm trying to migrate my signer lib to use the new signer interface, then i noticed the dependency requirement changed. Before, my signer lib didn't have to depend on `@polkadot/api` or `@polkadot/types`, extrinsic is passed in and i only need to get the key pair what it's asked for.
But now, I need SignaturePayload from `@polkadot/types`, which means future change from `@polkadot` might break my signer lib.

so, can we make SignerPayload support toU8a(), which return the same encoded u8a we passed to IKeyringPair.sign(). (cennznet/api/packages/types/src/primitive/Extrinsic/util.ts)

then i can remove the dependency of `@polkadot/types` from my lib.

ianhe8x:
A second thought, maybe rather than having function toU8a(), we just have u8a: Uint8Array.

=============
And I understand maybe it's inevitable for signer to instantiate api, since signer need to present to user what user's going to sign about, which requires `@polkadot/api|types`. And signer need a way to verify the given u8a is generated from the rest fields in the payload.

let's have some discussions here.

ianhe8x:
Allow me categorize signers into dumb signer and smart signer.

The one i mentioned in the first floor is a dumb signer which is supposed to be used with `@polkadot/api`, running in the same process. In this case I trust whatever u8a data that api passes to me, and I don't have interactions with end user. In this case, the extra u8a field is useful, and simplify my lib dependency.

For a smart signer, usually the SignPayload is transmitted via untrusted approachs. (cable, internet or something else), and it is very likely need user's interactions (to confirm). For these signers, they need to depend on `@polkadot/api | types` anyway.

jacogr:
The only suggestion is to the `signPayload` (as-is) and `signPayloadU8a` where the signer can implement either. API detects, uses or throws if none available. 

The suggested solution in the PR is not quite there since it doubles-up.

ianhe8x:
what if I revert my pr to the first version, which adds toU8a(), then it won't double data in the payload

jacogr:
I wouldn't mess about too much with the `toPayload` since effectively it is a poorly-named JSON.  (toJSON will not work at all in that case, well, it would, with some check overrides)

`signRaw({ type: 'payload' | 'data', data: '0x....' })` (also a serialized input, not 100% on `signRaw`, but haven't had coffee yet...)

- `data` - arbitrary data, identified by the `type`
[...]

PR #1253, opened 2019-08-02
SignerPayload support toU8a()

ianhe8x:
see #1252


packages/api/src/SignerPayload.ts
```diff
@@ -52,7 +52,10 @@ export default class SignerPayload extends Struct.with({
       method: method.toHex(),
       nonce: nonce.toHex(),
       tip: tip.toHex(),
-      version: version.toNumber()
+      version: version.toNumber(),
+      get encoded (): Uint8Array {
```
jacogr: Effectively we now have the same info in the same payload - twice. This doesn't quite work.

jacogr:
Replaced by #1256

PR #1258, opened 2019-08-02
Stripped metadata

jacogr:


jacogr:
@Tbaut & @yjkimjunior Do you have an appetite for this as used in the Mobile signer? It will need some extra elbow-grease here 

It won't make it for 0.90.1 - as of now, well, it proves that it can . strip things significantly, but since the Metadata, specifically `Type` was never meant to serialize again, so work required. 

Nothing critical, as couple of hours to get it to be sane and working, but rather close and forget about . it if we are not going down this route - instead of having yet another PR hanging in limbo.

Tbaut:
Dumb question, how likely is it that the metadata will change for things like balances.transfer?

I am on the fence with this upgradability of Signer. This include the metadata upgrade.

jacogr:
Generally (and I say generally since there are no checks for this), we always tried (in the past) to add new modules to the end. Same with methods in the actual modules. However this relies on developers following a specific pattern.

In most cases, things _should_ be safe. But there are no guarantees at all.

Tbaut:
Ok, if there are extraordinary changes happening, then we can do extraordinary updates to Signer. For the time being I'm for keeping things as simple as possible.

jacogr:
Ok, will close this one - we can re-visist at any point in time.

PR #1265, opened 2019-08-03
Pass online/offline status indicators with staking accounts

kwingram25:
This turned out to be a ton of work :)

Pass combined recentlyOffline / receivedHeartbeats object with all accounts retrieved for staking.
- Not sure of the overhead concerns; Alexander loads very slowly, but it already did anyways
- Alternative is to rely solely on parsing recent events? I'm working on an apps update that uses both this and event checking to get specific BlockNumber


packages/api-derive/src/staking/info.ts
```diff
@@ -104,94 +107,126 @@ function unwrapSessionIds (stashId: AccountId, validatorIds: AccountId[], auraId
...
+          of([
+            nominators,
+            sessionId,
+            nextSessionId
+          ]),
+          eraLength(api)(),
+          bestNumber(api)(),
+          recentlyOffline(api)(),
+          receivedHeartbeats(api)([stashId, controllerId, sessionId, nextSessionId]),
+          receivedHeartbeats(api)(nominators),
```
jacogr: Not for norminators.

packages/api-derive/src/staking/info.ts
```diff
@@ -104,94 +107,126 @@ function unwrapSessionIds (stashId: AccountId, validatorIds: AccountId[], auraId
...
+        return combineLatest([
+          of([
+            nominators,
+            sessionId,
+            nextSessionId
+          ]),
+          eraLength(api)(),
+          bestNumber(api)(),
+          recentlyOffline(api)(),
+          receivedHeartbeats(api)([stashId, controllerId, sessionId, nextSessionId]),
```
jacogr: Only for stash

packages/api-derive/src/staking/controllers.ts
```diff
@@ -5,25 +5,62 @@
...
           ])
         ),
+        switchMap(([recentlyOffline, stashIds, controllerOptions]): Observable<[DerivedRecentlyOffline, AccountId[], (AccountId | null)[], boolean[], boolean[]]> => {
+          const controllerIds = controllerOptions.map((option): AccountId | null => option.unwrapOr(null));
+          return combineLatest([
+            of(recentlyOffline),
+            of(stashIds),
+            of(controllerIds),
+            receivedHeartbeats(api)(stashIds),
+            receivedHeartbeats(api)(controllerIds)
```
jacogr: Not for controllers.

packages/api-derive/src/session/validators.ts
```diff
@@ -0,0 +1,45 @@
...
+    combineLatest([
+      recentlyOffline(api)(),
+      api.query.session.validators() as any as Observable<AccountId[]>
+    ])
+      .pipe(
+        switchMap(([recentlyOffline, validatorIds]): Observable<[DerivedRecentlyOffline, AccountId[], boolean[]]> =>
+          combineLatest([
+            of(recentlyOffline),
+            of(validatorIds),
+            receivedHeartbeats(api)(validatorIds)
```
jacogr: Here it is correct.

packages/api-derive/src/imOnline/receivedHeartbeats.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    return api.query.imOnline && api.query.imOnline.receivedHeartbeats
+      ? api.query.session.currentIndex<SessionIndex>()
+        .pipe(
+          switchMap((currentIndex): Observable<Bytes[]> =>
+            combineLatest(
+              addresses.map((address): Observable<Bytes> =>
+                address
+                  ? api.query.imOnline.receivedHeartbeats<Bytes>(currentIndex.toString(), address.toString())
```
jacogr: This should be a multi query.

packages/api-derive/src/staking/info.ts
```diff
@@ -203,8 +236,8 @@ export function info (api: ApiInterfaceRx): (_accountId: Uint8Array | string) =>
 
     return (
       api.queryMulti([
-        [api.query.staking.bonded, accountId], // try to map to controller
-        [api.query.staking.ledger, accountId] // try to map to stash
+        [api.query.staking.bonded, accountId.toString()], // try to map to controller
```
jacogr: Why the `toString` here? (And elsewhere)

packages/api-derive/src/staking/info.ts
```diff
@@ -104,94 +107,124 @@ function unwrapSessionIds (stashId: AccountId, validatorIds: AccountId[], auraId
...
+  return combineLatest([
+    // FIXME while we have 2.x and 1.x support, don't add this to .multi -
+    // should be added when only 2.x
+    api.query.aura && api.query.aura.authorities
+      ? api.query.aura.authorities<Vec<AccountId>>()
+      : of([] as AccountId[]),
+    api.queryMulti([
+      api.query.session.queuedKeys
+        ? [api.query.session.queuedKeys]
+        : [api.query.session.nextKeyFor, controllerId.toString()],
```
jacogr: Why the toString?

packages/api-derive/src/staking/info.ts
```diff
@@ -104,94 +107,124 @@ function unwrapSessionIds (stashId: AccountId, validatorIds: AccountId[], auraId
...
+        return combineLatest([
+          of([
+            nominators,
+            sessionId,
+            nextSessionId
+          ]),
+          eraLength(api)(),
+          bestNumber(api)(),
+          recentlyOffline(api)(),
+          receivedHeartbeats(api)([stashId]),
```
jacogr: We should really drop these completely.

packages/api-derive/src/session/validators.ts
```diff
@@ -0,0 +1,45 @@
...
+import { ApiInterfaceRx } from '@polkadot/api/types';
+import { DerivedRecentlyOffline, DerivedStakingAccount, DerivedStakingAccounts } from '../types';
+
+import { receivedHeartbeats } from '../imOnline';
+import { recentlyOffline } from '../staking';
+import { drr } from '../util/drr';
+import { addOnlineStatusToStakingAccount } from '../util/addOnlineStatusToStakingAccount';
+
+/**
+ * @description From the list of stash accounts, retrieve the list of controllers
```
jacogr: Not sure the description here matches.

jacogr:
Would def stick to only events and keep that on the UI. The recentlyOffline is good.

packages/api-derive/src/staking/info.ts
kwingram25: I was trying to figure out why the switchMap wouldn't go through and thought it was related to that :)

jacogr:
I merged it with master and cleaned up slightly. There is something here however that makes the apps UI unhappy - on master this shows up, while the `staking.validators` query only has the active account, not both.

image

Due to it using the `derive.staking.controllers` query, which seems to have broken compatibility.

kwingram25:
Yes, have a big apps update as well for things like this (though it will have to be changed some now)

jacogr:
The issue is that I don't want to break the existing derives - so adding more info is fine, making the derives break in incompatible ways, are not. (It is not an apps-specific thing, this is used by other teams). 

An example of this was reverting the `stashId` -> `stash` changes, that is an incompatible change, rather the `offline` was added to the structure.

packages/api/test/e2e/api-derive/promise-dev.spec.ts
```diff
@@ -73,7 +73,7 @@ describeE2E({
         }
 
         expect(info.accountId.eq(accountId)).toBe(true);
-        expect(info.controllerId.eq('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY')).toBe(true);
+        expect(info.controllerId!.eq('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY')).toBe(true);
```
jacogr: A side note, we need to not add any more of these, i.e. https://github.com/polkadot-js/api/blob/master/.eslintrc.js#L10

This is the TypeScript equivalent of raw `unwrap()` without a check in Rust. It is passable in tests, however we need to remove that rule override since it is a code smell. (Disallowed by default by the standard rules, e.g. the recommended set)

packages/api-derive/src/session/validators.ts
```diff
@@ -0,0 +1,34 @@
...
+import { addOnlineStatusToStakingAccount } from '../util/addOnlineStatusToStakingAccount';
+
+/**
+ * @description Retrieve the current session validators and any offline status reports
+ */
+export function validators (api: ApiInterfaceRx): () => Observable<DerivedStakingAccounts> {
+  return (): Observable<DerivedStakingAccounts> =>
+    combineLatest([
+      recentlyOffline(api)(),
+      api.query.session.validators() as any as Observable<AccountId[]>
```
jacogr: Ok, this is dangerous - this returns different information for 1.x vs 2.x

- 1.x - controllers
- 2.x - stashes

So this can only be used in the current form on 2.x chains. Can add it, but I would strongly suggest not using this in the apps UI at all since it will work and not work depending on chain.

PR #1266, opened 2019-08-03
Retrieve nextKeys, only deal with stash

jacogr:


jacogr:
The screw-up with this atm - on 1.x session.validators return the controllers on 2.x it returns the stash.

PR #1267, opened 2019-08-04
WIP: Adding of rollup version of api

retotrinkler:
### Added

- Add rollup dependencies and script to package.json
- Add rollup.api.js config file
- Add compilerOption ES2015 to ts config file

### Updated
- Update .gitignore


retotrinkler:
actually don't know if this worked. will close it to not spam your PR list

jacogr:
At the very least, what it did is uncover some circular-dep-horrors - so even just for that, it was worth it's weight in gold.

retotrinkler:
thanks :)

PR #1271, opened 2019-08-06
Add Wasm Provider to RPC-Provider

yjkimjunior:
Context: Pierre’s Substrate light client in browser -
Demo:  http://substrate-node.cleverapps.io/, Soruce: https://github.com/paritytech/substrate/commit/67e82c2f788ef339bef499b8a9274bfc84710769 - is in need of a UI so this PR is working on adapting @polkadot/rpc-provider to also handle connecting to the bundled wasm client.

There is a fair bit of code reuse between this and WSProvider, and it is more or less the same with the main difference being that instead of making RPCs on a WS connection, WasmProvider should call the matching wasm functions.

The wasm pkg included is meant to only be for testing, and it practice the bundled wasm would be passed in as a constructor argument.

Opening the draft PR early for potential feedback and collaboration, as I'm currently also a bit crunched on time trying to get the UOS stuff out the door.

### DIFF
--- a/packages/rpc-provider/src/index.ts
+++ b/packages/rpc-provider/src/index.ts
@@ -4,3 +4,4 @@
 
 export { default as HttpProvider } from './http';
 export { default as WsProvider } from './ws';
+export * from './wasm';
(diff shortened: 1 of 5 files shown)


yjkimjunior:
closing since https://github.com/axelchalon/substrate-browser/blob/master/src/WasmProviderLite.ts

Issue #1290, opened 2019-08-10
Type overrides for dependents


jacogr:
... logging in here since, well, need to think about this.

Basically, this doesn't work - 

- https://github.com/polkadot-js/api/blob/master/packages/api/src/Base.ts#L577
- https://github.com/polkadot-js/api/pull/1284/files#diff-95597f67ed68d7dffc27ba1f0501e5b1R585

The issue is the the types are injected, registered and anything dependent has a created type with the initial locked-in class.

jacogr:
Solved with lazy-creation & new injection logic. (The above PR laid most of the groundwork)

PR #1291, opened 2019-08-10
Lazily create registry types (also solves resolution order)

jacogr:
- Should solve https://github.com/polkadot-js/api/issues/1290
- The one big benefit here is that registration is not order-dependent anymore


packages/types/src/interfaces/babe/definitions.ts
```diff
@@ -5,12 +5,12 @@
...
   types: {
     BabeWeight: 'u64',
-    SlotNumber: 'u64',
     RawBabePreDigest: {
       slotNumber: 'SlotNumber',
       authorityIndex: 'u32', // AuthorityIndex (also in aura, not same size there)
       vrfOutput: 'H256', // should be '[u8; 32]' (generator support lacking here)
       vrfProof: 'H256' // should be '[u8; 32]'
-    }
+    },
```
jacogr: From here going forward, we are only re-arranging things to be alphabetical - since we can now assign in any order, make sure that this is actually the case (testing needs to be in-place, so if the ordering is an issue, this would create chaos...)

PR #1300, opened 2019-08-12
Add Forcing enum (staking)

jacogr:
As introduced in https://github.com/paritytech/substrate/pull/3374


packages/types/src/interfaces/staking/definitions.ts
```diff
@@ -14,6 +14,13 @@ export default {
       own: 'Compact<Balance>',
       others: 'Vec<IndividualExposure>'
     },
+    Forcing: {
```
jacogr: Only this, the rest is generated.

PR #1304, opened 2019-08-13
Payload Call as Bytes, test serialization

jacogr:
cc @yjkimjunior


yjkimjunior:
thanks!! no more padding :)

PR #1315, opened 2019-08-16
Remove deprecated Metadata getters

jacogr:
Additionally move some (low-hanging) type classes from Metadata to interface definitions


jacogr:
I've just started going through CodeClimate, all the F's and working my way downwards when I'm waiting for unrelated stuff to get in. Each one requires a bit of a re-think (for the better) in various areas.

PR #1320, opened 2019-08-17
Generate types from new contract ABI JSON

kwingram25:
Just starting, waiting on fixes to generated JSON as discussed with Robin
- Unclear where new class MetaRegistry should live - replacing or different from Metadata folder?


jacogr:
@kwingram25 As for the registry question - for now, keep it in . `package/api-contracts` - it will supplement the rest when we have updated metadata. As a start want to keep the changes as localized as possible

kwingram25:
that was a frustrating merge :)

jacogr:
@kwingram25 Sorry about that merge - but really needed to start making . some sense of createType, as a blob of code is was not really easily -

- extensible
- understandable
- testable

i.e. your changes are actually now more side piggy-backing on that, we just crossed wires somewhere in the middle.

packages/api-contract/src/Abi.ts
```diff
@@ -16,16 +17,41 @@ export default class ContractAbi implements InterfaceAbi {
...
 
-  public constructor (abi: ContractABI) {
-    validateAbi(abi);
+  public readonly version: AbiVersion = AbiVersion.v2;
 
-    this.abi = abi;
-    this.deploy = createMethod('deploy', abi.deploy);
+  public constructor (json: ContractABIV1 | ContractABIV2) {
+    let abi;
+    switch (json.version) {
```
jacogr: (ni) extra blank space after

packages/api-contract/src/Abi.ts
```diff
@@ -16,16 +17,41 @@ export default class ContractAbi implements InterfaceAbi {
...
+      case AbiVersion.v2:
```
jacogr: Like we actually do in the types constructors, I would probably guard against the constructor getting . to big here.

So rather split this with 2 static functions, both returning `[abi, deploy, messages]` and then . do something like . the following -

```js
if (json.version === AbiVersion.v2) {
  [this.abi, this.deploy, this.messages] = Abi.decodeV2(json);
} else {
  [this.abi, this.deploy, this.messages] = Abi.decodeV1(json);
}
```

packages/api-contract/src/ContractRegistry.ts
```diff
@@ -0,0 +1,59 @@
+import { ContractABI, ContractABIArg, ContractABIFnArg, ContractABIMethod, ContractABIMethodBase } from './types';
```
jacogr: (nip) file header missing

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,298 @@
+import * as t from '@polkadot/types/types';
```
jacogr: (nit) file header

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,298 @@
...
+    const sub = structFields.map(
+      (field: t.MetaTypeDefStructField): t.TypeDef => {
+        const { name: nameIndex, type: fieldTypeIndex } = field;
+        return {
+          ...this.typeDefFromMetaTypeAt(fieldTypeIndex),
+          name: this.stringAt(nameIndex)
+        };
+      }
+    );
+    return {
```
jacogr: (nit) extra blank line before, split blocks

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,298 @@
...
+        return {
+          info: t.TypeDefInfo.Tuple,
+          type: encodeTuple(sub),
+          sub
+        };
+      }
+
+      // builtin array
+      const { 'array.type': vecTypeIndex, 'array.len': vecLength } = id as t.MetaTypeIdArray;
+      if (vecTypeIndex) {
```
jacogr: (nit) split vars from ifs

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,298 @@
...
+          if (unitNameIndex) {
+            return {
+              info: t.TypeDefInfo.Plain,
+              name: this.stringAt(unitNameIndex),
+              type: 'Null'
+            };
+          }
+
+          const { 'tuple_struct_variant.name': tupleStructVariantNameIndex } = variant as t.MetaTypeDefEnumVariantTupleStruct;
+          if (tupleStructVariantNameIndex) {
```
jacogr: (nit) as above

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,298 @@
...
+
+      return {
+        info: t.TypeDefInfo.Enum,
+        type: encodeEnum(sub),
+        sub
+      };
+    }
+
+    // clike enum
+    const { 'clike_enum.variants': clikeEnumVariants } = def as t.MetaTypeDefClikeEnum;
```
jacogr: (nit) as above

jacogr: At this point this has turned into a real absolute monster function. Can we clean it up to be in manageable chunks, would rather not come back at some point to make CodeClimate happy

packages/api-contract/src/types.ts
```diff
@@ -68,12 +96,69 @@ export interface ContractABIFn extends ContractABIMeta {
   (...args: CodecArg[]): Uint8Array;
 }
 
+// export interface ContractABIArgV2 {
```
jacogr: We should rather forgo the commented out types - still marked WIP, so still ok, but something to keep in mind and just clean up before the flag is removed.

packages/types/src/codec/create/types.ts
```diff
@@ -32,11 +32,102 @@ export interface TypeDefExtVecFixed {
...
-  ext?: TypeDefExtVecFixed; // add additional here as required
+  ext?: TypeDefExtVecFixed | TypeDefExtEnumDiscriminant; // add additional here as required
   name?: string;
+  namespace?: string;
   type: string;
   sub?: TypeDef | TypeDef[];
 }
+
+export type TypeIndex = number;
+export type StringIndex = number;
```
jacogr: (nit) Extra blank like before, jsut aligning with the rest.

packages/types/src/codec/utils/encodeTypes.ts
```diff
@@ -18,25 +26,25 @@ export function encodeEnum (sub: TypeDef[]): string {
...
-      .map(({ name, type }: TypeDef): string => `"${name}": ${type}`)
+      .map((type: TypeDef): string => `"${type.name}": "${encodeTypeWithParams(type)}"`)
       .join(', ')
-  }}`;
+  }} }`;
 }
 
 export function encodeStruct (sub: TypeDef[]): string {
-  return `{${
+  return `{ "_enum": {${
```
jacogr: This can't be right or am I missing something, i.e. struct !== enum

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
+// Copyright 2017-2019 @polkadot/api-contract authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import * as t from '@polkadot/types/types';
```
jacogr: Import explicitly what we are using, i.e. `import { MetaRegistryJson, ... } from '@polkadot/types/types';`

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+
+import { assert } from '@polkadot/util';
+import { encodeEnum, encodeStruct, encodeTuple } from '@polkadot/types';
+
+const SPECIAL_TYPES = ['AccountId', 'Balance'];
```
jacogr: Not 100% sure what "special types" mean here? (So need to be more descriptive). However, assuming these are the types that have special display/encoding requirements, there are a couple more - `Address` & `AccountIndex`

(These latter 2 are not strictly in-use atm, but they do fall into exactly the same bucket)

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  private _types: t.MetaType[] = [];
+
+  public typeDefs: t.TypeDef[] = [];
+
+  public constructor ({ registry: { strings, types } }: t.MetaRegistryJson) {
+    this._strings = strings;
+    this._types = types;
+
+    this._types.forEach((_, index: number): void => this.setTypeDefAtIndex(index + 1));
```
jacogr: Just a short comment just above this of why we are doing this would help.

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+    );
+  }
+
+  public hasTypeDefAt (index: t.TypeIndex): boolean {
+    return !!this.typeDefs[index - 1];
+  }
+
+  public typeDefAt (index: t.TypeIndex, extra: Pick<t.TypeDef, never> = {}): t.TypeDef {
+    assert(this.hasTypeDefAt(index), `Invalid TypeDef index requested: ${index}`);
+    return {
```
jacogr: (nit) extra blank line before the return for readability.

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+      return {};
+    }
+
+    const {
+      'custom.name': nameIndex,
+      'custom.namespace': namespaceIndices,
+      'custom.params': paramsIndices
+    } = id as t.MetaTypeIdCustom;
+
+    if (nameIndex) {
```
jacogr: Probably would swap this check to be `if (!nameIndex) { return {} }`

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    return {};
+  }
+
+  private typeDefDefFields (metaType: t.MetaType, typeIndex?: t.TypeIndex): Pick<t.TypeDef, never> {
+    assert(!(metaType.def as t.MetaTypeDefUnion)['union.fields'], 'Invalid union type definition found');
+
+    let typeDef: Pick<t.TypeDef, never> = {};
+    if (this.isBuiltin(metaType)) {
```
jacogr: (nit) extra blank line before

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+      typeDef = this.typeDefForBuiltin(metaType, typeIndex);
+    }
+
+    if (this.isEnum(metaType)) {
```
jacogr: These should all be `else if` rather? (i.e. it can only be one of these)

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+      };
+    }
+
+    return {};
+  }
+
+  private typeDefDefFields (metaType: t.MetaType, typeIndex?: t.TypeIndex): Pick<t.TypeDef, never> {
+    assert(!(metaType.def as t.MetaTypeDefUnion)['union.fields'], 'Invalid union type definition found');
+
+    let typeDef: Pick<t.TypeDef, never> = {};
```
jacogr: Just `let typeDef: Pick<t.TypeDef, never>`;

(Then remove the `assert` below and make a final leg with `else` after the last `else if` where we throw.

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+    if (this.isStruct(metaType)) {
+      typeDef = this.typeDefForStruct(metaType.def as t.MetaTypeDefStruct);
+    }
+
+    if (this.isTupleStruct(metaType)) {
+      typeDef = this.typeDefForTupleStruct(metaType.def as t.MetaTypeDefTupleStruct);
+    }
+
+    assert(Object.keys(typeDef).length > 0, 'Error decoding metadata type');
+    return typeDef || {};
```
jacogr: As per the comment above, when we get here, it should be defined. We could generally actually just get rid of the assinments here and the just 

```js
  else if (this.isTupleStruct(metaType)) {
      return this.typeDefForTupleStruct(metaType.def as t.MetaTypeDefTupleStruct);
    }
```

packages/api-contract/src/RxBlueprint.ts
```diff
@@ -46,7 +46,7 @@ export default class Blueprint extends RxBase {
   public deployContract (endowment: number | BN, maxGas: number | BN, ...params: any[]): BlueprintCreate {
     const signAndSend = (account: IKeyringPair | string | AccountId | Address): BlueprintCreateResultSubscription => {
       return this.apiContracts
-        .create(endowment, maxGas, this.codeHash, this.abi.deploy(...params))
+        .create(endowment, maxGas, this.codeHash, this.abi.constructors[0](...params))
```
jacogr: We should really add a `TODO` here - we probably want to specify exactly the constructor to be used in the longer-term, not just default to the first.

packages/api-contract/src/validation.ts
```diff
@@ -1,50 +1,50 @@
...
-export function validateAbi (abi: ContractABI): void {
-  const unknownKeys = Object.keys(abi).filter((key): boolean => !['deploy', 'messages', 'name'].includes(key));
-
-  assert(unknownKeys.length === 0, `Unknown fields ${unknownKeys.join(', ')} found in ABI`);
-  assert(abi.deploy && abi.messages && isString(abi.name), 'ABI should have deploy, messages & name sections');
-
-  validateDeploy(abi);
-  validateMethods(abi);
-}
+//
```
jacogr: Just remove the file outright.

packages/types/src/codec/create/getTypeDef.ts
```diff
@@ -59,15 +59,15 @@ function _decodeStruct (value: TypeDef, type: string, subType: string): TypeDef
...
+    getTypeDef(parsed[name], { name })
   );
 
   return value;
 }
 
 // decode a fixed vector, e.g. [u8;32]
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-function _decodeFixedVec (value: TypeDef, type: string, subType: string): TypeDef {
+function _decodeFixedVec (value: TypeDef, type: string): TypeDef {
```
jacogr: If the unsued var is removed, the eslint comment can also go. (Would generally prefer to keep it though to keep all the signatures consistent, rather `_` it is not used)

packages/types/src/codec/utils/typeDefToString.ts
```diff
@@ -0,0 +1,18 @@
...
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { TypeDef } from '../types';
+
+import { isString } from '@polkadot/util';
+
+export function typeDefToString ({ params = [], type, displayName = type }: TypeDef): string {
+  const paramsText = params.map(param => typeDefToString(param)).join(', ');
+  if (['Option', 'Result', 'Vec'].includes(displayName)) {
```
jacogr: (nit) extra blank space before

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  private isBuiltin ({ def }: t.MetaType): boolean {
+    return def === 'builtin';
+  }
+
+  private typeDefForBuiltin ({ id }: t.MetaType, typeIndex?: t.TypeIndex): Pick<t.TypeDef, never> {
+    if (!id) {
+      throw new Error('Invalid builtin type metadata found');
+    }
```
amaurymartiny: could use assert here too

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+    );
+
+    return {
+      info: t.TypeDefInfo.Enum,
+      type: encodeEnum(sub),
+      sub
+    };
+  }
+
+  private isClikeEnum ({ def }: t.MetaType): boolean {
```
amaurymartiny: maybe a typeguard here would be useful? `: def is t.MetaTypeDefClikeEnum` (if that's possible, because `def` is nested). So that we can avoid some `as ...`

packages/api-contract/src/ContractRegistry.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  public convertAbi ({ contract, storage }: t.ContractABIPre): t.ContractABI {
+    return {
+      contract: this.convertContract(contract),
+      storage: this.convertStorage(storage)
+    };
+  }
+
+  public convertType ({ ty, display_name: displayNameIndices }: t.ContractABITypePre): TypeDef {
```
amaurymartiny: While scanning through quickly, I had a few of those moments where i needed to take a couple of minutes to understand what a function was doing.

Some small comments on functions with not-explicit names would be much appreciated!

packages/api-contract/src/MetaRegistry.ts
```diff
@@ -0,0 +1,333 @@
...
+    const { 'clike_enum.variants': clikeEnumVariants } = def as t.MetaTypeDefClikeEnum;
+    return !!clikeEnumVariants;
+  }
+
+  private typeDefForClikeEnum (def: t.MetaTypeDefClikeEnum): Pick<t.TypeDef, never> {
```
amaurymartiny: What is `Pick<T, never>` doing?

jacogr:
@amaurymartiny On ditching `TypeDef`, I wouldn't do it here since I'd rather not touch too much "other stuff outside scope", but yes, I had simmilar thoughts - we need to combine these. 

At the same time, don't want to move _too_ early on that, I'd rather first . see what the metadata version looks like - I'm worried that that may introduce "other" things.

amaurymartiny:
Agreed on what you said.

packages/types/src/codec/utils/encodeTypes.ts
kwingram25: You're right, misunderstood your decodeStruct

packages/api-contract/src/RxBlueprint.ts
kwingram25: Good call, put something in there I think is right but lmk

packages/api-contract/src/validation.ts
kwingram25: Missed that :)

packages/types/src/codec/utils/typeDefToString.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { TypeDef } from '../types';
+
+import { isString } from '@polkadot/util';
+
+export function typeDefToString ({ params = [], type, displayName = type }: TypeDef): string {
```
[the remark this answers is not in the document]
kwingram25: Oh nice! Looks about right to me, haven't dug in but what is setImports about/for?

packages/api-contract/src/MetaRegistry.ts
kwingram25: Indicates that it has some but maybe not all fields for TypeDef since some of them may come from ```typeDefIdFields```

packages/api-contract/src/MetaRegistry.ts
kwingram25: good idea but it looks like the destructuring does mess with it :(

packages/types/src/codec/utils/typeDefToString.ts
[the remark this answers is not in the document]
amaurymartiny: Basically, when generating e.g. https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaceRegistry.ts, we `setImports` all the needed imports into a global variable, and then fs.writeFile these imports on top.

So setImport is something really specific to the `build:interfaces` scripts

packages/api-contract/src/MetaRegistry.ts
amaurymartiny: So similar to `Partial<>`?

packages/types/src/codec/utils/typeDefToString.ts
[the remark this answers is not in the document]
kwingram25: Adjusted to use formatType and optionally pass in imports to be set, and starting from either a string (your version) or a TypeDef

packages/api-contract/src/MetaRegistry.ts
jacogr: Still needs an adjustment.

packages/api-contract/src/MetaRegistry.ts
jacogr: Still needs doing.

packages/api-contract/src/MetaRegistry.ts
jacogr: Still needs adjustment.

Issue #449, opened 2018-12-07
Subscribe to runtime version, update metadata on change


jacogr:
Process -

- API subscribes to runtimeVersion
- When the runtime version changes, retrieve metadata via `state_getMetadata`
- Inject

Basically we want the initialization that happens in https://github.com/polkadot-js/api/blob/master/packages/api/src/Base.ts#L456-L462 to happen on version updates

This is the one area where oo7 does it really well, i.e. `tie` to version, update on changes - https://github.com/paritytech/oo7/blob/master/packages/oo7-substrate/src/bonds.js#L207-L211

ltfschoen:
i tried the following but it gives error that `subscribeRuntimeVersion` is not a function. I also tried adding to polkadot-js/api/packages/api-observable/src/Calls.ts
```
  subscribeRuntimeVersion = (): Observable<Header | undefined> => {
    return this._api.chain.subscribeRuntimeVersion().pipe(
      defaultIfEmpty()
    );
  }
```

```
// Required imports
const { ApiPromise } = require('@polkadot/api');

async function main () {
  const api = await ApiPromise.create();
  
  // Make a subscription to the network head
  api.rpc.chain.subscribeRuntimeVersion((version) => {
    console.log(`Runtime version is: #${version}`);
  });
}

main().catch(console.error).finally(() => process.exit());
```

jacogr:
It is available here - https://github.com/polkadot-js/api/blob/master/packages/type-jsonrpc/src/chain.ts#L72 (Added at the end of last week - and it is being moved to `state` in Substrate)

jacogr:
Needs #459

PR #1326, opened 2019-08-20
Subscribe to runtime version, refresh metadata

jacogr:
Closes https://github.com/polkadot-js/api/issues/449

### DIFF
--- a/packages/api/src/base/Init.ts
+++ b/packages/api/src/base/Init.ts
@@ -104,9 +104,7 @@ export default abstract class Init<ApiType> extends Decorate<ApiType> {
       : await this._rpcCore.state.getMetadata().toPromise();
 
     // based on the node, inject specific types - this is very specific to known chains
-    this.registerTypes(
-      SPEC_TYPES[this._runtimeVersion.specName.toString()]
-    );
+    this.registerTypes(SPEC_TYPES[this._runtimeVersion.specName.toString()]);
 
     // HACK-ish Old EventRecord format for e.g. Alex, based on metadata versions
     if (metadata.version <= 3) {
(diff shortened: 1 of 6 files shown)


packages/api/src/base/Init.ts
```diff
@@ -94,33 +98,59 @@ export default abstract class Init<ApiType> extends Decorate<ApiType> {
...
+
     return metadata;
   }
 
-  private async initFromMeta (metadata: Metadata): Promise<boolean> {
-    // HACK-ish Old EventRecord format for e.g. Alex, based on \metadata format
-    if (metadata.version <= 3) {
-      this.registerTypes(TYPES_SUBSTRATE_1);
-    }
+  private runtimeVersionObservable (): Observable<Promise<boolean>> {
```
amaurymartiny: `Observable<Promise<boolean>>` is kind of weird, why not just Observable?

jacogr: Yes, exceptionally weird. First order of business is actually getting it to work since it is the last thing remaining for 0.90 - actually haven’t figured out how this _should_ look atm, more or less going through steps to see what works and what doesn’t. 

So atm -

- have subscriptions fixed
- actually get events on UI and API on upgrades (not a pleasant testing process at all)
- have some really weird horrors in block decoding (after update)

Hopefully can get some sort of semblance of working tomorrow and start looking how to best structure after that.

jacogr: EDIT: to clarify, depending on the testing, this may not make 0.90, but needs to be in before a first runtime upgrade in Kusama (latest 2 weeks delay after 0.9). 

So trying, but as mentioned this one is a real logistics nightmare (bumping, re-compilation, disabling modules, and the verifying against actual use, etc), so doing what can be done

jacogr:
Still needs to be done, but way outdated here.

PR #1329, opened 2019-08-21
Check Enum index before value on construct

### LINKED ISSUE #1327: export EnumConstructor
Having `error TS4020: 'extends' clause of exported class '...' has or is using private name 'EnumConstructor'.` when i tried to create a class via `extends Enum.with(...)`

export EnumConstructor should solve my problem.

### LINKED ISSUE #1328: When an EnumType's value is also an Enum
```
#Enum.ts
private static decodeEnum (def: TypesDef, value?: any, index?: number | Enum): Decoded {
    if (value instanceof Enum) {
// branch 1)
      return Enum.createValue(def, value._index, value.raw);
    } else if (index instanceof Enum) {
// branch 2)
      return Enum.createValue(def, index._index, index.raw);
    } else if (isNumber(index)) {
// branch 3)
      return Enum.createValue(def, index, value);
    }
...
  }
```
When i provide both value and index, I expect the code runs into branch 3), but it actually runs into branch 1) because the subType is also a Enum.

```
    const typeRegistry = getTypeRegistry();
    const NestedEnum = ClassOfUnsafe('{ "_enum": ["A1", "A2"]}');
    typeRegistry.register({NestedEnum});

    const E = ClassOfUnsafe('{"_enum": {"A": "NestedEnum"}}');

    const a1 = new NestedEnum(0);
    console.log(a1.toString()); // output: A1
    console.log(new E(0, 0).asA.toString()); // expect: A1, works
    console.log(new E(a1, 0).asA.toString()); // expect: A1, fail
// TypeError: Cannot read property 'toLowerCase' of undefined
// at Function.toLowerCase [as createViaJSON]

```

jacogr:
- Closes #1327 
- Closes #1328
- Remove Enum allowed as index (really cannot figure this one out as to applicability)
- Notch up test-coverage for all paths (lacking of that lead us to having to fix here)

cc @ianhe8x Does this fit your needs?

### DIFF
--- a/packages/types/src/codec/Enum.spec.ts
+++ b/packages/types/src/codec/Enum.spec.ts
@@ -234,20 +234,82 @@ describe('Enum', (): void => {
           new Enum(['foo', 'bar'], 1).eq('foo')
         ).toBe(false);
       });
+
+      it('has isNone set, with correct index (i.e. no values are used)', (): void => {
+        const test = new Enum(['foo', 'bar'], 1);
+
+        expect(test.isNone).toBe(true);
+        expect(test.index).toEqual(1);
+      });
     });
   });
 
-  describe('toRawType', (): void => {
-    it('has a sane output for basic enums', (): void => {
-      expect(
-        new Enum(['foo', 'bar']).toRawType()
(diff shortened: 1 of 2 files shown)


ianhe8x:
I believe so, thanks.

PR #1330, opened 2019-08-21
Cleanups & expand tests

jacogr:


packages/types/src/primitive/Generic/Vote.ts
```diff
@@ -4,12 +4,18 @@
...
+import { isBoolean, isNumber, isU8a, isUndefined } from '@polkadot/util';
 
 import { createType } from '../../codec/create';
 import U8aFixed from '../../codec/U8aFixed';
 import Bool from '../Bool';
 
+// eslint-disable-next-line @typescript-eslint/ban-types
+type InputTypes = boolean | number | Boolean | Uint8Array | {
+  aye: boolean;
+  conviction?: number | ('None' | 'Locked1x' | 'Locked2x' | 'Locked3x' | 'Locked4x' | 'Locked5x');
```
amaurymartiny: out of curiosity, why the parenthesis here?

packages/types/src/primitive/Generic/Vote.ts
```diff
@@ -26,7 +32,7 @@ export default class Vote extends U8aFixed {
 
   private _conviction: Conviction;
 
-  public constructor (value?: any) {
+  public constructor (value?: InputTypes) {
```
amaurymartiny: This might be a good idea for all type constructors actually, so we can remove the any

packages/types/src/primitive/Generic/Vote.ts
jacogr: lol, for my sanity - wanted to group the "string values" since they belong together. (And didn't want to define yet another type that is not actually exported.)

To be brutally honest, I'm still rally unhappy with Vote as it stands. This is the 2nd round of cleanups here and it still just doesn't quite get clean. (But each bit helps)

packages/types/src/primitive/Generic/Vote.ts
jacogr: Agreed. I'm on a drive to -

- expand tests (each time I touch them, I find bugs, aka check the VecFixed that just went it - it was buggy)
- cleanup CodeClimate, get rid of the duplication, get rid of the noise, make it code to be proud of

Well, I do these mostly while waiting for other stuff, just drives me batty that I'm still finding bugs atm.

jacogr: Tracking this over time - https://codeclimate.com/github/polkadot-js/api/trends/technical_debt

PR #1331, opened 2019-08-21
EdcsaSignature to [u8; 65] (no tuple hacks)

jacogr:


jacogr:
Don't want to mess this one up, so requires an e2e test for claiming again in the UI against the . claims module. (Nothing here should mess up at all, but not going to take that chance without a proper actual claim before merging)

jacogr:
Still works, shows up correctly.

PR #1350, opened 2019-08-29
Split Submittable into actual classes

### LINKED ISSUE #1045: Rewrite SubmittableExtrinsic.ts

jacogr:
- Replaces #1075 
- Closes #1045


packages/api/src/submittable/Extrinsic.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+// The default for 6s allowing for 5min eras. When translating this to faster blocks -
+//   - 4s = (10 / 15) * 5 = 3.33m
+//   - 2s = (10 / 30) * 5 = 1.66m
+const BLOCKTIME = 6;
+const ONE_MINUTE = 60 / BLOCKTIME;
+const DEFAULT_MORTAL_LENGTH = 5 * ONE_MINUTE;
+
+const _Extrinsic: Constructor<Extrinsic> = ClassOf('Extrinsic');
```
jacogr: Need to double-check this - I'm not convinced that in all cases this is lazily created. (The registry certainly tries, however it is not always a simple problem to solve)

If not... and it cannot be convinced to do so... an option is to move the whole `class Submittable ...` into the actual final exported function, so it will only be evaluated when the function is called. (Which is not great since we end up with a monster function there)

jacogr: Yeap, it works. I injected the type `{ Extrinsic: 'ExtrinsicUnknown' }` and as expected it fails.

jacogr:
Tested with local dev, Alex & FF. (with & without extension in all cases)

PR #1358, opened 2019-09-02
Fix redeemables

jacogr:
As reported in https://github.com/polkadot-js/apps/issues/1572


Tbaut:
Working well so far. Hacking a node now to get a short redeemable time to be able to do the whole process bond -> unbond -> redeem

jacogr:
Sorry about that, I went blind momentarily when I introduced it. Will pull into apps tonight.

PR #1359, opened 2019-09-02
use bash in copy-to-apps.sh

Tbaut:
Not sure if that would work on mac, my couple weeks old linux was not happy with the parenthesis, the curly brackets etc until I used `bash`..

Take it if your mac doesn't complain with that.


jacogr:
Should just extend to common & ui as well (same type of thing, not urgent, will probably fix if I remember when I play there anyway)

PR #1360, opened 2019-09-03
move SignerPayload to @polkadot/types

ianhe8x:
part of #1355 

SignerPayload is moved to `@polkadot/types` so it can be registered to typeRegistry.
Change the original SignerPayload to SignerPayloadJSON to avoid confusion.

Because of these, this is a breaking change.

PR #1365, opened 2019-09-04
Fix raw payload

### LINKED ISSUE #1364: inconsistent between toPayload() and toRaw()
I've got Invalid Transaction (0) error when upgrading sdk, after hours debuging it turn out because of the signer I used in my test is a rawSigner.
RawSigner should be able to do the job same as PayloadSigner, but I found they are feed with different data.
I've work out a fix for this.

ianhe8x:
Fix #1364 
base on my PR #1360 

the only place changed is in `SignerPayload.ts`
```
- const data = createType('ExtrinsicPayload', payload, { version: payload.version }).toHex();
+ const data = u8aToHex(createType('ExtrinsicPayload', payload, { version: payload.version }).toU8a(true));
```

### DIFF
--- a/packages/types/src/primitive/Extrinsic/SignerPayload.ts
+++ b/packages/types/src/primitive/Extrinsic/SignerPayload.ts
@@ -2,6 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
+import { u8aToHex } from '@polkadot/util';
 import { Address, Balance, BlockNumber, Call, ExtrinsicEra, Hash, Index, RuntimeVersion } from '../../interfaces';
 import Compact from '../../codec/Compact';
 import Struct from '../../codec/Struct';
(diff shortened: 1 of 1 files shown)


jacogr:
Actually the same as the little comment PR I put in last night. 100% correct. (Need to bump the actual tests on this, this can be properly checked against ed25519 known - and would have avoided wasted time)

packages/types/src/primitive/Extrinsic/SignerPayload.ts
```diff
@@ -63,7 +64,7 @@ export default class SignerPayload extends _Payload implements ISignerPayload {
    */
   public toRaw (): SignerPayloadRaw {
     const payload = this.toPayload();
-    const data = createType('ExtrinsicPayload', payload, { version: payload.version }).toHex();
+    const data = u8aToHex(createType('ExtrinsicPayload', payload, { version: payload.version }).toU8a(true));
```
jacogr:
```suggestion
    // NOTE Explicitly pass the bare flag to the method is encoded un-prefixed (non-decodable, for signing only)
    const data = u8aToHex(createType('ExtrinsicPayload', payload, { version: payload.version }).toU8a(true));
```

PR #1370, opened 2019-09-05
Alternative lazy Submittable

jacogr:
- Alternative to https://github.com/polkadot-js/api/pull/1369
- Follow-up on #1350

cc @ianhe8x


jacogr:
Confirmed working as expected by @ianhe8x - thanks for the full e2e testing there

Issue #1372, opened 2019-09-06
Workaround for function interface override (in SubmittableExtrinsicFunction)


ianhe8x:
I have an override in my Api class which extends ApiPromise.
```
    get tx(): CennznetSubmittableExtrinsics<'promise'> {
        return super.tx as CennznetSubmittableExtrinsics<'promise'>;
    }
```
where i got an error: 
```
Type 'ICennznetExtrinsic<"rxjs">' is not assignable to type 'Call'.
              Property '_meta' is missing in type 'ICennznetExtrinsic<"rxjs">'.
```
My ICennznetExtrinsic extends CennznetSubmittableExtrinsic which extends the original SubmittableExtrinsic.

I understand typescript complains because ICennznetExtrinsic doesn't extends Call, which is weird since it doesn't complain that on SubmittableExtrinsic.

```
export interface SubmittableExtrinsicFunction<ApiType> extends CallFunction {
  (...params: CodecArg[]): SubmittableExtrinsic<ApiType>;
}
``` 

========================
Forget about tsc, is it correct to say SubmittableExtrinsic extends Call? if not, we probably shouldn't extends CallFunction.

(recently edited)

ianhe8x:
Solve this via merging my ICennznetExtrinsic and CennznetSubmittableExtrinsic into one.
It seems my usage of typescript causing that.

PR #1373, opened 2019-09-06
fix SubmittableExtrinsicFunction override

ianhe8x:
fix for #1372

### DIFF
--- a/packages/api/src/types.ts
+++ b/packages/api/src/types.ts
@@ -147,7 +147,7 @@ export interface QueryableStorage<ApiType> {
   [index: string]: QueryableModuleStorage<ApiType>;
 }
 
-export interface SubmittableExtrinsicFunction<ApiType> extends CallFunction {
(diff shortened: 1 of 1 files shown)


packages/api/src/types.ts
```diff
@@ -147,7 +147,7 @@ export interface QueryableStorage<ApiType> {
   [index: string]: QueryableModuleStorage<ApiType>;
 }
 
-export interface SubmittableExtrinsicFunction<ApiType> extends CallFunction {
+export interface SubmittableExtrinsicFunction<ApiType> extends Pick<CallFunction, 'callIndex' | 'meta' | 'method' | 'section' | 'toJSON'> {
```
ianhe8x: instead of `Pick` maybe just declare it explicitly is more clear.
```
export interface SubmittableExtrinsicFunction<ApiType> {
  (...params: CodecArg[]): SubmittableExtrinsic<ApiType>;
  callIndex: Uint8Array;
  meta: FunctionMetadataV7;
  method: string;
  section: string;
  toJSON: () => any;
}
```
only this means another place to change when doing Metadata bump

ianhe8x:
canceled

Issue #108, opened 2018-06-28
Get documentation going again...


jacogr:
... and make sure we have proper docs here

~~https://github.com/polkadot-js/dev/issues/132~~

jacogr:
Closed in https://github.com/polkadot-js/api/pull/145

PR #1374, opened 2019-09-07
Things I wish I knew...

jacogr:
... before I started using the API. Chapter 1

Best way to actually read through atm is via https://github.com/polkadot-js/api/tree/jg-things-to-know/docs/start


docs/start/README.md
```diff
@@ -0,0 +1,11 @@
+# Getting started
+
+These sections should provide you will all the information needed to install the `@polkadot/api` package, understand the structure of the interfaces and allow you to start using it. For existing users this really should be titled "Things I wish I knew before I started using the api" - it really aims to close the gap to allow anybody to get to grips with using the packages.
```
Tbaut:
```suggestion
These sections should provide you with all the information needed to install the `@polkadot/api` package, understand the structure of the interfaces and allow you to start using it. For existing users this really should be titled "Things I wish I knew before I started using the api" - it really aims to close the gap to allow anybody to get to grips with using the packages.
```

docs/start/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Help us help others
+
+If you spot gaps in the information provided, or are uncertain about any specific area, please do [log an issue](https://github.com/polkadot-js/api/issues) or if you are that way inclined, make a pull-requst. We really want to have good documentation in these areas and allow people to be productive right from the start.
```
Tbaut:
```suggestion
If you spot gaps in the information provided, or are uncertain about any specific area, please do [log an issue](https://github.com/polkadot-js/api/issues) or if you are that way inclined, make a pull-request. We really want to have good documentation in these areas and allow people to be productive right from the start.
```

docs/start/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Ready? Steady? Go!
+
+Will all that said, let's get started... [What should be installed, and how should we do it?](install.md)
```
Tbaut:
```suggestion
With all that said, let's get started... [What should be installed, and how should we do it?](install.md)
```

docs/start/install.md
```diff
@@ -0,0 +1,27 @@
+# Installation
+
+Yes, it really is as simple as installing [from npm](https://www.npmjs.com/package/@polkadot/api), so not going to waste too much time with the bare basics, just install the API via
+
+`yarn @polkadot/api`
```
Tbaut: idk if it's required but I guess the add helps 
```suggestion
`yarn add @polkadot/api`
```

docs/start/install.md
```diff
@@ -0,0 +1,27 @@
...
+
+And it will be added and ready for use. The above will always install the latest stable release, which should allow you to connect to testsnets and local nodes that are tracking versioned releases for [Polkadot](https://github.com/paritytech/polkadot) and [Substrate](https://github.com/paritytech/substrate).
+
+## Betas
+
+For users who have a slightly higher appetite for risk, or are using bleeding-edge master branches of either Polkadot/Substrate, we also publish a beta version as soon as anything is merged into the master branch. This version really contains all the latest fixes and features, and is the version we actually use inside the polkadot-js projects - eating our own dogfood.
+
+To install a beta version, either to test or for support of a feature that is available in Substrate master (and has not yet made it to a stable api release), you can install it via the `@beta` tag, i.e.
+
+`yarn @polkadot/api@beta`
```
Tbaut: seems that is really is not required..

```suggestion
`yarn add @polkadot/api@beta`
```

docs/start/install.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Other dependencies
+
+In most cases, you don't need to do anything else apart from just installing `@polkadot/api` above. It has dependencies such as `@polkadot/types` which aresinstalled automatically alongside. When using `yarn` the dependencies are installed, flatenned, available for use and you will never run into issues with mismatched versions.
```
Tbaut:
```suggestion
In most cases, you don't need to do anything else apart from just installing `@polkadot/api` above. It has dependencies such as `@polkadot/types` which are installed automatically alongside. When using `yarn` the dependencies are installed, flattened, available for use and you will never run into issues with mismatched versions.
```

docs/start/install.md
```diff
@@ -0,0 +1,27 @@
...
+
+This means that by simply installing `@polkadot/api`, you will have access to utilities (crypto and normal), types, providers and even higher-order (derived) API functions. (We will get to all of these in follow-up sections)
+
+If you do however decide to explicitly install other packages (even through they are dependencies), please make sure that the versions inside the api repo always match, i.e. if you installed `@polkadot/api` `0.91.0-beta.22` and you have your own version of `@polkadot/types`, ensure that it is also `0.91.0-beta.22`.
```
Tbaut:
```suggestion
If you do however decide to explicitly install other packages (even though they are dependencies), please make sure that the versions inside the api repo always match, i.e. if you installed `@polkadot/api` `0.91.0-beta.22` and you have your own version of `@polkadot/types`, ensure that it is also `0.91.0-beta.22`.
```

docs/start/types.basics.md
```diff
@@ -0,0 +1,65 @@
...
+- `.toU8a()` - returns a `Uint8Array` representation of the encoded value (genrally exactly as passed to the node, where values are SCALE encoded)
+
+Additionally, the following getters and utilities are available -
+
+- `.isEmpty` - `true` if the value is an all-empy value, i.e. `0` in for numbers, all-zero for Arrays (or anything `Uint8Array`)
+- `.hash` - a `Hash` (once again with all the methods above) that is a `blake2-256` of the contained value
+
+## Working with numbers
+
+All numbers wraps and extends an instance of [bn.js](https://github.com/indutny/bn.js/). This means that in addition  to the interfaces defined above, it has some additional methods -
```
Tbaut:
```suggestion
All numbers wrap and extend an instance of [bn.js](https://github.com/indutny/bn.js/). This means that in addition  to the interfaces defined above, they have some additional methods -
```

docs/start/basics.md
```diff
@@ -0,0 +1,34 @@
...
+- [query](../substrate/storage.md) - All chain state, e.g. `api.query.balances.freeBalance(<accountId>)`.
+- [tx](../substrate/extrinsics.md) - All extrinsics, e.g. `api.tx.balances.transfer(<accountId>, <value>)`.
+
+Additionally the metadata also provides information on [events](../substrate/events.md), these are query-able via the `api.query.system.events()` interface and also appear on transactions... both these cases are detailed later.
+
+## Types
+
+The metadata details all the types for the various interfaces. At the moment (this is undergoing investigations and could improve in future versions of metadata), this also means that the types between the API and the node need to be aligned. For instance, by default Substrate defines a `BlockNumber` type as a `u32` and the API follows the Substrate defaults - if a chain has a different definitions, the API needs to be aware of this so it can actually decode (and encode) the type.
+
+At this point, just be aware of it, we will touch of types, custom chains and their impacts in a later section.
```
JoshOrndorff:
```diff
- At this point, just be aware of it, we will touch of types, custom chains and their impacts in a later section.
+ At this point, just be aware of it, we will touch on types, custom chains and their impacts in a later section.
```

docs/start/create.md
```diff
@@ -0,0 +1,68 @@
...
+
+We have the API installed and we understand what will actually be exposed, down the rabbit hole we go - let's create an actual API instance, and then take it from there -
+
+```js
+// import
+import { ApiPromise, WsProvider } from '@polkadot/api';
+
+// construct
+const wsProvider = new WsProvider('wss://poc-3.polkadot.io');
+const api = await ApiPromise.create({ provider: wsProvider });
```
JoshOrndorff: `SyntaxError: await is only valid in async function`

JoshOrndorff: Not sure the best way to handle that. Adding `async function main` or `async (() => { ... })()` adds extra complexity.

docs/start/create.md
```diff
@@ -0,0 +1,68 @@
...
+// import
+const { ApiPromise, WsProvider } = require('@polkadot/api');
+...
+```
+
+We are basing all our examples on the [ApiPromise](../examples/promise/README.md) version, however there is also an RxJS version of the API available. Since Promises are a part of the ES2015 specification, it covers the greater amount of use and is the one that will be used in 95% of the cases and should be familiar to 100% of all developers. However if you are in an environment where RxJs is standard, you could take a look at the [RxJS examples](../examples/rx/README.md) once you are familiar with the base concepts. For now... just ignore and focus on understanding the concepts.
+
+## Providers
+
+Focussing on the construction, any API requires a provider and we create one via the `const wsProvider = new WsProvider(...)`. (By default, if none is provided to the API it will construct a default `WsProvider` instance and connect to `ws://127.0.0.1:9944`, however in most cases we want to not just connect to a local node, and even if we wish to, we probably want to be explicit).
```
JoshOrndorff:
```diff
- Focussing on the construction, any API requires a provider and we create one via the `const wsProvider = new WsProvider(...)`. (By default, if none is provided to the API it will construct a default `WsProvider` instance and connect to `ws://127.0.0.1:9944`, however in most cases we want to not just connect to a local node, and even if we wish to, we probably want to be explicit).
+ Focusing on the construction, any API requires a provider and we create one via the `const wsProvider = new WsProvider(...)`. (By default, if none is provided to the API it will construct a default `WsProvider` instance and connect to `ws://127.0.0.1:9944`, however in most cases we want to not just connect to a local node, and even if we wish to, we probably want to be explicit).
```

[ngram](https://books.google.com/ngrams/graph?content=focussing%2C+focusing&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cfocussing%3B%2Cc0%3B.t1%3B%2Cfocusing%3B%2Cc0)

docs/start/create.md
```diff
@@ -0,0 +1,68 @@
...
+
+At this time the only provider type that is fully supported by the API is the WebSocket version. Polkadot/Substrate really comes alive with possibilities once you have access to bo-directional RPCs, such as what WebSockets provide. (It is possible to have some limited capabilities via bare-HTTP, but at this point WebSockets is the only fully-operational and supported version - alwaysd remember that is "just upgraded HTTP".)
```
JoshOrndorff:
```diff
- At this time the only provider type that is fully supported by the API is the WebSocket version. Polkadot/Substrate really comes alive with possibilities once you have access to bo-directional RPCs, such as what WebSockets provide. (It is possible to have some limited capabilities via bare-HTTP, but at this point WebSockets is the only fully-operational and supported version - alwaysd remember that is "just upgraded HTTP".)
+ At this time the only provider type that is fully supported by the API is the WebSocket version. Polkadot/Substrate really comes alive with possibilities once you have access to bi-directional RPCs, such as what WebSockets provide. (It is possible to have some limited capabilities via bare-HTTP, but at this point WebSockets is the only fully-operational and supported version - always remember that is "just upgraded HTTP".)
```

docs/start/create.md
```diff
@@ -0,0 +1,68 @@
...
+
+// do something
+console.log(api.genesisHash.toHex());
+```
+
+## Advanced creation
+
+There are more advanced cases where you would prefer to use the longer version, for instance: if you want to explicitly listen to events emitted, you probably want to attach to the API even before connecting to the chain. In these cases, create via `new`, attach listeners and then wait for the `isReady`.
+
+All instances Api implement an `EventEmitter` interface, with `on` handlers, which emit `connected`, `disconnected`, `ready` and `error` events, which allows you to listen to what actually happens on the transport layer.
```
JoshOrndorff:
```diff
- All instances Api implement an `EventEmitter` interface, with `on` handlers, which emit `connected`, `disconnected`, `ready` and `error` events, which allows you to listen to what actually happens on the transport layer.
+ All Api instances implement an `EventEmitter` interface, with `on` handlers, which emit `connected`, `disconnected`, `ready` and `error` events, which allows you to listen to what actually happens on the transport layer.
```

docs/start/api.consts.md
```diff
@@ -0,0 +1,34 @@
...
+
+// the amount required per byte on an extrinsic
+console.log(api.consts.balances.transactionByteFee.toNumber());
+```
+
+Since these are constants and defined by the metadata, it is not a call, but rather the values immediately available - as you'll see in subsequent sections, there is no need for `await` on these, it immediately returns the type and value for you to work with.
+
+## The API and types
+
+There is some magic applied by the API, for instance the `createFee` comes back as `Balance` type, however the API knows this type and makes a `Balance` object available, hence the `toNumber`. This is consistent in retrieving constants, making queries or event making transactions -
```
JoshOrndorff:
```diff
- There is some magic applied by the API, for instance the `createFee` comes back as `Balance` type, however the API knows this type and makes a `Balance` object available, hence the `toNumber`. This is consistent in retrieving constants, making queries or event making transactions -
+ There is some magic applied by the API, for instance the `createFee` comes back as `Balance` type, however the API knows this type and makes a `Balance` object available, hence the `toNumber`. This is consistent in retrieving constants, making queries or even making transactions -
```

docs/start/api.query.md
```diff
@@ -0,0 +1,47 @@
...
+  api.query.system.accountNonce(ADDR),
+  api.query.balance.freeBalance(ADDR)
+]);
+```
+
+## Parameters & return values
+
+As indicated in previous sections, any return value is always an object with a consistent interface that reflects the type being returned. In the above example, the timestamp is a `Moment` (a `u64` value), the nonce is an `Index` (a `u32` value) and the `Balance` is an underlying `u128`.
+
+Additionally we have provided some parameters for the query calls, specifically for the retrieval of the nonce and balance. It is important to note that the API will automatically convert any parameters into the correct type for encoding  and making calls, in this case the `AccountId` parameter could be specified as a ss58 address (as it was), an actual `AccountId` (retrieved via another call) or just a plain `Uint8Array` (or even hex-string representation) for a publicKey.
```
JoshOrndorff: I'm not  clear what the constraints on the type coercion are. If a query expects a u32 parameter, can I give it
* `4`
* `new BN(4)`
* `"4"`
* `"four"`

What happens when the API doesn't know how to convert the types (as is probably the case in my last example)? Or what happens if it knows how but there is overflow. Like if the underlying parameter type is a `u8` but I provide `10000`?

docs/start/create.md
jacogr: Yes, should just clarify some wrappers up-front. More-or-less around here as I do with the Promise thing. (All the subsequent examples should really try and focus on the meat instead of boilerplate)

jacogr: Added an expanded note here that we are actually wrapped in an `async` block (with an example). Hopefully this is enough... the intent is not to teach people JS

docs/start/api.query.md
jacogr: I am not convinced that we have the overflow checks correct in all places -  it should throw when converting from a too large value upon serialization.

It is not quite optimal to do this on all operations, for instance if we have a `const a = new u8(12)` and we do `a.muln(1024)` - it overflows. 

Bit out of scope here, but certainly requires an investigation for all to tighten up where it softly fails and does wrong things. (Loud complaining when it cannot do things is always . ok)

docs/start/api.rpc.md
```diff
@@ -0,0 +1,70 @@
...
+## Detour into derives
+
+The `api.derive` interfaces will be covered in a follow-up section, but since the above example deals with new head subscriptions, a quick detour is warranted. The derives are just helpers that define certain functions and combine results from multiple sources. For new headers, the following information is useful in certain scenarios -
+
+```js
+...
+const unsub = await api.derive.chain.subscribeNewHeads((lastHeader) => {
+  console.log(`#${lastHeader.number} was authored by ${lastHeader.author}`);
+});
+```
```
JoshOrndorff: I'm getting `#108 was authored by undefined`

JoshOrndorff: Ahh, that's because I was using the node template which doesn't have `api.query.session`

docs/start/api.query.multi.md
```diff
@@ -0,0 +1,60 @@
...
+
+// subscribe to the timestamp, our index and balance
+const unsub = await api.queryMulti([
+  api.query.timestamp.now,
+  [api.query.system.accountNonce, ADDR],
+  [api.query.balances.freeBalance, ADDR]
+], ([now, nonce, balance]) => {
+  console.log(`${now}: balance of ${balance} and a nonce of ${nonce}`);
+});
+```
```
JoshOrndorff: What's the syntax for making a multi-query when one of the queries expects multiple parameters like in `api.query.session.nextKeys`? My guess was:
```javascript
const unsub = await api.queryMulti([
  api.query.timestamp.now,
  [api.query.session.nextKeys, BYTES, ADDR]
], ([now, keys]) => {
  ...
});
```
But experimentation suggests it is
```javascript
const unsub = await api.queryMulti([
  api.query.timestamp.now,
  [api.query.session.nextKeys, [BYTES, ADDR]]
], ([now, keys]) => {
  ...
});
```

docs/start/api.query.multi.md
```diff
@@ -0,0 +1,60 @@
+# Multi queries
+
+In a number of applications, it is useful to monitor a number of like-queries at the same time. For instance, we may want to track the balances for a list of accounts we have. The `api.query` interfaces allows this via the `.multi` sub-call.
```
JoshOrndorff: What are the advantages of using `api.query.multi` vs making the regular queries we learned previously and using `Promise.all`?

docs/start/api.query.multi.md
JoshOrndorff: I like the way you explain a similar concept to `.at`:
> In the above example, we introduce the `.at(<hash>, [params])` query.

JoshOrndorff: FWIW, I would have guessed correctly on the first try if the example had been
```javascript
  const unsub = await api.queryMulti([
    [api.query.timestamp.now, []],
    [api.query.system.accountNonce, [ADDR]],
    [api.query.balances.freeBalance, [ADDR]]
  ], ([now, nonce, balance]) => {
    console.log(`${now}: balance of ${balance} and a nonce of ${nonce}`);
  });
```

docs/start/api.query.other.md
```diff
@@ -0,0 +1,75 @@
...
+In the above example, we introduce the `.at(<hash>, [params])` query. For all `.at` queries, the first parameter is always the block hash at which we want to make the query, in our example we use both the last retrieved block and the parent thereof. The params are optional as per the type of query made, for instance to retrieve the timestamp for a previous block, it would be -
+
+```js
+...
+
+// retrieve the timestampt for the  previous block
+const momentPrev = await api.query.timestamp.now.at(lastHdr.parentHash);
+```
+
+The `.at` queries are all single-shot, i.e. there are no subscription option to these, since the state for a previous block should be static. (This is true to a certain extent, i.e. when blocks have been finalized).
```
JoshOrndorff: Is there any way to listen for re-orgs?

docs/start/api.query.multi.md
JoshOrndorff: Oh, I see:

* `.multi` is for _subscribing_ to multiple queries with a single callback.
* `Promise.all` is for reading the _current state_ of multipe queries.

docs/start/keyring.md
```diff
@@ -0,0 +1,104 @@
...
+import { Keyring } from '@polkadot/keyring';
+
+// initialize the API as we would normally do
+...
+
+// create a keyring instance
+const keyring = new Keyring({ type: 'sr25519' });
+```
+
+In the above example, the import is self-explanatory. Upon creation we pass through a `type` which can have a value of either `ed25519` or `sr25519`, when not specified this would default to `ed25519`. This type parameter only applies to the default type of account created when no type is specified, it does not mean that the keyring can only store that type of account.
```
JoshOrndorff: Then the parameter should be called `defaultType` or similar.

JoshOrndorff: Here's a PR for that. https://github.com/polkadot-js/common/pull/468

docs/start/api.tx.subs.md
```diff
@@ -0,0 +1,61 @@
+# Transactions and subscriptions
+
+Previously we send simple transactions with the `api.tx` endpoints, in this section we will extend that to monitor the actual transactions for inclusion and also extend the monitoring for transaction events.
+
+## Transaction finalization
```
JoshOrndorff: The use of "finalization" here just means included in a block right? This might be confusing for dApp developers who are also thinking of finalization in the sense of grandpa or casper FFG where it means the block will never be reverted or orphaned by a re-org.

docs/start/api.tx.wrap.md
```diff
@@ -0,0 +1,44 @@
...
+As indicated in previous sections (we will cover types in more detail next), the API will format the inputs into the actual type required for submission. For primitives such as numbers, this is quite understandable, but it is worth spending at least one example on cases where an object is provides as an input. For instance, making a call to validate -
+
+```js
+...
+const txHash = await api.tx.staking.validate({
+  validatorPayment: 12345
+});
+```
+
+In the above example, all we need to provide is a call with the fields for the `ValidatorPrefs` object. (Any fields not defined will be set to the default for that type, i.e. all zero).
```
JoshOrndorff: I think this would benefit from a link to the actual struct that you're sending. https://github.com/paritytech/substrate/blob/master/srml/staking/src/lib.rs#L346-L354

Someone who's developing their dApp frontend, but doesn't know all the internals of staking on Polkadot or Substrate may be confused about where `validatorPayment` came from.

Alternatively you could show a sample rust struct and the corresponding call in javascript.

docs/start/types.basics.md
```diff
@@ -0,0 +1,65 @@
...
+
+All structures, a wrapping of an object containing a number of member variables, is an implementation of a standard JS `Map` object, so all the functions available on  a `Map` such as `.entries()` are available. Additionally it is decorated with actual getters for the fields.
+
+As an example, a `Header` will have getters for the `.parentHash`, `.number`, `.stateRoot`, `.extrinsicsRoot` and `.digest` fields. The same applies for all structures, as they are returned, each member will have an associated getter.
+
+## Working with enums
+
+Each enum has additional getters which are injected based on the fields wrapped. These take the form of `.is<Name>` and `.as<Name>` to allow you to check is the enum is a certain value or to retrieve the underlying value as a specific type.
+
+As an example, when an extrinsic is applied, the `Phase` enum has one of two states, `ApplyExtrinsic(u32)` or `Finalization`. In this case `.isApplyExtrinsic` would be `true` when an extrinsic is being applied, and `.asApplyExtrinsic` would return the phase value as a `u32`
```
JoshOrndorff: What would `.asApplyExtrinsic` return when the extrinsic is `Finalization`?

jacogr: It will throw.

jacogr: Actually needs some clarification here - on apply the u32 value is the index (in the block) of the extrinsic being applied

docs/start/api.query.multi.md
jacogr: `nextKeys` is one of the "funny" ones. Since it is a `DoubleMap`, the `[bytes, key]` needs to be a single variable, i.e. they are combined into 1 to produce the key. For `LinkedMap`s & `DoubleMap`s there still needs to be a specific section, there are a number of gotchas.

docs/start/api.query.multi.md
jacogr: `.multi` is also more effective since it bundles queries into a single RPC call, instead of using one per statement. That is one of the things I was hoping to make clear, but have failed at. Will retry :)

docs/start/api.query.other.md
jacogr: with `newHeads`, that is how this is constructed - https://polkadot.js.org/apps/#/explorer/forks

docs/start/api.tx.subs.md
jacogr: Kept the working to align with the actual enum, which is `Finalized`

docs/start/api.tx.wrap.md
jacogr: Made a claifying comment.

docs/start/api.tx.subs.md
jacogr: Clarified via comment.

docs/start/types.basics.md
jacogr: Clarified.

docs/start/api.query.multi.md
jacogr: Added these cases to the outstanding . issues in `.TODO.md`

jacogr:
Wrapping it right here and now. Not perfect, but ok as a start.

Issue #1392, opened 2019-09-13
Add support for available RPC queries 


jacogr:
As per https://github.com/paritytech/substrate/pull/3613

PR #1397, opened 2019-09-14
Add rpc_methods

### LINKED ISSUE #1342: Add account_nextIndex RPC
We probably want to start using this in sumittables as well - the only issue is that it is a non-default RPC, so _may_ not be there. (But will be in 99.9% of the cases). So potentially we want some detection and then use what is available either this or the nonce. (Detection happens up-front)

### LINKED ISSUE #1389: Add contracts_call RPC
https://github.com/paritytech/substrate/pull/3563

jacogr:
- `rpc_methods`, closes https://github.com/polkadot-js/api/issues/1392
- `account_nextIndex`, closes #1342 
- `contracts_call`, closes #1389 
- Filter rpc methods to only have those exposed that are in `rpc_methods` when the `isOptional` flag is set (this allows us to maintain backwards compatibility)

This means we can use `account_nextIndex` inside submittables, i.e.

```js
api.rpc.account.nextIndex
  ? api.rpc.account.nextIndex(address)
  : api.query.system.accountIndex(address)
```

### DIFF
--- a/packages/api/src/base/Decorate.ts
+++ b/packages/api/src/base/Decorate.ts
@@ -161,7 +161,7 @@ export default abstract class Decorate<ApiType> extends Events {
         (section as any)[methodName] = decorateMethod(method, { methodName });
 
         // add this endpoint mapping to our internal map - we use this for filters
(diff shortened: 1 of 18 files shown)


jacogr:
Tested alongside apps across a variety of nodes.

PR #1406, opened 2019-09-18
Correctly type `api.query.*.*`

amaurymartiny:
I'm exploring on how we could have `api.query.*.*` typed.

```typescript
  const bar = await api.query.foo.bar(); // Codec, same as before
  const bal = await api.query.balances.freeBalance(keyring.alice.address); // u128
```

Notes:
- output is in `api/src/query.types.ts`, **this file is totally optional**, we can remove it and there shouldn't be any additional thing to fix somewhere else in the code.
- the magic function is in `types/src/scripts/generateTypes/query.ts`, which takes a `Metadata` as input (using static.substrate for now).
- reorganized a bit `types/src/scripts/` into smaller files


packages/api/src/checkTypes.manual.ts
```diff
@@ -28,17 +28,18 @@ async function derive (api: ApiPromise): Promise<void> {
     console.log('current author:', header.author);
   });
 
-  await api.query.staking.intentions((intentions): void => {
-    console.log('intentions:', intentions);
+  await api.query.staking.validatorCount((count): void => {
```
amaurymartiny: removed `intentions` because it was underlined red

amaurymartiny:
There are still some ts errors to fix here and there, but `checkTypes.manual.ts` already works well.

packages/api/src/checkTypes.manual.ts
jacogr: This actually is a v1 thing - so it should not be underlined red, it can exist. It is a valid test actually, the ones that it doesn't know about, it shouldn't complain about.

We should also add some modules-not-defined, i.e. `api.query.parachains` - similar to the above, don't complain about modules not defined.

TL;DR -

- don't complain about methods not known, but allow the `<Type>` overrides
- don't complain about modules not known, also allow the generic overrides

amaurymartiny:
Modified types so that it doesn't break backward-compat at all:

```typescript
  const bar = await api.query.foo.bar(); // bar is Codec (unknown module)
  const bal = await api.query.balances.freeBalance(keyring.alice.address); // bal is u128
  const bal2 = await api.query.balances.freeBalance(keyring.alice.address, 'WRONG_ARG'); // bal2 is Codec (wrong args)
  const override = await api.query.balances.freeBalance<Header>(keyring.alice.address); // override is still available
  const oldBal = await api.query.balances.freeBalance.at('abcd', keyring.alice.address);
```

amaurymartiny:
Ready for review.

I would also remove `packages/api/src/query.types.ts` (and all generated files) away from CodeClimate, so that it doesn't pollute the maintainability score.

packages/api/test/e2e/api/promise-tx-eras.spec.ts
```diff
@@ -74,7 +74,7 @@ describeE2E({
     });
 
     it('fails on a transfer with invalid time', async (done): Promise<void> => {
-      const nonce = await api.query.system.accountNonce(keyring.alice.address) as Index;
+      const nonce = await api.query.system.accountNonce(keyring.alice.address);
```
amaurymartiny: I did a couple of these in the codebase, end-users would need to do this too, or else eslint complains "assertion is useless"

amaurymartiny:
Hmm, I can add some tests for the generation scripts, but I'm not sure they will be super relevant

packages/api/src/checkTypes.manual.ts
```diff
@@ -37,6 +37,14 @@ async function query (api: ApiPromise, keyring: TestKeyringMap): Promise<void> {
   const intentions = await api.query.staking.intentions();
   console.log('intentions:', intentions);
 
+  // api.query.*.* is well-typed
+  const bar = await api.query.foo.bar(); // bar is Codec (unknown module)
+  const bal = await api.query.balances.freeBalance(keyring.alice.address); // bal is u128
```
jacogr: The result should be `Balance` - not sure why this defaults to `U128` (Checked it, it is def `U128` as indicated here)

packages/api/src/checkTypes.manual.ts
```diff
@@ -96,7 +104,7 @@ async function tx (api: ApiPromise, keyring: TestKeyringMap): Promise<void> {
     });
 
   // with options and the callback
-  const nonce2 = await api.query.system.accountNonce<Index>(keyring.alice.address);
+  const nonce2 = await api.query.system.accountNonce(keyring.alice.address);
```
jacogr: Same here, it should be `Index`

packages/api/src/submittable/Submittable.ts
```diff
@@ -99,10 +99,10 @@ export default class Submittable<ApiType> extends _Extrinsic implements Submitta
   }
 
   // send with an immediate Hash result
-  public send (): SubmitableResultResult<ApiType>;
+  public send(): SubmitableResultResult<ApiType>;
```
jacogr: Not sure why the `send ()` was dropped, i.e. the space should really be there.

packages/api/src/types.ts
```diff
@@ -84,18 +91,22 @@ export type DecoratedRpc<ApiType, AllSections> = {
...
-  (arg1?: CodecArg, arg2?: CodecArg): Observable<Codec>;
-  <T extends Codec>(arg1?: CodecArg, arg2?: CodecArg): Observable<T>;
-  multi: <T extends Codec>(args: (CodecArg[] | CodecArg)[]) => Observable<T[]>;
+export interface StorageEntryBase<ApiType, F extends AnyFunction> {
+  at: (hash: Hash | Uint8Array | string, ...args: Parameters<F>) => PromiseOrObs<ApiType, ObsInnerType<ReturnType<F>>>;
+  creator: StorageEntry;
+  hash: (...args: Parameters<F>) => PromiseOrObs<ApiType, Hash>;
+  key: (...args: Parameters<F>) => string;
+  size: (...args: Parameters<F>) => PromiseOrObs<ApiType, u64>;
+  multi: ApiType extends 'rxjs' ? StorageEntryObservableMulti : StorageEntryPromiseMulti;
```
jacogr: Didn't see a multi test. However this does seem to be a `Codec[]` result -

`const balances = await api.query.balances.freeBalance.multi([keyring.alice.address, keyring.bob.address, keyring.eve.address]);`

packages/api/src/checkTypes.manual.ts
amaurymartiny: Unfortunately i think this is how TS resolves types, if it's a type alias, it'll take the source type.

https://www.typescriptlang.org/play/#code/C4TwDgpgBAglC8UB2BXAtgIwgJwNwCh8ATCAYwBsBDbaAMxSVOAEsB7JKWgCgEoAuWAXyl2AZ2BRKCTr1xQA9PKgADSsqjNRydFmwAaZKwkx8QA

amaurymartiny: One possible solution might to be replace all `type Balance = u128` by `interface Balance extends u128 {}`

packages/api/src/checkTypes.manual.ts
amaurymartiny: Just tried, even with `<Index>`, nonce2 is a u32. So related to type aliases

packages/api/src/types.ts
amaurymartiny: I added a .multi test. It should be a `u128[]`, but I cannot type that correctly, so it'll stay a `Codec[]` for now.

This works:
```
const balances = await api.query.balances.freeBalance.multi<Balance>([keyring.alice.address, keyring.bob.address, keyring.eve.address]);
```

packages/api/src/checkTypes.manual.ts
jacogr: Actually happy with that - that is actually the way we do it in createType via the registry now, should have put 2 and 2 together here that is is possibly the same issue.

packages/api/src/types.ts
jacogr: That is fine. There must be a way of doing this on the `.multi` though, `queryMulti` probably not. Not a crisis since this is 100x better already, but something to keep in mind.

packages/types/src/types.ts
```diff
@@ -154,6 +154,11 @@ export interface SignatureOptions {
 
 export type ArgsDef = Record<string, Constructor>;
 
+// A type alias for [Type1, Type2] & Codec, representing a tuple (Type1, Type2)
+// FIXME Implement this generic <Sub> on Tuple.ts itself.
+// eslint-disable-next-line @typescript-eslint/interface-name-prefix
+export type ITuple<Sub extends Codec[]> = Sub & Codec
```
amaurymartiny: Introduced this to replace `[Type1, Type2] & Codec`, since I cannot `interface extends ([Type1, Type2] & Codec)`

But as written, should be on Tuple itself ideally. Not straightforward though, as `Tuple extends Array<T>`

PR #1408, opened 2019-09-19
More fixes for examples

JamesLefrere:
Hello again,

I decided to run through all of the examples, and found a few errors while doing so :)

There's still an issue with calling `toPromise` for queries with for the RxJS API; it seems like the promise doesn't resolve, which makes me wonder whether the observable isn't completed? i.e. we should just `take` one value for this? 

I don't have a ton of experience with RxJS so I'm not sure about that, but this test I tried worked at the RPC level:

```typescript
  it('can consume the observable value as a promise', async (): Promise<void> => {
    const value = await rpc.system.chain().toPromise();
    expect(value).toEqual('mockChain'); // Defined in MockProvider
  });
```

(that isn't included in this PR, because it's probably not necessary).

Also, I was wondering about some improvements that could be added for these examples:

* Make the examples runnable on the docs site
* If we want to be really safe, add e2e tests for examples (just a script to run each example and read stdout)

**Changes**

* Add various fixes for the examples
* Add the workaround for proposals for `impl_version` 94
* TODO: fix `toPromise` issue for queries with RxJS [here](https://github.com/polkadot-js/api/compare/master...JamesLefrere:jl-more-examples-fixes?expand=1#diff-e9aa5e81e54e9f4eb42eb09e50f649fbR18)

Cheers

### DIFF
--- a/docs/examples/rx/03_listen_to_balance_change/index.js
+++ b/docs/examples/rx/03_listen_to_balance_change/index.js
@@ -22,7 +22,7 @@ async function main () {
       pairwise()
     )
     .subscribe((balance) => {
-      if (balance[0].eq('first')) {
+      if (balance[0] === 'first') {
         // Now we know that if the previous value emitted as balance[0] is `first`,
         // then balance[1] is the initial value of Alice account.
         console.log(`Alice ${Alice} has a balance of ${balance[1]}`);
--- a/docs/examples/promise/04_unsubscribe/index.js
+++ b/docs/examples/promise/04_unsubscribe/index.js
@@ -9,7 +9,7 @@ async function main () {
   const api = await ApiPromise.create();
 
   // Subscribe to chain updates and log the current block number on update.
-  const unsubscribe = await api.rpc.chain.subscribeNewHead((header) => {
+  const unsubscribe = await api.rpc.chain.subscribeNewHeads((header) => {
     console.log(`Chain is at block: #${header.number}`);
   });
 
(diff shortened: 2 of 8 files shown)


docs/examples/rx/03_listen_to_balance_change/index.js
```diff
@@ -22,7 +22,7 @@ async function main () {
       pairwise()
     )
     .subscribe((balance) => {
-      if (balance[0].eq('first')) {
+      if (balance[0] === 'first') {
```
JamesLefrere: I introduced this bug before without testing the example, sorry :)

jacogr: It happens :)

jacogr:
We have the examples runnable on the `polkadot-js/apps` side. The issue with them in the API is that for queries it works great (we can connect to a node), for transacting this is not quite as clear since you need funds.

The same with the e2e tests, we don't want to spin up nodes on CI nor really rely on external RPCs. So the closest we have for e2e . is the actual `api/e2e` tests that are run inside docker. The issue there is that work has basically stalled and it is not keeping up-to-date. (And doesn't run anywhere else as a separate nightly process on CI)

jacogr:
PS: The query won't resolve - the issue with the queries is that it uses the subscriptions interface, that never completes. So in this case the example should really not be using `toPromise()`. You are right, that is a bug, but rather in the example.

PR #1410, opened 2019-09-23
Update metadata (Babe updates)

jacogr:
Marked ~~WIP~~ On Ice, not because it is not ready, but rather since I need to think about the merge here - basically the internal digests break, so not sure how to handle this. (Primary is incompatible and there is no way to fallback)


packages/types/src/interfaces/babe/definitions.ts
```diff
@@ -15,17 +16,29 @@ export default {
       }
     },
     RawBabePreDigestPrimary: {
+      vrfOutput: 'VrfData',
+      vrfProof: 'VrfProof',
+      authorityIndex: 'u32', // AuthorityIndex (also in aura)
+      slotNumber: 'SlotNumber'
+    },
+    RawBabePreDigestPrimary0to159: {
```
jacogr: These are the problematic changes.

jacogr:
And back to WIP.... building with the latest-latest...

PR #1433, opened 2019-10-02
Don't decode and re-encode storage queries, single pass

jacogr:
This simplifies the logic around working with storage queries and the actual encoding & re-coding thereof -

- we don't decode the storage to re-encode it again, rather we perform a single pass in the output formatting, converting to the Uint8Array representation
- remove `:code` and StorageData specific hacks in `Bytes`, rather check the key and do it explicitly, input cleanups above caters for the actual results
- align the `Linkage` outputs with what is returned via storage queries (empty is not encoded)
- simplify the . isPedantic checks, not the input should always match with less passes

### DIFF
--- a/packages/rpc-core/src/index.ts
+++ b/packages/rpc-core/src/index.ts
@@ -4,7 +4,6 @@
 
 import { ProviderInterface } from '@polkadot/rpc-provider/types';
 import { RpcMethod } from '@polkadot/jsonrpc/types';
-import { StorageChangeSet } from '@polkadot/types/interfaces';
 import { AnyJson, Codec } from '@polkadot/types/types';
 import { RpcInterface } from './jsonrpc.types';
 import { RpcInterfaceMethod } from './types';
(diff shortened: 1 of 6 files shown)


packages/rpc-core/src/index.ts
```diff
@@ -282,98 +281,114 @@ export default class Rpc implements RpcInterface {
     );
   }
 
-  private formatOutput (method: RpcMethod, params: Codec[], result?: any): Codec | (Codec | null | undefined)[] {
-    const base = createTypeUnsafe(method.type, [result]);
+  private treatAsHex (key: StorageKey): boolean {
+    // :code is problematic - it does not have the length attached, which is
+    // unlike all other storage entries where it is indeed properly encoded
+    return ['0x3a636f6465'].includes(key.toHex());
+  }
```
jacogr: We have basically moved the check for . this from `Bytes` (where we try an determine what we are working with), to here where we are being explicit in what the "problematic" keys are

packages/rpc-core/src/index.ts
```diff
@@ -282,98 +281,114 @@ export default class Rpc implements RpcInterface {
...
 
+  private formatOutput (method: RpcMethod, params: Codec[], result?: any): Codec | Codec[] {
```
jacogr: There is no undefined, it always returns a type

PR #1448, opened 2019-10-04
Latest metadata (srml/utility)

jacogr:


packages/types/src/interfaces/utility/definitions.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+export default {
+  types: {
+    // Not 100% on this one this is defined as - result::Result<(), Error>
```
jacogr: Confirmed, the encoding here is correct - logging an issue to add a `Result` type

Issue #1449, opened 2019-10-04
Add Result codec type


jacogr:
As introduced here - https://github.com/polkadot-js/api/pull/1448/files#diff-7c87abb5204066c30053817e1823ae7dR7

(the enum is a good approximation, `0<success data encoded>` or `1<error data encoded>`)

PR #1451, opened 2019-10-05
Add Result type

jacogr:
Closes https://github.com/polkadot-js/api/issues/1449


packages/types/src/codec/Result.ts
```diff
@@ -0,0 +1,142 @@
...
+
+import { typeToConstructor } from './utils';
+import Base from './Base';
+
+/**
+ * @name Result
+ * @description
+ * A Result maps to the Rust Result type, that can either wrap a success or error value
+ */
+export default class Result<O extends Codec, E extends Codec> extends Base<O | E> {
```
amaurymartiny: So finally why not extending Enum?

packages/types/src/codec/Result.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import Enum from './Enum';
+
+/**
+ * @name Result
+ * @description
+ * A Result maps to the Rust Result type, that can either wrap a success or error value
+ */
+export default class Result<O extends Codec, E extends Codec> extends Enum {
+  public constructor (Ok: Constructor | InterfaceTypes, Error: Constructor | InterfaceTypes, value?: any) {
```
amaurymartiny: one small request: can we do `constructor (def: {Ok: ..., Error: ...}, value?: any)` here, instead of 3 args, to keep similar signature to Enum?

jacogr: That would mean it is order dependent and at some point somebody is going to "remove the useless" constructor. Additionally, `Result<O, E>` does actually take 2 params, not a single.

The `with` taking in `{ Ok: ..., Error: ... }` is already problematic here. (Which is required since it extends `Enum`, normally you would go `{ Error: ..., Ok: ... }` since that . is alpabetical)

jacogr:
@kwingram25 Should be ready for your use, hopefully covers all your contract cases as well.

PR #1462, opened 2019-10-09
Call.injectMetadata (align with Events)

jacogr:


packages/types/src/primitive/index.ts
```diff
@@ -38,3 +34,6 @@ export { default as u64 } from './U64';
 export { default as u128 } from './U128';
 export { default as u256 } from './U256';
 export { default as usize } from './USize';
+
+export * from './Generic';
+export * from './Extrinsic';
```
amaurymartiny: ah, yeah

jacogr:
"small" - this one created more havoc than I would have ever imagined... ;)

PR #1463, opened 2019-10-10
Add a `Decorated` class, which will be offline decorated Metadata

### LINKED ISSUE #1461: Allow for a completely offline API
It would be good to have a completely offline `api` object, for example to do offline extrinsics creation.

Requirements: 
- the metadata, genesisHash, blockNumber etc are retrieved by an external party, and can be inputs to the offline `api` object.
- no WsProvider at all, a.k.a completely offline

Output:
- be able to create an extrinsic, ready to be broadcasted to the network

---

### Possible hack today (it's a hack, not recommended)

Today, the solution to this is hacky: we use the `injectMethods` which will populate a global singleton object:

```typescript
import Call from '@polkadot/types/primitive/Generic/Call';
Call.injectMethods(extrinsics);
```

This is hacky, and we want to remove it: https://github.com/polkadot-js/api/issues/580. But with this hack we can

```typescript
createType('Extrinsic', someValue, { version: 3});
```

---

### Possible solutions:

- `new Api({ metadata })` will not create a WsProvider, and will decorate based on inputted metadata. 
    - `api.query.balances.freeBalance().then()` could just fail.
    - `api.query.balances.transfer(...).sign(...)` here we pass in blockNumber, genesisHash, specVersion here, will return a ready-to-be-broadcasted tx.
- `createType('Type', value, { metadata, extrinsicVersion, otherParams... })`
    - needs Type signature unification: https://github.com/polkadot-js/api/issues/1224
- instead of decorating `api`, we could decorate `Metadata`?
    - `metadata.query.balances.freebalance(...); // returns a StorageKey`
    -  `metadata` can be offline, and `api` is just a wrapper of promise/rx + provider around `metadata`

amaurymartiny:
addresses part of #1461 (solution 3 in the possible solutions)

- [x] Move all `Metadata` stuff to api-metadata
    - Note: api-metadata and types have circular dependency
- [x] Make tests pass in types and api-metadata
- [x] Created `Decorated` class
- [x] Everyone that uses anything from `api-metadata` of `types/Metadata` should instead just use `Decorated`

User-facing syntax:
```
const metadata = new Metadata(the_huge_blob_from_rpc);
metadata.tx.balances.transfer(to, amount); // will output a uint8array ready to be injected in an extrinsic
```


packages/api-metadata/src/Metadata/MetadataVersioned.ts
```diff
@@ -129,70 +129,70 @@ class MetadataEnum extends Enum {
...
-  public get isV6 (): boolean {
+  public get isV6(): boolean {
     return this.type === 'V6';
   }
 
   /**
    * @description `true` if V7
    */
-  public get isV7 (): boolean {
+  public get isV7(): boolean {
```
jacogr: Not sure why these keep slipping into PRs lately - this form is wrong. (Not just here and by you, but by a number of people)

jacogr: (Really seems like an auto-save thing since it mucks up whole files at once in all cases)

amaurymartiny: My guess is: people have auto-save one, and auto-save will use the .eslintrc.js. But eslint doesn't work on my computer (I guess on others neither), because e.g. `yarn lint` gives:
```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

same Vscode config works on smaller projects with eslint, but this one is too big. Not 100%, just a guess

jacogr: :(

amaurymartiny:
ready for review

packages/api-metadata/src/Decorated/index.ts
```diff
@@ -0,0 +1,3 @@
+import Decorated from './Decorated';
```
jacogr: (nit) copyright header

packages/api-metadata/src/Decorated/storage/fromMetadata/createFunction.ts
```diff
@@ -6,11 +6,11 @@ import { Codec } from '@polkadot/types/types';
 
 import BN from 'bn.js';
 import { Compact, U8a, createType, createTypeUnsafe } from '@polkadot/types/codec';
-import { StorageEntryMetadata, StorageEntryType } from '@polkadot/types/Metadata/v8/Storage';
 import { StorageEntry } from '@polkadot/types/primitive/StorageKey';
 import { assert, isNull, isUndefined, stringLowerFirst, stringToU8a, u8aConcat } from '@polkadot/util';
 
 import getHasher, { HasherFunction } from './getHasher';
+import { StorageEntryMetadata, StorageEntryType } from '../../../Metadata/v8/Storage';
```
jacogr: (nit) Just move this up a line

packages/api-metadata/src/Decorated/storage/fromMetadata/fromMetadata.spec.ts
```diff
@@ -7,23 +7,25 @@ import '@polkadot/types/injector';
 import testingPairs from '@polkadot/keyring/testingPairs';
 import { u8aToHex } from '@polkadot/util';
 
-import storage from '../static';
+import Decorated from '../../Decorated';
+import json from '../../../Metadata/static';
```
jacogr: (nit) Swap order

packages/api-metadata/src/Decorated/types.ts
```diff
@@ -4,6 +4,7 @@
 
 import { Codec } from '@polkadot/types/types';
 import { ModuleConstantMetadataLatest } from '@polkadot/types/interfaces';
+import { StorageEntry } from '@polkadot/types/primitive/StorageKey';
```
jacogr: (nit) Move up a couple of lines

packages/api-metadata/src/Metadata/types.ts
```diff
@@ -1,9 +1,9 @@
-// Copyright 2017-2019 @polkadot/types authors & contributors
+// Copyright 2017-2019 @polkadot/api-metadata authors & contributors
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { Vec } from '../codec';
-import { Codec } from '../types';
+import Vec from '@polkadot/types/codec/Vec';
+import { Codec } from '@polkadot/types/types';
```
jacogr: (nit) Probably want to flip these with types on top.

packages/api-metadata/src/Metadata/v3/toV4.ts
```diff
@@ -1,10 +1,10 @@
-// Copyright 2017-2019 @polkadot/types authors & contributors
+// Copyright 2017-2019 @polkadot/api-metadata authors & contributors
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { createType, Option, Vec } from '../../codec';
+import { createType, Option, Vec } from '@polkadot/types/codec';
+import StorageHasher from '@polkadot/types/primitive/StorageHasher';
```
jacogr: (nit) Probably want this up a line

packages/rpc-core/src/formatting.spec.ts
```diff
@@ -7,15 +7,13 @@ import '@polkadot/types/injector';
...
-import fromMetadata from '@polkadot/api-metadata/storage/fromMetadata';
-import { Storage } from '@polkadot/api-metadata/storage/types';
-import Metadata from '@polkadot/types/Metadata';
-import rpcMetadataV3 from '@polkadot/types/Metadata/v3/static';
-import rpcMetadataV4 from '@polkadot/types/Metadata/v4/static';
-import rpcMetadataV5 from '@polkadot/types/Metadata/v5/static';
-import rpcMetadataV6 from '@polkadot/types/Metadata/v6/static';
-import rpcMetadataV7 from '@polkadot/types/Metadata/v7/static';
-import rpcMetadataV8 from '@polkadot/types/Metadata/v8/static';
+import Decorated from '@polkadot/api-metadata/Decorated';
```
jacogr: Looking at the naming here I'm not convinced. :( Decorated can be anything.

packages/types/src/codec/Tuple.spec.ts
```diff
@@ -8,9 +8,10 @@ import { VoteThreshold } from '../interfaces/elections';
 import { BlockNumber } from '../interfaces/runtime';
 import { CodecTo } from '../types';
 
+import Metadata from '@polkadot/api-metadata/Metadata';
```
jacogr: I don't believe we should muddle things - we should only have a single class exposed from `api-metadata`  and work with that throughout. So would really suggest `import Metadata from '@polkadot/api-metadata';` where this is the holder object.

packages/types/src/primitive/Extrinsic/v2/Extrinsic.spec.ts
```diff
@@ -5,19 +5,20 @@
 import '../../../injector';
 
 import BN from 'bn.js';
-import extrinsics from '@polkadot/api-metadata/extrinsics/static';
+import Decorated from '@polkadot/api-metadata/Decorated';
+import Metadata from '@polkadot/api-metadata/Metadata';
```
jacogr: Same comment as above, all the files between the previous comment and this one - we should really just work with the wrapper.

packages/types/src/primitive/Extrinsic/v2/Extrinsic.spec.ts
```diff
@@ -5,19 +5,20 @@
...
 import Call from '../../Generic/Call';
 import Extrinsic from './Extrinsic';
 
+const decorated = new Decorated(rpcMetadata);
 const keyring = testingPairs({ type: 'ed25519' }, false);
 
 describe('ExtrinsicV2', (): void => {
   beforeEach((): void => {
-    Call.injectMetadata(new Metadata(metadataStatic));
+    Call.injectMetadata(new Metadata(rpcMetadata));
```
jacogr: Inject the wrapper.

packages/types/src/primitive/Generic/Call.ts
```diff
@@ -5,13 +5,15 @@
 import { FunctionArgumentMetadataLatest, FunctionMetadataLatest } from '../../interfaces/metadata';
 import { AnyU8a, ArgsDef, CallFunction, Codec, IMethod } from '../../types';
 
-import extrinsicsFromMeta from '@polkadot/api-metadata/extrinsics/fromMetadata';
+// FIXME Here we unfortunately cannot use Decorated, because that already calls
```
jacogr: This will go away if we move the `findCall` off from here and only available on the wrapper. (So in here we never decorate anything else, only the wrapper knows how that stuff works.) So happy, it is short-term.

packages/types/src/scripts/MetadataMd.ts
```diff
@@ -5,13 +5,13 @@
 import '../injector';
 
 import fs from 'fs';
+import Metadata from '@polkadot/api-metadata/Metadata';
```
jacogr: This should really be the wrapper object.

packages/types/src/primitive/Extrinsic/v2/Extrinsic.spec.ts
amaurymartiny: I think injecting the wrapper won't work, because the wrapper calls `extrinsicFromMetadata`, which already needs Call

jacogr: Ok, then let's mark this for once we extract `findCall` onto the metadata wrapper.

amaurymartiny:
Would be nice to call this `@polkadot/metadata` actually

jacogr:
Happy to break/rename that as part of the 0.96 series (we are breaking compat, so may as well get all the grumblies out of the way)

Issue #1097, opened 2019-07-10
Observables should not fire twice the exact same value


amaurymartiny:
**Repro:**

```javascript
      api.derive.staking.info(stashId, (result) => {
        ++count;

        console.error('***', count, JSON.stringify(result));
      }).catch(console.error);
```

**Expected:**

```
 *** 7 {"accountId":"5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY","controllerId":"5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY","nextSessionId":"5FA9nQDVg267DEd8m1ZypXLBnvN7SFxYwV7ndqSYGiN9TTpu","nominators":[],"redeemable":"0","rewardDestination":1,"stakers":{"total":"0x0000000000000000002386f26fc10000","own":"0x0000000000000000002386f26fc10000","others":[]},"stakingLedger":{"stash":"5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY","total":"0x0000000000000000002386f26fc10000","active":"0x0000000000000000002386f26fc10000","unlocking":[]},"stashId":"5GNJqTPyNqANBkUVMN1LPPrxXnFouWXoe2wNSmmEoLctxiZY","validatorPrefs":{"unstakeThreshold":3,"validatorPayment":0}}

...

 *** 8 something slightly different

```

**Actual:**

```
[...]

PR #1469, opened 2019-10-10
Update drr equality

jacogr:
Closes https://github.com/polkadot-js/api/issues/1097

### DIFF
--- a/packages/api-derive/src/util/drr.ts
+++ b/packages/api-derive/src/util/drr.ts
@@ -14,7 +14,8 @@ function isEqual (prev: any, next: any): boolean {
   if (isNull(prev) || isNull(next)) {
     return next === prev;
   } else if (Array.isArray(next)) {
-    return !next.some((value: any, index): boolean => isEqual(prev[index], value));
+    return next.length === prev.length &&
+      !next.some((value: any, index): boolean => isEqual(prev[index], value));
   } else if (next.eq) {
     return next.eq(prev);
   }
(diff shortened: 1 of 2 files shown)


packages/api-derive/src/util/drr.ts
```diff
@@ -4,12 +4,25 @@
...
 import { catchError, distinctUntilChanged, publishReplay, refCount } from 'rxjs/operators';
 import { Observable } from 'rxjs';
-import { logger } from '@polkadot/util';
+import { isFunction, isNull, logger } from '@polkadot/util';
 
 type DrrResult = <T> (source$: Observable<T>) => Observable<T>;
 
 const l = logger('drr');
 
+function isEqual (prev: any, next: any): boolean {
```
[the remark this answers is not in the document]
jacogr: Swapped the logic-tree to exactly that.

packages/rpc-core/src/index.ts
```diff
@@ -250,6 +250,10 @@ export default class Rpc implements RpcInterface {
             );
         };
       }).pipe(
+        // Duplicated in api-derive/util/drr
+        distinctUntilChanged((a: any, b: any): boolean =>
```
amaurymartiny: If i'm not mistaken, `a` and `b` can either be `Codec | Codec[] | null` , and nothing else. I don't think we need to do JSON.stringify, we can just use `.eq` for Codec, `map(.eq)` for Codec[] or `===` for null.

(in api-derive we do return some more complex stuff, so JSON.stringify might be the easiest solution)

jacogr: Yes, started that way. However, it is just overkill since have have proper toJSON

https://github.com/polkadot-js/api/commit/45dc8e766f076a7123af5745dd811f43323c8f9c

PR #1474, opened 2019-10-11
Basic BTreeMaps codec support

### LINKED ISSUE #1473: Support BTreeMaps
[BTreeMaps](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html) are supported in module and are part of the [SCALE codec implementation](https://github.com/paritytech/parity-scale-codec/search?q=BTreeMap&unscoped_q=BTreeMap), but there's currently no JavaScript-level support for them.

Edit: There's a PR for this: https://github.com/polkadot-js/api/pull/1474

yourheropaul:
closes #1473

This adds a simple class to decode and encode [BTreeMaps](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html), which can be used for values in module storage (with a caveat; see below).

### New Codec class

A new class, `BTreeMap<K,V>` , available in the `types` package. Once the package is built, it can be included in TypeScript with

```
import { BTreeMap } from '@polkadot/types/codec';
```
or

```
import BTreeMap from '@polkadot/types/codec/BTreeMap';
```

The class can either be used with explicit `Codec`-extending types or using its `with` static function:

```
import { u32, u64 } from '@polkadot/types';

// Explicit instantiation with key and value types
const a = new BTreeMap(u32, u64, [data or map]?)

// Really explicit key and values types for maximum type safety
const b = new BTreeMap<u32, u64>(u32, u64, [data or map]?)

// using `with` constructor. Also providex maximum type safety
const c = new BTreeMap.with(u64,u32)([data or map]?)
```

This would typically be used in some other data structure:

```
 class MyStruct extends Struct {
  constructor (value?: any) {
    super({
      value: BTreeMap.with(u64, Text),
    }, value);
  }
}
```

## Caveat: no string type support

Unlike structs, tuples, vectors, sets and so forth, there's current no way of instantiating a BTreeMap from a string. This is partly due to the runtime metadata not including type information for BTreeMaps (see https://github.com/paritytech/substrate/issues/3795).

It's still possible to use BTreeMaps in all other conditions: as field in structs, types in tuples, etc.

### DIFF
--- a/packages/types/src/codec/BTreeMap.spec.ts
+++ b/packages/types/src/codec/BTreeMap.spec.ts
@@ -1,14 +1,14 @@
+// Copyright 2017-2019 @polkadot/types authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
 import '../injector';
-import BTreeMap from './BTreeMap';
 import Text from '../primitive/Text';
 import U32 from '../primitive/U32';
-import { CodecTo } from '../types';
+import Struct from './Struct';
 import { CodecTo } from '../types';
 
-import Text from '../primitive/Text';
-import U32 from '../primitive/U32';
 import BTreeMap from './BTreeMap';
-import Struct from './Struct';
 
 const expectedMap = new Map<Text, U32>();
 expectedMap.set(new Text('bazzing'), new U32(69));
--- a/packages/types/src/codec/BTreeMap.spec.ts
+++ b/packages/types/src/codec/BTreeMap.spec.ts
@@ -10,34 +10,65 @@ import { CodecTo } from '../types';
 
 import BTreeMap from './BTreeMap';
 
-const expectedMap = new Map<Text, U32>();
-expectedMap.set(new Text('bazzing'), new U32(69));
+const mockU32TextMap = new Map<Text, U32>();
+mockU32TextMap.set(new Text('bazzing'), new U32(69));
+const mockU32TextMapString = '{"bazzing":69}';
+const mockU32TextMapObject = { bazzing: 69 };
+const mockU32TextMapHexString = '0x041c62617a7a696e6745000000';
+const mockU32TextMapUint8Array = Uint8Array.from([4, 28, 98, 97, 122, 122, 105, 110, 103, 69, 0, 0, 0]);
+
+const mockU32U32Map = new Map<U32, U32>();
+mockU32U32Map.set(new U32(1), new U32(2));
+mockU32U32Map.set(new U32(23), new U32(24));
+mockU32U32Map.set(new U32(28), new U32(30));
+mockU32U32Map.set(new U32(45), new U32(80));
+const mockU32U32MapString = '{"1":2,"23":24,"28":30,"45":80}';
+const mockU32U32MapObject = { 1: 2, 23: 24, 28: 30, 45: 80 };
+const mockU32U32MapHexString = '0x10043102000000083233180000000832381e00000008343550000000';
+const mockU32U32MapUint8Array = Uint8Array.from([16, 4, 49, 2, 0, 0, 0, 8, 50, 51, 24, 0, 0, 0, 8, 50, 56, 30, 0, 0, 0, 8, 52, 53, 80, 0, 0, 0]);
 
 describe('BTreeMap', (): void => {
   describe('decoding', (): void => {
-    const testDecode = (type: string, input: any): void =>
+    const testDecode = (type: string, input: any, output: string): void =>
       it(`can decode from ${type}`, (): void => {
         const s = new BTreeMap(Text, U32, input);
 
-        expect(s.toString()).toBe('{"bazzing":69}');
+        expect(s.toString()).toBe(output);
       });
 
-    testDecode('map', expectedMap);
-    testDecode('hex', '0x041c62617a7a696e6745000000');
-    testDecode('Uint8Array', Uint8Array.from([4, 28, 98, 97, 122, 122, 105, 110, 103, 69, 0, 0, 0]));
+    testDecode('map', mockU32TextMap, mockU32TextMapString);
+    testDecode('hex', mockU32TextMapHexString, mockU32TextMapString);
+    testDecode('Uint8Array', mockU32TextMapUint8Array, mockU32TextMapString);
+
+    testDecode('map', mockU32U32Map, mockU32U32MapString);
+    testDecode('hex', mockU32U32MapHexString, mockU32U32MapString);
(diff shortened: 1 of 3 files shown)


jacogr:
Brilliant! Thank you.

xlc:
I am a bit unsure about the name `BTreeMap` here. On Rust side, the underlying implementation is indeed sorted `BTreeMap`. But here at JS side the implementation it is just a JS map. It is definitely out of the scope to implement a `BTreeMap` data structure in JS, but I am mostly worried about the potential confusion caused by mismatched behaviour.

jacogr:
The CI test failures are caused by a (hidden to forks) token. Will merge a fix for that into master, so it will go away. (Will just link through here - had the same on another repo yesterday as well)

I understand your concerns, but believe the naming is fine since the codec is generally used to map directly from the available SCALE types, so devs will know what to look for. And once we get the creating via type-string in as well, i.e. use can specify 

```js
  types: {
    MyMap: 'BTreeMap<Text, u32>`
  }
})
```

It makes even more sense since at this level we certainly want to try and keep the gymnastics of swapping between the Rust-defs and JS-devs as small as possible.

yourheropaul:
I assume I still need to fix the codeclimate errors though, @jacogr?

jacogr:
@yourheropaul The PR is out for the update, so once that goes through I'll merge and a master merge in here should return things to normality.

For the CC issues, I would certainly look at the first 2 issues to just reduce the complexity - we generally take these as a "guide of where to look" rather than a "you must do this or else" rule. (Hence not a "it must pass" check) 

CC is a pain, but generally do show at least where there could be problems coming back to stuff 6 months down the line. So in this case if you are up for it, would probably just split the "monster decoder" into separate functions to keep it small and to-the-point.

packages/types/src/codec/BTreeMap.spec.ts
```diff
@@ -0,0 +1,76 @@
+import '../injector';
```
jacogr: (nit) Add the copyright header

packages/types/src/codec/BTreeMap.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+import BTreeMap from './BTreeMap';
+import Text from '../primitive/Text';
+import U32 from '../primitive/U32';
+import { CodecTo } from '../types';
+import Struct from './Struct';
```
jacogr:
```suggestion
import { CodecTo } from '../types';

import Text from '../primitive/Text';
import U32 from '../primitive/U32';
import BTreeMap from './BTreeMap';
import Struct from './Struct';
```

packages/types/src/codec/BTreeMap.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+  describe('encoding', (): void => {
+    const testEncode = (to: CodecTo, expected: any): void =>
+      it(`can encode ${to}`, (): void => {
+        const s = new BTreeMap(Text, U32, expectedMap);
+        expect(s[to]()).toEqual(expected);
+      });
+
+    testEncode('toHex', '0x041c62617a7a696e6745000000');
+    testEncode('toJSON', { bazzing: 69 });
+    testEncode('toU8a', Uint8Array.from([4, 28, 98, 97, 122, 122, 105, 110, 103, 69, 0, 0, 0]));
```
jacogr: I think we need one test to just reflect for than one entry as well, so make sure the . in/out is fine. (Possibly "steal" one from the scale-codec, so at least we know we have parity on both sides)

packages/types/src/codec/BTreeMap.ts
```diff
@@ -0,0 +1,183 @@
+import { Compact, U8a } from '.';
```
jacogr: (nit) Copyright header

jacogr:
```suggestion
import Compact from './Compact';
import U8a from './U8a';
```

Rationale for sticking to single imports - the `BTreeMap` will be in `index.ts` as well and we want to be explicit to avoid any chance of circular deps. (We have some issues in this regard)

packages/types/src/codec/BTreeMap.ts
```diff
@@ -0,0 +1,183 @@
...
+import { compareMap, decodeU8a } from './utils';
+import { AnyJson, Constructor, Codec, IHash } from '../types';
```
jacogr: Put the types import first, then the rest.

Additionally - first the package imports, then our won.

packages/types/src/codec/BTreeMap.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * @description The length of the value when encoded as a Uint8Array
+   */
+  public get encodedLength (): number {
+    let len = Compact.encodeU8a(this.size).length;
+    this.forEach((v: V, k: K) => {
+      len += v.encodedLength + k.encodedLength;
+    });
+    return len;
```
jacogr:
```suggestion
    return this.reduce((len, v: V, k: K) => {
      return len + v.encodedLength + k.encodedLength;
    }, Compact.encodeU8a(this.size).length);
```

packages/types/src/codec/index.ts
```diff
@@ -18,6 +18,7 @@ export { default as Struct } from './Struct';
 // export { default as StructAny } from './StructAny';
 export { default as Tuple } from './Tuple';
 export { default as Vec } from './Vec';
+export { default as BTreeMap } from './BTreeMap';
```
jacogr: (nit) Alphabetical

packages/types/src/codec/BTreeMap.spec.ts
yourheropaul: I took the fixture from https://github.com/paritytech/parity-scale-codec/blob/4a559de5f6fece90205a159e008b86304d4c8750/src/encode_like.rs#L130. It seems arbitrary, but at least now the JavaScript and the Rust tests use the same fixtures.

yourheropaul:
On the subject of type-strings (given in the example above and repeated here):

> ```js
> const api = await Api.create({
>   types: {
>     MyMap: 'BTreeMap<Text, u32>`
>   }
> })
> ```

We may not be able (or want to) to implement BTreeMaps with key and value types from the metadata (see discussion at https://github.com/paritytech/substrate/issues/3795) which means that there may be an inherent contradiction with string types. I guess we'll have to wait and see how or if to proceed with implementing them on the JS side.

jacogr:
As per this comment - https://github.com/polkadot-js/api/issues/1476#issuecomment-541084843 - we would need full support to get part the "types lose info" point. (The type parser needs to understand that `BTreeMap` needs generics and that they contain required info)

PS: The same thing actually happened with `Result<T, E>` prior to us doing full support.

PR #1483, opened 2019-10-14
Allow BTreeMap.with('u32', 'Text') (not only classes)

jacogr:


jacogr:
The actual `createType` now generates via strings - so this path is the one being tested.

PR #1494, opened 2019-10-15
Split Extrinsic into base and interface

jacogr:


jacogr:
Closing, needs to be done, but don't have a gap to solve conflicts after V4

PR #1498, opened 2019-10-19
Extrinsic v4

jacogr:
- Support for `MultiSignature` as per https://github.com/paritytech/substrate/pull/3861
- Assumes https://github.com/paritytech/substrate/pull/3904 is available

Note: The v2, v3, v4 needs a serious cleanup - there is a bunch of duplicated stuff now, which is certainly not great. (However need the updated support, so kick the can down the road a tiny bit)


jacogr:
Tested against the apps UI, substrate merged, here we go...

PR #1503, opened 2019-10-22
Exclude build folders

jacogr:
@amaurymartiny Still cannot reproduce your out-of-memory, but try this one - I had an OOM on the apps UI, this solved it there

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,10 @@
 {
   "extends": "./node_modules/@polkadot/dev/config/tsconfig",
-  "compilerOptions":
-  {
+  "exclude": [
+    "build/**/*",
+    "**/build/**/*"
+  ],
+  "compilerOptions": {
     "baseUrl": ".",
     "paths":
     {


amaurymartiny:
Yes, might be a good idea. Let me pull and run master/this branch to compare

jacogr:
It is in .eslintignore, however the typescript-eslint uses the tsconfig... Anyway, even with dirty build folders, eslint doesn't give me issues in this repo but like I mentioned, in apps this config change fixed it (at the very least, can't hurt)

jacogr:
Just wtf is on your machine? My specs are effectively the same. Does a `yarn clean && yarn lint` help in your case? (Which would suggest we are still pulling in `build/*` somewhere...)

There are some typescript-eslint CPU issues on VS Code as well (unfortunately not fixed in yestreday's 2.5.0 yet, but seems to have a PR)

amaurymartiny:
Alright, found the culprit. I still had the old `_book` directory in my root folder. I deleted that folder, and now linting works

PR #1504, opened 2019-10-23
Allow user-define RPC interfaces

jacogr:
Effectively this means we can add new RPC definitions as per user request, i.e.

```js
  provider: ...,
  types: ...,
  rpc: {
    testing: [
      {
        name: 'foo',
        params: [
          {
            name: 'bar',
            type: 'u32
          }
        ],
        type: 'Balance'
      }
    ]
  }
});
```

Format of the rpc map (if not obvious from the above) -

```js
<sectionA>: [
  {
    name: '...' // methodname
    type: '...' // returntype
    params: [
      { name: 'blah', type: 'u32' },
      { name: 'bleh', type: 'Hash', isOptional: true }
    ] // array of params
  }
] // array of methods includes,
<sectionB>: [...]
```

These are not really useable via the apps UI (it will display them, but then bail when selecting), but does appear in the `api.rpc.*` list (provided, obviously, it is included in the `rpc_methods` RPC)

cc @gnunicorn

### DIFF
--- a/packages/rpc-core/src/index.ts
+++ b/packages/rpc-core/src/index.ts
@@ -131,10 +131,10 @@ export default class Rpc implements RpcInterface {
 
   private createInterfaces<Section extends keyof RpcInterface> (interfaces: Record<string, RpcSection>, userBare: UserRpc): void {
     // these are the base keys (i.e. part of jsonrpc)
-    this.sections.concat(...Object.keys(interfaces));
+    Array.prototype.push.apply(this.sections, Object.keys(interfaces));
 
     // add any extra user-defined sections
-    this.sections.concat(...Object.keys(userBare).filter((key): boolean => !this.sections.includes(key)));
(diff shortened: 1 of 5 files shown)


packages/rpc-core/src/index.ts
```diff
@@ -123,21 +127,55 @@ export default class Rpc implements RpcInterface {
     return `${Rpc.signature(method)}:: ${error.message}`;
   }
 
-  private createInterface<Section extends keyof RpcInterface> (section: Section): RpcInterface[Section] {
-    const { methods } = interfaces[section];
+  private createInterfaces<Section extends keyof RpcInterface> (interfaces: Record<string, RpcSection>, userBare: UserRpc): void {
+    // these are the base keys (i.e. part of jsonrpc)
+    Array.prototype.push.apply(this.sections, Object.keys(interfaces));
```
amaurymartiny: Out of curiosity, why not `this.sections.push()`?

Issue #1506, opened 2019-10-23
Storage queries should error on too few and too many args


jacogr:
Currently it does not complain when sending too many args.

PR #1507, opened 2019-10-23
Validate storage args

jacogr:
Closes https://github.com/polkadot-js/api/issues/1506


jacogr:
Aware of that. You don't have the information at the "lower-level" at all. At that point, you only have already trimmed the args. So e.g. if you passed `fn(123, 435)` to a map, you would only get the first through.

The reason why I have not (yet) removed the low-level checks is that I'm not mucking about inside api-metadata at all atm, at least not until we have the metadata stuff merged.

PR #1532, opened 2019-10-30
Cleanup staking info

jacogr:
The most important here is probably 

- V1/V2 split (this makes it easy to actually optimize V2 and yank V1 as/when)
- V2 queueKeys are not part of multi (this is for everybody, it really was crazy having it as part of the multi)

### DIFF
--- a/packages/api-derive/src/accounts/idAndIndex.ts
+++ b/packages/api-derive/src/accounts/idAndIndex.ts
@@ -30,6 +30,9 @@ export type AccountIdAndIndex = [AccountId?, AccountIndex?];
  * ```
  */
 export function idAndIndex (api: ApiInterfaceRx): (address?: Address | AccountId | AccountIndex | string | null) => Observable<AccountIdAndIndex> {
+  const idToIndexCall = idToIndex(api);
+  const indexToIdCall = indexToId(api);
+
   return (address?: Address | AccountId | AccountIndex | string | null): Observable<AccountIdAndIndex> => {
     try {
       // yes, this can fail, don't care too much, catch will catch it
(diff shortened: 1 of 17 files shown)


packages/types/src/codec/create/createType.ts
```diff
@@ -17,13 +17,27 @@ function u8aHasValue (value: Uint8Array): boolean {
 // With isPedantic, actually check that the encoding matches that supplied. This
 // is much slower, but verifies that we have the correct types defined
 function checkInstance<T extends Codec = Codec, K extends string = string> (value: Uint8Array, created: FromReg<T, K>): void {
-  // For option, we are not adding  the initial byte (this is via storage)
-  const crHex = created.toHex(true);
+  // the underlying type created.toRawType()
+  const rawType = created.toRawType();
+
+  // ignore bytes completely - this is probably a FIXME, since these are somewhat
```
jacogr: This is probably "HACK"

 I wish I could spend time making this completely robust atm, but alas... it will need to wait

jacogr:
Current status ... just need V1 working (V2 is working).

packages/api-derive/src/accounts/idAndIndex.ts
```diff
@@ -30,6 +30,9 @@ export type AccountIdAndIndex = [AccountId?, AccountIndex?];
  * ```
  */
 export function idAndIndex (api: ApiInterfaceRx): (address?: Address | AccountId | AccountIndex | string | null) => Observable<AccountIdAndIndex> {
+  const idToIndexCall = idToIndex(api);
+  const indexToIdCall = indexToId(api);
```
jacogr: Did this everywhere now - so these don't get re-created.

PR #1535, opened 2019-11-01
Sub/unsub cleanups

jacogr:


packages/api-derive/src/util/memo.ts
```diff
@@ -0,0 +1,13 @@
+
+// Copyright 2017-2019 @polkadot/api-derive authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import memoizee from 'memoizee';
+
+export function memo <T extends Function> (fn: T): T {
```
amaurymartiny: Can put this one too in rpc-core, it's the same code as https://github.com/polkadot-js/api/blob/master/packages/rpc-core/src/index.ts#L298-L306

packages/api-derive/src/accounts/info.ts
```diff
@@ -7,39 +7,44 @@ import { AccountId, AccountIndex, Address, Balance } from '@polkadot/types/inter
...
+      : of(undefined);
+}
+
 /**
  * @name info
  * @description Returns aux. info with regards to an account, current that includes the accountId, accountIndex and nickname
  */
 export function info (api: ApiInterfaceRx): (address?: AccountIndex | AccountId | Address | string | null) => Observable<DeriveAccountInfo> {
   const idAndIndexCall = idAndIndex(api);
+  const nickCall = memo(retrieveNick(api));
```
amaurymartiny: Do you see a difference with memo like this?

It's unclear what memo does here: `retrieveNick(api)` creates a new function everytime, so everytime we cache a new function.

Intuitively i would do:
```
const retrieveNickMemo = memo(retrieveNick);
const nickCall = memo(retrieveNickMemo(api));
```

packages/api-derive/src/util/memo.ts
jacogr: Almost the same... however length is not added.

packages/api-derive/src/accounts/info.ts
jacogr: Yes, that memo dropped the connections. I had a log inside the `retrieveNick`, initially it fired each time, then... well, it stopped. (It is most probably unneeded atm after the refCountDelay, have not checked, on 3 things at once)

packages/rpc-core/src/rxjs/refCountDelay.ts
```diff
@@ -0,0 +1,47 @@
...
+    let trackerConnection: Subscription;
+    let subscriptionTracker = subscribeUpdates.pipe(
+      // scan  for updates
+      scan((total, change) => change + total, 0),
+      switchMap((count) => {
+        // when we have a zero count, schedule a unsubscribe
+        return count === 0
+          ? timer(delay, asapScheduler).pipe(
+            tap((): void => {
+              source.connect().unsubscribe();
```
amaurymartiny: This looks a bit weird, I would write:

```typescript
    // We are using connect, cast it
    const source = _source as ConnectableObservable<T>;
    let sourceSubscription: Subscription | null = null;

    const subscribeUpdates = new Subject<number>();
    let trackerConnection: Subscription;
    const subscriptionTracker = subscribeUpdates.pipe(
      // scan  for updates
      scan((total, change) => change + total, 0),
      switchMap((count) => {
        // when we have a zero count, schedule a unsubscribe
        return count === 0
          ? timer(delay, asapScheduler).pipe(
            tap((): void => {
              sourceSubscription && sourceSubscription.unsubscribe();
              trackerConnection.unsubscribe();
            })
          )
          : NEVER;
      })
    );

    const onNewSubscriber = () => {
      sourceSubscription = source.connect();
      trackerConnection = subscriptionTracker.subscribe();
      subscribeUpdates.next(1);

      return {
        unsubscribe: (): void => {
          subscribeUpdates.next(-1);
        }
      };
    };
```

PR #1540, opened 2019-11-03
Contracts with Data return

jacogr:


jacogr:
Confirmed working by @kwingram25

PR #1584, opened 2019-11-18
Make api.const.*.* typed

### LINKED ISSUE #1490: Cleanup/add api.consts.<s>.<m> typings
We really should need to do this - https://github.com/polkadot-js/api/pull/1489/files#diff-2c17a9848687821518cc295f7baa4d99R20-R25

amaurymartiny:
fixes #1490

### DIFF
--- a/packages/api/src/consts.types.ts
+++ b/packages/api/src/consts.types.ts
@@ -13,83 +13,83 @@ declare module '@polkadot/metadata/Decorated/types' {
   export interface Constants {
     babe: {
       [index: string]: Codec;
-      epochDuration: u64 & ConstantCodec;
-      expectedBlockTime: Moment & ConstantCodec;
+      epochDuration: u64;
+      expectedBlockTime: Moment;
     };
     timestamp: {
       [index: string]: Codec;
-      minimumPeriod: Moment & ConstantCodec;
+      minimumPeriod: Moment;
     };
     balances: {
       [index: string]: Codec;
-      existentialDeposit: Balance & ConstantCodec;
-      transferFee: Balance & ConstantCodec;
-      creationFee: Balance & ConstantCodec;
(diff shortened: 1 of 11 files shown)


packages/api/src/types.ts
```diff
@@ -2,8 +2,9 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-// `query.types` adds module augmentation so that `api.query.*.*` gets typed.
-// Import that file to make these types effective.
+// These two files add module augmentation so that `api.{query,consts}.*.*`
+// get typed. Import them to make these types effective.
+import './consts.types';
```
jacogr: :D

packages/api/src/consts.types.ts
```diff
@@ -0,0 +1,95 @@
...
+import { Gas } from '@polkadot/types/interfaces/contracts';
+import { VoteIndex } from '@polkadot/types/interfaces/elections';
+import { SessionIndex } from '@polkadot/types/interfaces/session';
+import { EraIndex } from '@polkadot/types/interfaces/staking';
+import { Codec } from '@polkadot/types/types';
+
+declare module '@polkadot/metadata/Decorated/types' {
+  export interface Constants {
+    babe: {
+      [index: string]: Codec;
```
amaurymartiny: The reason why this is not a ConstantCodec is that with ConstantCodec, users cannot do `api.const.a.b as Balance` without a TS error "Balance has not enough overlapping with ConstantCodec"

jacogr: That is fine. Really don't see any issue with it as it stands. (It actually simplifies things somewhat, since we use Codec everywhere so it it not a new type that comes in)

PR #1518, opened 2019-10-26
Allow UInt<bitLength>

jacogr:
Closes https://github.com/polkadot-js/api/issues/1516

It still has some serious issues where it detects classOf Balance for effectively everything.

PR #1592, opened 2019-11-19
Pass registry to all codec types

jacogr:
WIP, mucking about.


packages/types/src/types.ts
```diff
@@ -86,6 +86,11 @@ export interface Codec {
    */
   isEmpty: boolean;
 
+  /**
+   * @description The registry associated with this object
+   */
+  readonly registry: Registry;
```
amaurymartiny: I don't think users should do ever `api.type('u32',2 ).registry`, would argue to remove this and put protected in class

packages/types/src/primitive/U8.ts
```diff
@@ -12,7 +12,7 @@ import UInt from '../codec/UInt';
  * An 8-bit unsigned integer
  */
 export default class U8 extends UInt {
-  constructor (value?: AnyNumber) {
-    super(value, 8);
+  constructor (registry: Registry, value?: AnyNumber) {
```
amaurymartiny: Overall I'd suggest `constructor ({ registry }: { registry: Registry })`. With this signature it'd be extensible to pass ss58 id to Address and metadata to Event&Method

jacogr: The metadata ref. will probably be added to the registry, that is the idea, but I'm not there yet, so dunno. 

I started off with a `context: { registry: Registry }`, but that creates untold messes inside the actual registry where the values are passed though to `createType`. So, the calling context here for any type-related stuff will be the registry, and nothing more. (Which can be extended as the need arrises)

packages/types/src/types.ts
jacogr: Ok, won't remove, but can add protected since it is only internally to the object and it's derivates.

jacogr: ... adding `protected` is not actually allowed in the interface definition. It is exceptionally useful having it as "Codec", since we need to ensure that all types do implement it. so, rock, hard place.

packages/types/src/codec/Compact.spec.ts
```diff
@@ -109,52 +109,52 @@ describe('Compact', (): void => {
...
       expect(
-        new Compact('BlockNumber', 0xfffffff9).encodedLength
+        new (Compact.with('BlockNumber'))(registry, 0xfffffff9).encodedLength
       ).toEqual(5);
     });
 
     it('has the correct encodedLength for constructor values (class BlockNumber)', (): void => {
       expect(
-        new Compact(ClassOf('BlockNumber'), 0xfffffff9).encodedLength
+        new (Compact.with(ClassOf(registry, 'BlockNumber')))(registry, 0xfffffff9).encodedLength
```
amaurymartiny: Looking at these, we could put Compact, Uint, Vec, Tuple, Struct etc as abstract classes, and enforce the usage `.with()` which will create a real class

amaurymartiny: Throwing ideas again, an alternative syntax would be `new Compact.with(registry.BlockNumber)`, on in general `new (api.)registry.BlockNumber('234')`

packages/types/src/primitive/Extrinsic/SignerPayload.spec.ts
```diff
@@ -2,16 +2,20 @@
...
 import Metadata from '@polkadot/metadata/Metadata';
 import rpcMetadata from '@polkadot/metadata/Metadata/static';
 
 import { TypeRegistry } from '../../codec';
 import SignerPayload from './SignerPayload';
 
+const registry = new TypeRegistry();
+
+// eslint-disable-next-line no-new
+new Metadata(registry, rpcMetadata);
```
amaurymartiny: Could we do `const registry = new TypeRegistry(meta: Metadata | string)` (string being rpc). This way we could remove `setMetadata` from Metadata's constructor, and make it private.

amaurymartiny: hmm, maybe not, because Metadata needs registry already

packages/types/src/codec/Compact.spec.ts
jacogr: I would prefer them to be used that way, yes. Probably after https://github.com/polkadot-js/api/pull/1518 (which has it's own issues, but opens the way for removing the `U{8,16,32,64,128,256}` classes (and allows stuff like `U2048`)

packages/types/src/primitive/Extrinsic/SignerPayload.spec.ts
jacogr: So the above is only in tests - and it is a bit of magic, as per the comment suppression. I was thinking about doing that, but it actually only helps in tests - in the Api itself the registry is created right-up-front before we even have metadata, so we still need to inject or call `setMetadata` (Originally I have an explicit set, now metadata does it).

Your approach may be cleaner, at least for "it not just doing stuff" and will certainly clean up  the tests.

jacogr: Still creates a small issue in the Decorated tests, but can call `setMetadata` there explicitly like we have in the Api then. Ok, will follow the `meta?: string | Uint8Array` and see where it gets us.

amaurymartiny: Yeah, actually I don't think metadata should go inside registry. Ideally, I would see a `class Context` which has this interface:

```
constructor(registry: Registry, metadata?: Metadata);
registry: Registry;
metadata?: Metadata;
setMetadata: (meta: string|8a);
```

In tests that don't need metadata, we just `new Context`. In tests with calls/events, we do an additional `context.setMetadata()`

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 0.97.0-beta.x
+
+- **Breaking change** Add the passing on an explicit `Registry` to all calls to `createType` as well as all codec classes. If you are upgrading form an older version, use the [migration guide](UPGRADING.md) to walk through the changes required.
+- **Breaking change** The metadata classes have been moved from `@polkadot/types` into `@polkadot/metadata`. If you explicitly create `Metadata` classes, you now would need to do so with the `import { Metadata } from '@polkadot/metadata';`
+- `api.derive.*` now has built-in memomization
+- Various fixes and cleanups to `pi.derive.*` including era length progress and vesting calculations
```
amaurymartiny:
```suggestion
- Various fixes and cleanups to `api.derive.*` including era length progress and vesting calculations
```

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 0.97.0-beta.x
+
+- **Breaking change** Add the passing on an explicit `Registry` to all calls to `createType` as well as all codec classes. If you are upgrading form an older version, use the [migration guide](UPGRADING.md) to walk through the changes required.
+- **Breaking change** The metadata classes have been moved from `@polkadot/types` into `@polkadot/metadata`. If you explicitly create `Metadata` classes, you now would need to do so with the `import { Metadata } from '@polkadot/metadata';`
```
amaurymartiny: We don't export Metadata, should be `import Metadata from '@polkadot/metadata';` (and this one is the decorated now, which should be the user-facing one).

So would also add a note about `Metadata.metadata`

UPGRADING.md
```diff
@@ -2,7 +2,91 @@
...
 
-While we try to keep the user-facing interfaces as stable as possible, sometimes you just need to make additions to move forward and improve things down the road, as painful as they may be. Like you, we are also users of the API, and eat our own dogfood - and as such, feel any pains introduced first.
+While we try to keep the user-facing interfaces as stable as possible, sometimes you just need to make additions to move forward and improve things down the road, as painful as they may be. Like you, we are also users of the API, and eat our own dog food - and as such, feel any pains introduced first.
+
+
+## 0.97.1 (and newer), from 0.90.1 (and older)
+
+The 0.97 series lays the groundwork to allow type registration  to be ties to a specific chain and a specific Api instance. In the past, 2 Api instances in the same process would share types, which mean that you could not connect to 2 independent chains with different types. This is very problematic for Polkadot chains, where the ide is to connect to multiple chains.
```
amaurymartiny:
```suggestion
The 0.97 series lays the groundwork to allow type registration  to be ties to a specific chain and a specific Api instance. In the past, 2 Api instances in the same process would share types, which mean that you could not connect to 2 independent chains with different types. This is very problematic for Polkadot chains, where the idea is to connect to multiple chains.
```

UPGRADING.md
```diff
@@ -2,7 +2,91 @@
...
+
+...
+const registry = new TypeRegistry();
+const blockNumber = createType(registry, 'BlockNumber', 12345);
+const api = await ApiPromise.create({ registry });
+```
+
+### Extrinsic metadata
+
+In some applications, the undocumented `findFunction` has been used to determine the Api has the metadata for a specific extrinsic. The has been exposed on to of `GenericCall`, and it typically used in applications such as signers. Along with the compulsory registry, the above functions have been moved to the `Registry` itself, so if you previously had -
```
amaurymartiny:
```suggestion
In some applications, the undocumented `findFunction` has been used to determine the Api has the metadata for a specific extrinsic. The has been exposed on top of `GenericCall`, and it typically used in applications such as signers. Along with the compulsory registry, the above functions have been moved to the `Registry` itself, so if you previously had -
```

jacogr:
Well, it works. Not perfect, the CC here is f-ing out of control. Decent start.

PR #1620, opened 2019-11-29
Perbill/Permill with helpers

jacogr:
Adds Perbill/Permill with toPercentate/fromPercentage helpers


jacogr:
Need to think through this one, it adds extra types :( - less types, more definitions. Maybe alongside `Uint<..>` add a `PerX<...>`

PR #1623, opened 2019-12-01
Add 6x to conviction

jacogr:


jacogr:
Ok, this still needs work and improvement - however need to roll this out to Kusama ASAP for the next democracy support. So next round will start cleaning up.

PR #1626, opened 2019-12-02
Add signAsync to SubmittableExtrinsic

### LINKED ISSUE #1624: The `sign` method of the `Submittable` class doesn't support signing via external signers
Conversely, `signAndSend` does support external signers

c410-f3r:
Fixes #1624 

To avoid breaking changes, a new method was added instead of changing
the `sign` signature to be async.

### DIFF
--- a/packages/api/src/promise/Api.spec.ts
+++ b/packages/api/src/promise/Api.spec.ts
@@ -78,4 +78,16 @@ describe('ApiPromise', (): void => {
       ).toEqual(SIG);
     });
   });
+
+  describe('decorator.signAsync', (): void => {
+    it('signs a transfer using an external signer', async (): Promise<void> => {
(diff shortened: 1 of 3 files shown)


jacogr:
Happy with the approach. Had the same in mind a month-or-so ago, never got around to it. What I do suggest is a slight rework -

- make is `signAsync`
- make it work with both pairs & external signers

That way we can just starting to nudge people into that direction as a one-stop-shop. (Additionally, it allows the possibility to fill in the options as well as we do with the async signAndSend - that we can leave for future enhancement)

c410-f3r:
Function name changed from `signWithSigner` to `signAsync`.

@jacogr The current implementation works with either pairs or external signers

c410-f3r:
Thanks!

PR #1639, opened 2019-12-04
Add BTreeSet<V> Codec

satellitex:
This PR is that we will use a default type `BTreeSet<V>` in PolkadotJS.
- Test of BTreeSet implements Codec.
- Type Defined of BTreeSet implements Codec.


jacogr:
Thank you!

Have had a brief look, and it looks good. Will do a proper review once the linting issues has been sorted.

jacogr:
To get it to work properly, i.e. by padding `BTreeSet<u32>` as a string type registrations, the following is also needed -

- Exclude it from santize so it keeps the generic params - https://github.com/polkadot-js/api/blob/master/packages/types/src/codec/create/sanitize.ts#L8
- Allow it to decode into a typedef, https://github.com/polkadot-js/api/blob/master/packages/types/src/codec/create/getTypeDef.ts#L106

satellitex:
@jacogr Thank you for reviewing. And I added implementation pointed by you.

jacogr:
Generally looks good, we just need to take care of the testing & linting failures.

satellitex:
@jacogr Sorry for fixing late. I passed testing and listing.
But, I don't have an idea to fix codeclimate. What's about you?

jacogr:
Sorry about not getting to this today - had some npmjs downtime which had a serious knock-on effect. 

The CC issues are fine, it is guide, rather than fixed rules. 

Will all that said, I have it in the queue for the morning.

PR #1640, opened 2019-12-04
Adjust contract rpc formatData

kwingram25:


jacogr:
Closing in favor for #1643 which assumes that the RPC does the right thing, no work-arounds.

PR #1645, opened 2019-12-06
Use TypeDef[] for TypeDef.sub

kwingram25:


packages/api-contract/src/MetaRegistry.spec.ts
```diff
@@ -17,8 +17,6 @@ function compare (meta: MetaRegistry, other: any): void {
   try {
     expect(meta.typeDefs).toEqual(other);
   } catch (error) {
-    console.error(JSON.stringify(meta.typeDefs));
```
jacogr: We actually do want this - so in the case of stuff going wrong, we can actually see the expected output - and can actually easily adjust as needed.

jacogr: PS And it will restore the linting.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules/typescript/lib"
+}
```
jacogr: Probably need .vscode in .gitignore

jacogr: Still need to whack this file.

packages/metadata/src/Metadata/util/validateTypes.ts
```diff
@@ -20,7 +20,7 @@ function extractTypes (types: string[]): any[] {
       case TypeDefInfo.Compact:
       case TypeDefInfo.Option:
       case TypeDefInfo.Vec:
-        return extractTypes([(decoded.sub as TypeDef).type]);
+        return extractTypes([(decoded.sub as TypeDef[])[0].type]);
```
jacogr: Can just combine this with the ones below, i.e. like in Tuple just iterate through the array, less moving parts and logic.

jacogr:
Closing, very outdated by now.

PR #1694, opened 2020-01-02
Remove v9->v10 hack

amaurymartiny:
https://github.com/polkadot-js/api/blob/774f41e6db2f323e61690073fc7f60168f2296d0/packages/metadata/src/Metadata/Metadata.ts#L27-L36

- [x] Update static.polkadot.ts to Kusama 1038 so that tests pass


jacogr:
Will merge when CI passes, was actually just coming back to do exactly that. (But still waiting on `OpenTip` to pass)

PR #1697, opened 2020-01-03
Enum toJSON returns toString for a basic enum

amaurymartiny:
Before: for a basic Enum (non parametrized), toJSON returned toNumber, i.e. the index

After: it returns toString.

Why? Looks better, we can also move StorageHasher into interfaces. @jacogr i don't think this breaks anything right?


jacogr:
It actually does break. I do agree it looks better.

The issue is (I believe, need to double check), that the JSON representation of a c-like enum is the index from the Rust side. The `toJSON` needs compatibility withy that.

jacogr:
Wanted o post the folow-up yesterday and then didn' - been thinking abou a `toFormattted` that -

- formats enums nicely
- formats balances
- ... anything else that makes it human-readable

Never got around to it.

amaurymartiny:
Yes, agreed, that would be a good idea. I'm closing this for now then.

packages/types/src/codec/Enum.spec.ts
```diff
@@ -191,9 +191,9 @@ describe('Enum', (): void => {
     testDecode('Uint8Array', Uint8Array.from([0]), 'foo');
     testDecode('Uint8Array', Uint8Array.from([1]), 'bar');
 
-    testEncode('toJSON', 1);
+    testEncode('toJSON', 'bar');
     testEncode('toNumber', 1);
-    testEncode('toString', 'bar');
+    testEncode('toString', 'Bar');
```
amaurymartiny: Note: I didn't change the `.toString()`, that stays the same

jacogr:
A conflict seems to have snuck in.

PR #1698, opened 2020-01-06
Make submittable signAsync Observable-aware

jacogr:


jacogr:
cc @amaurymartiny Would just appreciate a check here, I am sure I'm being rusty in mangling Observables.

packages/api/src/submittable/createClass.ts
```diff
@@ -59,15 +59,24 @@ export default function createClass <ApiType extends ApiTypes> ({ api, apiType,
...
     //
     // also supports signing through external signers
-    public async signAsync (account: IKeyringPair, optionsOrNonce: Partial<SignerOptions>): Promise<this> {
-      if (this._api.signer) {
-        const options = this._makeSignOptions(this._optionsOrNonce(optionsOrNonce), {});
-        await this._signViaSigner(account.address, options, null);
-      } else {
-        this.sign(account, optionsOrNonce);
-      }
+    public signAsync (account: IKeyringPair, optionsOrNonce: Partial<SignerOptions>): SubmittableThis<ApiType, this> {
```
jacogr: cc @amaurymartiny - Here specifically. (Tests seems to work, just not 100% I'm doing the best here with the various return paths)

amaurymartiny: decorateMethod looks good, the return paths could be simplified into e.g.

```typescript
          return (
            this._api.signer
              ? from(this._signViaSigner(account.address, this._makeSignOptions(optionsWithNonce, {}), null))
              : of(0).pipe(tap(()=> this.sign(account, optionsWithNonce)))
          ).pipe(
            map(()=>this)
          )
```

aka `switchMap((): Observable<this> => from(Promise.resolve(this))) === map(()=>this)`

PR #1702, opened 2020-01-06
Add enum _alias, rename isFinalized to isInBlock

### LINKED ISSUE #1684: Rename isFinalized to isInBlock
However, on this one we need to ensure we have backwards compat - so if we don't have a class-interface for this, we should add it. (RPC still returns finalized)

kwingram25:
Closes #1684


jacogr:
Closing based on this https://github.com/paritytech/substrate/pull/4930/files

PR #1705, opened 2020-01-10
Fix `*.sub.map is not a function` when initing API

kigawas:


jacogr:
I have not seen reports of this out in the wild. Closing as-is - like I mentioned initially, if this does happen, it means we have a bug elsewhere where things are not constructed properly, a Tuple should always have a sub as Array, https://github.com/polkadot-js/api/blob/master/packages/types/src/codec/create/getTypeDef.ts#L82

Issue #1679, opened 2019-12-18
Update RPC methods for childStorage


Stefie:
There have been breaking changes introduced with https://github.com/paritytech/substrate/pull/4162

The `child_storage` rpc functions now expect two additional parameters, namely 
```
child_info: StorageKey,
child_type: u32,
```

See:
https://github.com/paritytech/substrate/pull/4162/files#diff-d0e6e9d0dec379376c4381b7b12a0b4fR63

PR #1709, opened 2020-01-10
Updated child storage parameters

Stefie:
Fixes https://github.com/polkadot-js/api/issues/1679

I ran `yarn run docs` because I noticed the docs are checked in, but happy to revert the files if that's not the workflow.

It can't really be tested at the moment, because the only two pallets using child_storage are pallet_contracts in substrate and the crowdfunding module for parachains in Kusama.


jacogr:
No docs, please.

Stefie:
Ok, reverted.

Stefie:
If that is all that is needed to get the new childStorage parameters in, it would be great if we could merge it! We're making use of the JS API for contract development and it's currently broken and not really usable for us with newer Substrate versions because of the additional parameters.

jacogr:
???

This is nowhere near merge-able -

- Linting fails
- Build fails

We are not breaking master, if this gets merged nothing will publish.

Stefie:
Yup, sorry, that was actually a quick shot... The linting failed on the tests, which were broken before and Amaury told me that they're most likely going to be removed soon anyways, that's why I didn't pay closer attention.
I'm testing it with with ink! and contracts now to help to pin down @Robbepop's bug

PR #1726, opened 2020-01-19
Fix api.derive[collective] returning undefined error

kwingram25:


packages/api-derive/src/collective/proposals.ts
```diff
@@ -3,28 +3,30 @@
...
+              hashes && hashes.length
+                ? combineLatest(
+                  hashes.map((hash): Observable<Option<Proposal>> => api.query[section].proposalOf<Option<Proposal>>(hash))
+                )
+                : of([]),
               api.query[section].voting.multi<Option<Votes>>(hashes)
             ]);
           }),
-          map(([hashes, proposals, votes]): DerivedCollectiveProposals => {
+          map(([hashes = [], proposals = [], votes]): DerivedCollectiveProposals => {
```
jacogr: Why are the default required here?

packages/api-derive/src/collective/proposals.ts
```diff
@@ -3,28 +3,30 @@
...
       ? api.query[section].proposals()
         .pipe(
-          switchMap((hashes: Hash[]): Observable<[Hash[], Option<Proposal>[], Option<Votes>[]]> => {
+          switchMap((hashes: Hash[] = []): Observable<[Hash[], Option<Proposal>[], Option<Votes>[]]> => {
             return combineLatest([
               of(hashes),
-              combineLatest(
-                hashes.map((hash): Observable<Option<Proposal>> => api.query[section].proposalOf(hash))
-              ),
+              hashes && hashes.length
```
jacogr: Only `hashes.length`, it does get set above. (Should not be, but that is a different comment). The length-only check is sane/fine.

packages/api-derive/src/collective/proposals.ts
```diff
@@ -3,28 +3,30 @@
...
 import { map, switchMap } from 'rxjs/operators';
-import { Option } from '@polkadot/types';
 
 export function proposals (api: ApiInterfaceRx, section: 'council' | 'technicalCommittee'): () => Observable<DerivedCollectiveProposals> {
   return (): Observable<DerivedCollectiveProposals> =>
     api.query[section]
       ? api.query[section].proposals()
         .pipe(
-          switchMap((hashes: Hash[]): Observable<[Hash[], Option<Proposal>[], Option<Votes>[]]> => {
+          switchMap((hashes: Hash[] = []): Observable<[Hash[], Option<Proposal>[], Option<Votes>[]]> => {
```
jacogr: Under what conditions are hashes empty? This means there is an issue elsewhere since the query _always_ should return a `Codec` (in this case a `Vec`, which may have no elements)

packages/api-derive/src/collective/proposals.ts
kwingram25: holdover from other fix attempt :)

jacogr:
Replaced by https://github.com/polkadot-js/api/pull/1732 (Sorry, had a fix for the `.multi` stuff, so mucked-around in this file and then also added the `.length`)

PR #1753, opened 2020-01-27
Fill signAsync with expanded chain options

jacogr:
Should address the concern in https://github.com/polkadot-js/apps/pull/2203#issuecomment-578897811 (`signAsync` being completely self-standing here, same logic as `signAndSend`)


mzolkiewski:
Awesome, I actually started working on this myself but your implementation is better :)

PR #1758, opened 2020-01-28
Bump deps

jacogr:


SCaptainCAP:
This PR breaks compatibility with webpack because the next lines:
```javascript
try {
  pkgJson = require('./package.json');
} catch (error) {
  pkgJson = require('../package.json');
}
```

SCaptainCAP:
To fix this you should add try/catch to the second require

Issue #1742, opened 2020-01-23
Unable to connect + Uncatchable api error


jak-pan:
I'm trying to connect to my unsynced node with api 1.0.0-beta.17 
```
target=#750431 (25 peers), best: #520014
```
I'm getting 
![Screen Shot 2020-01-23 at 16 42 13](https://user-images.githubusercontent.com/3409250/72999295-74938080-3dff-11ea-9667-bd96dcc9c5bd.png)

which is uncatchable with both of these

```javascript 
try {
  let api = await ApiPromise.create().catch(e => {})
} catch (e) {}
```

It also doesn't emit error on 
```javascript
provider.on('error',(e)=>{})
```

It results in stuck application since I can't react on anything that happened here. It also happens when you get error while connecting to ``wsprovider`` but it can be handled by wrapping api creation inside another promise, which I can reject. It is not ideal, but it works.

When trying to connect to ``wss://cc3-5.kusama.network/`` I get this on the other hand

![Screen Shot 2020-01-23 at 16 41 42](https://user-images.githubusercontent.com/3409250/72999309-79583480-3dff-11ea-839f-8dc55b605d44.png)

jacogr:
In this case you got very unlucky on Kusama. The issue is that at some point metadata v9 was actually broken - a type was added without changing the metadata version. (The bump to metadata v10 fixed that - retrospectively)

However, despite the brokenness there, going to keep this issue as logged and see what we can do around the error handling so it properly propagates.

jacogr:
@jak-pan With the changes from @laec is this still an issue for you?

jak-pan:
This looks like a good fix. It should be documented however to catch apiPromise.create() since it can crash now.

jacogr:
Closing, documentation updates logged.

PR #1763, opened 2020-01-28
Make sure awaiting on an api object won't stuck forever on error

laec:
I ran into the same issue as #1742 when connecting to our custom chain, this seems to fixed it by rejecting the promise on error.


xlc:
FYI https://github.com/polkadot-js/api/pull/563

Issue #714, opened 2019-02-18
Generate TypeScript interfaces from Metadata


jacogr:
Basically would allow generation of interfaces as specified by metadata on a per chain basis, replacing the full class definitions as we currently have.

(After https://github.com/polkadot-js/api/pull/712, original idea for generation by @xlc )

jacogr:
When metadata v6 (reflection) is in-place.

PR #1766, opened 2020-01-29
Add polkadot-types-from-{chain,defs}

jacogr:
Closes https://github.com/polkadot-js/api/issues/714 (Can be expanded when/if we get on-chain types)

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -3,6 +3,10 @@
   "version": "1.0.0-beta.37",
   "description": "Implementation of the Parity codec",
   "main": "index.js",
+  "bin": {
+    "polkadot-types-from-chain": "./scripts/polkadot-types-from-chain.js",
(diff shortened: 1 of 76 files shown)


packages/types/package.json
```diff
@@ -3,6 +3,10 @@
   "version": "1.0.0-beta.37",
   "description": "Implementation of the Parity codec",
   "main": "index.js",
+  "bin": {
+    "polkadot-types-from-chain": "./scripts/polkadot-types-from-chain.js",
+    "polkadot-types-from-defs": "./scripts/polkadot-types-from-defs.js"
+  },
```
amaurymartiny: Why not put these in api? They are only relevant when using api.

jacogr: I thought to put them there - but since it is all type extraction, left them in types.

Issue #1779, opened 2020-01-30
Allow unsubscription from wsprovider.on()


jak-pan:
It's really hard to handle disconnects, reconnects or changing servers cleanly right now. We managed to be pretty close by using manual connect to ``wsProvider`` and doing some hacks with ``wsProvider.on('error', cb)``, ``wsProvider.on('connected', cb)`` and ``wsProvider.on('disconnected', cb)``

However returning unsubscribe function like when subscribing to API queries would be very helpful in this case. I've already tried to implement this and it solved most of our problems. I'll make it a PR. 

If that's not acceptable at least returning EventEmitter would be helpful but that might not be as user friendly.

https://github.com/polkadot-js/api/blob/de9f033d994bebe2145e5cd0155f77ef8471a3c0/packages/rpc-provider/src/ws/Provider.ts#L172

jacogr:
Happy to take a look.

jak-pan:
Thanks! That was quick. :) 

P.S. Should I start branch in this repo next time?

jacogr:
Closed in #1780 

For that you need commit access - happy to give it out for consistent contributors. So lets ice that until we have a couple more PRs under the belt and I'll gladly expand the pool with you included. Love contributions for actual users more than anything else.

PR #1780, opened 2020-01-30
Added unsubscribe return function to wsProvider.on()

jak-pan:
solves #1779

### DIFF
--- a/packages/rpc-provider/src/http/index.ts
+++ b/packages/rpc-provider/src/http/index.ts
@@ -84,8 +84,11 @@ export default class HttpProvider implements ProviderInterface {
    * @description HTTP Provider does not have 'on' emitters. WebSockets should be used instead.
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  public on (type: ProviderInterfaceEmitted, sub: ProviderInterfaceEmitCb): void {
+  public on (type: ProviderInterfaceEmitted, sub: ProviderInterfaceEmitCb): () => void {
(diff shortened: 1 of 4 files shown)


jacogr:
The idea looks sane. (And is much simpler to use than `off`). The interfaces needs to change throughout though (ProviderInterface, Http, Mock (?) - all indicated by the linting/build failures.

packages/rpc-provider/src/http/index.ts
```diff
@@ -84,8 +84,11 @@ export default class HttpProvider implements ProviderInterface {
    * @description HTTP Provider does not have 'on' emitters. WebSockets should be used instead.
    */
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  public on (type: ProviderInterfaceEmitted, sub: ProviderInterfaceEmitCb): void {
+  public on (type: ProviderInterfaceEmitted, sub: ProviderInterfaceEmitCb): () => void {
     l.error('HTTP Provider does not have \'on\' emitters, use WebSockets instead');
+    return (): void => {
+      // noop
```
jak-pan: Is this feasible or do I allow ProviderInterface to also return `void` or  do we split http / ws provider interface type defs?

jacogr: No, it is fine as-is. If we also return void than the called needs to check if it is empty or not - which just complicates things. As it stands, he/she/it can just ignore the return if no action is taken.

PR #1794, opened 2020-02-02
Override for Vote type (not a struct, not caught)

jacogr:


amaurymartiny:
I'm a bit late to the party. But looking at this PR, having `Vote extends Struct` (with custom encode/decode) would fit mentally better in my mind.

PR #1829, opened 2020-02-10
Expose type-only interfaces (e.g. Tuple from types/types)

jacogr:


jacogr:
Closed here, needs some thought first with other codec types.

Issue #1840, opened 2020-02-13
Add errors from metadata to document output


jacogr:
Like we have events, tx, storage, ... do the same for the errors (section, name, documentation)

PR #1841, opened 2020-02-13
Add errors to documentation

axelchalon:
* Add errors to docs, close https://github.com/polkadot-js/api/issues/1840
* Fix broken anchor links; in the {constants,storage,extrinsics,events} docs pages, Vuepress would generate for e.g. the `## imOnline` section an `imonline` anchor; however our TOC links to `#imOnline`

Also, fwiw I was thinking of adding [`docs/substrate`](https://github.com/polkadot-js/api/tree/master/docs/substrate) to the .gitignore; since the docs there are outdated and not automatically regenerated, it doesn't really make sense to have them versioned. what do you think?


jacogr:
For those, we can probably delete them as they are there and then .gitignore - since they are not committed by GH, the fact that they are outdated is just creating more issues.

PR #1864, opened 2020-02-15
Refine types of `entries` function for queries on double maps

monitz87:
This PR:

- Makes the `entries` function in `StorageEntryBase` accept an optional argument
- Introduces special `AugmentedQueryDoubleMap` and `StorageEntryDoubleMap` types that refine the type of the `entries` function argument
- Modifies the autogen script to detect if the storage entry is a double map and apply the aforementioned types if so
- Updates the autogenerated types to apply these changes

I thought about adding a type parameter to `AugmentedQuery` and `StorageEntryBase` to avoid repeating code, but it seemed hacky. This is the alternative solution (let me know if you prefer it)

It would look something like this
```
export type AugmentedQuery<ApiType extends ApiTypes, F extends AnyFunction, DoubleMapFirstKeyType extends any = false> = MethodResult<ApiType, F> & StorageEntryBase<ApiType, F, DoubleMapFirstKeyType>
```

And then in `StorageEntryBase`
```
export interface StorageEntryBase<ApiType extends ApiTypes, F extends AnyFunction, DoubleMapFirstKeyType extends any = false> {
  at: (hash: Hash | Uint8Array | string, ...args: Parameters<F>) => PromiseOrObs<ApiType, ObsInnerType<ReturnType<F>>>;
  creator: StorageEntry;
  entries: (arg?: DoubleMapFirstKeyType extends false ? any : DoubleMapFirstKeyType) => PromiseOrObs<ApiType, [StorageKey, ObsInnerType<ReturnType<F>>][]>;
  hash: (...args: Parameters<F>) => PromiseOrObs<ApiType, Hash>;
  key: (...args: Parameters<F>) => string;
  keyPrefix: () => string;
  size: (...args: Parameters<F>) => PromiseOrObs<ApiType, u64>;
  multi: ApiType extends 'rxjs' ? StorageEntryObservableMulti : StorageEntryPromiseMulti;
}
```

### DIFF
--- a/packages/api/src/types/augment/query.ts
+++ b/packages/api/src/types/augment/query.ts
@@ -97,7 +97,7 @@ declare module '@polkadot/api/types/storage' {
       /**
        * The set of open multisig operations.
        **/
-      multisigs: AugmentedQuery<ApiType, (key1: AccountId | string | Uint8Array, key2: Raw | string | Uint8Array) => Observable<Option<Multisig>>> & QueryableStorageEntry<ApiType>;
+      multisigs: AugmentedQueryDoubleMap<ApiType, (key1: AccountId | string | Uint8Array, key2: Raw | string | Uint8Array) => Observable<Option<Multisig>>, AccountId | string | Uint8Array> & QueryableStorageEntry<ApiType>;
     };
     babe: {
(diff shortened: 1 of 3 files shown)


jacogr:
Absolutely brilliant! Will take a look a bit later today.

PR #1876, opened 2020-02-17
Make bn.js typings into a regular dependency instead of a devDependency

monitz87:
Right now projects that use polkadot are forced to manually install `@types/bn.js`, otherwise they will get compilation errors because the typings are not found (the `AbstractInt` class exposes a `BN` instance in its `toBn` method). This PR fixes that


monitz87:
Just for reference, it appears that in previous versions, `@polkadot/util` had `@types/bn.js` as a dependency, but it was removed, causing the current issue

jacogr:
Yes, this PR is what caused the cascading effect - https://github.com/polkadot-js/common/pull/531/files

However the above actually pointed out some severe holes through the chain (dev -> util -> api -> others), for instance when using yarn pnp (as per yarn2), the packages are not just flattened and "available to all", it needs to be explicit. This was certainly not the case, e.g. in here where the dependency didn't appear anywhere.

The general rule is that "if a package does appear in the *.d.ts, it needs to be a normal dep", so this PR is correct. There will probably be 1 or 2 others "somewhere", was expecting it and needs to be dealt with on a case-by-case basis.

Issue #1678, opened 2019-12-18
Add manual seal RPCs (& associated types)


jacogr:
Introduced in https://github.com/paritytech/substrate/pull/4143

PR #1894, opened 2020-02-19
Add manual seal RPCs

axelchalon:
reference: https://github.com/paritytech/substrate/blob/master/client/consensus/manual-seal/src/rpc.rs
close https://github.com/polkadot-js/api/issues/1678


jacogr:
Thanks a million.

PR #1899, opened 2020-02-19
All heads subscription

### LINKED ISSUE #1898: Add support for all heads subscription
Introduced in https://github.com/paritytech/substrate/pull/4979

jak-pan:
Fixes #1898 when [#4979](https://github.com/paritytech/substrate/pull/4979) will be approved.

sorry for the history, probably best to squash.


jak-pan:
Looks like that, it looks very strange since current master is at .8 and it should compare my head which is also .8. I’ll rebase next time to avoid this. 

The code was tested with 4979 so it should be good when it merges.

PR #1934, opened 2020-02-24
fix: don't format "unorthodox" query input types

monitz87:
This PR:
- Allows enums and other complex types to be passed as inputs
- Adds better support to tuples in `getSimilarTypes` (`Tuple<[a, b]>` => `Tuple<[a, b]> | [a | ... similar_to_a ... , b | ... similar_to_b ...]`)
- Adds support for double maps with a dummy first argument

# Explanation:

## Complex ("unorthodox") types:

The current autogen fails on storage calls that include enums (and probably other complex types) as inputs

**Example storage call:**

```
{
  "name": "MultiSigSigners",
  "modifier": "Default",
  "type": {
    "DoubleMap": {
      "hasher": "Blake2_256",
      "key1": "AccountId",
      "key2": "Signatory",
      "value": "Signatory",
      "key2Hasher": "Blake2_256"
    }
  },
  "fallback": "0x000000000000000000000000000000000000000000000000000000000000000000",
  "documentation": [
    " Signers of a multisig. (mulisig, signer) => signer."
  ]
}
```

**Definitions:**

```
"Signatory":{
  "_enum": {
      "Identity": "IdentityId",
      "AccountKey": "AccountKey"
  }
},
"IdentityId": "H256",
"AccountKey": "[u8;32]",
```

Since `getSimilarTypes` will expand `Signatory` to
```
[ 'Signatory',
  '{ identity: any }',
  '{ accountKey: any }',
  'string',
  'Uint8Array' ]
```

The next call to `formatType` with `{ identity: any }` as an argument will throw
```
/Users/jere/Code/polkadot-types-test/node_modules/@polkadot/types/create/getTypeDef.js:57
    parsed = JSON.parse(type);
                  ^
SyntaxError: Unexpected token i in JSON at position 1
    at JSON.parse (<anonymous>)
    at Array._decodeStruct (/Users/jere/Code/polkadot-types-test/node_modules/@polkadot/types/create/getTypeDef.js:57:19)
    at getTypeDef (/Users/jere/Code/polkadot-types-test/node_modules/@polkadot/types/create/getTypeDef.js:147:21)
    at formatType (/Users/jere/Code/polkadot-types-test/node_modules/@polkadot/typegen/util/formatting.js:157:38)
    at similarTypes2.map.type (/Users/jere/Code/polkadot-types-test/node_modules/@polkadot/typegen/generate/query.js:60:73)
    at Array.map (<anonymous>)
    at entrySignature (/Users/jere/Code/polkadot-types-test/node_modules/@polkadot/typegen/generate/query.js:60:39)
    at concat.concat.storage.unwrap.items.map.storageEntry (/Users/jere/Code/polkadot-types-test/node_modules/@polkadot/typegen/generate/query.js:84:32)
    at Array.map (<anonymous>)
    at Array.map (/Users/jere/Code/polkadot-types-test/node_modules/@polkadot/types/codec/AbstractArray.js:156:27)
```

I fixed that by leaving complex types as-is (skipping the call to `formatType`)

## Null first key in double maps:

Example storage call:

```
{
  "name": "EventTopics",
  "modifier": "Default",
  "type": {
    "DoubleMap": {
      "hasher": "Blake2_256",
      "key1": "()",
      "key2": "Hash",
      "value": "Vec<(BlockNumber,EventIndex)>",
      "key2Hasher": "Blake2_256"
    }
  },
  "fallback": "0x00",
  "documentation": [
    " Mapping between a topic (represented by T::Hash) and a vector of indexes",
    " of events in the `<Events<T>>` list.",
    "",
    " The first key serves no purpose. This field is declared as double_map just",
    " for convenience of using `remove_prefix`.",
    "",
    " All topic vectors have deterministic storage locations depending on the topic. This",
    " allows light-clients to leverage the changes trie storage tracking mechanism and",
    " in case of changes fetch the list of events of interest.",
    "",
    " The value has the type `(T::BlockNumber, EventIndex)` because if we used only just",
    " the `EventIndex` then in case if the topic has the same contents on the next block",
    " no notification will be triggered thus the event might be lost."
  ]
}
```

The type generation doesn't fail but the resulting type is
`eventTopics: AugmentedQueryDoubleMap<ApiType, (key1: () | [], key2: Hash | string | Uint8Array) => Observable<Vec<ITuple<[BlockNumber, EventIndex]>>>, () | []>;`

Which isn't valid typescript

The solution was to convert `()` to `null` which produces
`eventTopics: AugmentedQueryDoubleMap<ApiType, (key1: null, key2: Hash | string | Uint8Array) => Observable<Vec<ITuple<[BlockNumber, EventIndex]>>>, null>;`

### DIFF
--- a/packages/api/src/augment/query.ts
+++ b/packages/api/src/augment/query.ts
@@ -319,7 +319,7 @@ declare module '@polkadot/api/types/storage' {
        * Records information about the maximum slash of a stash within a slashing span,
        * as well as how much reward has been paid out.
        **/
-      spanSlash: AugmentedQuery<ApiType, (arg: ITuple<[AccountId, SpanIndex]>) => Observable<SpanRecord>> & QueryableStorageEntry<ApiType>;
+      spanSlash: AugmentedQuery<ApiType, (arg: ITuple<[AccountId, SpanIndex]> | [AccountId | string | Uint8Array, SpanIndex | AnyNumber | Uint8Array]) => Observable<SpanRecord>> & QueryableStorageEntry<ApiType>;
       /**
        * The earliest era for which we have a pending, unapplied slash.
        **/
@@ -361,7 +361,7 @@ declare module '@polkadot/api/types/storage' {
        * The first key is always `DEDUP_KEY_PREFIX` to have all the data in the same branch of
        * the trie. Having all data in the same branch should prevent slowing down other queries.
        **/
-      keyOwner: AugmentedQueryDoubleMap<ApiType, (key1: Bytes | string | Uint8Array, key2: ITuple<[KeyTypeId, Bytes]>) => Observable<Option<ValidatorId>>, Bytes | string | Uint8Array> & QueryableStorageEntry<ApiType>;
+      keyOwner: AugmentedQueryDoubleMap<ApiType, (key1: Bytes | string | Uint8Array, key2: ITuple<[KeyTypeId, Bytes]> | [KeyTypeId | AnyNumber | Uint8Array, Bytes | string | Uint8Array]) => Observable<Option<ValidatorId>>, Bytes | string | Uint8Array> & QueryableStorageEntry<ApiType>;
     };
     democracy: {
       [index: string]: QueryableStorageEntry<ApiType>;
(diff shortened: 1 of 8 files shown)


monitz87:
Fair enough. I still have to fix some stuff that isn't building properly so I'll refactor while I'm at it

packages/api/src/augment/query.ts
```diff
@@ -779,7 +779,7 @@ declare module '@polkadot/api/types/storage' {
        * First account is the account to be recovered, and the second account
        * is the user trying to recover the account.
        **/
-      activeRecoveries: AugmentedQueryDoubleMap<ApiType, (key1: AccountId | string | Uint8Array, key2: AccountId | string | Uint8Array) => Observable<Option<ActiveRecovery>>, AccountId | string | Uint8Array> & QueryableStorageEntry<ApiType>;
+      activeRecoveries: AugmentedQueryDoubleMap<ApiType, (key1: AccountId | string | Uint8Array, key2: AccountId | string | Uint8Array) => Observable<Option<ActiveRecovery>>> & QueryableStorageEntry<ApiType>;
```
jacogr: Good catch, I completely missed this one :(

PR #1938, opened 2020-02-25
Make Strict Mode Strict Again

monitz87:
This PR:
- Fixes the RegExp in `formatType` to cover another edge case
- Makes the typegen's strict mode **REALLY** strict by eliminating all index signatures

(For some reason the commit history got weird and it looks like I'm trying to merge commits that have already been merged)

# Explanation

## `formatType` RegExp:
The last part of the regexp (`^\[([^;]+|)?[^;]+\]`) was intended to match strings like `[a | b, c | d]` while ignoring `[a;b]`. It didn't take into account cases like `[a | b, c | { d?: any; e?: any }]` (which get produced by structs and should be matched by the RegExp). This solution is designed to properly cover those cases while still not matching `[a;b]`.

## Actual Strict mode:
The only thing I wasn't sure whether to eliminate on strict mode was the `(extrinsic: Uint8Array | string): SubmittableExtrinsic<ApiType>;` signature from `SubmittableExtrinsics`, since it felt a bit like removing functionality. Still, it's kind of generic and I was tempted to remove it as well. Any thoughts?


jacogr:
Thank you. Will take a peek in a couple of hours. 

—strict should actually be the default for “metadata from chain”, so it matches exactly. 

Will take a peek at the extrinsic issue, the submittables are always a bit “special”

packages/api/src/augment/tx.ts
```diff
@@ -1622,4 +1623,9 @@ declare module '@polkadot/api/types/submittable' {
       vestOther: AugmentedSubmittable<(target: LookupSource | Address | AccountId | AccountIndex | string | Uint8Array) => SubmittableExtrinsic<ApiType>>;
     };
   }
+
+  export interface SubmittableExtrinsics<ApiType extends ApiTypes> extends AugmentedSubmittables<ApiType> {
+    (extrinsic: Uint8Array | string): SubmittableExtrinsic<ApiType>;
```
jacogr: So even in strict mode, this line is useful - it allows you to create an extrinsic directly from source. (Ad it actually does get used by some teams, it is a relatively late addition)

So would probably move this back. 

As for the next line, this we can augment in here on top of `SubmittableExtrinsics`

Or the other way around - always populate this interface, but in non-strict mode add the line below as well.

packages/typegen/src/generate/query.ts
```diff
@@ -110,18 +110,28 @@ function generateForMeta (registry: Registry, meta: Metadata, dest: string, extr
...
+        types: ['ApiTypes']
       }
     ]);
     const interfaceStart = [
       "declare module '@polkadot/api/types/storage' {",
       indent(2)('export interface AugmentedQueries<ApiType> {\n')
     ].join('\n');
-    const interfaceEnd = `\n${indent(2)('}')}\n}`;
+    const interfaceEnd = `\n${indent(2)('}')}\n\n`;
+    const queryableStorageInterface = indent(2)('export interface QueryableStorage<ApiType extends ApiTypes> extends AugmentedQueries<ApiType> {')
```
jacogr: Weird, not seeing this appear in the generated. So probably will be fixed on CI with a generate run. No harm.

packages/api/src/augment/tx.ts
jacogr: Ok, was wrong about this - exactly as I explained - so all 100%.

Issue #1649, opened 2019-12-07
Tuple construction error


c410-f3r:
The constructor of a custom `Struct` that is part of a custom `Tuple` is being called with a passed instance of itself for every new `Tuple` object.  `super` should re-use the same passed instance instead of trying to re-create yet another instance.

```ts
import { Struct, Tuple, TypeRegistry, u128 } from "@polkadot/types";
import { Registry } from "@polkadot/types/types";

class Foo extends Struct {
  constructor(registry: Registry, value: number) {
    // Error! value is either `number` or `Foo` at run-time
    super(registry, { fooField: "u128" }, { fooField: new u128(registry, value) });
  }
}

class Bar extends Tuple {
  constructor(registry: Registry, value: Foo) {
    // For some reason, `super` calls `Foo` constructor and pass the same `Foo` as an argument
    super(registry, [ Foo ], [ value ]);
  }
}

const registry = new TypeRegistry();
registry.register("Foo", Foo);
registry.register("Bar", Bar);
let bar = new Bar(registry, new Foo(registry, 123));
```

axelchalon:
Just went through all the base codec types and noticed that the behaviour when `param instanceof Constructor` is inconsistent for `Tuple` and `Enum`. will make a PR to fix.
```js
// uses param directly if param instanceof Constructor
new Option(registry, Constructor, param)
new Vec(registry, Constructor, [param]) // even though we also have VecAny
new BTreeSet(registry, Constructor, new Set([param]))
new BTreeMap(registry, keyConstructor, valueConstructor, new Map([keyParam, valueParam]))
new VecFixed(registry, Constructor, 1, [param])
new Struct(registry, {a: Constructor}, {a: param})
// always instantiates Constructor with param
new Tuple(registry, [Constructor], [param])
new Enum(registry, {a: Constructor}, {a: param})
new Enum(registry, {a: Constructor}, param, 0)
new Result(registry, okConstructor, errConstructor, {Ok: param})
```

PR #1954, opened 2020-02-26
Make Enum/Tuple constructor use value directly if instance

axelchalon:
- Make the Tuple, Enum (& Result, which extends Enum) constructor use `value` directly if it's an instanceof the correct type, i.e. align their behaviour with that of all the other base codecs (cf https://github.com/polkadot-js/api/issues/1649#issuecomment-591515933), close https://github.com/polkadot-js/api/issues/1649
- Add tests for this behaviour for all base types
- Fix BTreeSet equality which was broken (compareSet converted sets to array with .entries() instead of .values(), so each value was compared against [value, value])


c410-f3r:
Thanks!

Issue #845, opened 2019-04-16
Ensure correct metadata for parsing (archive nodes, historical queries)


afenrir:
There are several blocks throughout the chain that produce the following error when you try to fetch them:

```
2019-04-16 20:34:56        RPC-CORE: getBlock (hash: Hash): SignedBlock:: Source is too large
ExtError: getBlock (hash: Hash): SignedBlock:: Source is too large
    at call (/Users/self/dev/polkadot/statd/node_modules/@polkadot/rpc-core/index.js:134:15)
    at processTicksAndRejections (internal/process/next_tick.js:81:5)
```

This code will recreate the problem:

```javascript
    const hash = new Hash(
        '0x6f6f9bba0eed8e3ae9446c37eee763f93118b52a315a7b46090453ba6288da1f'
    );
    api.rpc.chain.getBlock(hash)

```

Also, you will notice there is no `then` or `catch` on the `getBlock` promise. That is because the above error does not actually land in the catch block. It gets handled and logged by the library itself, which makes it more difficult for me to handle the error in my code.

jacogr:
It must be the runtime upgrade blocks. (Sudo + setCode) Cannot think off hand where the error originates, but def. an issue and it should just work. 

(Strangely I have pulled a number of the upgrade blocks before without issues on multiple chains, so it feels like something new - or maybe I just got lucky earlier on and it just hit some limit lately)

afenrir:
To add a little more information, the most recent instance of this happening is just prior to block 580,000. However, prior to that, there are perhaps another 100 or so errors of this type. If you need a larger sample of block hashes that can reproduce this issue I am happy to grab those for you.

jacogr:
I am currently sync-ing with archive mode, so will be looking at this one as well as https://github.com/polkadot-js/api/issues/846 ASAP. (They have been hanging for way too long)

jacogr:
Ok, this one is problematic... basically the staking interfaces has changed radically. It seems to fail on this extrinsic -

```
 meta:
       FunctionMetadata [Map] {
         'id' => <BN: 2>,
         'name' => [String: 'unbond'],
         'arguments' => Vector [
           [FunctionArgumentMetadata],
           _Type: [Function: FunctionArgumentMetadata] ],
         'documentation' => Vector [
           [String: ' Schedule a portion of the stash to be unlocked ready for transfer out after the bond'],
           [String: ' period ends. If this leaves an amount actively bonded less than'],
[...]

PR #1979, opened 2020-02-29
Add .range([from, to?], ...args) on storage queries

### LINKED ISSUE #1980: Add `.between(fromBlock, toBlock)` to `StorageEntryBase`
As discussed in riot chat, this function would work like `.at(blockHash)` but return results in a range of blocks

jacogr:
- Closes #1980 

The only issue here is that it still will create issues with historic queries when the runtime has been upgraded. In these cases the indexes may have swapped and types can occur for older blocks which may be mismatching with current. (So it certainly is not fail-safe when types are adjusted, e.g. https://github.com/polkadot-js/api/issues/845)


jacogr:
Closes #1980

PR #1997, opened 2020-03-03
Add Github Action for Matrix release bot

Stefie:
This action will publish all pre-released, edited and published releases to a dedicated release notes Matrix room using https://github.com/marketplace/actions/matrix-message

### DIFF
--- a/.github/workflows/release-bot.yml
+++ b/.github/workflows/release-bot.yml
@@ -1,0 +1,18 @@
+name: Pushes release updates to a pre-defined Matrix room
+on:
+  release:
+    types:
+      - edited
+      - prereleased
+      - published
+jobs:
+  ping_matrix:
+    runs-on: ubuntu-latest
+    steps:
+      - name: send message
+        uses: s3krit/matrix-message-action@v0.0.2
+        with:
+          room_id: ${{ secrets.MATRIX_ROOM_ID }}
+          access_token: ${{ secrets.MATRIX_ACCESS_TOKEN }}
+          message: "**${{github.event.repository.full_name}}:** A release has been ${{github.event.action}}<br/>Release version [${{github.event.release.tag_name}}](${{github.event.release.html_url}})<br/><br/>***Description:***<br/>${{github.event.release.body}}<br/>"
+          server: "matrix.parity.io"


jacogr:
All ops-related questions -

0. What is this for? Why is this not comprehensive in coverage, this seems random. It is an honest question, since it is tough to keep everything in mind across everything, the only sanity that I have is consistency.

1. All devops is currently done by me on this org. So how were these tokens "already added" to settings?

2. This is a very-optional thing. Matrix can be like a yo-yo at the best of times. So as it stands the package is not really something I would add, honestly don't care about matrix downtimes adding (yet another) failure status and something to look at.

.github/workflows/release-bot.yml
```diff
@@ -0,0 +1,18 @@
+name: Pushes release updates to a pre-defined Matrix room
+on:
+  release:
+    types:
+      - edited
+      - prereleased
+      - published
```
jacogr: Only published status, if this is something that we want. (Not convinced)

Stefie: Ok, no problem, I can change it to that.

Stefie:
> 0. What is this for? Why is this not comprehensive in coverage, this seems random. It is an honest question, since it is tough to keep everything in mind across everything, the only sanity that I have is consistency. (This is an extra "thing" on "some")

It is for a Matrix/ Riot channel that people at Parity as well as external users can subscribe to. The plan is to automatically push the release notes of  all core projects relevant for the Polkadot/ Substrate ecosystem new versions to this channel. That's why I only added the Action to the lower level projects and not to the UI projects like https://github.com/polkadot-js/apps

> 1. All devops is currently done by me on this org. So how were these tokens "already added" to settings?

Copy & Paste error, sorry. I had to open this PR in 10 repos and forgot to delete that line. He only added them in the @paritytech repos

> 2. This is a very-optional thing. Matrix can be like a yo-yo at the best of times. So as it stands the package is not really something I would add, honestly don't care about matrix downtimes adding (yet another) failure status and something to look at.

Do you think that a Matrix failure could cause a failure status? @s3krit can you say something about it?
Unfortunately the Matrix  RSS bot that needs no configuration at all can't handle the GitHub atom feeds properly, that's why the devops team came up with the Matrix Bot in combination with Github actions

jacogr:
As discussed, sadly cannot merge this. Open for revisiting this in the future if my concerns are addressed. As it stands it is extra inconsistencies.

s3krit:
>Matrix can be like a yo-yo at the best of times.

This isn't really true. Outside of planned maintenance events, the matrix server doesn't really go down. Below is a graph of the last 21 days. The one event on that graph was actually an issue with the connection to the prometheus server. Don't confuse downtime with messages in encrypted channels being slow to send to users (which is actually a client-side issue).

downtime

It's also worth noting that this workflow would only run when a release is published - the chances of the matrix server being down, and someone publishing a release at the exact same time is fairly negligible, I believe.

jacogr:
We can just agree to disagree - I am not adding anything that depends on yet another service (not critical) and throws up a failure state. 

I have enough to look at.

PR #2053, opened 2020-03-15
Use /usr/bin/env bash instead of /bin/bash

sorpaas:
POSIX only specifies `/usr/bin/env` but not `/bin/bash`. Using the former increases compatibility with OSes that do not ship with the later, for example, NixOS.


jacogr:
This is a bit of a “special hack dev script” so compat was never much of a concern. Food to have this either way. Thank you.

PR #2083, opened 2020-03-24
ValidateUnsigned extension (no payload impact)

jacogr:
As per https://github.com/paritytech/substrate/pull/5006
Also included in https://github.com/paritytech/substrate/pull/4517

(It is a noop from an extrinsic perspective, only effect here is that the API will not warn when it gets found)


jacogr:
Closing, unused.

PR #2087, opened 2020-03-24
Jd refine query multi types

monitz87:
This PR improves types on `.multi` calls so that arguments and return types match the specific storage query function


monitz87:
Seems like there are some cases I didn't take into consideration. Ignore this until further notice

Issue #2112, opened 2020-03-28
Parsing Issue with Custom Type in `Constants` view


shawntabrizi:
There seems to be some parsing error when using a custom struct in the `constants` tab of the Polkadot JS Apps.

There might be a bug in Polkadot JS Apps, but in general here is the steps to get what I think you want.

We have a custom type:
```
#[derive(Encode, Decode)]
struct Rule {
    name: Vec<u8>,
    version: u32,
    ops: Vec<u32>,
}
```

We place a `const` in the `decl_module` so it exposes it in the metadata:
```
         /// Rules for the PoE
        const rule: Rule = Rule {
            name: b"rule 1".to_vec(),
            version: 1,
            ops: vec![2]
        };
```

Then we start our chain. In Polkadot JS Apps, I register a new custom type:
```
{
  "Rule": {
    "name": "Vec<u8>",
    "version": "u32",
    "ops": "Vec<u32>"
  }
}
```

Then I go look at the constants in the pallet:

<img width="1421" alt="image" src="https://user-images.githubusercontent.com/1860335/77825708-6ae20e00-710b-11ea-893b-1c2da81d3b0e.png">

It seems it does not correctly detect the length prefix. When i parse the bytes on my own, it works:

<img width="1421" alt="image" src="https://user-images.githubusercontent.com/1860335/77825726-8816dc80-710b-11ea-89d0-0c294b580dc7.png">

shawntabrizi:
```
          constants: [
            {
              name: rule,
              type: Rule,
              value: 0x1872756c652031010000000402000000,
              documentation: [
                 Rules for the PoE
              ]
            },
 ```

Let me know if you need anything else.

jacogr:
Should be in apps in around 20 mins - if that doesn't solve it, can re-open, but should be correct now. (Should have been fixed originally instead of hacked around for bytes-only)

shawntabrizi:
can confirm it is fixed

<img width="1271" alt="image" src="https://user-images.githubusercontent.com/1860335/77833130-4736bc00-713b-11ea-9387-77c7925e6509.png">

thanks!

PR #2113, opened 2020-03-28
Handle double-encoded Bytes in fallback/constants

jacogr:
Closes https://github.com/polkadot-js/api/issues/2112

The issue is that the metadata has a prefix (since the data stream for constants are `Bytes`), but we want the data after the prefix (which could once again contain a prefix, e.g. for `Vec<u8>`), passed into createType.

This affects both constants and actually storage fallbacks - in the latter this has not bitten (since most/all are default, however in non-default cases the fallback value would be incorrectly applied).

### DIFF
--- a/packages/metadata/src/Decorated/consts/fromMetadata/index.ts
+++ b/packages/metadata/src/Decorated/consts/fromMetadata/index.ts
@@ -5,12 +5,10 @@
 import { Registry } from '@polkadot/types/types';
 import { Constants, ConstantCodec, ModuleConstants } from '../../types';
 
-import { createTypeUnsafe } from '@polkadot/types/create';
-import { stringCamelCase } from '@polkadot/util';
+import { createTypeUnsafe } from '@polkadot/types';
+import { hexToU8a, stringCamelCase } from '@polkadot/util';
 
 import Metadata from '../../../Metadata';
-
-const AS_STRIPPED = ['Bytes'];
 
 /** @internal */
 export default function fromMetadata (registry: Registry, metadata: Metadata): Constants {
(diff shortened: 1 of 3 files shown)


packages/metadata/src/Decorated/consts/fromMetadata/index.ts
```diff
@@ -23,19 +21,10 @@ export default function fromMetadata (registry: Registry, metadata: Metadata): C
...
-        AS_STRIPPED.includes(type)
-          ? meta.value.toU8a(true)
-          : meta.value
-      ]);
-
-      // This is not a perfect idea, however as it stands with number-only constants on the metadata
-      // does not have any effect. However, this could become problematic in cases where items are
-      // exposed that contain their own metadata. As of now, the compatibility with current, e.g.
-      // storage is the driving factor, one consistent way of handling interfaces
+      const codec: ConstantCodec = createTypeUnsafe(registry, type, [hexToU8a(meta.value.toHex())]);
```
jacogr: Remove the workaround we had, now -

- go to hex (which doesn't have the prefix
- convert that to the raw Uint8Array (which may have an _embedded_ prefix)

We cannot use Raw here, since that would assume no prefix added, so we need to go to a pure Uint8Array

packages/rpc-core/src/index.ts
```diff
@@ -370,7 +370,13 @@ export default class Rpc implements RpcInterface {
       );
     }
 
-    return createTypeUnsafe(this.registry, type, [isEmpty ? meta.fallback : input], true);
+    return createTypeUnsafe(this.registry, type, [
+      isEmpty
+        ? meta.fallback
+          ? hexToU8a(meta.fallback.toHex())
```
jacogr: Apply as per constants

PR #2125, opened 2020-04-01
Base centrifuge types

jacogr:
cc @philipstanislaus - need your help here to fill in the blanks with regards to versions. (If you don't need historic compat and the starting version is ok, this is fine)


philipstanislaus:
Thanks @jacogr! Starting version is fine, we have just reset our testnets.

PR #2204, opened 2020-04-24
remove unnecessary parentheses and duplicate types

monitz87:
## Parentheses

Plain enum similar types were being wrapped in parentheses, causing the type generator to misjudge their type when part of a tuple.

Example:

```
"MyEnum": {
  _enum: ['A', 'B', 'C']
}
```

Was being converted to `MyEnum | ('A' | 'B' | 'C' ) | ...`, which was causing `setImports` to mistake it for a nested tuple, which caused an error later on

 Since `A | (B | C)` is equal  to `A | B | C`, my solution simply removes the parentheses.

##  Duplicate types

When overriding a type (such as `Weight`), the typegen was creating entries for both occurrences in the augmented type registry. This PR fixes that by storing the types that have already been read and only write down the ones that haven't

### DIFF
--- a/packages/typegen/src/generate/interfaceRegistry.ts
+++ b/packages/typegen/src/generate/interfaceRegistry.ts
@@ -19,7 +19,8 @@ export function generateInterfaceTypes (importDefinitions: { [importPath: string
     Object.entries(importDefinitions).reduce((acc, def) => Object.assign(acc, def), {} as object);
 
     const imports = createImports(importDefinitions);
-    const definitions = imports.definitions;
+    const definitions = { ...imports.definitions, polymesh: { rpc: {}, types: { Weight: 'u32' } } } as object;
+
     const primitives = Object
       .keys(primitiveClasses)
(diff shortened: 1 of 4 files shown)


packages/typegen/src/generate/interfaceRegistry.ts
```diff
@@ -19,7 +19,8 @@ export function generateInterfaceTypes (importDefinitions: { [importPath: string
     Object.entries(importDefinitions).reduce((acc, def) => Object.assign(acc, def), {} as object);
 
     const imports = createImports(importDefinitions);
-    const definitions = imports.definitions;
+    const definitions = { ...imports.definitions, polymesh: { rpc: {}, types: { Weight: 'u32' } } } as object;
```
jacogr, line 22: The `polymesh` seems to have snuck in here.

monitz87: Yeah, was using it to test the duplicate type thing

PR #2209, opened 2020-04-26
allow override build-in rpc

xlc:
the default RPC defs may not align with custom chain, user should be able to override it.


xlc:
actually, reading code again, I think it was correct? the filter is avoid duplicated keys

PR #2222, opened 2020-04-30
Fix extrinsic ECDSA signatures

akru:
**Motivation**

The extrinsic signing method uses a raw public key for the `signer` field but it doesn't work for ECDSA.
This PR adds address derivation for ECDSA signers into sign method.


akru:
Oh, yes, thank you @jacogr! Fixed!

jacogr:
As always, thanks a lot! Great effort, all round.

PR #2230, opened 2020-05-04
Allow keyPrefix to accept an additional argument for double maps

### LINKED ISSUE #2213: get key prefix for a double map with first key argument
### Goal:

Update `keyPrefix` for double map storage to take one argument as the first key and return the key prefix.

### Background

Currently polkadot.js provides two methods to work with the storage key value:

`keyPrefix()` that return the storage prefix.
`key(args)` that return the whole storage key.

None of it able to provide key prefix for a double map with a given first key item.

### Actions:

Modify `keyPrefix`, change signature to `(key1?: CodecArg): string`

https://github.com/polkadot-js/api/blob/e77050b9f59a0b772d1d0b8c5c1da488a7c36810/packages/api/src/base/Decorate.ts#L347

Check if it is double map and have key1, if false, keep existing behavior. else return prefix with key1 happened.

How the key prefix for double map with a key can be found here

https://github.com/polkadot-js/api/blob/e77050b9f59a0b772d1d0b8c5c1da488a7c36810/packages/api/src/base/Decorate.ts#L466-L472

Update interface

https://github.com/polkadot-js/api/blob/e77050b9f59a0b772d1d0b8c5c1da488a7c36810/packages/api/src/types/storage.ts#L41

Provide appropriate unit tests.

---

@jacogr please tag this if you are happy about this.

Tags:
`bounty-awaiting-approval`
`bounty-S`

KiChjang:
Fixes #2213.

### DIFF
--- a/packages/api/src/base/Decorate.ts
+++ b/packages/api/src/base/Decorate.ts
@@ -15,14 +15,13 @@ import { map, switchMap, take, tap, toArray } from 'rxjs/operators';
 import decorateDerive, { ExactDerive } from '@polkadot/api-derive';
 import { memo } from '@polkadot/api-derive/util';
 import DecoratedMeta from '@polkadot/metadata/Decorated';
-import getHasher from '@polkadot/metadata/Decorated/storage/fromMetadata/getHasher';
 import RpcCore from '@polkadot/rpc-core';
 import { WsProvider } from '@polkadot/rpc-provider';
 import { Metadata, Null, Option, Raw, Text, TypeRegistry, u64 } from '@polkadot/types';
 import Linkage, { LinkageResult } from '@polkadot/types/codec/Linkage';
 import { DEFAULT_VERSION as EXTRINSIC_DEFAULT_VERSION } from '@polkadot/types/extrinsic/constants';
 import StorageKey, { StorageEntry, unwrapStorageType } from '@polkadot/types/primitive/StorageKey';
-import { assert, compactStripLength, isNull, isUndefined, logger, u8aConcat, u8aToHex } from '@polkadot/util';
+import { assert, compactStripLength, logger, u8aToHex } from '@polkadot/util';
 
 import { createSubmittable } from '../submittable';
 import augmentObject from '../util/augmentObject';
@@ -345,14 +344,7 @@ export default abstract class Decorate<ApiType extends ApiTypes> extends Events 
       u8aToHex(compactStripLength(creator(creator.meta.type.isDoubleMap ? [arg1, arg2] : arg1))[1]);
 
     decorated.keyPrefix = (key1?: Arg): string =>
-      creator.meta.type.isDoubleMap && !isUndefined(key1) && !isNull(key1)
-        ? this.createType('Raw', u8aConcat(
-          creator.keyPrefix,
-          getHasher(creator.meta.type.asDoubleMap.hasher)(
-            this.createType(creator.meta.type.asDoubleMap.key1.toString() as 'Raw', key1).toU8a()
-          )
-        )).toHex()
-        : u8aToHex(creator.keyPrefix);
+      u8aToHex(creator.keyPrefix(key1));
 
     decorated.range = decorateMethod((range: [Hash, Hash?], arg1?: Arg, arg2?: Arg): Observable<[Hash, Codec][]> =>
       this._decorateStorageRange(decorated, [arg1, arg2], range));
@@ -462,7 +454,7 @@ export default abstract class Decorate<ApiType extends ApiTypes> extends Events 
           .subscribeStorage<[[Codec, Linkage<Codec>]]>([[creator, ...args]])
           .pipe(map(([data]): [Codec, Linkage<Codec>] => data))
         : this._rpcCore.state
-          .subscribeStorage<[LinkageResult]>([creator.iterKey])
+          .subscribeStorage<[LinkageResult]>([creator.keyPrefix()])
           .pipe(switchMap(([key]): Observable<LinkageResult> => getNext(head = key)))
     );
   }
@@ -470,14 +462,7 @@ export default abstract class Decorate<ApiType extends ApiTypes> extends Events 
   private _retrieveMapKeys ({ iterKey, meta }: StorageEntry, arg?: Arg): Observable<StorageKey[]> {
     assert(iterKey && (meta.type.isMap || meta.type.isDoubleMap), 'keys can only be retrieved on maps, linked maps and double maps');
 
-    const headKey = this.createType('Raw', u8aConcat(
-      iterKey,
-      meta.type.isDoubleMap && !isUndefined(arg) && !isNull(arg)
-        ? getHasher(meta.type.asDoubleMap.hasher)(
-          this.createType(meta.type.asDoubleMap.key1.toString() as 'Raw', arg).toU8a()
-        )
-        : new Uint8Array([])
-    )).toHex();
+    const headKey = iterKey(arg).toHex();
     const startSubject = new BehaviorSubject<string>(headKey);
 
     return this._rpcCore.state.getKeysPaged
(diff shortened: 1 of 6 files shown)


KiChjang:
Unsure where unit tests are located, originally thought `api.spec.ts` would have them, but they do not look like tests.

jacogr:
Thanks a lot for this! Will look in detail a bit later today.

packages/api/src/base/Decorate.ts
```diff
@@ -344,8 +344,15 @@ export default abstract class Decorate<ApiType extends ApiTypes> extends Events
     decorated.key = (arg1?: Arg, arg2?: Arg): string =>
       u8aToHex(compactStripLength(creator(creator.meta.type.isDoubleMap ? [arg1, arg2] : arg1))[1]);
 
-    decorated.keyPrefix = (): string =>
-      u8aToHex(creator.keyPrefix);
+    decorated.keyPrefix = (key1?: Arg): string =>
```
jacogr, line 347: The problem here is that we now have duplication of code, as per https://github.com/polkadot-js/api/blob/master/packages/api/src/base/Decorate.ts#L466-L473

So from a maintenance pov we need to ensure that we maintain exactly the same in 2 places. This is problematic.

KiChjang: Can extract this into a function call.

KiChjang:
@jacogr It looks to me that the `iterKey` field on `StorageEntry` is optional, but `keyPrefix` isn't. What exactly is the difference between the two? And should what you said about `iterKey` instead be implemented on `keyPrefix`?

jacogr:
Yes, `iterKey` is optional since it only appears on maps and doublemaps, so in the case of a normal state entry it does not exist. So it really is the key-used-for-iteration.

So what we should do -

- change `iterKey?: Uint8array` to `iterKey?: (arg?: any) => Uint8Array`
- in `kyePrefix` it just becomes `(arg?: any) => iterKey(arg)` (with the conversions to hex as per current)

KiChjang:
@jacogr Ok, I just pushed a new commit up to move the key generation logic into the lower layers, please take a look.

packages/metadata/src/Decorated/storage/fromMetadata/createFunction.ts
```diff
@@ -157,7 +157,7 @@ function extendHeadMeta (registry: Registry, { meta: { documentation, name, type
     type: registry.createType('StorageEntryTypeLatest', registry.createType('PlainTypeLatest', type.isMap ? type.asMap.key : type.asDoubleMap.key1), 0)
   });
 
-  return registry.createType('StorageKey', iterFn, { method, section });
+  return (arg?: any) => registry.createType('StorageKey', iterFn(arg), { method, section });
```
xlc, line 160: One small issue I see here is  when no arguments,`StorageKey` will need to be recreated. Previously it was cached.

```
const prefixKey = registry.createType('StorageKey', iterFn(), { method, section });
return (arg?: any) => !isUndefined(arg) && !isNull(arg)
  ? registry.createType('StorageKey', iterFn(arg), { method, section })
  : prefixKey;
```

packages/metadata/src/Decorated/storage/fromMetadata/createFunction.spec.ts
```diff
@@ -104,6 +104,15 @@ describe('createFunction', (): void => {
         (): Uint8Array => storageFn(['5DXUeE5N5LtkW97F2PzqYPyqNkxqSWESdGSPTX6AvkUAhwKP'])
       ).toThrow(/requires two arguments/);
     });
+
+    it.only('accepts an optional parameter for key construction', (): void => {
```
xlc, line 108:
```suggestion
    it('accepts an optional parameter for key construction', (): void => {
```

packages/metadata/src/Decorated/storage/fromMetadata/createFunction.ts
```diff
@@ -157,7 +157,11 @@ function extendHeadMeta (registry: Registry, { meta: { documentation, name, type
     type: registry.createType('StorageEntryTypeLatest', registry.createType('PlainTypeLatest', type.isMap ? type.asMap.key : type.asDoubleMap.key1), 0)
   });
 
-  return registry.createType('StorageKey', iterFn, { method, section });
+  const prefixKey = registry.createType('StorageKey', iterFn, { method, section });
+
+  return (arg?: any): StorageKey => !isUndefined(arg) && !isNull(arg)
+    ? registry.createType('StorageKey', iterFn(arg), { method, section })
+    : prefixKey;
```
jacogr, line 164:
```suggestion
  return (arg?: any): StorageKey =>
    !isUndefined(arg) && !isNull(arg)
      ? registry.createType('StorageKey', iterFn(arg), { method, section })
      : prefixKey;
```

Just slight OCD, i.e. my eyes follow it better together.

packages/metadata/src/Decorated/storage/fromMetadata/createFunction.spec.ts
```diff
@@ -104,6 +104,15 @@ describe('createFunction', (): void => {
         (): Uint8Array => storageFn(['5DXUeE5N5LtkW97F2PzqYPyqNkxqSWESdGSPTX6AvkUAhwKP'])
       ).toThrow(/requires two arguments/);
     });
+
+    it('accepts an optional parameter for key construction', (): void => {
+      expect(storageFn.iterKey).not.toBeUndefined();
+
+      const result = storageFn.iterKey!('5DXUeE5N5LtkW97F2PzqYPyqNkxqSWESdGSPTX6AvkUAhwKP');
```
jacogr, line 111: No `!` please. (Trying to rid the codebase of them, sadly we have a handful - even in tests, tests are the actual reason we have not been able to switch on the rule, i.e. https://github.com/polkadot-js/api/blob/master/.eslintrc.js#L16)

packages/api/src/base/Decorate.ts
```diff
@@ -455,22 +454,15 @@ export default abstract class Decorate<ApiType extends ApiTypes> extends Events
           .subscribeStorage<[[Codec, Linkage<Codec>]]>([[creator, ...args]])
           .pipe(map(([data]): [Codec, Linkage<Codec>] => data))
         : this._rpcCore.state
-          .subscribeStorage<[LinkageResult]>([creator.iterKey])
+          .subscribeStorage<[LinkageResult]>([creator.keyPrefix()])
```
jacogr, line 457:
```suggestion
          .subscribeStorage<[LinkageResult]>([creator.iterKey()])
```

The reason is the `iterKey` is meant to be the iteration key for maps explicitly. (Whereas `keyPrefix` does also exist for non-map types)

packages/metadata/src/Decorated/storage/fromMetadata/createFunction.ts
```diff
@@ -219,7 +227,7 @@ export default function createFunction (registry: Registry, itemFn: CreateItemFn
     extendDoubleMap(registry, itemFn, storageFn);
   }
 
-  storageFn.keyPrefix = storageFn.iterKey || compactStripLength(storageFn())[1];
+  storageFn.keyPrefix = (arg?: any): Uint8Array => (storageFn.iterKey && storageFn.iterKey(arg)) || compactStripLength(storageFn())[1];
```
jacogr, line 230:
```suggestion
  storageFn.keyPrefix = (arg?: any): Uint8Array =>
    (storageFn.iterKey && storageFn.iterKey(arg)) || compactStripLength(storageFn())[1];
```

Small readability change. (This and the previous one is hard to follow, just due to what we are dealing with here)

packages/metadata/src/Decorated/storage/fromMetadata/createFunction.spec.ts
KiChjang: I'm not sure how I can do so, because the TypeScript compiler is not smart enough to know that I've already checked for it to be non-null or non-undefined in a statement above.

packages/metadata/src/Decorated/storage/fromMetadata/createFunction.spec.ts
```diff
@@ -4,7 +4,7 @@
 
 import { Text, TypeRegistry } from '@polkadot/types';
 import { StorageEntry } from '@polkadot/types/primitive/StorageKey';
-import { stringToU8a, u8aConcat, u8aToHex } from '@polkadot/util';
+import { stringToU8a, u8aConcat, u8aToHex, assert } from '@polkadot/util';
```
jacogr, line 7:
```suggestion
import { assert, stringToU8a, u8aConcat, u8aToHex } from '@polkadot/util';
```

Issue #2012, opened 2020-03-09
Make options for `.signAsync` optional


xlc:
### Goal:

Update `signAsync` to make the second argument optional, and default to `{}`

### Background:

Currently this code will crash:

`const signed = await api.system.remark().signAsync(pair, options);` when `options` is `undefined`, it gives this error

```
(node:78914) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'nonce' of undefined
    at Submittable.value (/Users/xiliangchen/projects/playground/node_modules/@polkadot/api/submittable/createClass.js:59:42)
    at Submittable.value (/Users/xiliangchen/projects/playground/node_modules/@polkadot/api/submittable/createClass.js:129:99)
    at /Users/xiliangchen/projects/playground/node_modules/@polkadot/api/submittable/createClass.js:251:110
    at /Users/xiliangchen/projects/playground/node_modules/@polkadot/api/promise/Api.js:78:14
    at Submittable.signAsync (/Users/xiliangchen/projects/playground/node_modules/@polkadot/api/submittable/createClass.js:251:170)
    at /Users/xiliangchen/projects/playground/index.ts:28:33
    at step (/Users/xiliangchen/projects/playground/index.ts:33:23)
    at Object.next (/Users/xiliangchen/projects/playground/index.ts:14:53)
    at /Users/xiliangchen/projects/playground/index.ts:8:71
    at new Promise (<anonymous>)
```

### Actions:

Make second argument optional

https://github.com/polkadot-js/api/blob/e77050b9f59a0b772d1d0b8c5c1da488a7c36810/packages/api/src/submittable/createClass.ts#L72

Update interface

https://github.com/polkadot-js/api/blob/e77050b9f59a0b772d1d0b8c5c1da488a7c36810/packages/api/src/submittable/types.ts#L55

Provide appropriate unit tests.

----

@jacogr please tag this if you are happy about this.

Tags:
`bounty-awaiting-approval`
`bounty-XS`

jacogr:
Small one, it is one of those "don't do that", which crashy, it does indicate how to fix so no immediately critical..

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 50.0 DAI (50.0 USD @ $1.0/DAI)  attached to it as part of the web3foundation fund.__

[...]

PR #2231, opened 2020-05-04
Make options for `.signAsync` optional

developerfred:
#2012

I made the second parameter optional

### DIFF
--- a/packages/api/src/submittable/createClass.ts
+++ b/packages/api/src/submittable/createClass.ts
@@ -60,7 +60,7 @@ export default function createClass <ApiType extends ApiTypes> ({ api, apiType, 
     }
 
     // sign a transaction, returning the this to allow chaining, i.e. .sign(...).send()
-    public sign (account: IKeyringPair, optionsOrNonce: Partial<SignerOptions>): this {
+    public sign (account: IKeyringPair, optionsOrNonce?: Partial<SignerOptions>): this {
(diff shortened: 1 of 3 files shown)


jacogr:
Thanks a lot for this! Will look in detail a bit later today.

packages/api/src/submittable/createClass.ts
```diff
@@ -69,10 +69,10 @@ export default function createClass <ApiType extends ApiTypes> ({ api, apiType,
     // signs a transaction, returning `this` to allow chaining. E.g.: `sign(...).send()`
     //
     // also supports signing through external signers
-    public signAsync (account: AddressOrPair, optionsOrNonce: Partial<SignerOptions>): SubmittableThis<ApiType, this> {
+    public signAsync (account: AddressOrPair, optionsOrNonce?: Partial<SignerOptions>): SubmittableThis<ApiType, this> {
       return decorateMethod(
         (): Observable<this> =>
-          this.#observeSign(account, optionsOrNonce).pipe(mapTo(this))
+          this.#observeSign(account, {}).pipe(mapTo(this))
```
jacogr, line 75: This means we will never pass the options through. This needs to be `optionOrNonce`, as was passed. (The signature would change with default initialization)

developerfred: ok!

Issue #2226, opened 2020-05-01
Enhance WsProvider to support multiple endpoints


xlc:
### Goal:

Enhance `WsProvider` to allow it support multiple endpoints. On reconnect, switch to next endpoint in a round robin way.

### Background:

Currently `WsProvider` only takes one endpoint, and incase this particular node is down, it will lose access to the network. While it is possible to detect disconnect event and create a new api instance, the existing subscription will be lost.

### Actions:

Modify `WsProvider` to change type of `endpoint` to `string | string[]`.

Update `WsProvider#connect` to use next endpoint.

Provide appropriate unit tests.

----

@jacogr please tag this if you are happy about this. The tags are come from W3F bounties doc.

Tags:
`bounty-awaiting-approval`
`bounty-XS`

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 70.0 DAI (70.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/polkadot-js/api/2226/4275).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/polkadot-js/api/2226/4275).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $87,763.33 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 266 years, 6 months from now.
Please review their action plans below:


**1) [hoani](https://gitcoin.co/hoani) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/polkadot-js/api/2226/4275?mutate_worker_action=approve&worker=hoani) | [reject worker](https://gitcoin.co/issue/polkadot-js/api/2226/4275?mutate_worker_action=reject&worker=hoani))_.**

Have used the polkadot API quite a bit, and this seems like a great enhancement. It looks like I would need to start at the rpc-provider and go from there.

[...]

PR #2234, opened 2020-05-05
Multiple Endpoint support for WsProvider

hoani:
This PR allows `WsProvider` to take an array of endpoint strings as an alternative to a single endpoint string.
```js
const provider = new WsProvider(["ws://node-1:9944", "ws://node-2:9944"]);
```
The provider will round-robin through its available endpoints on each reconnect.

## Changes
* Fixed the original tests in `connect.spec.ts`
  * these never actually tested if we were connected to the server properly - `expect(provider.isConnected()).not.toBe(true);` is NOT the opposite of `expect(provider.isConnected()).toBe(false);
  * **[Concern]** Added a delay to allow for the provider to connect to the `mock-socket`'s `Server`
    * Open to alternatives, this solution feels really fragile (though better than the previous tests)
* Allow the user to set `endpoint` arg as a string or array of strings
  * Throws an error if the array is empty
  * Throws an error if any of the endpoints are not `ws://` or `wss://`
  * Private `#endpoint` is now an array of strings
  * Private `#endpointNext` number added to track the next index of `#endpoint[]`
  * Round-robin implemented for `WsProvider.connect()`
* Added missing unit tests to `index.spec.ts` to check for valid/invalid endpoint strings.
* Added unit tests to `index.spec.tx` to cover valid/invalid endpoint arrays
* Added unit tests to `connect.spec.ts` for endpoint array functionality

Closes https://github.com/polkadot-js/api/issues/2226

### DIFF
--- a/packages/rpc-provider/src/ws/Provider.ts
+++ b/packages/rpc-provider/src/ws/Provider.ts
@@ -136,11 +136,12 @@ export default class WsProvider implements WSProviderInterface {
       const WS = await getWSClass();
 
       this.#websocket = new WS(this.#endpoints[this.#endpointNext]);
-      this.#endpointNext = (this.#endpointNext + 1) % this.#endpoints.length;
       this.#websocket.onclose = this.#onSocketClose;
       this.#websocket.onerror = this.#onSocketError;
       this.#websocket.onmessage = this.#onSocketMessage;
       this.#websocket.onopen = this.#onSocketOpen;
+
+      this.#endpointNext = (this.#endpointNext + 1) % this.#endpoints.length;
     } catch (error) {
       l.error(error);
     }
--- a/packages/rpc-provider/src/ws/Provider.ts
+++ b/packages/rpc-provider/src/ws/Provider.ts
@@ -77,7 +77,7 @@ export default class WsProvider implements WSProviderInterface {
 
   #autoConnectMs: number;
 
-  #endpointNext: number;
+  #endpointIndex: number;
 
   #isConnected = false;
 
@@ -103,7 +103,7 @@ export default class WsProvider implements WSProviderInterface {
     this.#eventemitter = new EventEmitter();
     this.#autoConnectMs = autoConnectMs || 0;
     this.#coder = new Coder();
-    this.#endpointNext = 0;
+    this.#endpointIndex = -1;
     this.#endpoints = endpoints;
     this.#websocket = null;
 
@@ -135,13 +135,13 @@ export default class WsProvider implements WSProviderInterface {
     try {
       const WS = await getWSClass();
 
-      this.#websocket = new WS(this.#endpoints[this.#endpointNext]);
+      this.#websocket = new WS(this.#endpoints[this.#endpointIndex]);
       this.#websocket.onclose = this.#onSocketClose;
       this.#websocket.onerror = this.#onSocketError;
       this.#websocket.onmessage = this.#onSocketMessage;
       this.#websocket.onopen = this.#onSocketOpen;
 
-      this.#endpointNext = (this.#endpointNext + 1) % this.#endpoints.length;
+      this.#endpointIndex = (this.#endpointIndex + 1) % this.#endpoints.length;
     } catch (error) {
       l.error(error);
     }
@@ -281,7 +281,7 @@ export default class WsProvider implements WSProviderInterface {
 
   #onSocketClose = (event: CloseEvent): void => {
     if (this.#autoConnectMs > 0) {
-      l.error(`disconnected from ${this.#endpoints} code: '${event.code}' reason: '${event.reason}'`);
+      l.error(`disconnected from ${this.#endpoints[this.#endpointIndex]} code: '${event.code}' reason: '${event.reason}'`);
     }
 
     this.#isConnected = false;
@@ -376,7 +376,7 @@ export default class WsProvider implements WSProviderInterface {
   #onSocketOpen = (): boolean => {
     assert(!isNull(this.#websocket), 'WebSocket cannot be null in onOpen');
 
-    l.debug((): any[] => ['connected to', this.#endpoints]);
+    l.debug((): any[] => ['connected to', this.#endpoints[this.#endpointIndex]]);
 
     this.#isConnected = true;
 
--- a/packages/rpc-provider/src/ws/Provider.ts
+++ b/packages/rpc-provider/src/ws/Provider.ts
@@ -132,7 +132,7 @@ export default class WsProvider implements WSProviderInterface {
       const WS = await getWSClass();
 
       this.#websocket = new WS(this.#endpoint[this.#endpointNext]);
-      this.#endpointNext++;
+      this.#endpointNext = (this.#endpointNext + 1) % this.#endpoint.length;
       this.#websocket.onclose = this.#onSocketClose;
       this.#websocket.onerror = this.#onSocketError;
       this.#websocket.onmessage = this.#onSocketMessage;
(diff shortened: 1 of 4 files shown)


jacogr:
Thanks a lot for this. Will take a detailed look through a bit later today once CI is happy.

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -75,6 +75,8 @@ export default class WsProvider implements WSProviderInterface {
 
   readonly #waitingForId: Record<string, JsonRpcResponse> = {};
 
+  #endpointNext: number;
```
jacogr, line 78: Please just move this down to below `autoConnectMs` (keeping it alphabetical in the mutable section)

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -84,17 +86,22 @@ export default class WsProvider implements WSProviderInterface {
...
    * @param {boolean} autoConnect Whether to connect automatically or not.
    */
-  constructor (endpoint: string = defaults.WS_URL, autoConnectMs: number | false = 1000) {
-    assert(/^(wss|ws):\/\//.test(endpoint), `Endpoint should start with 'ws://', received '${endpoint}'`);
+  constructor (endpoint: string | string[] = defaults.WS_URL, autoConnectMs: number | false = 1000) {
+    endpoint = (typeof endpoint === 'string') ? [endpoint] : endpoint;
+    assert(endpoint.length !== 0, 'WsProvider requires at least one Endpoint');
+    endpoint.forEach((endpoint) => {
+      assert(/^(wss|ws):\/\//.test(endpoint), `Endpoint should start with 'ws://', received '${endpoint}'`);
+    });
```
jacogr, line 97:
```suggestion
    const endpoints = isString(endpoint)
      ? [endpoint]
      : endpoint;
    
    assert(endpoints.length !== 0, 'WsProvider requires at least one Endpoint');
    
    endpoints.forEach((endpoint) => {
      assert(/^(wss|ws):\/\//.test(endpoint), `Endpoint should start with 'ws://', received '${endpoint}'`);
    });
```

(`isString` from `@polkadot/util`)

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -84,17 +86,22 @@ export default class WsProvider implements WSProviderInterface {
...
 
     this.#eventemitter = new EventEmitter();
     this.#autoConnectMs = autoConnectMs || 0;
     this.#coder = new Coder();
     this.#endpoint = endpoint;
     this.#websocket = null;
+    this.#endpointNext = 0;
```
jacogr, line 104: Can just move this up alphabetical as well.

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -65,7 +65,7 @@ const l = logger('api-ws');
 export default class WsProvider implements WSProviderInterface {
   readonly #coder: Coder;
 
-  readonly #endpoint: string;
+  readonly #endpoint: string[];
```
jacogr, line 68:
```suggestion
  readonly #endpoints: string[];
```

Since it is now always multiple.

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -124,7 +131,8 @@ export default class WsProvider implements WSProviderInterface {
     try {
       const WS = await getWSClass();
 
-      this.#websocket = new WS(this.#endpoint);
+      this.#websocket = new WS(this.#endpoint[this.#endpointNext]);
+      this.#endpointNext = (this.#endpointNext + 1) % this.#endpoint.length;
```
jacogr, line 135:
```suggestion
      this.#endpointNext = (this.#endpointNext + 1) % this.#endpoints.length;
```

(Just aligning with the `s/endpoint/endpoints/` rename above)

jacogr: Would also suggest moving this below all `this.#websocket` stuff (just keeping the grouping together)

packages/rpc-provider/src/ws/Provider.ts
jacogr: Alternatively use `Array.isArray` and just swap the `[endpoint]` and `endpoint`.

hoani: I went with `Array.isArray`, I felt like that read better than isString()

hoani:
Thanks for your review @jacogr, good call on `endpoint` -> `endpoints`

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -268,7 +281,7 @@ export default class WsProvider implements WSProviderInterface {
 
   #onSocketClose = (event: CloseEvent): void => {
     if (this.#autoConnectMs > 0) {
-      l.error(`disconnected from ${this.#endpoint} code: '${event.code}' reason: '${event.reason}'`);
+      l.error(`disconnected from ${this.#endpoints} code: '${event.code}' reason: '${event.reason}'`);
```
jacogr, line 284:
```suggestion
      l.error(`disconnected from ${this.#endpoints[this.#endpointNext]} code: '${event.code}' reason: '${event.reason}'`);
```

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -363,7 +376,7 @@ export default class WsProvider implements WSProviderInterface {
   #onSocketOpen = (): boolean => {
     assert(!isNull(this.#websocket), 'WebSocket cannot be null in onOpen');
 
-    l.debug((): any[] => ['connected to', this.#endpoint]);
+    l.debug((): any[] => ['connected to', this.#endpoints]);
```
jacogr, line 379:
```suggestion
    l.debug((): any[] => ['connected to', this.#endpoints[this.#endpointNext]]);
```

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -135,13 +135,13 @@ export default class WsProvider implements WSProviderInterface {
...
 
-      this.#websocket = new WS(this.#endpoints[this.#endpointNext]);
+      this.#websocket = new WS(this.#endpoints[this.#endpointIndex]);
       this.#websocket.onclose = this.#onSocketClose;
       this.#websocket.onerror = this.#onSocketError;
       this.#websocket.onmessage = this.#onSocketMessage;
       this.#websocket.onopen = this.#onSocketOpen;
 
-      this.#endpointNext = (this.#endpointNext + 1) % this.#endpoints.length;
+      this.#endpointIndex = (this.#endpointIndex + 1) % this.#endpoints.length;
```
jacogr, line 144: Just put this before 138, otherwise that is -1

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -77,7 +77,7 @@ export default class WsProvider implements WSProviderInterface {
 
   #autoConnectMs: number;
 
-  #endpointNext: number;
+  #endpointIndex: number;
```
jacogr, line 80: +1 on the naming :)

packages/rpc-provider/src/ws/Provider.ts
hoani: Doh! On it =)

hoani:
Nice catch! I should have noticed when I changed `endpoint` -> `endpoints` I have pushed a change changing `endpointNext` to `endpointIndex` it makes the constructor feel a bit messier since the initial value of `endpointIndex = 0`, but functionally it is the same.

hoani:
Awesome! Thanks for keeping me on my toes, this was fun 😄 - I also got kudos from one of my work mates for working on this 😉

PR #2254, opened 2020-05-07
Added validatorsLen to Heartbeat struct

arjanz:
I noticed recently introduced `validatorsLen` was missing, didn't yet added the `OverrideVersionedType`.


jacogr:
Thanks an absolute million! I'll add the versioning. Another update I missed along the way :(

PR #2279, opened 2020-05-13
Update evm definitions to include vicinity

drewstone:
Not sure if this needs to be included to get rid of an unknown event error but this is not included in the types definitions.


packages/types/src/interfaces/evm/definitions.ts
```diff
@@ -18,6 +18,10 @@ export default {
       address: 'H160',
       topics: 'Vec<H256>',
       data: 'Bytes'
+    },
+    Vicinity: {
+      gas_price: 'U256',
```
jacogr, line 23:
```suggestion
      gasPrice: 'U256',
```

PR #2329, opened 2020-06-03
Adding some extra notes on changes in the specname

g2udevelopment:
Adding some extra notes to the faq about custom types when using node-template and an alternative specName.


jacogr:
Thanks a million for this, really appreciated to make the road smoother for the next guy/gal.

PR #2336, opened 2020-06-07
remove unnecessary checks

xlc:
No idea why those checks are needed. They blocks us having types like

```
{
	Foo: {
		_enum: {
			A: 'Null',
			B: 'Option<Balance>'
		}
	},
	Bar: 'Result<Option<Balance>, Balance>',
}

```


jacogr:
The example can be done, just with more effort. 

So really happy with this, however we have had anything out of this list breaking due to gen issues in the past. So it has been adjusted with additional when confirmed working. (Worst we can have is breaking generators)

So a blanket removal without each one at least use internally or as a confirmed test is going to be problematic.

xlc:
I guess one of the root issue is those code doesn't have any tests...

jacogr:
100%, that is the underlying issue - the only tests are for the internal types since the TS output is the thing that breaks.

PR #2342, opened 2020-06-10
Add SudoBalances to Proxy Types

joepetrowski:
Ref: https://github.com/paritytech/polkadot/blob/v0.8.6/runtime/polkadot/src/lib.rs#L766L772


jacogr:
Already done - it is spec-specific, see

- https://github.com/polkadot-js/api/blob/master/packages/types-known/src/spec/westend.ts#L32
- https://github.com/polkadot-js/api/blob/master/packages/types-known/src/spec/polkadot.ts#L23

PR #2346, opened 2020-06-10
impovement: better typing of entries related funcs

monitz87:


jacogr:
As always, thanks an absolute million!

PR #2347, opened 2020-06-10
Metadata bump

emostov:
Update `v11/static-substrate.json` and `v11/static.ts` based off of metadata from polkadot `v0.8.7`.


jacogr:
Thanks for this. However the `substrate` metadata is always generated from substrate master, not Polkadot.

PR #2356, opened 2020-06-16
Update contract metadata type definitions

ascjones:
`scale-info` has finally been released, and `ink!` has now migrated to using it instead of `type-metadata`. This PR updates the type definitions, and also pulls in storage layout changes.

# todo
- [ ] Define remaining storage layout variants
- [x] Wait until https://github.com/paritytech/scale-info/pull/14 merged
- [ ] Test with actual UI


jacogr:
Tadaaa! Saw it merged, really happy with that. My only concern here is the last point, I'm unsure about how the UI will handle changes, but hopefully it doesn't end up being a nightmare.

jacogr:
Closing, seems to be replaced by @kwingram25 's efforts.

PR #2361, opened 2020-06-17
Range queries return [Codec, Hash]

jacogr:


jacogr:
Closing this one. `.range` is actually going to be problematic going forward since the whole block range is now an unsafe RPC. So need to de-prioritize this one completelt.

Issue #2368, opened 2020-06-20
Memory leak


faboweb:
The api object has a memory leak. This is consistent across many versions.

Here you see the memory consumption. The consumption consistently drops when disconnecting and reconnecting.

![image](https://user-images.githubusercontent.com/5869273/85206318-cffe6780-b2e6-11ea-86c7-5bc5df1deccc.png)

Is there a clean up we are missing somewhere?

jacogr:
Without actual usage information, this is a bit of a needle in a haystack. Having said that theye are 2 places that do not clean up -

- where staking derives are used, some of them have an in-memory cache (no TTL)
- subscriptions to multiple value do have a cache (to fill in unchanged), no TTL there

faboweb:
We run all requests for our wallet through the API. So we get all data a user needs. We run the API on the server so we don't init the API on the client as it is too slow to query data and we use IP whitelisted nodes.

We only use one block subscription. So that is probably not the issue.

Should there be a TTL for all data to prevent the API to cause crashes?

jacogr:
This would be my immediate thought as to a culprit - https://github.com/polkadot-js/api/blob/5c5db2709b93451bc695adaf8a88dbbe32813337/packages/rpc-core/src/index.ts#L438-L441

(That comment has been there since it was introduced - which is why it jumps to the front of the mind as a possible culprit)

And LRU cache there would make sense (not quite straight-forward to do since it needs to keep type + values in mind. However even just keeping it on a per-connection basis and clearing values on on unsub would already have a jump in usability.

dillu24:
We are also experiencing memory leaks lately. If we connect our [API server](https://github.com/SimplyVC/polkadot_api_server/blob/master/src/server.js ) with a high number of nodes, we are noticing that the heap size is always increasing such that our server eventually crashes.

After comparing heap dumps I am noticing that there a lot of entries related to the `state_storage` and Types such as `Registry`.

jacogr:
The above would point to the cache, yes. As an aside, are you unsubscribing when the sub is no longer needed?

dillu24:
Currently what we do is the following:

1. When our API server starts, it connects with all the nodes the user has inputted in the config file.
2. Whenever the user calls an endpoint, for example the `/api/query/imOnline/authoredBlocks`, the corresponding function from the `polkadot-js/ap`i is called (in this case `api.query.imOnline.authoredBlocks`). 
3. The connection with the Nodes providers is left as open in case of further requests.

[...]

PR #2378, opened 2020-06-23
Do not use subscriptions on single subs

jacogr:
Swaps single-shot queries to properly use the getStorage RPC, as intended.

- Closes https://github.com/polkadot-js/api/issues/1039
- Part of addressing #2368

### DIFF
--- a/packages/rpc-core/src/index.ts
+++ b/packages/rpc-core/src/index.ts
@@ -232,9 +232,10 @@ export default class Rpc implements RpcInterface {
     const subName = `${section}_${subMethod}`;
     const unsubName = `${section}_${unsubMethod}`;
     const subType = `${section}_${updateType}`;
+    let memoized: null | RpcInterfaceMethod & memoizee.Memoized<RpcInterfaceMethod> = null;
 
-    const creator = (isRaw: boolean) => (...values: any[]): Observable<any> => {
-      return new Observable((observer: Observer<any>): VoidCallback => {
+    const creator = (isRaw: boolean) => (...values: any[]): Observable<unknown> => {
+      return new Observable((observer: Observer<unknown>): VoidCallback => {
         // Have at least an empty promise, as used in the unsubscribe
         let subscriptionPromise: Promise<number | void> = Promise.resolve();
 
@@ -248,7 +249,7 @@ export default class Rpc implements RpcInterface {
           const params = this._formatInputs(def, values);
           const paramsJson = params.map((param): AnyJson => param.toJSON());
 
-          const update = (error?: Error | null, result?: any): void => {
+          const update = (error?: Error | null, result?: unknown): void => {
             if (error) {
               logErrorMessage(method, def, error);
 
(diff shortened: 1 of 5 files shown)


packages/rpc-core/src/index.ts
```diff
@@ -232,9 +232,10 @@ export default class Rpc implements RpcInterface {
     const subName = `${section}_${subMethod}`;
     const unsubName = `${section}_${unsubMethod}`;
     const subType = `${section}_${updateType}`;
+    let memoized: null | RpcInterfaceMethod & memoizee.Memoized<RpcInterfaceMethod> = null;
 
-    const creator = (isRaw: boolean) => (...values: any[]): Observable<any> => {
-      return new Observable((observer: Observer<any>): VoidCallback => {
+    const creator = (isRaw: boolean) => (...values: any[]): Observable<unknown> => {
+      return new Observable((observer: Observer<unknown>): VoidCallback => {
```
jacogr, line 238:
```suggestion
    const creator = (isRaw: boolean) => (...values: unknown[]): Observable<any> => {
      return new Observable((observer: Observer<any>): VoidCallback => {
```

packages/rpc-core/src/index.ts
```diff
@@ -248,7 +249,7 @@ export default class Rpc implements RpcInterface {
           const params = this._formatInputs(def, values);
           const paramsJson = params.map((param): AnyJson => param.toJSON());
 
-          const update = (error?: Error | null, result?: any): void => {
+          const update = (error?: Error | null, result?: unknown): void => {
```
jacogr, line 252:
```suggestion
          const update = (error?: Error | null, result?: any): void => {
```

jacogr:
@dillu24 This is one step that addresses a long-standing issue. Single-shot queries now (correctly) use a non-subscriptions interface. (It actually has an additional benefit that it also has less RPC overhead.) Once merged (it will be 1.21.0-beta.0) would appreciate a test from your side as well against beta to see if it improves. What you should see

- It should still return valid results like it has been doing (This has been tested with scripts, so don't expect any issues)
- You should see less dangling subscriptions (The derive endpoints will still create underlying subscriptions)
- As a bonus, it should actually be faster for `await <endpoint>` queries (less RPC traffic)

It will not "fix" the linked issue completely, this is just one immediate long-outstanding part. (So more work remains, however this is a step)

dillu24:
@jacogr I have deployed the API server and I am noticing the following:

- The function results are still valid
- The memory usage percentage is becoming stable or at least increasing very very slowly, and even decreasing sometimes, something which was not being done previously (mem usage was always increasing at a faster rate till the application slows down and crashes).

I will continue to monitor the app and let you know of any issues that I encounter.

Will the subscription in derive endpoints be fixed? We currently use the `api.derive.staking.validators`

Some snapshots from heap comparisons: 

image

image

jacogr:
Yes, having to eat the elephant one-bite at a time.

PR #2379, opened 2020-06-23
Fixup subscriptions for promise (w/ callback)

jacogr:
- Part of #2368
- Cleanups for #2378 - everything went through non-subs, even where not intended. Move the sub/no-sub check to the API endpoint specifically


packages/api/src/base/Decorate.ts
```diff
@@ -178,7 +178,7 @@ export default abstract class Decorate<ApiType extends ApiTypes> extends Events
     augmentObject('query', this._decorateStorage(decoratedMeta.query, this._decorateMethod), this._query, fromEmpty);
     augmentObject('consts', decoratedMeta.consts, this._consts, fromEmpty);
 
-    // rx
+    // underlying rxjs, explicitly apply the rxjs type for storage
```
jacogr, line 181:
```suggestion
    // rx
```

PR #2380, opened 2020-06-23
Ensure subscriptions always unsubscribe as required

jacogr:
Part of #2368 

There is an issue where it does not (on derives being very apparent) close properly on single-shot results. Not only single-shot, but all subscriptions never fired unsubscribe over the wire. (After the swap to string ids for subscriptions this snuck in)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.21.0-beta.x
 
+- Fix subscriptions never unsubscribing after the id type swap in 1.18.1
 - Promise API will default to using `getStorage` on non-subscription calls, reducing RPC overhead
 
 ## 1.20.1 Jun 22, 2020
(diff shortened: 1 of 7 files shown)


jacogr:
cc @dillu24 This fixes the root cause that snuck in in 1.18 - nothing was ever unsubscribed. So once merged (1.21.0-beta.2), would appreciate a look. It will do the right thing now even on derives (those actually acted as the explicit testcase).

Will have a 1.21 later today with this included (there is a signed extension change that needs to go ASAP as well)

dillu24:
@jacogr I tried to run the latest beta (1.21.0-beta.3) and ran into this problem:

image

jacogr:
Weird... tested it all though staking on the apps UI. Ok, will take a peek, sorry for that.

dillu24:
@jacogr I am currently running my API server on a VM, it seems that memory is being released successfully so far.

Will leave it running and let you know of any issues.

PR #2405, opened 2020-07-02
fix: broken links in readme

yjkimjunior:


yjkimjunior:
erm, so here fixing the README path breaks the doc generation, and vice versa?

jacogr:
It is there for the doc generation, yes. There are some tricks that can be done to hide completely in GH-mode which can probably be explored. (Since md is just HTML, can wrap less useful content in a div, hide by default, show in docs. Would need to apply everywhere)

jacogr:
No easy fix, tries a couple of options in #2409 & #2410 

So it will have to stay non-navigatable until a better idea comes along.

jacogr:
I guess the only option left here is o not use the README.md in the package for the docs, but rather craft 2 seperate versions and use appropriately. (There would be some tweaks to the doc building process, which will also affect the common & ui repos)

PR #2423, opened 2020-07-13
[NEW] Support for custom headers with providers

brad-larson:
This allows the user to supply arbitrary headers when constructing the
providers.

### DIFF
--- a/packages/rpc-provider/src/ws/Provider.ts
+++ b/packages/rpc-provider/src/ws/Provider.ts
@@ -67,6 +67,8 @@ export default class WsProvider implements WSProviderInterface {
 
   readonly #endpoints: string[];
 
+  readonly #headers: Record<string, string>;
+
   readonly #eventemitter: EventEmitter;
 
   readonly #handlers: Record<string, WsStateAwaiting> = {};
@@ -89,7 +91,7 @@ export default class WsProvider implements WSProviderInterface {
    * @param {string | string[]}  endpoint    The endpoint url. Usually `ws://ip:9944` or `wss://ip:9944`, may provide an array of endpoint strings.
    * @param {boolean} autoConnect Whether to connect automatically or not.
    */
-  constructor (endpoint: string | string[] = defaults.WS_URL, autoConnectMs: number | false = 1000) {
+  constructor (endpoint: string | string[] = defaults.WS_URL, autoConnectMs: number | false = 1000, headers: Record<string, string> = {}) {
     const endpoints = Array.isArray(endpoint)
       ? endpoint
       : [endpoint];
@@ -105,6 +107,7 @@ export default class WsProvider implements WSProviderInterface {
     this.#coder = new Coder();
     this.#endpointIndex = -1;
     this.#endpoints = endpoints;
+    this.#headers = headers;
     this.#websocket = null;
 
     if (autoConnectMs > 0) {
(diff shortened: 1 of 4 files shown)


packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -138,7 +141,9 @@ export default class WsProvider implements WSProviderInterface {
 
       const WS = await getWSClass();
 
-      this.#websocket = new WS(this.#endpoints[this.#endpointIndex]);
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore - WS may be an instance of w3cwebsocket, which supports headers
```
brad-larson, line 145: There is likely a more elegant way to do this - but I wasn't sure which approach was preferred. Happy to update in that case

jacogr:
Thank you, took a really quick couple of minutes glance and it seems fine, will do a proper look-through ASAP.

TIL ... standard (browser) WebSocket does not support custom headers. (In all honesty, did actually know about this, but it has been a while since actively woking with it so it has not been front-of mind.) So effectively it is only an option in the Node.js environment. Not against it at all in this (more-limited) case, it is just disappointing.

packages/rpc-provider/src/ws/Provider.ts
jacogr: So this is where not doing this 100% in `getWSClass` bytes us. So generally, it would be -

```js
function getWSClass (): typeof WebSocket | typeof W3CWebSocket { ...} 

...

this.websocket = WS instance WebSocket
  ? new WS(...)
  : new WS(...)
```

Not the end of the world (although we know TS-overrides a a bit smelly), but the imp[act otherwise if probably much worse. So the comment here is probably the lesser of all evils and should rather be left as-is until the typing of `getWsClass` is done properly. (And not hacked about via `as unknown as ...` - which created the above)

TL;DR Certainly not perfect, but follows the lead of an issue introduced elsewhere - so with the comment, should be all good.

packages/rpc-provider/src/ws/Provider.ts
```diff
@@ -138,7 +141,9 @@ export default class WsProvider implements WSProviderInterface {
...
+      this.#websocket = new WS(this.#endpoints[this.#endpointIndex], null, null, this.#headers);
```
jacogr, line 146:
```suggestion
      this.#websocket = new WS(this.#endpoints[this.#endpointIndex], undefined, undefined, this.#headers);
```

type definition is `constructor(
        url: string,
        protocols?: string | string[],
        origin?: string,
        headers?: http.OutgoingHttpHeaders,
        requestOptions?: object,
        IClientConfig?: IClientConfig,
    );`

jacogr: Would also suggest, looking at the above -

```js
import { isChildClass } from '@polkadot/util`;

...
this.#websocket = typeof WebSocket !== 'undefined' && isChildClass(WebSocket, WS)
  ? new WS(this.#endpoints[this.#endpointIndex])
  // // @ts-ignore
  : new WS(this.#endpoints[this.#endpointIndex], undefined, undefined, this.#headers);
```

this should (hopefully, maybe) fix the test failures as well?

brad-larson: Updated!

jacogr:
Thank you very much, really appreciate it!

iicc1:
Thanks for your contribution. I was wondering how to use the API with Auth Basic.
I think this should be documented in the docs, I spent hours trying to figure out how to correctly add the headers until I saw the test file of this pull.
In case someone is wondering, here it is how it is done:
`const provider = new WsProvider('ws/s://ip:port', retry, { Authorization: 'Basic xxxx' })`

jacogr:
PRs to the docs are welcome and always appreciated.

PR #2438, opened 2020-07-20
support rpc import definitions

aniiantt:
`generateRpcTypes` should support custom rpcs. So `generateRpcTypes` should be able to pass in user-defined definitions, just like `generateInterfaceTypes` and `generateTsDef`.


jacogr:
Thank you fr this! Yes, this is a missing piece of the puzzle. Will look through and test today.

PR #2475, opened 2020-08-06
Add Centrifuge missing types

mikiquantum:


packages/types-known/src/spec/centrifuge-chain.ts
```diff
@@ -6,27 +6,46 @@ import { OverrideVersionedType } from '@polkadot/types/types';
...
+    leafHash: 'H256',
+    sortedHashes: 'H256'
+  },
   ReferendumInfo: 'ReferendumInfoTo239',
   StakingLedger: 'StakingLedgerTo240',
   Weight: 'u32'
 };
 
 const versioned: OverrideVersionedType[] = [
-  {
```
mikiquantum, line 17: We moved all the environments to 229, no need for backwards compatibility.

PR #2514, opened 2020-08-25
Update definitions.ts

kianenigma:
Will fix the decoding of these types. Sadly cannot provide a compact test case.


jacogr (commented):
Thank you, but the encoding for pre and post is actually 100% equivalent. So it won't make a difference in encoding/decoding. 

(The same type of definition structure has been used for the older format as well since it actually make working with it more consistent)

kianenigma:
>  but the encoding for pre and post is actually 100% equivalent

I thought so as well, but believe me, it does. I will prepare a repo in which you can inspect and link it here.

jacogr:
That is why tests are needed - i.e. the test failures introduced by you changes on the old type are based on actual decoding of on-chain data.

jacogr:
Anyway, if the new change does fix your issue, you do need to revert the changes on the old one (which does work and where the tests are now failing). So happy to try to unblock you, when this stuff appears on-chain and there is actual data can actually dig in.

kianenigma:
But one thing to note is that the test failure itself is a proof that this statement is wrong: 

> Thank you, but the encoding for pre and post is actually 100% equivalent.

I added two more tests to demonstrate the change. The bytes that I decode are coming from test case `solution_struct_works_with_and_without_compact` within substrate, which you can double check. Since we don't have live data of the new compact type, this is as good as it gets. 

Given this, I will revert the other changes to the legacy type to unblock the PR, but I do want to raise that something is def. off here: what I said in the beginning of this comment should be investigated. 

PS. I hope to have adhered sufficiently to your style guide.

jacogr:
I looked at the data, not sure why it was changed - as specified it comes directly from on-chain.

Happy for more tests, but prefer not breaking the old, rather adding new.

kianenigma:
kk, then we should be green now.

jacogr (approved):
Thank you!

PR #2519, opened 2020-08-25
add TransactionPriority def

aniiantt:
Hi, we upgraded the Laminar and were ready to plug it into the Rococo test network, but found there was no way to run it on the Polkadot Apps because a type definition was missing.

`TransactionPriority: u64`
https://github.com/paritytech/substrate/blob/0a391e4913a18ca78c1e08b220c3db120086595a/primitives/runtime/src/transaction_validity.rs#L25


jacogr (approved):
Thank you! We generally add the types as they are used in Substrate in one of the modules, so in this case you are the first..

Issue #2527, opened 2020-08-28
Query historical metadata without contaminating registry


ianhe8x:
There're cases we want to query metadata of a past block for comparison and don't want to have any side effects for such query. But with current implementation, it actually call `registry.setMetadata()`.
I believe this is an unexpected behavior for most people.

https://github.com/polkadot-js/api/blob/9c1327ea22e1e8600fa8d216ce36f6618ffdb093/packages/metadata/src/Metadata/MetadataVersioned.ts#L44

jacogr:
Should be done as part of #845. As it stands, it is actually expected behavior, since the registry contains all the info. Until the API had native historical-query support, it is non-optimal.

PR #2534, opened 2020-08-29
Add swapRegistry(hash?) to initlize clean registry

jacogr:
- Closes https://github.com/polkadot-js/api/issues/2527
- Part of #845

Usage -

```js
const HASH = '0xa8b5c7c2656d303d9b34ff44726d7091645f15b9e9e661e2d0ade1f520366bc2';

// initialize registry (once-off per spec version, will re-use if swapped again)
api.swapRegistry(HASH);

console.log(await api.query.system.timestamp.at(HASH));

// swap to default (current metadata)
api.swapRegistry();

console.log(await api.query.system.timestamp());
```

`.history(HASH)` queries will do this automatically


jacogr:
After https://github.com/polkadot-js/api/pull/2535 `swapRegistry(...)` does not need to be called for `.at` queries, it is done automatically.

Issue #2541, opened 2020-08-30
`api.derive.chain.getHeader()` fails with repeated calls to same block; api version >= 1.31.beta.17


emostov:
When querying `api.derive.chain.getHeader()` with an historic block hash (block from an older spec version) consecutively, it fails with the error `Unable to decode Vec on index 0 Tuple: failed on 0:: Invalid character`. I assume this is related to the optimization introduced here: https://github.com/polkadot-js/api/commit/637069d723bf9a8afcd42af8b3162a20737ff501

Here is a link to some code that illustrates the issue: https://github.com/emostov/substrate-tx-seed/blob/1-31-beta-17/src/main.ts (example is just in main.ts, rest of repo irrelevant, except for package info)

And here is that same code for posterity: 
```
async function main(): Promise<void> {
	const wsProvider = new WsProvider('wss://rpc.polkadot.io');
	const api = await ApiPromise.create({
		provider: wsProvider,
	});

	const hash1378238 =
		'0xbc38431a6a4e0cdaadd3b34d1fd1d6692cb08ff07d58bb1726101d341f100892'; // polkadot block number 1378238
	const firstTry1378238 = await api.derive.chain.getHeader(hash1378238);
	console.log('This works: (firstTry1378238) ', firstTry1378238?.toHuman());

	const secondTry1378238 = await api.derive.chain.getHeader(hash1378238);
	console.log('This works: (secondTry1378238) ', secondTry1378238?.toHuman());

	const hash1 =
		'0xc0096358534ec8d21d01d34b836eed476a1c343f8724fa2153dc0725ad797a90'; // polkadot block number 1

	const firstTry1 = await api.derive.chain.getHeader(hash1);
	console.log('This works: (firstTry1)', firstTry1?.toHuman());

	const secondTry1 = await api.derive.chain.getHeader(hash1);
	// Logs to console: Unable to decode Vec on index 0 Tuple: failed on 0:: Invalid character
	console.log(
		'This will not log to console (secondTry1)',
		secondTry1?.toHuman()
	);
}
```

Worth noting I consistently get this error with polkadot block 1, but with other low number blocks, consecutive calls seem to work sometimes.

TheGoldenEye:
I can confirm the error, it also occurs in my code

PR #2543, opened 2020-08-31
Ensure non-destructive registry swap

jacogr:
Closes https://github.com/polkadot-js/api/issues/2541 - ensure that the call works with a non-destructive registry as set for it specifically.

### DIFF
--- a/packages/api/src/base/Init.ts
+++ b/packages/api/src/base/Init.ts
@@ -216,7 +216,7 @@ export default abstract class Init<ApiType extends ApiTypes> extends Decorate<Ap
 
   private _adjustBundleTypes (registry: Registry, chain: Text, specName: Text): void {
     // adjust known type aliases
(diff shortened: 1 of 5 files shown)


packages/api/src/base/Init.ts
```diff
@@ -222,9 +214,9 @@ export default abstract class Init<ApiType extends ApiTypes> extends Decorate<Ap
     ).subscribe();
   }
 
-  private _adjustBundleTypes (chain: Text, specName: Text): void {
+  private _adjustBundleTypes (registry: Registry, chain: Text, specName: Text): void {
     // adjust known type aliases
-    this.registry.knownTypes.typesAlias = getSpecAlias(this.registry, chain, specName);
+    registry.knownTypes.typesAlias = getSpecAlias(this.registry, chain, specName);
```
jacogr, line 219:
```suggestion
    registry.knownTypes.typesAlias = getSpecAlias(registry, chain, specName);
```

PR #2548, opened 2020-08-31
Explicit delay for per-hash single queries

jacogr:


packages/rpc-core/src/index.ts
```diff
@@ -221,6 +222,8 @@ export default class Rpc implements RpcInterface {
     };
 
     const creator = (isRaw: boolean) => (...values: any[]): Observable<any> => {
+      const isDelayed = (hashIndex !== -1 && !!values[hashIndex]) || (cacheIndex !== -1 && !!values[hashIndex]);
```
jacogr, line 225:
```suggestion
      const isDelayed = (hashIndex !== -1 && !!values[hashIndex]) || (cacheIndex !== -1 && !!values[cacheIndex]);
```

PR #2566, opened 2020-09-01
add vestingRemainingBlock in DeriveBalancesAll

niklabh:
https://github.com/polkadot-js/apps/issues/3561


jacogr:
This cane be done on the using-api layer. There is no need to add it here. The end block is available.

PR #2574, opened 2020-09-03
staking.currentEra returns Option

jacogr:


emostov:
Is there legacy staking pallet code where `CurrentEra` is not an `Option<EraIndex>`? I have some code that assumes it is always an Option and I am wondering if it might break on some historic queries (yet to encounter this).

jacogr:
Yes. It is an issue for some older chains, e.g. Crab network, hence the change.

PR #2577, opened 2020-09-03
Add `isReadyOrError` when connecting to API

shawntabrizi:
https://github.com/polkadot-js/api/commit/51423e20b5297a18bd467b9285b2a496d8c71775 changed the behavior `api.isReady`. In the case of an error, a `reject` is returned, ending the promise.

This is not nice for developers who want to allow the promise to last forever until the connection is made. (Someone who wants exactly the opposite of what the PR above does).

So rather than have the API choose to prefer one kind of user, we expose two functions, one for each behavior.

* `api.isReady()` returns to its old behavior, keeping the promise open until it is resolved and the connection is made.
* `api.isReadyOrError()` has this new behavior, rejecting the promise in the case of some error connecting to the endpoint.

cc @laech

### DIFF
--- a/packages/api/src/promise/Api.spec.ts
+++ b/packages/api/src/promise/Api.spec.ts
@@ -97,7 +97,7 @@ describe('ApiPromise', (): void => {
       }
 
       try {
-        await new ErrorApiPromise().isReady;
+        await new ErrorApiPromise().isReadyOrError;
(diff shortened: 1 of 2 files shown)


jacogr (approved):
Thank you.

Will indicate this clearly in the CHANGELOG as breaking change.

PR #2582, opened 2020-09-04
Support new contracts Schedule

jacogr:
As per https://github.com/paritytech/substrate/pull/7017


athei:
The `Schedule` was changed before merge in the referenced PR. Forgot to inform you about that.

jacogr:
Sorry I missed that - I tried to check in all through the process, so it must have been one of the alerts I didn't check.

Issue #2620, opened 2020-09-17
Corrupt data from `query.system.events.at` after making historic query


emostov:
Relates to: https://github.com/paritytech/substrate-api-sidecar/issues/290

Code to reproduce: (also have actual repo here, using api `1.34.0-beta.10.`: https://github.com/emostov/substrate-tx-seed/blob/lost-events/src/main.ts)

```
import { ApiPromise, WsProvider } from '@polkadot/api';
import { Vec } from '@polkadot/types';
import { EventRecord } from '@polkadot/types/interfaces';

async function main(): Promise<void> {
	const wsProvider = new WsProvider('wss://rpc.polkadot.io');
	const api = await ApiPromise.create({
		provider: wsProvider,
	});

	const mapEvents = (events: Vec<EventRecord>) => {
		// console.log(events.toHuman());
		return events
			.map(({ event: { data, section, method } }) => {
				if (section && method) {
					// eslint-disable-next-line @typescript-eslint/restrict-template-expressions
					return `${section}.${method} - data: ${data.toString()}`;
				} else {
					return 'WARN: event missing section and method';
				}
			})
			.join(',\n');
	};

	const hash1622398_1 = await api.rpc.chain.getBlockHash(1622398);
	const events1622398_1 = await api.query.system.events.at(hash1622398_1);
	console.log('Block 1622398, attempt 1');
	console.log(mapEvents(events1622398_1), '\n');

	const hash1 = await api.rpc.chain.getBlockHash(1);
	const events1 = await api.query.system.events.at(hash1);
	console.log('Block 1');
	console.log(mapEvents(events1), '\n');

	const hash1622398_2 = await api.rpc.chain.getBlockHash(1622398);
	const events1622398_2 = await api.query.system.events.at(hash1622398_2);
	console.log('Block 1622398, attempt 2');
	console.log(mapEvents(events1622398_2), '\n');

	process.exit();
}

main().catch(console.error);
```

console output:
```
Block 1622398, attempt 1
system.ExtrinsicSuccess - data: [{"weight":159133000,"class":"Mandatory","paysFee":"Yes"}],
system.NewAccount - data: ["12Sz3YJih3NFjFUYGeGB5LnqrZNyytEpWaoBwP5tMSHuxaAV"],
balances.Endowed - data: ["12Sz3YJih3NFjFUYGeGB5LnqrZNyytEpWaoBwP5tMSHuxaAV",73409000000000],
[...]

Issue #2612, opened 2020-09-15
Decoration errors


mariopino:
Hi,

Tested @polkadot/api v1.31.2 in PolkaStats backend, crawling Kusama full chain and got some errors I want to share:

```
const blockHash = await api.rpc.chain.getBlockHash(4061641);
blockEvents = await api.query.system.events.at(blockHash);

Error: createType(Vec<EventRecord>):: Vec length 960823552 exceeds 32768
    at o (https://polkadot.js.org/apps/polkadot.01.cde4c033.js:1:69825)
    at t.default._formatStorageData (https://polkadot.js.org/apps/polkadot.01.cde4c033.js:1:342495)
    at t.default._formatOutput (https://polkadot.js.org/apps/polkadot.01.cde4c033.js:1:341740)
    at l (https://polkadot.js.org/apps/polkadot.01.cde4c033.js:1:339890)
```

```
const blockHash = await api.rpc.chain.getBlockHash(4061636);
blockEvents = await api.query.system.events.at(blockHash);

Error: createType(Vec<EventRecord>):: Unable to create Enum via index 162, in ApplyExtrinsic, Finalization, Initialization
    at o (https://polkadot.js.org/apps/polkadot.01.cde4c033.js:1:69825)
    at t.default._formatStorageData (https://polkadot.js.org/apps/polkadot.01.cde4c033.js:1:342495)
    at t.default._formatOutput (https://polkadot.js.org/apps/polkadot.01.cde4c033.js:1:341740)
    at l (https://polkadot.js.org/apps/polkadot.01.cde4c033.js:1:339890)
```

**NOTE:** Also can be reproduced with api v1.34.0-beta.3 at https://polkadot.js.org/apps/#/js

jacogr:
Weird, both works via the apps UI explorer.

mariopino:
Strange, I got the errors 1 hour ago in JS console using the same API version (v1.34.0-beta.3), but now there is no error! Closing it then ...

PR #2621, opened 2020-09-17
Explicit tie of metadata to registry (not implicit)

jacogr:
Introduced alongside the tx-wrapper and problematic ever since. All metadata instances are always ties the registry that created them to that instance. 

Fixes https://github.com/polkadot-js/api/issues/2612 & Closes https://github.com/polkadot-js/api/issues/2620


packages/metadata/src/Metadata/MetadataVersioned.ts
```diff
@@ -40,8 +40,6 @@ export default class MetadataVersioned extends Struct {
       magicNumber: MagicNumber,
       metadata: 'MetadataAll'
     }, value as Map<unknown, unknown>);
-
-    registry.setMetadata(this);
```
jacogr, line 44: Problematic line :( Not sure why this seemed like a good idea.

PR #2627, opened 2020-09-20
Add support to extract PoW author

sorpaas:
It's directly encoded as the PreRuntime.


packages/types/src/generic/ConsensusEngineId.ts
```diff
@@ -12,8 +12,9 @@ import U32 from '../primitive/U32';
 const CID_AURA = 0x61727561; // 'aura'
 const CID_BABE = 0x45424142; // 'BABE'
 const CID_GRPA = 0x4b4e5246; // 'FRNK' (don't ask, used to be afg1)
+const CID_POW = 0x5f776f70; // 'pow_'
```
jacogr, line 15: Checked, correct in reverse order.

jacogr (approved):
Thank you. Actually a bit sceptical about this since it non-standard, but actually quite happy to support this properly.

PR #2631, opened 2020-09-21
Check for erasStakersClipped in staking eraStakingExposure

### LINKED ISSUE #2630: Cater for older chain for derive eraStakingExposure
As discussed https://github.com/polkadot-js/apps/pull/3714#discussion_r490943875

Tbaut:
closes #2630

### DIFF
--- a/packages/api-derive/src/staking/erasExposure.ts
+++ b/packages/api-derive/src/staking/erasExposure.ts
@@ -44,15 +44,17 @@ export function _eraExposure (instanceId: string, api: ApiInterfaceRx): (era: Er
 
     return cached
       ? of(cached)
-      : api.query.staking.erasStakersClipped.entries(era).pipe(
-        map((stakers): DeriveEraExposure => {
-          const value = mapStakers(era, stakers);
+      : api.query.staking.erasStakersClipped
+        ? api.query.staking.erasStakersClipped.entries(era).pipe(
+          map((stakers): DeriveEraExposure => {
+            const value = mapStakers(era, stakers);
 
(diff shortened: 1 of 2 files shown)


packages/api-derive/src/staking/erasExposure.ts
```diff
@@ -44,15 +44,17 @@ export function _eraExposure (instanceId: string, api: ApiInterfaceRx): (era: Er
...
-          !withActive && deriveCache.set(cacheKey, value);
+            !withActive && deriveCache.set(cacheKey, value);
 
-          return value;
-        })
-      );
+            return value;
+          })
+        )
+        : of();
```
jacogr, line 57: Cannot be undefined here, need to be empty (the signature specified `DeriveEraExposure`)

Tbaut: Interresting that TS didn't complain in VSC. Just for my info, how should I do this? because something like `of({}) as DeriveEraExposure` would also not work given that the properties aren't optional, the type is defined as follow:
```
DeriveEraExposure {
  era: EraIndex;
  nominators: DeriveEraNominatorExposure;
  validators: DeriveEraValidatorExposure;
}
```

jacogr: Generally in this case the props should be optional. The issue with that is obviously that it mucks up all the way down the line, i.e. evrybody needs to do the `.?` everywhere.

Actually the optional fields in the derives are all due to support for older chains, and it adds maintenance overhead everywhere. (And then it still goes wrong in some cases)

PR #2641, opened 2020-09-24
Filters for undefined stragglers

drewstone:


packages/api-derive/src/staking/waitingInfo.ts
```diff
@@ -17,7 +17,9 @@ export function waitingInfo (instanceId: string, api: ApiInterfaceRx): () => Obs
     ]).pipe(
       switchMap(([{ nextElected }, stashes]): Observable<DeriveStakingWaiting> => {
         const elected = nextElected.map((a) => a.toString());
-        const waiting = stashes.filter((v) => !elected.includes(v.toString()));
+        const waiting = stashes
+          .filter(v => !!v)
+          .filter((v) => !elected.includes(v.toString()));
```
jacogr, line 22: This should not been needed when everything operates properly. If it still occurs after the type updates, something is wrong somewhere.

PR #2645, opened 2020-09-24
fix(typegen): add support for VecFixed in enums

monitz87:
Enums like

```
SomeEnum: {
  _enum: {
    VecValue: '[u8; 32]',
    OtherValue: '[u8;64]',
  }
}
```

Weren't being supported properly by the typegen. This PR fixes that


jacogr (approved):
Thank you!

PR #2648, opened 2020-09-24
types: add IdentityJudgment proxy type variant

andresilva:
We were missing the `IdentityJudgment` variant which is defined for polkadot runtimes. This isn't actually correct though since on the polkadot runtime we skip one index in the enum:

- Kusama/Westend: https://github.com/paritytech/polkadot/blob/master/runtime/kusama/src/lib.rs#L791
- Polkadot: https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L778

### DIFF
--- a/packages/types/src/interfaces/proxy/definitions.ts
+++ b/packages/types/src/interfaces/proxy/definitions.ts
@@ -15,7 +15,7 @@ export default {
       delay: 'BlockNumber'
     },
     ProxyType: {
-      _enum: ['Any', 'NonTransfer', 'Governance', 'Staking']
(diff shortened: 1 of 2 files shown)


jacogr:
We do have the overrides for -

- Kusama, https://github.com/polkadot-js/api/blob/master/packages/types-known/src/spec/kusama.ts#L10-L12
- Polkadot, https://github.com/polkadot-js/api/blob/master/packages/types-known/src/spec/polkadot.ts#L10-L13

andresilva:
Didn't notice those, sorry for the noise!

PR #2656, opened 2020-09-27
add secondary_vrf field to babe EpochAuthorship

andresilva:
https://github.com/paritytech/substrate/blob/master/client/consensus/babe/rpc/src/lib.rs#L180


jacogr (approved):
Thank you.

PR #2671, opened 2020-10-02
Remove broken link

roccomuso:
Don't know where the package has been moved


jacogr (approved):
Thank you! (It has been combined into the types, so each module has types & RPC endpoints, aligning what we find on Substrate)

PR #2673, opened 2020-10-02
remove broken link

roccomuso:
removing another dead link


jacogr (commented):
This is actually not a dead link, this is used for doc generation.

PR #2692, opened 2020-10-08
Introduce OriginCaller (& apply to Scheduler)

### LINKED ISSUE #2691: PalletsOrigin / OriginCaller & Scheduled type updated
With https://github.com/paritytech/substrate/pull/6387, it is now possible to store and expose `PalletsOrigin` type. It is generated by `construct_runtime` macro similar to `Call` type so polkadot.js needs some customized code to generate it.

Also just found out `Scheduled` type definition is not updated (or did I missed something?).

This means polkadot.js won't be able to decode `Agenda` correctly.

https://github.com/paritytech/substrate/blob/81f3e7caf351c27aab518ee3b640c06b1a4392bf/frame/scheduler/src/lib.rs#L161-L162

https://github.com/paritytech/substrate/blob/81f3e7caf351c27aab518ee3b640c06b1a4392bf/frame/scheduler/src/lib.rs#L126-L136

https://github.com/polkadot-js/api/blob/2faf68c4d2ddb1fef710f5ebf5ef8ce2088b923c/packages/types/src/interfaces/scheduler/definitions.ts#L16-L21

jacogr:
Closes #2691 

Sadly needs a lot of elbow-grease, the proc macro stuff is an absolute mess to try and bring in.

### DIFF
--- a/packages/metadata/src/Metadata/v12/toLatest.ts
+++ b/packages/metadata/src/Metadata/v12/toLatest.ts
@@ -8,6 +8,8 @@ import { getModuleTypes } from '@polkadot/types-known';
 import { Type } from '@polkadot/types/primitive';
 import { stringCamelCase } from '@polkadot/util';
 
+// Since we don't have insight into the origin specification, we can only define what we know about
+// in a pure Substrate/Polkadot implementation, any other custom origins won't be handled at all
 const KNOWN_ORIGINS: Record<string, string> = {
   Council: 'CollectiveOrigin',
   System: 'SystemOrigin',
--- a/packages/metadata/src/Metadata/v12/toLatest.ts
+++ b/packages/metadata/src/Metadata/v12/toLatest.ts
@@ -1,12 +1,18 @@
 // Copyright 2017-2020 @polkadot/metadata authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { EventMetadataV12, EventMetadataLatest, FunctionMetadataV12, FunctionMetadataLatest, MetadataV12, MetadataLatest, ModuleMetadataLatest, StorageMetadataV12, StorageMetadataLatest, StorageEntryMetadataLatest } from '@polkadot/types/interfaces/metadata';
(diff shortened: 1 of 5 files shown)


packages/metadata/src/Metadata/v12/toLatest.ts
```diff
@@ -1,13 +1,19 @@
...
 
-import { EventMetadataV12, EventMetadataLatest, FunctionMetadataV12, FunctionMetadataLatest, MetadataV12, MetadataLatest, ModuleMetadataLatest, StorageMetadataV12, StorageMetadataLatest, StorageEntryMetadataLatest } from '@polkadot/types/interfaces/metadata';
+import { EventMetadataV12, EventMetadataLatest, FunctionMetadataV12, FunctionMetadataLatest, MetadataV12, MetadataLatest, ModuleMetadataLatest, ModuleMetadataV12, StorageMetadataV12, StorageMetadataLatest, StorageEntryMetadataLatest } from '@polkadot/types/interfaces/metadata';
 import { Registry, OverrideModuleType } from '@polkadot/types/types';
 
 import { getModuleTypes } from '@polkadot/types-known';
 import { Type } from '@polkadot/types/primitive';
 import { stringCamelCase } from '@polkadot/util';
 
+const KNOWN_ORIGINS: Record<string, string> = {
```
jacogr, line 11:
```suggestion
// Since we don't have insight into the origin specification, we can only define what we know about
// in a pure Substrate/Polkadot implementation, any other custom origins won't be handled at all
const KNOWN_ORIGINS: Record<string, string> = {
```

packages/metadata/src/Metadata/v12/toLatest.ts
```diff
@@ -81,12 +89,39 @@ function convertStorage (registry: Registry, { items, prefix }: StorageMetadataV
   });
 }
 
+// generate & register the OriginCaller type
+function registerOriginCaller (registry: Registry, modules: ModuleMetadataV12[]): void {
```
xlc, line 93: can you somehow allow us to insert entries into `KNOWN_ORIGINS`? an additional parameter to registry should be enough.

PR #1662, opened 2019-12-10
Updated ABI definitions & parsing

jacogr:
- Closes #1660 Closes #1549 
- Implement types definitions for https://github.com/paritytech/ink/pull/296
- Implement type definitions for https://github.com/type-metadata/type-metadata/pull/36

PR #2693, opened 2020-10-08
Types for ink! 3.0

kwingram25:


copy-to-apps.sh
```diff
@@ -2,24 +2,4 @@
...
-copy_folder "api"
-copy_folder "api-contract"
-copy_folder "api-derive"
-copy_folder "metadata"
-copy_folder "rpc-core"
-copy_folder "rpc-provider"
-copy_folder "types"
-copy_folder "types-known"
-copy_folder "typegen"
+./copy-to.sh "apps"
```
jacogr, line 5: This should be reverted. There is a more generic version that will be added to dev. (Since it is duplicated everywhere)

packages/api-contract/src/base/Contract.ts
```diff
@@ -33,50 +34,52 @@ export type ContractCallResult<CallType extends ContractCallTypes> = CallType ex
 export default class Contract<ApiType extends ApiTypes> extends BaseWithTxAndRpcCall<ApiType> {
   public readonly address: Address;
 
-  constructor (api: ApiObject<ApiType>, abi: ContractABIPre | Abi, decorateMethod: DecorateMethod<ApiType>, address: string | AccountId | Address) {
+  constructor (api: ApiObject<ApiType>, abi: AnyJson | InkAbi, decorateMethod: DecorateMethod<ApiType>, address: string | AccountId | Address) {
     super(api, abi, decorateMethod);
 
-    this.address = this.registry.createType('Address', address);
+    this.address = this.api.registry.createType('Address', address);
```
jacogr, line 40: It should point to the own registry here, it always needs to in any type. The API one is not explicitly required to point to latest. It does move around based on the actual queries we are making, i.e. non-static and known types may be removed.

packages/api-contract/src/base/Contract.ts
```diff
@@ -3,14 +3,15 @@
...
+import { AnyJson, Codec, CodecArg, IKeyringPair } from '@polkadot/types/types';
+import { ApiObject, ContractCallOutcome, InkMessage } from '../types';
 
 import BN from 'bn.js';
 import { Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { SubmittableResult } from '@polkadot/api';
-import Abi from '../Abi';
+import { assert } from '@polkadot/util';
+import InkAbi from '../InkAbi';
```
jacogr, line 14:
```suggestion
import { assert } from '@polkadot/util';

import InkAbi from '../InkAbi';
```

packages/api-contract/src/base/Contract.ts
```diff
@@ -33,50 +34,52 @@ export type ContractCallResult<CallType extends ContractCallTypes> = CallType ex
...
+    const message = this.abi.messages[messageIndex];
 
     return {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
       send: this.decorateMethod(
         as === 'rpc' && this.hasRpcContractsCall
           ? (account: IKeyringPair | string | AccountId | Address): ContractCallResult<'rpc'> =>
             this._rpcContractsCall(
-              this.registry.createType('ContractCallRequest', {
+              this.api.registry.createType('ContractCallRequest', {
```
jacogr, line 56: As above.

packages/api-contract/src/base/Contract.ts
```diff
@@ -33,50 +34,52 @@ export type ContractCallResult<CallType extends ContractCallTypes> = CallType ex
...
                 dest: this.address.toString(),
                 gasLimit,
-                inputData: fn(...params),
+                inputData: message(...params),
                 origin: account,
                 value
               })
             ).pipe(map((result: ContractExecResult): ContractCallOutcome =>
-              this._createOutcome(result, this.registry.createType('AccountId', account), def, params)
+              this._createOutcome(result, this.api.registry.createType('AccountId', account), message, params)
```
jacogr, line 64: As above.

packages/api-contract/src/base/Contract.ts
```diff
@@ -33,50 +34,52 @@ export type ContractCallResult<CallType extends ContractCallTypes> = CallType ex
...
     let output: Codec | null = null;
 
     if (result.isSuccess) {
       const { data } = result.asSuccess;
 
       output = message.returnType
-        ? formatData(this.registry, data, message.returnType)
-        : this.registry.createType('Raw', data);
+        ? formatData(this.api.registry, data, message.returnType.type)
+        : this.api.registry.createType('Raw', data);
```
jacogr, line 82: As above.

packages/api-contract/src/base/Contract.ts
jacogr: (The constructor is still fine, however inside the class it becomes problematic). So generally in both cases, always point to own as set.

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -0,0 +1,173 @@
+// Copyright 2017-2020 @polkadot/types authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// const layout = {
+//   InkLayoutArray: {
+//     cells_per_elem: 'u32',
```
jacogr, line 6: Always camelCase

jacogr (commented):
Some small comments from an initial read-through, will give a proper walk-through in a while.

Can merge it when the apps UI has support for this variant, would not like to break compat without any fallbacks. (Since this goes in immediately)

EDIT: Since it contains the commits from https://github.com/polkadot-js/api/pull/1662 a bit difficult to look at since not sure where new changes has been made. (Would have preferred a PR into that branch)

packages/types/src/interfaces/contractsAbi/definitions.ts
kwingram25: This is directly based on the ink! 3.0 abi format, designed to be created from json...not sure how to use camelcase instead, though it's not necessary for now

jacogr: Ummm... then that is wrong/weird there. Seems there are some camelCase in that JSON and some not. Having said that, I believe the actual JSON paring does take care of anything non-camel, well, it _used_ to, will need a check. (If it doesn't, can't leave it as-is, but then will apply a fix there)

jacogr: Yeap, here, it attempts to... https://github.com/polkadot-js/api/blob/master/packages/types/src/codec/Struct.ts#L17-L20

jacogr: Ok, well, it _used_ to, doesn't seem to work. Need to apply a fix there.

kwingram25:
Simplifies matters!

jacogr (dismissed):
This needs to match the Rust definitions. 

(a) SCALE is order-dependent
(b) Not having it aligned (in cases where maybe SCALE is not used), it is a maintenance nightmare

jacogr:
I've compared it against my branch (your pre-revert version) and it seems sane in what it does and is nicely tested as well, which helps a lot. So the only issue is really the type ordering, which is remaining here.

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -0,0 +1,173 @@
...
+      version: 'Text'
+    },
+    InkContractSource: {
+      compiler: 'Text',
+      hash: '[u8; 32]',
+      language: 'Text'
+    },
+    InkProject: {
+      contract: 'InkContractContract',
+      metadata_version: 'Text',
```
jacogr, line 166: This one appears in the JSON files as camelCase as well?

packages/types/src/interfaces/contractsAbi/definitions.ts
ascjones: https://github.com/paritytech/ink/pull/514

packages/types/src/interfaces/contractsAbi/definitions.ts
ascjones: Should be fixed in latest master of `cargo-contract`

ascjones:
Mentioned in a comment already, but here is a small `ink!` PR to fix the remaining snake cases https://github.com/paritytech/ink/pull/514

jacogr:
@ascjones Thank you. The API was actually supposed to handle it transparently as well, and it did at some point. (Went missing, was actually re-added as a result of this - so all good, it uncovered a small issue anyway)

copy-to-apps.sh
```diff
@@ -4,7 +4,7 @@
 
 function copy_folder () {
   SRC="packages/$1/build"
-  DST="../apps/node_modules/@polkadot/$1"
+  DST="../canvas-ui/node_modules/@polkadot/$1"
```
jacogr, line 7: ?

packages/types/src/codec/Enum.ts
```diff
@@ -64,7 +64,11 @@ function decodeFromJSON (registry: Registry, def: TypesDef, key: string, value?:
 
   assert(index !== -1, `Cannot map Enum JSON, unable to find '${key}' in ${keys.join(', ')}`);
 
-  return createFromValue(registry, def, index, value);
+  try {
+    return createFromValue(registry, def, index, value);
+  } catch (error) {
+    throw new Error(`Enum(${key}):: ${(error as Error).message}`);
+  }
```
jacogr, line 71: It is fine in general since it is included in https://github.com/polkadot-js/api/pull/1662, however not sure why it made in in here since this was reverted to be the types.

packages/types/src/codec/utils/encodeTypes.ts
```diff
@@ -106,7 +106,7 @@ function encodeTuple (typeDef: Pick<TypeDef, any>): string {
   return `(${
     sub
       // eslint-disable-next-line @typescript-eslint/no-use-before-define
-      .map((type: TypeDef): string => encodeType(type))
+      .map((type: TypeDef): string => displayType(type))
```
jacogr, line 109: Is this intended? This has got an impact on more than just contracts and actually is now inconsistent with all the rest, see for instance Struct which uses the sub-encoding, which does `encodeType`.

packages/types/src/create/types.ts
```diff
@@ -34,6 +34,10 @@ export interface TypeDefExtEnumDiscriminant {
   discriminant: number;
 }
 
+export interface TypeDefExtStructAlias {
```
jacogr, line 37: As above, looks fine since it is in https://github.com/polkadot-js/api/pull/1662, but not sure how it made it back with the type-on after the revert?

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -0,0 +1,174 @@
...
+const layout = {
+  InkLayoutKey: '[u8; 32]',
+  InkStorageLayout: {
+    _enum: {
+      Cell: 'InkLayoutCell',
+      Hash: 'InkLayoutHash',
+      Array: 'InkLayoutArray',
+      Struct: 'InkLayoutStruct',
+      Enum: 'InkLayoutEnum'
+    }
```
jacogr, line 14: Nice, got these in. Can you just make the top-level keys alphabetical, since it will create conflicts with the impl. branch as-is. (It needs to be able to be checked against what is there, atm cannot quite do it to evaluate)

jacogr: i.e. `InkLayoutKey` after `InLayoutCell` (and obviously `InkStorageLayout` much later).

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -0,0 +1,174 @@
...
+  InkDiscriminant: 'u32',
+  InkLayoutEnum: {
+    dispatchKey: 'InkLayoutKey',
+    variants: 'BTreeMap<InkDiscriminant, InkLayoutStruct>'
+  }
+};
+
+const spec = {
+  InkPath: 'Vec<Text>',
+  InkDisplayName: 'InkPath',
```
jacogr, line 55: As above, alphabetical since it will create unneeded conflicts with the impl branch. (These are always tricky since (a) needs to be done manually & (b) no checking since the sorting is switched off)

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -0,0 +1,174 @@
...
+    params: 'Vec<MtLookupTypeId>',
+    path: 'Vec<Text>'
+  },
+  MtTypeDef: {
+    _enum: {
+      Array: 'MtTypeDefArray',
+      Composite: 'MtTypeDefComposite',
+      Primitive: 'MtTypeDefPrimitive',
+      Sequence: 'MtTypeDefSequence',
+      Slice: 'MtTypeDefSlice',
```
jacogr, line 116: There is no slice.

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -0,0 +1,174 @@
...
+      version: 'Text'
+    },
+    InkContractSource: {
+      compiler: 'Text',
+      hash: '[u8; 32]',
+      language: 'Text'
+    },
+    InkProject: {
+      contract: 'InkContractContract',
+      metadata_version: 'Text',
```
jacogr, line 167: This is still incorrect although marked as resolved, see https://github.com/polkadot-js/api/pull/1662/files#diff-db27842186eb038cb3e29484bc7e5888R147-R156

packages/types/src/interfaces/contracts/definitions.ts
```diff
@@ -90,7 +90,7 @@ export default {
     },
     ContractExecResultSuccess: {
       flags: 'u32',
-      data: 'Vec<u8>',
+      data: 'Bytes',
```
jacogr, line 93: Can be used interchangeably, so ok.

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -0,0 +1,174 @@
...
+  MtTypeDefPrimitive: {
+    // this enum definition is mapped in api-contracts/inkTypes.ts
+    _enum: ['Bool', 'Char', 'Str', 'U8', 'U16', 'U32', 'U64', 'U128', 'I8', 'I16', 'I32', 'I64', 'I128']
+  },
+  MtTypeDefSequence: {
+    type: 'MtLookupTypeId'
+  },
+  MtTypeDefSlice: {
+    type: 'MtLookupTypeId'
+  },
```
jacogr, line 137: As above, slice became sequence.

jacogr (changes requested):
Some mis-matches and things that snuck in to be addressed.

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -0,0 +1,174 @@
...
+  InkDiscriminant: 'u32',
+  InkLayoutEnum: {
+    dispatchKey: 'InkLayoutKey',
+    variants: 'BTreeMap<InkDiscriminant, InkLayoutStruct>'
+  }
+};
+
+const spec = {
+  InkPath: 'Vec<Text>',
```
jacogr, line 54: I believe this is `MtPath` (from scale-info). Not sure the `Mt` is self-descriptive as to where it comes from, but well, since I caused that issue (seemed like a good idea at the time), would just keep, move, rename the item.

jacogr (approved):
Thank you, seems sane compared to the version I extracted.

PR #2694, opened 2020-10-08
Handle Struct decoding from snake_case

jacogr:
As per https://github.com/polkadot-js/api/pull/2693#discussion_r501629874


packages/types/src/codec/Struct.spec.ts
```diff
@@ -99,6 +99,16 @@ describe('Struct', (): void => {
     expect(s.toString()).toEqual('{"txt":"fubar","foo":0,"bar":0}');
   });
 
+  it.only('decodes from a snake_case input', (): void => {
```
jacogr, line 102:
```suggestion
  it('decodes from a snake_case input', (): void => {
```

PR #2705, opened 2020-10-11
ContractRegistry always has own registry instance

jacogr:


packages/api-contract/src/ContractRegistry.ts
```diff
@@ -100,6 +100,8 @@ export default class ContractRegistry {
       ),
       ...typeDef
     });
+
+    return created;
```
jacogr, line 104:
```suggestion
```

packages/api-contract/src/ContractRegistry.ts
```diff
@@ -84,8 +85,7 @@ export default class ContractRegistry {
     }
 
     const displayName = path.pop()?.toString();
-
-    return withTypeString({
+    const created = withTypeString({
```
jacogr, line 88:
```suggestion

    return withTypeString({
```

PR #2716, opened 2020-10-12
Adjusts the priority of type override

aniiantt:
https://github.com/polkadot-js/apps/pull/3857

### DIFF
--- a/packages/types-known/src/index.ts
+++ b/packages/types-known/src/index.ts
@@ -51,10 +51,10 @@ export function getSpecTypes ({ knownTypes }: Registry, chainName: Text | string
   return {
     ...filterVersions(typesSpec[_specName], _specVersion),
     ...filterVersions(typesChain[_chainName], _specVersion),
-    ...filterVersions(knownTypes.typesBundle?.spec?.[_specName]?.types, _specVersion),
-    ...filterVersions(knownTypes.typesBundle?.chain?.[_chainName]?.types, _specVersion),
     ...(knownTypes.typesSpec?.[_specName] || {}),
-    ...(knownTypes.typesChain?.[_chainName] || {})
+    ...(knownTypes.typesChain?.[_chainName] || {}),
+    ...filterVersions(knownTypes.typesBundle?.spec?.[_specName]?.types, _specVersion),
+    ...filterVersions(knownTypes.typesBundle?.chain?.[_chainName]?.types, _specVersion)
   };
 }
 


jacogr:
The bundle types are handled the same as the typeSpec, i.e. they are meant to be the same as those, although user-specified so applies after. After this the intent is that the specified chain and specified spec will override those. So the intent of the original is correct as imagined.

Why doesn't it apply in your case?

aniiantt:
> The bundle types are handled the same as the typeSpec, i.e. they are meant to be the same as those, although user-specified so applies after. After this the intent is that the specified chain and specified spec will override those. So the intent of the original is correct as imagined.
> 
> Why doesn't it apply in your case?


For example, In typeSpec, I specify Acala types, but at the same time I want to specify the types of older versions of Acala via SpecVersion. I expect the following code to work:
```javascript
new ApiPromise({
  typesSpec: {
    TokenSymbol: {
      _enum: ["ACA", "AUSD", "DOT", "XBTC", "LDOT", "RENBTC"],
    },
    CurrencyId: {
      _enum: {
        Token: "TokenSymbol",
        DEXShare: "(TokenSymbol, TokenSymbol)",
      },
    },
  },
  typesBundle: {
    spec: {
      acala: [
        {
          minmax: [0, 10],
          types: {
            CurrencyId: {
              _enum: ["ACA", "AUSD", "DOT", "XBTC", "LDOT", "RENBTC"],
            },
          },
        },
      ],
    },
  },
});
```
I want to be able to use a different CurrencyId type if specVersion is less than 10. Now because typesSpec overrides typesBundle, it doesn't work.

jacogr:
So in the above, what you need to do -

```js
new ApiPromise({
  typesSpec: {
    acala: {
      TokenSymbol: {
        _enum: ["ACA", "AUSD", "DOT", "XBTC", "LDOT", "RENBTC"],
      },
    },
  },
  typesBundle: {
    spec: {
      acala: [
        {
          minmax: [0, 10],
          types: {
            CurrencyId: {
              _enum: ["ACA", "AUSD", "DOT", "XBTC", "LDOT", "RENBTC"],
            },
          },
        },
       {
         minmax: [11, undefined,
         types: {
           CurrencyId: {
             _enum: {
              Token: "TokenSymbol",
              DEXShare: "(TokenSymbol, TokenSymbol)",
            },
           }
         }  
       }
      ],
    },
  },
});
```

So when using the bundles, always specify the ranges.

aniiantt:
I see. So in the real world, I might write it this way.  It's not intuitive, but it works.

```javascript
// Lots of type definitions
const acaTypes = {
  // The definition of the current currencyid
  CurrencyId: {
    _enum: {
     Token: "TokenSymbol",
     DEXShare: "(TokenSymbol, TokenSymbol)",
   },
  }
  ...
  ...
  ...
}

const { CurrencyId, ...restTypes } = acaTypes;

const api = new ApiPromise({
  typesSpec: restTypes,
  typesBundle: {
    spec: {
      acala: [
        {
          minmax: [0, 10],
          types: {
            CurrencyId: {
              _enum: ["ACA", "AUSD", "DOT", "XBTC", "LDOT", "RENBTC"],
            },
          },
        },
        {
          minmax: [11, undefined],
          types: {
            CurrencyId,
          },
        },
      ],
    },
  },
});

```

jacogr:
Indeed.

I'm not against the override here, but really trying to figure out what to do with all these `types*` injections since it has gotten way out of hand. There seem to be 20 options now. And your use has just indicated how confusing it is - in my mind as done, it was either of the options, not both. But what you did (as adjusted) does work and does make sense, as I linked even just using one becomes difficult. But still ... way too many options, ways and interactions here.

If it does get cleaned it will be a clean break in a major version. But it will def. move towards the versioned-types only.

Issue #2469, opened 2020-08-04
Get storage entries for a particular block


xlc:
I would like to query historical account balances for all accounts at a particular block, something like

```
const accounts = await api.query.system.account.at(blockHash).entries()
```

But looks like not possible with current polkadot.js.

One way is add `blockHash` to `.entries()` and pass it to `getKeysPaged`

jacogr:
Same would apply to `.keys` (as well as manual paging). Since these both take optional params, the `.at` pattern probably works better (and it consistent)

PR #2746, opened 2020-10-19
Add keysAt & entriesAt

jacogr:
Closes https://github.com/polkadot-js/api/issues/2469

~~Especially for entriesAt, this PR leaves huge gaping hole - the metadata/types at that point is wrong. (Needs to be fixed here or before the next stable release, i.e. before it gets out in the wild)~~ This is handled, `isHistoric` is set on the RPCs


packages/api/src/base/Decorate.ts
```diff
@@ -359,15 +359,23 @@ export default abstract class Decorate<ApiType extends ApiTypes> extends Events
     if (creator.iterKey && (creator.meta.type.isMap || creator.meta.type.isDoubleMap)) {
       decorated.entries = decorateMethod(
         memo(this.#instanceId, (doubleMapArg?: Arg): Observable<[StorageKey, Codec][]> =>
-          this._retrieveMapEntries(creator, doubleMapArg)));
+          this._retrieveMapEntries(creator, null, doubleMapArg)));
+
+      decorated.entries = decorateMethod(
```
jacogr, line 364:
```suggestion
      decorated.entriesAt = decorateMethod(
```

PR #2752, opened 2020-10-20
Solang requires u256 and i256 types

seanyoung:
Tested with latest solang (current master).

Signed-off-by: Sean Young <sean@mess.org>

### DIFF
--- a/packages/types/src/interfaces/scaleInfo/definitions.ts
+++ b/packages/types/src/interfaces/scaleInfo/definitions.ts
@@ -41,7 +41,7 @@ export default {
       variants: 'Vec<SiVariant>'
     },
     SiTypeDefPrimitive: {
-      _enum: ['Bool', 'Char', 'Str', 'U8', 'U16', 'U32', 'U64', 'U128', 'I8', 'I16', 'I32', 'I64', 'I128']
+      _enum: ['Bool', 'Char', 'Str', 'U8', 'U16', 'U32', 'U64', 'U128', 'U256', 'I8', 'I16', 'I32', 'I64', 'I128', 'I256']
     },
     SiTypeDefSequence: {
       type: 'SiLookupTypeId'


packages/types/src/interfaces/scaleInfo/definitions.ts
```diff
@@ -41,7 +41,7 @@ export default {
       variants: 'Vec<SiVariant>'
     },
     SiTypeDefPrimitive: {
-      _enum: ['Bool', 'Char', 'Str', 'U8', 'U16', 'U32', 'U64', 'U128', 'I8', 'I16', 'I32', 'I64', 'I128']
+      _enum: ['Bool', 'Char', 'Str', 'U8', 'U16', 'U32', 'U64', 'U128', 'U256', 'I8', 'I16', 'I32', 'I64', 'I128', 'I256']
```
jacogr, line 44: Yes, I was actually wondering about this - I'm assuming the same changes are made on the scale-info side?

seanyoung: If you mean https://crates.io/crates/scale-info then yes, @Robbepop said so.

Unfortunately there are bunch of changes needed to that crate for solang, but that's another matter.

jacogr: The above needs to match with https://github.com/paritytech/scale-info/blob/master/src/ty/mod.rs#L166-L193

seanyoung: So yes, I'll make a PR for scale-info

jacogr (commented):
Just worried about the mismatch with scale-info (generally it is advisable to add enum entries at the end so it doesn't change behavior, but this _should_ be fine atm... hopefully since we are in a text-like json format)

Could you please supply me with an ABU with this (and WASM is possible), would love to add it as a testcase so I can make sure solang tests are run...

jacogr (approved):
Thank you very much for this! 

(I just double-checked the ABIs to ensure it is indeed string, so middle-order additions  won't break anything)

Will push through to the apps UI as soon as CI is done.

PR #2777, opened 2020-10-24
Add `inputData` when call bare_call runtime API

thashimoto1998:
This PR support input_data arg of bare_call. 
bare_call -> line 636~650 
(https://github.com/paritytech/substrate/blob/master/frame/contracts/src/lib.rs)

### DIFF
--- a/packages/api-contract/src/base/Contract.ts
+++ b/packages/api-contract/src/base/Contract.ts
@@ -98,11 +98,21 @@ export default class Contract<ApiType extends ApiTypes> extends Base<ApiType> {
     };
   }
 
+  #inputData = (messageOrId: AbiMessage | string | number, params: CodecArg[]): Uint8Array => {
+    const message = this.abi.findMessage(messageOrId);
+
+    return encodeMessage(this.registry, message, params);
+  }
+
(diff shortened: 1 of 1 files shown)


jacogr:
The contracts wrapper should ensure you don't need this, i.e. you should not need to deal with raw encoding . If a need does exists, the encodeMessage is exposed.

thashimoto1998:
``
 If a need does exists, the encodeMessage is exposed.
``
Yes, I use bare_call API so please expose encodeMessage.

jacogr:
So in 2.3.1 you can import `encodeMessage` from `@polkadot/api-contract/util` - however that is bound to change since there is a branch to add a `toU8a(...)` & `fromU8a(...)` to the events and message interfaces on the Abi itself.

There is no difference at all between `contract.tx.<messageName>` and `api.tx.contract.call(...)`, both return exactly the same SubmittableExtrinsic. (Although the former will also gain contract events as a decoded event, not there yet, but a reason for the `from/to` changes mentioned above)

thashimoto1998:
I got it. Thanks a lot!

PR #2786, opened 2020-10-26
fix(typegen): dedupe similar types

monitz87:


jacogr (approved):
As always, thank you very much!

PR #2821, opened 2020-11-06
Adjust max allowed call gas (5 * 10 ^ 12)

jacogr:
Part of https://github.com/polkadot-js/apps/issues/3979

### DIFF
--- a/packages/api-contract/src/base/Contract.ts
+++ b/packages/api-contract/src/base/Contract.ts
@@ -18,6 +18,7 @@ import Abi from '../Abi';
 import { applyOnEvent, formatData } from '../util';
 import Base from './Base';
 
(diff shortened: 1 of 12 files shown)


packages/api-contract/src/base/Contract.ts
```diff
@@ -18,6 +18,7 @@ import Abi from '../Abi';
 import { applyOnEvent, formatData } from '../util';
 import Base from './Base';
 
+const MAX_CALL_GAS = new BN(5_000_000_000_000).subn(1);
```
jacogr, line 21:
```suggestion
// As per Rust, 5 * GAS_PER_SEC
const MAX_CALL_GAS = new BN(5_000_000_000_000).subn(1);
```

PR #2846, opened 2020-11-13
Fix encodeEnum, add ABI enum test

kwingram25:


packages/api-contract/test/contracts/ink/enum.json
```diff
@@ -0,0 +1,528 @@
+{
```
jacogr, line 1: Duplicated, see `user/enumExample.*`

jacogr:
It is output-wise equivalent to the current master version which works. 

I do think the approach of JSON.stringify here is better than the fix that just added "'s. Having said that, will probably replace this one with one where there are actually tests with enums with nested Vec & enums, to ensure it works. (This is what the stringify buys us, but it is not checked at all, so it may revert at some point, so would like the ensure the correct version is in)

PR #2876, opened 2020-11-27
Allow typegen for custom RPCs

holygits:
Not sure how to actually execute this from this project 😓 
tried `.bin/polkadot-types-from-chain` but has some babel complaints

---
edit:
`yarn build` 🤦‍♀️

Not sure if I need to make use of metadata or not

### DIFF
--- a/packages/typegen/src/generate/rpc.ts
+++ b/packages/typegen/src/generate/rpc.ts
@@ -3,6 +3,7 @@
 
 import Handlebars from 'handlebars';
 
+import { Metadata } from '@polkadot/metadata';
 import staticData from '@polkadot/metadata/static';
 import { TypeRegistry } from '@polkadot/types/create';
 import { Definitions } from '@polkadot/types/types';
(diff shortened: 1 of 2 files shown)


packages/typegen/src/generate/rpc.ts
```diff
@@ -111,9 +113,17 @@ export function generateRpcTypes (importDefinitions: { [importPath: string]: Rec
   });
 }
 
-export function generateDefaultRpc (): void {
+export function generateDefaultRpc (dest = 'packages/api/src/augment/rpc.ts', metadata = staticData, extraTypes: Record<string, Record<string, { types: Record<string, any> }>> = {}): void {
+  const registry = new TypeRegistry();
+
+  registerDefinitions(registry, extraTypes);
+
+  registry.setMetadata(new Metadata(registry, metadata));
```
jacogr, line 121: As suggested, no, metadata is not needed as per the original implementation here. RPCs are not communicated over metadata.

jacogr:
Generally happy to merge, would just like the unused/unneeded Metadata reverted. Apart from that, 100% sane.

jacogr (approved):
Thank you for this. Really appreciate it.

PR #2895, opened 2020-12-02
Update parachain types

jacogr:
As per https://github.com/paritytech/cumulus/pull/252#discussion_r533959842

### DIFF
--- a/packages/types/src/interfaces/parachains/definitions.ts
+++ b/packages/types/src/interfaces/parachains/definitions.ts
@@ -42,6 +42,11 @@ export default {
       validatorIndices: 'BitVec'
     },
     AuctionIndex: 'u32',
+    BackedCandidate: {
+      candidate: 'CommittedCandidateReceipt',
+      validityVotes: 'Vec<ValidityAttestation>',
+      validatorIndices: 'BitVec'
+    },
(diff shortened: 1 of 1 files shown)


packages/types/src/interfaces/parachains/definitions.ts
```diff
@@ -42,32 +42,51 @@ export default {
...
       newValidationCode: 'Option<ValidationCode>',
-      processedDownwardMessages: 'u32'
+      headData: 'HeadData',
+      processedDownwardMessages: 'u32',
+      hrmpWatermark: 'BlockNumber'
     },
-    CandidateReceipt: {
-      parachainIndex: 'ParaId',
+    CandidateDescriptor: {
+      paraId: 'u32',
```
JoshOrndorff, line 65: Should this be `paraId: 'ParaId',?

JoshOrndorff: And same thing below in HrmpChannelId

jacogr: I saw that, copied it as supplied - will do a proper once-over once it actually lands properly to align 100% with Polkadot.

Issue #2930, opened 2020-12-08
Support derive injection via typesBundle (as RPC)


jacogr:
- In the same vein as https://github.com/polkadot-js/api/pull/2927
- See background at https://github.com/polkadot-js/apps/pull/4171#issuecomment-740948438

pr0fedt:
Would this suffice or should I also expose function like _getSpecDerives_ in _packages/types-known_ package?

jacogr:
Don't believe we need to expose the `getSpecDerives` at all - I think what you have there fits the bill.

jacogr:
Closed in #2932 - thank you very much!

PR #2932, opened 2020-12-09
Types bundle derive

pr0fedt:
Support derive injection via typesBundle
https://github.com/polkadot-js/api/issues/2930


jacogr (approved):
lgtm. Thanks an absolute million for this.

PR #2952, opened 2020-12-14
Don't log full error

dvdplm:
In `txwrapper` we can't always know if a payload is a signing payload or a signed tx so we first try one and then the other. When the former fails users get a scary-looking error. If upstream logs the error there's little we can do about it lest we [monkey patch `console.error`](https://github.com/paritytech/txwrapper-core/pull/22/files#diff-ea2d1c5845dd0e2fb9814e98a3464c2c63a43922983391e98fe9c20c328e979cR55) which is nasty af.

Hence this PR that removes the log here so that downstream can choose what to do with the error message.

rel. https://github.com/paritytech/txwrapper-core/issues/20


jacogr:
Not quite convinced. Have been trying to remove the logging in all the specific codec parts and only rely on this as a single point of all type-decoding logs. As examples - Call used to have specific logging on failures, as did Struct, etc. 

Will have to see what the impact is.

jacogr:
This will work and solve the need,

`createTypeUnsafe<T extends Codec = Codec, K extends string = string> (registry: Registry, type: K, params: any[] = [], isPedantic?: boolean): T`

changes to

`createTypeUnsafe<T extends Codec = Codec, K extends string = string> (registry: Registry, type: K, params: any[] = [], { isPedantic, withoutLog }: UnsafeOptions = {}): T`

Obviously the `l.error` line becomes `!withoutLog && l.error(...)`

Then call `createTypeUnsafe` directly, the only difference is really the params compared to `createType`, it is passed as an explicit array.

Would be happy with an adjustment like that that solves the need, without breaking the existing behavior.

jacogr (changes requested):
As suggested, not quite convinced as-is, but if the commented changes are applied, would be good since it -

- will allow the specific raised issue to be addressed
- not change the API logging to have nothing (since it is localised in the single place now ... well, there are some places left which has not been caught, Rome was not built in a day)

dvdplm:
Cool, thank you. Will proceed as indicated.

packages/api-contract/src/util.ts
```diff
@@ -17,7 +17,7 @@ type ContractEvents = 'CodeStored' | 'ContractExecution' | 'Instantiated';
 type TOptions = BlueprintOptions | ContractOptions;
 
 export function formatData (registry: Registry, data: Raw, { type }: TypeDef): Codec {
-  return createTypeUnsafe(registry, type, [data], true);
+  return createTypeUnsafe(registry, type, [data], { isPedantic: true, withoutLog: false });
```
jacogr, line 20:
```suggestion
  return createTypeUnsafe(registry, type, [data], { isPedantic: true });
```

packages/rpc-core/src/index.ts
```diff
@@ -468,7 +468,7 @@ export class RpcCore implements RpcInterface {
 
       if (!isEmpty) {
         try {
-          inner = createTypeUnsafe(registry, type, [input], true);
+          inner = createTypeUnsafe(registry, type, [input], { isPedantic: true, withoutLog: false });
```
jacogr, line 471:
```suggestion
          inner = createTypeUnsafe(registry, type, [input], { isPedantic: true });
```

packages/rpc-core/src/index.ts
```diff
@@ -484,7 +484,7 @@ export class RpcCore implements RpcInterface {
             ? hexToU8a(meta.fallback.toHex())
             : undefined
           : input
-      ], true);
+      ], { isPedantic: true, withoutLog: false });
```
jacogr, line 487:
```suggestion
      ], { isPedantic: true });
```

packages/types/src/create/createType.ts
```diff
@@ -59,13 +64,13 @@ function initType<T extends Codec = Codec, K extends string = string> (registry:
 // argument here can be any string, which, when it cannot parse, will yield a
 // runtime error.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export function createTypeUnsafe<T extends Codec = Codec, K extends string = string> (registry: Registry, type: K, params: any[] = [], isPedantic?: boolean): T {
+export function createTypeUnsafe<T extends Codec = Codec, K extends string = string> (registry: Registry, type: K, params: any[] = [], { isPedantic, withoutLog }: UnsafeOptions = { }): T {
```
jacogr, line 67:
```suggestion
export function createTypeUnsafe<T extends Codec = Codec, K extends string = string> (registry: Registry, type: K, params: any[] = [], { isPedantic, withoutLog }: UnsafeOptions = {}): T {
```

jacogr (commented):
Thank you, looks great and hopefully solves your issue.

Some small comments just to not pass through the `withoutLog` each time. Otherwise, does what is says on the tin.

jacogr (approved):
Brilliant, thank you.

PR #2959, opened 2020-12-16
Adjust rxjs imports to cjs-only

### LINKED ISSUE #2957: @polkadot/x-rxjs doesn't work with Node LTS from .mjs
It does work on 15.x, but not LTS. (The same named-module issues that is was fixing on 15.x, occurs on the older now)

jacogr:
Closes #2957

This is horrible, but until rxjs supports Node mjs, nothing much that can be done. And they _do_ have esm, it just doesn't play nicely together.


jacogr:
Also a dead-end.

Issue #2904, opened 2020-12-02
@polkadot/metadata package has undeclared dependency to @polkadot/api


APshenkin:
https://github.com/polkadot-js/api/blob/master/packages/metadata/src/decorate/types.ts#L4

This dependency is missing in @polkadot/metadata package.json.

Also if you take a look on export from @polkadot/api 
https://github.com/polkadot-js/api/blob/master/packages/api/src/types/consts.ts

You can see that there is no `QueryableConsts` dependency.

This dependency can be found here https://github.com/polkadot-js/api/blob/master/packages/api/src/augment/consts.ts#L429 that redeclare `@polkadot/api/types/consts` module (what??? is that legal?? :))

And all works only if you `import @polkadot/api` in your application.

jacogr:
Yes, it is legal - it is used for augmenting the API with user-defined types. So basically when you do that it stacks them on-top of each other. Quite  weird TS thing, but actually comes in quite helpful.

As for this specific circular dep, yes it is known, have not yet found a way to consistently (without mess/fuss) remove metadata <-> api. (There are a handful of those left at this point)

PR #2971, opened 2020-12-21
Adjust for last remaining circular deps

jacogr:
Closes https://github.com/polkadot-js/api/issues/2904

`yarn polkadot-dev-circular` picks up none remaining, still needs a check against rollup (generally found that slightly better)


packages/typegen/scripts/polkadot-types-internal-interfaces.js
```diff
@@ -14,6 +14,7 @@ try {
     plugins: [
       ['module-resolver', {
         alias: {
+          '^@polkadot/metadata/package-info.json': './packages/metadata/package.json',
```
jacogr, line 17:
```suggestion
```

PR #2976, opened 2020-12-22
Add frontier RPCs (& types for them)

jacogr:


jacogr:
There are still some issues with the serialization over RPC, this is an attempt to align on the frontier layer - https://github.com/paritytech/frontier/pull/254

Issue #2965, opened 2020-12-18
Events support in @polkadot/typegen


tomusdrw:
AFAICT there is no support for events typing generation in `@polkadot/typegen`:
https://polkadot.js.org/docs/api/examples/promise/typegen/
while it seems quite useful.

jacogr:
Overall there is no generation for events at all - the Event type is only a generic codec, so even for Polkadot/Kusama any use of the events is to cast them. (It only gets generated from metadata)

jacogr:
So, how would you expect this to work? Interested since this has been in the back of my mind for some time. So basically, code-wise,

```js
// returns Vec<EventRecord>
const records = await api.query.system.events();

// here Even is a generic Event, data is a Tuple - types can only be applies when the
// compiler actually knows exactly what is in the storage (unlike tx/rpc/storage where
// we know up-front based on the actual call made)
records.forEach(({ phase, event: { data } }) => {
 ...
});
```

So TS just cannot apply typing since it just doesn't know what is inside the `Vec<EventRecord>`. This is horrible, since for instance in balance,transfer you would need to manually recast `data` to `ITuple<[AccountId, Balance]>` when using. So the generation help that coupld be applied would be something like -

```js
// generated based on metadata
type EventDataBalancesTransfer = ITuple<[AccountId, Balance]>;

// use still needs to cast, but at least there is an alias...
const [accountId, balance] = event.data as EventDataBalancesTransfer;

// where this is what is done currently
const [accountId, balance] = event.data as  ITuple<[AccountId, Balance]>;
```

savudani8:
I was able to generate event types for the parachain metadata using the code in PR https://github.com/polkadot-js/api/pull/2981.

I didn't need to specify custom event types, as they were already part of the `EventMetadataV12` property of the Metadata https://github.com/polkadot-js/api/blob/333313fd6abf2820f00a555b46edb108e63a3848/packages/types/src/interfaces/metadata/types.ts#L233

The result can be observed here:
![Screenshot from 2020-12-23 16-02-11](https://user-images.githubusercontent.com/23065004/103015417-4e5af100-4538-11eb-9f64-8fa8318eee5c.png)

jacogr:
Yes, the generation is not the issue, that is really straight forward. 

My questions are around the use. Since it is not able to be auto-applied, so it is still being casted. I am guessing that applying and casting as „generated“ removes some errors, eg. When the Tuple changes for 3 to 2.

[...]

PR #2981, opened 2020-12-23
Auto-generate event types

savudani8:
PR for issue https://github.com/polkadot-js/api/issues/2965

### DIFF
--- a/packages/typegen/src/generate/events.ts
+++ b/packages/typegen/src/generate/events.ts
@@ -40,7 +40,7 @@ function generateForMeta (meta: Metadata, dest: string, extraTypes: Record<strin
 
             return {
               docs: documentation,
-              name: stringCamelCase(name),
+              name: name,
               type: args.toString()
             };
           });
--- a/packages/typegen/src/generate/events.ts
+++ b/packages/typegen/src/generate/events.ts
@@ -1,0 +1,87 @@
+// Copyright 2017-2020 @polkadot/typegen authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import Handlebars from 'handlebars';
+
+import staticData from '@polkadot/metadata/static';
+import { Metadata } from '@polkadot/metadata';
+import { TypeRegistry } from '@polkadot/types/create';
+import * as defaultDefs from '@polkadot/types/interfaces/definitions';
+import { stringCamelCase } from '@polkadot/util';
+
+import { createImports, compareName, formatType, readTemplate, registerDefinitions, setImports, writeFile } from '../util';
+
+const template = readTemplate('events');
+const generateForMetaTemplate = Handlebars.compile(template);
+
+/** @internal */
+function generateForMeta (meta: Metadata, dest: string, extraTypes: Record<string, Record<string, { types: Record<string, any> }>>, isStrict: boolean): void {
+  writeFile(dest, (): string => {
+    const allTypes: Record<string, Record<string, { types: Record<string, unknown> }>> = { '@polkadot/types/interfaces': defaultDefs, ...extraTypes };
+    const imports = createImports(allTypes);
+    const allDefs = Object.entries(allTypes).reduce((defs, [path, obj]) => {
+      return Object.entries(obj).reduce((defs, [key, value]) => ({ ...defs, [`${path}/${key}`]: value }), defs);
+    }, {});
+
+    const modules = meta.asLatest.modules
+      .sort(compareName)
+      .filter((mod) => mod.events.isSome)
+      .map((mod) => {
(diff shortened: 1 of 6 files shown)


packages/typegen/src/generate/events.ts
```diff
@@ -0,0 +1,87 @@
...
+
+        const items = events
+          .sort(compareName)
+          .map(({ args, documentation, name }) => {
+            const argTypes = args.map((type) => formatType(allDefs, type.toString(), imports))
+            setImports(allDefs, imports, argTypes);
+
+            return {
+              docs: documentation,
+              name: stringCamelCase(name),
```
jacogr, line 43: Since there are types-only, I would keep it to non-camelCase, i.e. most TS types are generally `SomeType` vs `someType`.

packages/typegen/src/generate/events.ts
```diff
@@ -0,0 +1,87 @@
...
+              type: args.toString()
+            };
+          });
+
+        return {
+          items,
+          name: stringCamelCase(name)
```
jacogr, line 50: +1 This one is fine, we make all module names `camelCase` everywhere.

jacogr (commented):
I love these PRs that solves somebody's bugbear. Thanks a lot for this. 

I have some questions on use (in the linked-issue), but overall quite happy with what is provided here. (Only one small comment on type casing)

jacogr:
Since these cannot be augmented, I probably would not expose it in the augmented space. We can probably place it somewhere else and then do -

```js
import type { BalancesTransferEvent } from '@polkadot/api/events`;

...
const events = await api.query.system.events();

events.forEach(({ method, section, event }) => {
  if (section === 'balances' && method === 'transfer')) {
    const [accountId, value] = event.data as BalancesTransferEvent;
    ...
  }
});
```

... or something similar

jacogr (commented):
Ok, so at this point, the PR while no breaking anything, also doesn't actually help the API along. Let me explain - since the API has no exposed events interfaces on `api.event.*` (which the augmentation is generally used for), nothing her is actually being used. Additionally, as already explained, the data that comes back from system.events is only known at runtime.

However there are 2 scenarios where this type of generation (in a slightly different form), can add value.

Scenario 1:

As explained earlier, instead of generating a structure like above (which is difficult to use as types), we generate `<Section><Method>Event` types into e.g. `augmented/events` and those can be applied manually via `as`. It has some benefit.

Scenario 2:

Here we actually introduce `api.event.*` and it acts as a TypeScript type guard, asserting the type. (This actually makes it useful in pure JS as well) So basically, looking at a `balances.Transfer` event, the augmentation would look something like when in use -

```js
const events = await api.query.system.events();

events.forEach((record) => {
  // check for transfer
  if (api.events.balances.transfer(record)) {
    // record here is casted as a Transfer
    // Since asserted (see the augmentation definition), it means this is valid types, aka
    // in TS from/to === AccountId and value === Balance
    const [from, to, value] = record.event.data;
  } else if (api.events.system.newAccount(record)) {
    // record here is nicely casted as a NewAccount
  }
});
```

So it basically solves 2 issues -

1. We actually works with API interfaces everywhere instead of matching on strings
2. Via TS type guards the data has assigned types in side the use of it

To pull this together, the following in needed. For the `augment/event` interfaces, it would look something like -

```js
declare module '@polkadot/api/types/events' {
  export interface AugmentedEvents<ApiType> {
    balances: {
      transfer: (record: EventRecord) => record is IEventRecord<ITuple<[AccountId, Balance]>>;
      ...
    }
    ...
```

On the API base itself, like query/consts/tx/derives are decorated, each metadata entry for events also adds decoration, the implementation itself is very straight-forward -

```js
// generic asserting, non-augmented - the generated augmentation makes it specific
decorated = (record: EventRecord): record is IEventRecord<GenericEventData> =>
  record.section === sectionName &&& record.method === methodName;
```

So then the missing pieces are just the `IEventRecord` type which is be be implemented by `types/codec/generic/Event` in both Event. The base interfaces just follows what is there already -

```js
interface IEventRecord<T extends ITuple> {
  phase: Phase;
  event: IEvent<T>;
  method: string;
  section: string;
  topics: Vec<Hash>;
}

interface IEvent<T extends ITuple> {
  index: EventId;
  data: T & {
    meta: EventMetadataLatest;
    method: string;
    section: string;
  }
}
``

savudani8:
Thanks for the comprehensive response @jacogr ! I’ll look into it in January

jacogr:
No issues, if you do have a gap and feel up to it. Shout if you are stuck.

I think scenario 2 can actually solve your "string matching issue" (so it helps both JS & TS users) and have everything typed on use after the type guard. So seems the most comprehensive (I have been wrong before).

PR #2983, opened 2020-12-25
fix MetadataVersioned.asV12

yjhmelody:


jacogr (approved):
Good catch. Thank you.

PR #2993, opened 2020-12-28
add client version and sh3 decorators for web3

### LINKED ISSUE #2989: Add decorators for web3_clientVersion, web3_sha3
Instantiating the api fires `API/INIT: RPC methods not decorated: web3_clientVersion, web3_sha3` for me.

You can assign this to me if you want

joelamouche:
closes #2989

### DIFF
--- a/packages/api/src/augment/rpc.ts
+++ b/packages/api/src/augment/rpc.ts
@@ -479,5 +479,15 @@ declare module '@polkadot/rpc-core/types.jsonrpc' {
        **/
       version: AugmentedRpc<() => Observable<Text>>;
     };
+    web3: {
+      /**
+       * Returns current client version.
+       **/
+      clientVersion: AugmentedRpc<() => Observable<string>>;
+      /**
+       * Returns sha3 of the given data
+       **/
+      sha3: AugmentedRpc<(_:Bytes) => Observable<H256>>;
+    };
   }
 }
--- a/packages/api/src/augment/rpc.ts
+++ b/packages/api/src/augment/rpc.ts
@@ -153,6 +153,10 @@ declare module '@polkadot/rpc-core/types.jsonrpc' {
        **/
       chainId: AugmentedRpc<() => Observable<U64>>;
       /**
+       * Returns current client version.
+       **/
+      clientVersion: AugmentedRpc<() => Observable<Text>>;
+      /**
        * Returns block author.
        **/
       coinbase: AugmentedRpc<() => Observable<H160>>;
@@ -268,6 +272,10 @@ declare module '@polkadot/rpc-core/types.jsonrpc' {
        * Sends transaction; will block waiting for signer to return the transaction hash
        **/
       sendTransaction: AugmentedRpc<(tx: EthTransactionRequest | { from?: any; to?: any; gasPrice?: any; gas?: any; value?: any; data?: any; nonce?: any } | string | Uint8Array) => Observable<H256>>;
+      /**
+       * Returns sha3 of the given data
+       **/
+      sha3: AugmentedRpc<(_: Bytes | string | Uint8Array) => Observable<H256>>;
       /**
        * Used for submitting mining hashrate.
        **/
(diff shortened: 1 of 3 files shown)


packages/api/src/augment/rpc.ts
```diff
@@ -479,5 +479,15 @@ declare module '@polkadot/rpc-core/types.jsonrpc' {
        **/
       version: AugmentedRpc<() => Observable<Text>>;
     };
+    web3: {
```
jacogr, line 482: This is auto-generated.

The area to add them is here - https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/eth/rpc.ts

Would follow the net example with an `aliasSection` (keeping all the frontier RPCs in a single place). Once `yarn build:interfaces` is run (as on CI), the above is created.

packages/rpc-core/src/index.ts
```diff
@@ -122,6 +122,8 @@ export class RpcCore implements RpcInterface {
 
   public readonly system!: RpcInterface['system'];
 
+  public readonly web3!: RpcInterface['web3'];
```
jacogr, line 125: This should not be needed when adding to the above location, only when a new top-level interface folder is created. (Part of the reason net is included in the eth structure above with an explicit alias - apart from keeping things together consistently)

joelamouche:
updated. Thanks for the help :)

packages/types/src/interfaces/eth/rpc.ts
```diff
@@ -26,9 +26,24 @@ const netRpc: DefinitionsRpc = {
...
+  clientVersion: {
+    aliasSection: 'web3',
+    description: 'Returns current client version.',
+    params: [],
+    type: 'String'
+  },
+  sha3: {
+    aliasSection: 'web3',
+    description: 'Returns sha3 of the given data',
+    params: [{ name: '_', type: 'Bytes' }],
```
jacogr, line 39:
```suggestion
    params: [{ name: 'data', type: 'Bytes' }],
```

packages/api/src/augment/rpc.ts
```diff
@@ -152,6 +152,10 @@ declare module '@polkadot/rpc-core/types.jsonrpc' {
        * Returns the chain ID used for transaction signing at the current best block. None is returned if not available.
        **/
       chainId: AugmentedRpc<() => Observable<U64>>;
+      /**
+       * Returns current client version.
+       **/
+      clientVersion: AugmentedRpc<() => Observable<Text>>;
```
jacogr, line 158: Had a slight bug here affecting the TS types, this is fixed in https://github.com/polkadot-js/api/pull/2994 so it will end up in the correct web3 section.

packages/types/src/interfaces/eth/rpc.ts
jacogr: Just prefer this not being unnamed, since where we use TS (like in VSCode) it is nice when it shows the actual field in addition to the type. This also pulls through to things like the apps UI where these definitions are pulled-in, i.e. it allows you to see some descriptive name for the field 

(I've seen in frontier it has been slightly "lazy" where RPC params are unnamed, at least in the outer layers)

joelamouche: I agree with you. I was just afraid it could somehow break code

packages/api/src/augment/rpc.ts
joelamouche: ha ok cool. yes I was thinking it was weird net and web3 ended up in the eth section

packages/rpc-core/src/index.ts
jacogr: Ok... I was wrong in this... Now that I fixed the generation elsewhere, this is indeed needed. Sorry about that.

jacogr (approved):

jacogr:
Thank you!

packages/rpc-core/src/index.ts
joelamouche: np ^^

PR #3022, opened 2021-01-05
fix(typegen): support vecs of tuples in enums

monitz87:
Enums with vectors of tuples as members weren't being processed properly by the typegen

This definition:

```
MyEnum: {
  _enum: {
    Something: 'Vec<(A, B)>'
  }
}
````

Gets converted to:

```
/** @name MyEnum */
export interface MyEnum extends Enum {
  readonly isSomething: Vec<(A, B)>;
}
```

Which isn't valid typescript


jacogr:
This looks good. However will try and find a test case for it in the existing interfaces to run through. (There is nothing doubt a couple of them)

jacogr (approved):
lgtm.

PR #3030, opened 2021-01-07
Fix EthTransaction to ethereum crate

jacogr:
As per https://github.com/polkadot-js/apps/issues/4335

For some weird reason the fields were alphabetically sorted.

However, not the issue. The issue is that the type comes from a non-frontier crate, externally. After adjusting -

image


joelamouche:
Does the order matter?
I remember that the linter enforced that

jacogr:
Yes, the order matters - it is a binary encoding, so fields needs to be in the exact order. (So all interfaces have a "ignore sorting" eslint header)

PR #3033, opened 2021-01-08
Add API option for chain-specific signedExtensions

jacogr:
... not quite sure where the issue for this went.


packages/types/src/extrinsic/signedExtensions/types.ts
```diff
@@ -6,8 +6,8 @@ import type { InterfaceTypes } from '../../types';
 export type ExtTypes = Record<string, keyof InterfaceTypes>;
 
 export type ExtInfo = {
-  extra: ExtTypes;
-  types: ExtTypes;
+  extrinsic: ExtTypes;
```
jacogr, line 9: Renamed since it _may_ be confusing for chain devs without proper context.

PR #3047, opened 2021-01-13
toBigInt() uses primitive type `bigint`

### LINKED ISSUE #3045: use bigint instead of BigInt
BigInt is constructor of primitive type bigint, like Number() to number.

Adding two BigInt variables will lead to a compiler error: `TS2365: Operator '+' cannot be applied to types 'BigInt' and 'BigInt'.`

Therefore, toBigInt() should return `bigint`

ianhe8x:
Fixes #3045

### DIFF
--- a/packages/types/src/codec/AbstractInt.ts
+++ b/packages/types/src/codec/AbstractInt.ts
@@ -146,7 +146,7 @@ export abstract class AbstractInt extends BN implements Codec {
   /**
    * @description Returns a BigInt representation of the number
    */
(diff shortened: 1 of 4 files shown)


jacogr (approved):
Thank you.

Will add some tests after this,

```
it('has a working toBigInt', (): void => {
    expect(
      new UInt(registry, 4567).toBigInt() + BigInt(1)
    ).toEqual(BigInt(468));
  });
  ```

PR #3054, opened 2021-01-14
feat(types): add correct tuple type to storage key

monitz87:


jacogr:
Thank you... this is a very, very, very old oversight. Could you just look at the linting issues as well, there are some areas where the assertion is now not needed (as it should be).

monitz87:
Sure thing, I'll look into it

jacogr (approved):
Brilliant, thank you!

Issue #3057, opened 2021-01-15
Bounties derive -> treasury/council could not be there either


jacogr:
This line - https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/bounties/bounties.ts#L40

This is reported - https://github.com/polkadot-js/apps/pull/4419#issue-555385805

cc @ekowalsk

jacogr:
Could just leave as-is and fix on the UI layer as well, i.e. don't try when council is not available. (BTW to test, connect to the Acala PC2 testchain under Rococo, and the issue shows up)

jacogr:
Closing, addressed.

PR #3062, opened 2021-01-15
Handle no council for bounties proposals

ekowalsk:
This is a partial fix of #3057, since we also need to fix usage of api.consts.treasury from apps

### DIFF
--- a/packages/api-derive/src/bounties/bounties.ts
+++ b/packages/api-derive/src/bounties/bounties.ts
@@ -42,7 +42,7 @@ export function bounties (instanceId: string, api: ApiInterfaceRx): () => Observ
       switchMap(() =>
         combineLatest([
           bountyBase.bounties.keys(),
-          api.derive.council.proposals()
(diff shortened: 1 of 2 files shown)


packages/api-derive/src/bounties/bounties.ts
```diff
@@ -37,7 +37,7 @@ export function bounties (instanceId: string, api: ApiInterfaceRx): () => Observ
   return memo(instanceId, (): Observable<DeriveBounties> =>
     combineLatest([
       bountyBase.bountyCount<BountyIndex>(),
-      api.query.council.proposalCount<ProposalIndex>()
+      api.query.council ? api.query.council.proposalCount<ProposalIndex>() : of(0)
```
jacogr, line 40: I believe line 45 also needs adjustment? (It won't be decorated if no council is present)

jacogr: api.derive.council ? api.derive.council.proposals() : of([])

jacogr (approved):
Thank you!

PR #3082, opened 2021-01-20
Modify extractAuthor to support Moonbeam

joelamouche:
This modification will allow displaying the author of a block for a parachain/blockchain that uses Consensus digest with an h160 address for the author.

@jacogr let me know if that covers all the other parachains

### DIFF
--- a/packages/api-derive/src/type/util.ts
+++ b/packages/api-derive/src/type/util.ts
@@ -18,7 +18,7 @@ export function extractAuthor (digest: Digest, sessionValidators: AccountId[] = 
       }
     });
   } else {
-    // Aura and Babe use PreRunti;e digest now
+    // Aura and Babe use PreRuntime digest now
     const [pitem] = digest.logs.filter(({ type }) => type === 'PreRuntime');
 
     if (pitem) {
--- a/packages/api-derive/src/type/util.ts
+++ b/packages/api-derive/src/type/util.ts
@@ -4,23 +4,29 @@
 import type { AccountId, Digest } from '@polkadot/types/interfaces';
 
 export function extractAuthor (digest: Digest, sessionValidators: AccountId[] = []): AccountId | undefined {
-  const [pitem] = digest.logs.filter(({ type }) => type === 'PreRuntime');
+  let author;
+  const citems = digest.logs.filter(({ type }) => type === 'Consensus');
 
-  // extract from the substrate 2.0 PreRuntime digest
-  if (pitem) {
-    const [engine, data] = pitem.asPreRuntime;
+  if (citems.length > 0) {
+    citems.forEach((citem) => {
+      // extract author from the consensus (substrate 1.0, digest)
+      const [, data] = citem.asConsensus;
 
-    return engine.extractAuthor(data, sessionValidators);
+      if (data.length === 20) {
+        // This is used by Moonbeam with an h160 address for the author
+        author = data.toString();
+      }
+    });
   } else {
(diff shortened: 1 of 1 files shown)


packages/api-derive/src/type/util.ts
```diff
@@ -4,23 +4,29 @@
 import type { AccountId, Digest } from '@polkadot/types/interfaces';
 
 export function extractAuthor (digest: Digest, sessionValidators: AccountId[] = []): AccountId | undefined {
-  const [pitem] = digest.logs.filter(({ type }) => type === 'PreRuntime');
+  let author;
```
jacogr, line 7: Not sure I would swap the order around - the most-used is `PreRuntime` (and most recent), so it should go first.

Additionally we have lost the `return engine.extractAuthor(data, sessionValidators);` from the citems now. In reality the specific log for Moonbeam should probably be in there?

joelamouche: >Not sure I would swap the order around - the most-used is PreRuntime (and most recent), so it should go first

If you look at the logs in https://github.com/PureStake/moonbeam/pull/159#discussion_r560418535 you'll see that Moonbeam also has a PreRuntime log but it doesn't work. So we should first check for Consensus before we look at PreRuntime.

>Additionally we have lost the return engine.extractAuthor(data, sessionValidators); from the citems now. In reality the specific log for Moonbeam should probably be in there?

`return engine.extractAuthor(data, sessionValidators)` doesn't work for Moonbeam. Since all other uses of Consensus have been removed, I assumed it would be simple to just drop it. I can reestablish it for non-h160 addresses if you think it's useful

JoshOrndorff: > Moonbeam also has a PreRuntime log

That's only true when you're talking about the standalone node with Aura. The parachain node (the primary moonbeam product) does not have a PreRuntime digest, although it may in the future). For example, look at the live testnet.

image

jacogr: For instance, by just returning here we have actually broken Kulupu which also uses the Consensus digest.

The actual extraction needs to happen here - https://github.com/polkadot-js/api/blob/master/packages/types/src/generic/ConsensusEngineId.ts

(Aura, Babe & Kulupu POW is done there)

jacogr: TL;DR this logic needs to be kept in-tact, it will determine the type of digest and defer to the `ConsensusEngineId` for extracting.

joelamouche: But if we keep this logic, this won't work for the standalone, which would have a PreRuntime log, but where the relevant information is in the Consensus log

jacogr: The logic is as follow, directl from the source, commented here ...

```js
  // try to find any PreRuntime digests here  
  const [pitem] = digest.logs.filter(({ type }) => type === 'PreRuntime');

  if (pitem) {
   // if we have some PreRuntime logs, follow this path
    ...
  } else {
    // we didn't have any of the previous type, now try to find Consensus logs
    const [citem] = digest.logs.filter(({ type }) => type === 'Consensus');
    ...
  }
```

So if doesn't require PreRuntime, it is one of the two.

jacogr: I'm happy to swap the oder above, as long as we don't try and make the extraction here as well.

jacogr:
Replaced by #3108

PR #3108, opened 2021-01-25
Adapted to display moonbeam author

joelamouche:
replacement of https://github.com/polkadot-js/api/pull/3082/files

- swap Consensus-preRuntime check
- add non-aura non-babe author extraction

### DIFF
--- a/packages/api-derive/src/type/util.ts
+++ b/packages/api-derive/src/type/util.ts
@@ -4,19 +4,19 @@
 import type { AccountId, Digest } from '@polkadot/types/interfaces';
 
 export function extractAuthor (digest: Digest, sessionValidators: AccountId[] = []): AccountId | undefined {
-  const [pitem] = digest.logs.filter(({ type }) => type === 'PreRuntime');
+  const [citem] = digest.logs.filter(({ type }) => type === 'Consensus');
 
   // extract from the substrate 2.0 PreRuntime digest
-  if (pitem) {
-    const [engine, data] = pitem.asPreRuntime;
+  if (citem) {
+    const [engine, data] = citem.asConsensus;
 
     return engine.extractAuthor(data, sessionValidators);
   } else {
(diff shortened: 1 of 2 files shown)


packages/api-derive/src/type/util.ts
```diff
@@ -4,19 +4,19 @@
 import type { AccountId, Digest } from '@polkadot/types/interfaces';
 
 export function extractAuthor (digest: Digest, sessionValidators: AccountId[] = []): AccountId | undefined {
-  const [pitem] = digest.logs.filter(({ type }) => type === 'PreRuntime');
+  const [citem] = digest.logs.filter(({ type }) => type === 'Consensus');
 
   // extract from the substrate 2.0 PreRuntime digest
-  if (pitem) {
-    const [engine, data] = pitem.asPreRuntime;
+  if (citem) {
```
jacogr, line 10: Small nitpick, we probably want to drop the comments on the previous line and on line 17 now.

packages/types/src/generic/ConsensusEngineId.ts
```diff
@@ -104,6 +108,11 @@ export class GenericConsensusEngineId extends u32 {
       return this._getPowAuthor(bytes);
     }
 
+    // Moonbeam is neither Aura nor Babe nor Pow and uses h160 addresses
+    if (bytes.length === 20) {
+      return this._getH160Author(bytes);
```
jacogr, line 113: +1

joelamouche:
Is there a way to check codeclimate conditions with the cli before pushing?

jacogr:
Ignore CC, it really is a guide rather than a requirement. In this case can break it down into move pieces but the code would then actually lose maintainability since you can follow it 100% atm. In a lot of cases it does help, in some cases (like this), the alternative is worse.

jacogr (approved):
Thank you!

PR #3113, opened 2021-01-27
Show prior locks from democracy.locks

Tbaut:
I'd need some hand holding here.
- Making many of the `DeriveDemocracyLock` possibly undefined creates less mess than I thought cf some small changes in apps https://github.com/polkadot-js/apps/pull/4516
- this early return `if (prior.length)` is certainly not the right way to do. I'm blocked with the rxjs syntax. `concat` seems to be deprecated, what should I use to add the potential result of what's in `prior` to potentially existing `votes[]`

### DIFF
--- a/packages/api-derive/src/democracy/locks.ts
+++ b/packages/api-derive/src/democracy/locks.ts
@@ -46,16 +46,23 @@ function delegateLocks (api: ApiInterfaceRx, { balance, conviction, target }: Vo
         referendumId,
         unlockAt: unlockAt.isZero()
           ? unlockAt
-          : referendumEnd.add(api.consts.democracy.enactmentPeriod.muln(LOCKUPS[conviction.index])),
-        vote: api.registry.createType('Vote', { aye: vote.isAye, conviction })
+          : referendumEnd?.add(api.consts.democracy.enactmentPeriod.muln(LOCKUPS[conviction.index])) || new BN(0),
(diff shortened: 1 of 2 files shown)


packages/api-derive/src/democracy/locks.ts
```diff
@@ -46,18 +46,25 @@ function delegateLocks (api: ApiInterfaceRx, { balance, conviction, target }: Vo
...
 }
 
-function directLocks (api: ApiInterfaceRx, { votes }: VotingDirect): Observable<DeriveDemocracyLock[]> {
-  if (!votes.length) {
+function directLocks (api: ApiInterfaceRx, { prior, votes }: VotingDirect): Observable<DeriveDemocracyLock[]> {
+  if (!votes.length && !prior.length) {
     return of([]);
   }
 
+  if (prior.length) {
```
jacogr, line 61: I finally have a testcase for this - months later.

The above is not quite correct. So when unlocked the direct vote is there in all cases. So this path would always be followed.

- when a proxy is removed, it has non-zero values
- with just normal clearing, it has 0/0 values (both balance and unlockAt is 0)

So in the case with a proxy removed, it looks like -

image

And in the case of stuff just cleared, it looks like

image

jacogr:
Going to close this one. Really appreciate it and was up and down in the last 2 days on "should this be here and be treated like the others". For this specifica case, I actually just changed the way things are displayed UI-wise, mostly due to the issues listed.

So this derive stays as "votes only locks".

PR #3127, opened 2021-01-29
Prevent logging errors when making an http connection

Tbaut:
When connecting with a HttpProvider, I got those logs. This PR removes the errors.

```
2021-01-29 13:03:54        API/INIT: Api will be available in a limited mode since the provider does not support subscriptions
2021-01-29 13:03:54        API-HTTP: HTTP Provider does not have 'on' emitters, use WebSockets instead
2021-01-29 13:03:54        API-HTTP: HTTP Provider does not have 'on' emitters, use WebSockets instead
2021-01-29 13:03:54        API-HTTP: HTTP Provider does not have 'on' emitters, use WebSockets instead
```


jacogr (approved):
Thank you.

PR #3148, opened 2021-02-04
fix userExtensions type

ntduan:
The type of `userExtensions` should be `ExtDef` instead of `Record<string, ExtDef>`. Because the type of ExtDef is already `Record<string, ExtInfo>`. 
The reason why typescript doesn't report an error is because the type of allExtensions in this case is `Record<string, ExtInfo>`. So the type of `allExtensions[key]` is `ExtInfo`, but it should actually be `ExtInfo | undefined`.
image

And I think `userExtensions `should be in front of `allExtensions`?

### DIFF
--- a/packages/api/src/types/index.ts
+++ b/packages/api/src/types/index.ts
@@ -60,7 +60,7 @@ export interface ApiOptions extends RegisteredTypes {
   /**
    * @description Any chain-specific signed extensions that are now well-known
    */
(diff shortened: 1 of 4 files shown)


jacogr:
Tank you. the answers are "yes" and "yes" - it has ben included as-is in https://github.com/polkadot-js/api/pull/3142 as well. (Happy to merge this as a subset)

jacogr (approved):
Thank you.

PR #3150, opened 2021-02-05
Add eth filter RPC endpoints and FilterChanges type.

jnaviask:
Adds support for changes in this PR: https://github.com/paritytech/frontier/pull/274.

Does not add FilterPool-related types as they're not used in RPC, but we could add those in the future if needed.


jacogr (approved):
Thanks an absolute million for this. There may be tweaks, but as you said, rather have something that you can tweak based on usage, thank nothing at all.

PR #3193, opened 2021-02-21
better support of C-like enum

### LINKED ISSUE #3191: Better support of C-like enum
Let's say I have this enum defined in Rust
```
enum Test {
  A = 1,
  B = 10,
  C = 255
}
```

I need a better way to define it in polkadot.js. Make an array with 256 items with bunch dummies is not a good idea.

Maybe we can support syntax like

```
Test: {
  _enum: {
    A: 1,
    B: 10,
    C: 255
  }
}
```

xlc:
Closes #3191

Not the most efficient solution but should be good enough.

### DIFF
--- a/packages/types/src/codec/Enum.spec.ts
+++ b/packages/types/src/codec/Enum.spec.ts
@@ -394,4 +394,31 @@ describe('Enum', (): void => {
       );
     });
   });
+
+  describe('C-like enum', (): void => {
+    it('handles C-like enum', (): void => {
+      const Test = Enum.with({
+        A: 0,
+        B: 1,
+        C: 10,
+        D: 255
+      });
+
+      const testA = new Test(registry, 'A');
+
+      expect(testA.toNumber()).toEqual(0);
+
+      const testB = new Test(registry, 'B');
+
+      expect(testB.toNumber()).toEqual(1);
+
+      const testC = new Test(registry, 'C');
+
+      expect(testC.toNumber()).toEqual(10);
+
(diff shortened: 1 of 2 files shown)


jacogr:
Have no issues, but especially for your case, the one in https://github.com/polkadot-js/api/pull/3194 would be cleaner. (Your is certainly acceptable, i.e. it does basically automate the manual steps that have been applied in 3-4 places)

Having said that... your is more intrusive and requires much less effort for initial testing, where the other one would need a bit more burn-in. So basically would suggest going with your solution and then looking a bit more at the other one to remove the inefficiencies here.

jacogr:
Ok, dug through a bit - the missing part is the part, so as it stands it will actually create some issues. It is more-or-less these parts - https://github.com/polkadot-js/api/pull/3194/files#diff-bbf81c73859be56ad50f2b62127235662f7da3842624728a8a58ebe71fdde5cbR90

Without this the actual definition parsing and typegen does fail, e.g. this change doesn't actually work - https://github.com/polkadot-js/api/pull/3194/files#diff-bbf81c73859be56ad50f2b62127235662f7da3842624728a8a58ebe71fdde5cbR90

Probably will close this and go straight for the other version. Will run it through additional tests tomorrow, but thus far it actually does seem to do the trick.

jacogr:
@xlc Will try to merge the other one tomorrow - was meant to be today, but ran out of time with releases.

xlc:
Thanks. In that case I will close this one.

PR #3197, opened 2021-02-21
clean ws-provider's states on disconnected

### LINKED ISSUE #3179: Promise won't resolve nor reject after disconnection
I'm using ApiPromise here, it might be also a problem for ApiRx, though I haven't tested there.
To reproduce the issue, just do a slow api query (e.g rpc.chain.getBlock that contains runtime upgrade),
and drop the connection before it resolves. Then even after the connection is recovered, the previous promise is still hanging there, never be resolved or rejected.

ianhe8x:
Fixes #3179

### DIFF
--- a/packages/rpc-provider/src/ws/index.ts
+++ b/packages/rpc-provider/src/ws/index.ts
@@ -42,6 +42,13 @@ const RETRY_DELAY = 1000;
 
 const l = logger('api-ws');
 
+function eraseRecord<T> (record: Record<string, T>, cb?: (item: T) => void): void {
+  for (const key of Object.keys(record)) {
+    if (cb) cb(record[key]);
+    delete record[key];
+  }
+}
+
 /**
  * # @polkadot/rpc-provider/ws
  *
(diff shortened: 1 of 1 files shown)


packages/rpc-provider/src/ws/index.ts
```diff
@@ -42,6 +42,13 @@ const RETRY_DELAY = 1000;
 
 const l = logger('api-ws');
 
+function eraseRecord<T> (record: Record<string, T>, cb?: (item: T) => void): void {
+  for (const key of Object.keys(record)) {
+    if (cb) cb(record[key]);
+    delete record[key];
+  }
+}
```
jacogr, line 50:
```suggestion
  Object.keys(record).forEach((key): void => {
    if (cb) {
      cb(record[key]);
    }
    
    delete record[key];
  });
}
```

Prefer not using `of`

packages/rpc-provider/src/ws/index.ts
```diff
@@ -334,12 +341,18 @@ export class WsProvider implements ProviderInterface {
...
-      l.error(`disconnected from ${this.#endpoints[this.#endpointIndex]}: ${event.code}:: ${event.reason || getWSErrorString(event.code)}`);
+      l.error(error.message);
     }
 
     this.#isConnected = false;
     this.#emit('disconnected');
+    // reject all hanging requests
+    eraseRecord(this.#handlers, (handler) => handler.callback(error, undefined));
+    eraseRecord(this.#waitingForId);
+    eraseRecord(this.#subscriptions);
```
jacogr, line 355: This all depends. If auto-reconnect, you actually expect it not to fail. Subscriptions especially - in that case it should be re-created when re-connected.

The waiting for Id & single-shot is problematic, since they won't receive an id.

ianhe8x: even auto reconnected, we would need to re-fire the requests, will it worth all the efforts and ok that we hide these details from user? 
Also, who knows how long it might takes to reconnect, in the end we have to chose between fast fail or add timeout guards

jacogr: This is a major breaking change - the API since pre-1.0 always re-connects subscriptions. So along with it operating differently, it also means that codes needs to be rewritten to cater for the new behavior.

ianhe8x: Oh, I didn't realize that, will check the code in the morning, this pr is only to prevent hanging promises, don't mean to change the behavior

jacogr: So I believe it is fine for the other 2 cases, i.e. the non-connected subs.

jacogr (approved):
Thank you.

Issue #3202, opened 2021-02-22
API is not able to handle hashers different to H256.


hbulgarini:
Hey , when we try to access to a block information on a chain that contains a H512 hasher type: 
 Network -> Explorer -> Chain Info -> any block

i get  the following error:
```
Unable to retrieve the specified block details. Unable to retrieve header and parent from supplied hash
```

Please take into account that we are adding already the custom type:
`  "BlockHash": "H512"`

Doing a research on the api code we detected that the hash is being hardcoded to H256:
https://github.com/polkadot-js/api/blob/d2ff350db963600fb5256152ef6946856e1bbb96/packages/types/src/types/registry.ts#L112 
https://github.com/polkadot-js/api/blob/d2ff350db963600fb5256152ef6946856e1bbb96/packages/types/src/codec/Struct.ts#L228

Here is the link to our custom chain:
[...]

PR #3206, opened 2021-02-23
Allow for CodecHash override & hasher option

jacogr:
Closes https://github.com/polkadot-js/api/issues/3202

Usage:

```js
import { u8aConcat } from '@polkadot/util';
import { blake2AsU8a, keccakAsU8a } from '@polkadot/util-crypto';
import { ApiPromise } from '@polkadot/api';

// create a custom hasher (512 bits, combo of blake2 and keccak)
function hasher (data) {
  return u8aConcat(
    blake2AsU8a(data),
    keccakAsU8a(data)
  );
}

// create custom api with our hasher as well as type override (H256 normally)
  // custom hasher everywhere
  hasher,
  types: { 
    // our global hash (e.g. Header parentHash returns Hash)
    Hash: 'H512'
  }
});
```

### DIFF
--- a/packages/types/src/interfaces/runtime/definitions.ts
+++ b/packages/types/src/interfaces/runtime/definitions.ts
@@ -43,6 +43,7 @@ export default {
       digestLevels: 'u32'
     },
     ConsensusEngineId: 'GenericConsensusEngineId',
+    CodecHash: 'H256',
     Digest: {
       logs: 'Vec<DigestItem>'
     },
--- a/packages/types/src/interfaces/runtime/definitions.ts
+++ b/packages/types/src/interfaces/runtime/definitions.ts
@@ -43,7 +43,7 @@ export default {
       digestLevels: 'u32'
     },
     ConsensusEngineId: 'GenericConsensusEngineId',
-    CodecHash: 'H256',
+    CodecHash: 'Hash',
     Digest: {
       logs: 'Vec<DigestItem>'
     },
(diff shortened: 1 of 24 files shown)


packages/types/src/interfaces/runtime/definitions.ts
```diff
@@ -43,6 +43,7 @@ export default {
       digestLevels: 'u32'
     },
     ConsensusEngineId: 'GenericConsensusEngineId',
+    CodecHash: 'H256',
     Digest: {
```
hbulgarini, line 47: @jacogr , is this the part of the code that defaults the hasher to H256 in case no custom hasher is specified?

jacogr: If there is no override, it will use that type.

hbulgarini (approved):
The change looks good to me. Is there any way we can include the setHasher in the docs?

jacogr:
Sure. I will eventually get there - docs is not high on the list. It generally gets done when I have some time between the rest of the repos. (And generally only when a specific item’s support costs outweighs the costs of doing the docs)

hbulgarini:
> Sure. I will eventually get there - docs is not high on the list. It generally gets done when I have some time between the rest of the repos. (And generally only when a specific item’s support costs outweighs the costs of doing the docs)

thank you! anyway , support tasks works as documentation in some way :)

PR #3223, opened 2021-02-26
modified api-derive balances to allow use of custom balance storages …

pr0fedt:
…via derives; current implementation only works with standard call signatures

### DIFF
--- a/packages/api-derive/src/balances/account.ts
+++ b/packages/api-derive/src/balances/account.ts
@@ -55,7 +55,13 @@ function queryBalancesFree (api: ApiInterfaceRx, accountId: AccountId): Observab
 }
 
 function queryBalancesAccount (api: ApiInterfaceRx, accountId: AccountId, modules: string[] = ['balances']): Observable<Result> {
-  const balances = modules.map((m): QueryableStorageMultiArg<'rxjs'> => [api.query[m].account, accountId]);
+  const balances = modules.map(
+    (m): QueryableStorageMultiArg<'rxjs'> => [
+      (api.derive[m as 'balances'] as unknown as ApiInterfaceRx['query']['balances'])?.account ?? api.query[m].account,
+      accountId
+    ]
+  );
+
   const extract = (data: AccountData[]) =>
     data.map(({ feeFrozen, free, miscFrozen, reserved }): BalanceResult => [free, reserved, feeFrozen, miscFrozen]);
 
--- a/packages/api-derive/src/balances/all.ts
+++ b/packages/api-derive/src/balances/all.ts
@@ -114,9 +114,13 @@ function queryOld (api: ApiInterfaceRx, accountId: AccountId): Observable<Result
 
 // current (balances, vesting)
 function queryCurrent (api: ApiInterfaceRx, accountId: AccountId, balanceInstances: string[] = ['balances']): Observable<ResultBalance> {
-  const lockCalls = balanceInstances.map((m) => api.query[m].locks);
+  const lockCalls = balanceInstances.map(
+    (m): ApiInterfaceRx['query']['balances']['locks'] | undefined =>
+      (api.derive[m as 'balances'] as unknown as ApiInterfaceRx['query']['balances'])?.locks ?? api.query[m as 'balances']?.locks
+  );
+
   const lockEmpty = lockCalls.map((c) => !c);
-  const lockQueries = lockCalls.filter((c) => c).map((c): QueryableStorageMultiArg<'rxjs'> => [c, accountId]);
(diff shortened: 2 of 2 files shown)


packages/api-derive/src/balances/all.ts
```diff
@@ -114,9 +114,13 @@ function queryOld (api: ApiInterfaceRx, accountId: AccountId): Observable<Result
...
 function queryCurrent (api: ApiInterfaceRx, accountId: AccountId, balanceInstances: string[] = ['balances']): Observable<ResultBalance> {
-  const lockCalls = balanceInstances.map((m) => api.query[m].locks);
+  const lockCalls = balanceInstances.map(
+    (m): ApiInterfaceRx['query']['balances']['locks'] | undefined =>
+      (api.derive[m as 'balances'] as unknown as ApiInterfaceRx['query']['balances'])?.locks ?? api.query[m as 'balances']?.locks
+  );
+
   const lockEmpty = lockCalls.map((c) => !c);
-  const lockQueries = lockCalls.filter((c) => c).map((c): QueryableStorageMultiArg<'rxjs'> => [c, accountId]);
+  const lockQueries = lockCalls.filter((c) => c).map((c): QueryableStorageMultiArg<'rxjs'> => [c!, accountId]);
```
jacogr, line 123: Will need to think about this PR :) 

However comment here - really don't want `<var>!` anywhere in the codebase. At some point they are going to fail. Here there is no actual filter that says "only valid", so it is an assumption.

pr0fedt: There is a filter "(c) => c" for truthy values, so I think "!" operator cannot do much harm

jacogr: Really don't want them in the codebase at all. 

There are some that have been introduced that we need to specifically work around and that I have not managed to get out (see https://github.com/polkadot-js/api/blob/master/.eslintrc.js#L28, although I believe I cleared at leats some of them recently, need to re-check) and I really don't want to add more.

pr0fedt: using type predicate as a filter value seems to have done the trick

jacogr:
Not ignoring, sorry for me delay. Just need to think this through so will circle back tomorrow. 

It looks like an interesting “hack/approach”, so don’t have major issues atm - just don’t want to rush.

jacogr (approved):

packages/api-derive/src/balances/account.ts
```diff
@@ -55,7 +55,13 @@ function queryBalancesFree (api: ApiInterfaceRx, accountId: AccountId): Observab
 }
 
 function queryBalancesAccount (api: ApiInterfaceRx, accountId: AccountId, modules: string[] = ['balances']): Observable<Result> {
-  const balances = modules.map((m): QueryableStorageMultiArg<'rxjs'> => [api.query[m].account, accountId]);
+  const balances = modules.map(
+    (m): QueryableStorageMultiArg<'rxjs'> => [
+      (api.derive[m as 'balances'] as unknown as ApiInterfaceRx['query']['balances'])?.account ?? api.query[m].account,
```
jacogr, line 60: Ummm.... now after merge, I don't see how this works - this points to an existing derive entry, so basically will always have a circular ref to itself. So it will never do the underlying query.

jacogr: ... need more coffee first thing in the morning.

pr0fedt: But does not _api_ object have a reference the instance that called at runtime? That way it could be used to call our custom derives. The trick for multiple currencies in our chain is that they are resolved by a single pallet, that's why I thought that a good idea was to create derives for all our currencies to bind account ballance parameter, eg. `api.derive.btc.account = (accountId) => api.query.balances.account(accountId, "Btc")`

jacogr: The idea is good - for the derive through since there is an exitsing `api.derive.balances.account` it will fallback to that one in all cases. (Since that is the default for chains without any instances, only the `balances` being queried). the derive gets checked first, it is always available and will be used, no queries.

So it is probably just as simple as a rename here, something like `api.derive.balances.customAccount` & `api.derive.balances.customLocks` (aligning these two and ensuring no overrides of existing).

pr0fedt: thanks, will update now

jacogr: Sorry about this - I read, thought, read again and then on merge saw it. So my lack of being awake made this slightly a longer process.

pr0fedt: https://github.com/polkadot-js/api/pull/3248

PR #3248, opened 2021-03-01
Use derive customAccount and customLocks for balance overrides

pr0fedt:


jacogr (approved):
Thank you. Believe it should tick the boxes now.

PR #3261, opened 2021-03-04
Added ReAlis Network types

Daelon02:
Added type for out nft-pallet


jacogr (commented):
The API doesn't ship with chain-specific types - it only has the definitions for the base Substrate (all palettes) and base Polkadot (all palettes).Trying to manage all the chains in one place will yield some interesting challenges since there will be massive conflicts between.

For customisation, see https://polkadot.js.org/docs/api/start/typescript.user

(The above is how other chains publish their types, e.g. Sora, Acala, Edgeware, etc.)

PR #3263, opened 2021-03-05
Make unknownTypes externally accessible

ii-ii-ii:
Hi, I'm working on something like a blockchain browser. I will encounter a lot of types of problems when parsing blocks. I think it would help a lot to debug if I could get unknownTypes. It's like getting knownTypes. 

I know unknownTypes will have logs printed, but we shouldn't depend on logs.

### DIFF
--- a/packages/types/src/create/registry.ts
+++ b/packages/types/src/create/registry.ts
@@ -193,8 +193,8 @@ export class TypeRegistry implements Registry {
     return this.#knownTypes;
   }
 
-  public get unknownTypes (): Map<string, boolean> {
-    return this.#unknownTypes;
+  public get unknownTypes (): string[] {
+    return [...this.#unknownTypes.keys()];
   }
 
   public get signedExtensions (): string[] {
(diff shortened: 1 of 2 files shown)


jacogr (commented):
Happy with this generally. 

We just need to add this to the Registry interface as well, ie. when publicly accessible, it should be on the interface.

packages/types/src/create/registry.ts
```diff
@@ -193,6 +193,10 @@ export class TypeRegistry implements Registry {
     return this.#knownTypes;
   }
 
+  public get unknownTypes (): Map<string, boolean> {
+    return this.#unknownTypes;
```
jacogr, line 197: Think it will make more sense not as Map. It really is an internal representation.

```suggestion
  public get unknownTypes (): string[] {
    return [...this.#unknownTypes.keys()];
```

With the relevant change in the interface as well. The TypeRegistry is bound to change, so really don't want the internal structures to bleed through interface use since that yields breaking changes.

packages/types/src/types/registry.ts
```diff
@@ -98,6 +98,7 @@ export interface Registry {
   readonly chainSS58: number | undefined;
   readonly chainTokens: string[];
   readonly knownTypes: RegisteredTypes;
+  readonly unknownTypes: Map<string, boolean>;
```
jacogr, line 101:
```suggestion
  readonly unknownTypes: string[];
```

As above.

jacogr (approved):
Thank you.

PR #3266, opened 2021-03-05
add missing types for election provider

kianenigma:
https://knowyourmeme.com/memes/but-its-honest-work


kianenigma:
Seems like some scripts (in `typegen`) need to be executed to take my types and generate ts from it. Not sure if that will happen in CI or I should do it.

jacogr:
CI will generate the typescript interfaces and commit them. 

For "Phase", it is a conflicting type name. So in this case it would need an alias, see e.g. https://github.com/polkadot-js/api/blob/master/packages/types-known/src/modules.ts#L61

So something like `Phase: 'ElectionPhase'` in the above location and then inside the definitions as added all occurences renamed from `Phase` to `ElectionPhase`.

jacogr:
Thank you. (Which is basically just a copy from the added phase with the correct alias - will give credit here credit is due in the CHANGELOG)

Replaced by #3280

PR #3347, opened 2021-03-22
Abi constructor takes different parameters

vminkov:
The Abi constructor used to take the registry as a parameter, now it takes registry.getChainProperties() and this change introduced hard to debug behaviour. 

Fixed the README example


jacogr (approved):
Thank you.

Issue #3392, opened 2021-04-13
When the rpc cannot connect, the program will crash.


ntduan:
In nodejs 15, if an UnhandledPromiseRejection exception is thrown, the program will crash.

My Code: 
```javascript
import { ApiPromise, WsProvider } from "@polkadot/api";

new ApiPromise({
  provider: new WsProvider("ws://127.0.0.1:8888"),
});
```

Result: 
```javascript
2021-04-13 17:11:16          API-WS: disconnected from ws://127.0.0.1:8888: 1006:: connection failed
node:internal/process/promises:245
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<_Event>".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```
It looks like this is an internal promise reject that is not being caught. I think there should be a way for the user to catch the error. This will prevent the program from crashing.

mario-purestake:
I was experiencing the same when connecting to the Amber Tesnet because the endpoint seemed to be down and the polkadot API was stuck after trying to init the api instance. Now that it's resolved, I tested on localhost with the node not running and I got the same result.

My code:

```javascript
const { ApiPromise, WsProvider } = require("@polkadot/api");
const { version } = require("./node_modules/@polkadot/api/package.json");

async function main() {
  console.log(`API version: ${version}\n`);

  console.log(`Checking API in localhost...`);

  // Initialise the provider to connect to the local node
  const provider = new WsProvider("wss://localhost:9944");
  console.log(` ✔ Provider OK`);

  const blockFetched = new Promise(async (resolve, reject) => {
    // Create the API and wait until ready
    let api = await ApiPromise.create({ provider });

    api.on("disconnected", () => {
      const error = "API got disconnected";
      reject(error);
    });

    api.on("error", (error) => {
      reject(error);
    });

    console.log(` ✔ API OK`);

    // Get latest block number
    const { number } = await api.rpc.chain.getHeader().catch((error) => {
      reject(error);
    });

    console.log(` ✔ Current block in the localhost network is ${number}`);

    resolve(number);
  });

  await blockFetched.catch((error) => {
    console.log(` ✘ Error on localhost. Error: `, error);
  });
}

// log unhandledRejection
process.on("unhandledRejection", (error) => {
  console.log(error);
});

main()
[...]

PR #3395, opened 2021-04-13
Explicitly handle isReadyOrError (when only using isReady)

jacogr:
Closes https://github.com/polkadot-js/api/issues/3392

### DIFF
--- a/packages/api/src/types/index.ts
+++ b/packages/api/src/types/index.ts
@@ -69,6 +69,10 @@ export interface ApiOptions extends RegisteredTypes {
    * @description The source object to use for runtime information (only used when cloning)
    */
   source?: ApiBase<any>;
+  /**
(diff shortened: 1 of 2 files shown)


packages/api/src/types/index.ts
```diff
@@ -69,6 +69,10 @@ export interface ApiOptions extends RegisteredTypes {
    * @description The source object to use for runtime information (only used when cloning)
    */
   source?: ApiBase<any>;
+  /**
+   * @description Throws an error when the initial connection fails (same as isReadyOrError)
+   */
+  throwOnConnect?: boolean;
```
xlc, line 75: throw on connect sounds like throw error when connected
but I don't have a better name to suggest...

jacogr:
There are no change to the create so use as normal.

PR #3492, opened 2021-05-05
parachains: update HostConfiguration type

andresilva:
https://github.com/paritytech/polkadot/blob/master/runtime/parachains/src/configuration.rs#L37-L175


jacogr (approved):
Thank you.

PR #3511, opened 2021-05-10
Backward compatibility thing

KarishmaBothara:
Support querying historic block where system runtime module exist, however account does not exist in it...


jacogr (approved):
Thank you.

PR #3568, opened 2021-05-25
Changed `FundInfo` definition

arjanz:
Changed `firstSlot` and `lastSlot` to `firstPeriod` and `lastPeriod`, reference:  [Polkadot definition source](https://github.com/paritytech/polkadot/blob/master/runtime/common/src/crowdloan.rs#L141)


jacogr (approved):
Thank you.

PR #3625, opened 2021-06-07
Add Kusama 9040 upgrade block

jacogr:


ntduan:
Does this pr have a big impact?  We are using v4.12.1 now.  We are preparing kusama crowdloan.  If there is no updated version, will it cause type parsing errors?

jacogr:
No, there are no type differences between 9030 and 9040. (This is more an accounting entry, with no types impact)

PR #3639, opened 2021-06-11
Fix typo and add missing field to unique types

hamidra:
fixing a typo and also seems the 'attribute': 'u32' is missing from ClassDetails struct.
https://github.com/paritytech/substrate/blob/41ab01a8cb2a43f7d743778c066ad91453e0c883/frame/uniques/src/types.rs#L54


jacogr (approved):
Thank you, I appreciate it :)

"interesting" typo there...

Issue #3657, opened 2021-06-15
Support Accounts without AccountData


apopiak:
This code assumes `AccountData` to be present in `AccountInfo` if there is a nonce.

https://github.com/polkadot-js/api/blob/f97a6d041827e64c96e6274bc1e2c3ea275fcd5f/packages/api-derive/src/balances/account.ts#L87-L89

This assumption does not hold for chains that set `type AccountData = ();` (e.g. Compound Gateway [here](https://github.com/compound-finance/gateway/blob/0d8b3f4e6bea7fda6c458e9cdce9c0f79e382cc7/runtime/src/lib.rs#L214)).

jacogr:
Closed in https://github.com/polkadot-js/api/pull/3658

PR #3658, opened 2021-06-15
handle empty account data/missing balance gracefully

apopiak:
take a stab at resolving #3657

### DIFF
--- a/packages/api-derive/src/balances/account.ts
+++ b/packages/api-derive/src/balances/account.ts
@@ -24,12 +24,14 @@ type DeriveCustomAccount = ApiInterfaceRx['derive'] & {
 }
 
 function getBalance (api: ApiInterfaceRx, [freeBalance, reservedBalance, frozenFee, frozenMisc]: BalanceResult): DeriveBalancesAccountData {
+  const votingBalance = api.registry.createType('Balance', freeBalance ? freeBalance.toBn() : 0);
+
   return {
     freeBalance,
     frozenFee,
     frozenMisc,
     reservedBalance,
-    votingBalance: api.registry.createType('Balance', freeBalance.toBn())
+    votingBalance
   };
 }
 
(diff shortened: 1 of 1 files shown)


jacogr:
Thank you.! 

On initial glance, it seems sane. Will take a proper look in the morning.

packages/api-derive/src/balances/account.ts
```diff
@@ -84,14 +86,22 @@ function queryCurrent (api: ApiInterfaceRx, accountId: AccountId): Observable<Re
...
         ? (infoOrTuple as AccountInfo).data
         : (infoOrTuple as [Index, AccountData])[1];
 
-      return [
-        (infoOrTuple as AccountInfo).nonce || (infoOrTuple as [Index, AccountData])[0],
-        [[free, reserved, feeFrozen, miscFrozen]]
-      ];
+      const nonce = (infoOrTuple as AccountInfo).nonce || (infoOrTuple as [Index, AccountData])[0];
+
+      if (data) {
```
jacogr, line 95: Not 100% sure about this part... with `AccountData = ()` it basically is still an object, in this case the `Null` object since it has not data.

What does work in this case is `if (!data.isEmpty) { /* handle normal case */ } else { /* everything set to zero */ }`

jacogr, line 95:
```suggestion
      if (data && !data.isEmpty) {
```

Handle both -

- data is set to something like `()`
- structure does not contain data

packages/api-derive/src/balances/account.ts
```diff
@@ -23,13 +23,19 @@ type DeriveCustomAccount = ApiInterfaceRx['derive'] & {
   }
 }
 
+function zero (api: ApiInterfaceRx) {
+  return api.registry.createType('Balance');
+}
+
 function getBalance (api: ApiInterfaceRx, [freeBalance, reservedBalance, frozenFee, frozenMisc]: BalanceResult): DeriveBalancesAccountData {
+  const votingBalance = api.registry.createType('Balance', freeBalance.toBn());
+
```
jacogr, line 32:
```suggestion
```

Just reverting this. It is certainly a bit cleared, but at the same time since it is not used anywhere, guess it doesn't need to create a variable. (Although the optimizing compiler should take care of it...)

(Goes with the suggestion just below)

packages/api-derive/src/balances/account.ts
```diff
@@ -23,13 +23,19 @@ type DeriveCustomAccount = ApiInterfaceRx['derive'] & {
...
   return {
     freeBalance,
     frozenFee,
     frozenMisc,
     reservedBalance,
-    votingBalance: api.registry.createType('Balance', freeBalance.toBn())
+    votingBalance
```
jacogr, line 38:
```suggestion
    votingBalance: api.registry.createType('Balance', freeBalance.toBn())
```

As part of the change commented on above...

jacogr (commented):
One small comment (neither here nor there), but otherwise lgtm. (Not tested, but don't seems completely sane)

jacogr (approved):
Thank you.

PR #3669, opened 2021-06-17
Do not rely on balances pallet or particular type for AccountInfo

jflatow:


packages/api-derive/src/tx/signingInfo.ts
```diff
@@ -19,9 +19,13 @@ interface Result {
 }
 
 function latestNonce (api: ApiInterfaceRx, address: string): Observable<Index> {
-  return api.derive.balances.account(address).pipe(
```
jacogr, line 22: This cannot be dropped - it means others are not available.

kylezs:
Right now I'm encountering an issue that this PR seems to be looking to solve. There's an assumption made by this api client that the chain must use the `balances` pallet. The same assumption is being made by [`substrate-subxt`](https://github.com/paritytech/substrate-subxt), which is what brought me here.

PR #3694, opened 2021-06-25
fixes #3693 Fix uniques return types

### LINKED ISSUE #3693: Uniques InstanceDetails is missing InstanceMetadatas
If I am understanding this correctly then there's a mistake here where Class is using `instanceMetadatas` instead of `classMetadatas`, same with destroy it's using `instanceMetadatas`, and lastly `instanceDetails` missing `instanceMetadatas` alltogether

https://github.com/polkadot-js/api/blob/dfc546955fdbfb657246ae0b88be1bb06c0a383c/packages/types/src/interfaces/uniques/types.ts#L44

if my assumption is correct then let me know and I can create a PR

-----------------

On a similar note, I am having trouble quering class metadata, the only way I found is this

```
const nft = await polkadotApi.query.uniques.class('test');
const s = polkadotApi.registry.createType(
  'ClassMetadata',
  nft.unwrap().instanceMetadatas.toHex(),
);
console.log(s.toJSON());
```

Surely there's a better way?

And also even tho I did setClassMetadata, this still returns 0x

image

Yuripetusko:


packages/types/src/interfaces/uniques/definitions.ts
```diff
@@ -37,14 +38,12 @@ export default {
     },
     ClassMetadata: {
       deposit: 'DepositBalance',
-      name: 'Vec<u8>',
-      information: 'Vec<u8>',
+      data: 'Vec<u8>',
```
jacogr, line 41: These should fix what you are seeing with the wrong decoding...

Issue #3697, opened 2021-06-25
Polkadot-api fails to read historical Kusama blocks


dzhelezov:
This is a duplicate of  https://github.com/polkadot-js/apps/issues/5685 feel free to close either.

I am trying to use the latest polkadot-api (4.15.1) to access historical Kusama blocks, down to the genesis. But I get decoding errors trying to read e.g. block 0x2df84d4c6bb8441f7a1702b4589dc33f8dc43a0794a4df446c74110636456989

```
2021-06-25 16:39:40             VEC: Unable to decode on index 2 createType(ExtrinsicV4):: createType(Call):: Call: failed decoding imOnline.heartbeat:: Struct: failed on args: {"heartbeat":"Heartbeat","_signature":"Signature"}:: decodeU8a: failed at 0x8cd6df15d379c0c2… on _signature: H512:: Expected at least 64 bytes (512 bits), found 60 bytes
2021-06-25 16:39:40        RPC-CORE: getBlock(hash?: BlockHash): SignedBlock:: createType(SignedBlock):: Struct: failed on block: {"header":"Header","extrinsics":"Vec<Extrinsic>"}:: Struct: failed on extrinsics: Vec<Extrinsic>:: createType(ExtrinsicV4):: createType(Call):: Call: failed decoding imOnline.heartbeat:: Struct: failed on args: {"heartbeat":"Heartbeat","_signature":"Signature"}:: decodeU8a: failed at 0x8cd6df15d379c0c2… on _signature: H512:: Expected at least 64 bytes (512 bits), found 60 bytes
```
I understand that this is due to some legacy types incompatibility, but cannot find a spec bundle or other type def info for Kusama, Polkadot, and other chains. My goal is to read the whole block history.

jacogr:
Please don't log duplicates - there is one of me and multiple repos that I'm looking at and working on, so this really just add lots of noise and actually make me ignore them).

dzhelezov:
Ok closed the other one

PR #3698, opened 2021-06-25
Add polkadot 9050 upgrade block

jacogr:
Also adds missing HeartbeatTo244 type for historic Kusama blocks


jacogr:
Closes https://github.com/polkadot-js/api/issues/3697

Issue #3687, opened 2021-06-23
Type generation from `v14` metadata with `scale-info` types


ascjones:
https://github.com/paritytech/substrate/pull/8615 introduces a new version of substrate metadata which includes `scale-info` type information. 

Here is an example of the json metadata generated from a `polkadot` dev instance from this branch https://github.com/paritytech/polkadot/pull/3336.

There is already some code for contracts https://github.com/polkadot-js/api/blob/master/packages/api-contract/src/MetaRegistry.ts which generates types from this metadata so this could be a starting point.

It would be good to get some early feedback on the format/contents of this new metadata while it is still easy to change. The `v14` metadata structure can be seen here: https://github.com/paritytech/frame-metadata/blob/main/frame-metadata/src/v14.rs.

PR #3759, opened 2021-07-12
Parsing for metadata v14

jacogr:
As per https://github.com/paritytech/substrate/pull/8615 
Closes https://github.com/polkadot-js/api/issues/3687

Dependencies:

- [x] https://github.com/paritytech/scale-info/pull/111
- [x] https://github.com/paritytech/scale-info/pull/112
- [x] https://github.com/paritytech/scale-info/pull/114
- [x] https://github.com/paritytech/frame-metadata/pull/19
- [x] https://github.com/paritytech/frame-metadata/pull/23

Since we have checkmarks above, may as well add some more pertinent ones -

- [x] linting & building is all happy
- [x] testing is all happy
- [x] it works on v13...
- [ ] it works on earlier than v13...
- [x] it works on v14


packages/types-known/src/knownOrigins.ts
```diff
@@ -0,0 +1,26 @@
...
+  // (2) Defaults from Polkadot
+  //
+  Xcm: 'XcmOrigin',
+  XcmPallet: 'XcmOrigin',
+  //
+  // (3) Defaults from Acala
+  //
+  Authority: 'AuthorityOrigin',
+  GeneralCouncil: 'CollectiveOrigin'
+  /* eslint-enable sort-keys */
```
jacogr, line 25:
```suggestion

// We want predictive ordering (manually managed)
/* eslint-disable sort-keys */

// FIXME: Need some sort of solution for specifying these
// Since we don't have insight into the origin specification, we can only define what we know about
// in a pure Substrate/Polkadot implementation, any other custom origins won't be handled at all
export const knownOrigins: Record<string, string> = {
  //
  // (1) Defaults from Substrate
  //
  Council: 'CollectiveOrigin',
  System: 'SystemOrigin',
  TechnicalCommittee: 'CollectiveOrigin',
  //
  // (2) Defaults from Polkadot
  //
  Xcm: 'XcmOrigin',
  XcmPallet: 'XcmOrigin',
  //
  // (3) Defaults from Acala
  //
  Authority: 'AuthorityOrigin',
  GeneralCouncil: 'CollectiveOrigin'
```

Issue #3768, opened 2021-07-13
Inconsistent Junction Variant Keys


dt665m:
I've found an inconsistency in the XCM types that caused some confusion during testing.  The type defined for AccountKey20Junction https://github.com/polkadot-js/api/blob/ddf6f574f616d28cc0c59354baaf58208a776274/packages/types/src/interfaces/xcm/definitions.ts#L157 has "index" as it's second enum struct field, but the field on the polkadot end is "key" [here](https://github.com/paritytech/polkadot/blob/master/xcm/src/v0/junction.rs#L112)

This caused some confusion because for some reason, in javascript, if you set the value "index" like:
```
{
  X1: {
    AccountKey20: {
      network: { Any: "null" },
      index: "0x...."
    }
  }
}
```
 it deserializes properly in pallet-xcm / Rust Junction enum.  If you try to set "key" on the js side like 
```
{
  X1: {
    AccountKey20: {
      network: { Any: "null" },
      key: "0x...."
    }
  }
}
```
it gets deserialized as a the null account key ([0;20]) on the rust side.

What is most confusing about this is if you omit "index" or "key", polkadotjs api will happily serialize the payload with a null account.  If you omit "network", the usual complaint of incorrect shape is thrown.  @jacogr seems like there is a type naming mismatch here and maybe some scale codec nuance that I'm not privy to (does the js side handle enum struct variants using positional indexes like tuples?)  Why doesn't the js api complain if I use "key" instead of "index" or better yet, how come "index" can be omitted? 

Tested versions:
4.14.1 and 5.0.1

jacogr:
When a field is not specified, the default is used. Unknown fields are ignored. 

Not sure why the fields are different, quite possibly a rename at some point. (The serialization does not include names, o lay values so the key names are more for “human use”)

dt665m:
> When a field is not specified, the default is used. Unknown fields are ignored.
> 
> Not sure why the fields are different, quite possibly a rename at some point. (The serialization does not include names, o lay values so the key names are more for “human use”)

That makes sense.  So I'm assuming "network" cannot be omitted because there is no default for that?  Omitting that field causes the whole junction to fail serialization.  Is this as simple of a change as just renaming the file in the type defs?  Happy to PR.

jacogr:
You can just PR the rename in packages/types/interfaces/…

jacogr:
Closed in the linked PR.

PR #3770, opened 2021-07-13
realign AccountKey20Junction to polkadot xcm's Junction enum

dt665m:
fix typings per discussion in #3768


jacogr:
As per https://github.com/paritytech/polkadot/blob/master/xcm/src/v0/junction.rs#L112

jacogr (approved):
Thank you.

Issue #3636, opened 2021-06-11
BTreeSet and BTreeMap encoding


Lezek123:
There seems to be a long-running issue with the encoding of `BTreeSet` / `BTreeMap`.
Currently tested with `@polkadot/api` version: `4.2.1`

**The issue**
If one of the runtime extrinsics takes either `BTreeSet` or `BTreeMap` as argument, it can be provided as JavaScript `array` / `Set` (`BTreeSet`) or `Map` (`BTreeMap`), but we need to make sure that:
- the values are sorted in ascending order (in case of `BTreeSet`)
- the keys are sorted in ascending order (in case of `BTreeMap`)

Otherwise the call will fail with `ExtrinsicStatus:: 1010: Invalid Transaction: Transaction has a bad signature`.

So for example, if we assume the extrinsic argument is a `BTreeSet`, then:
```
await api.tx.module.method([1, 2, 3]).signAndSend(ALICE) // This will work
await api.tx.module.method([3, 2, 1]).signAndSend(ALICE) // This will throw an error
```

Those errors can be hard to spot and parse and require additional workarounds in `@polkadot/api` consumer applications.

**How to reproduce**

Unfortunately I don't have a minimum repository example, but this can be observed with [Joystream](https://github.com/Joystream/joystream) node:

1.
```
docker pull joystream/node
docker run -p 127.0.0.1:9944:9944 -p 127.0.0.1:9933:9933 joystream/node --dev --alice --validator --log runtime --rpc-port 9933 --unsafe-rpc-external --unsafe-ws-external --rpc-methods Unsafe
```
2. Go to https://testnet.joystream.org/#/settings
3. Set endpoint to `Local node`
4. Set interface operation mode to `Fully featured`
5. Hit `Save`
6. Go to `Developer`
7. Paste the json types from https://raw.githubusercontent.com/Joystream/joystream/master/types/augment/all/defs.json (this will prevent the current workaround for this issue from taking effect)
8. Go to https://testnet.joystream.org/#/js
9. Paste the following script:
```
const ALICE = '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY';

const fillOpeningTx = api.tx.contentDirectoryWorkingGroup.fillOpening(
  1, // openingId,
  [1, 2, 3], // applicants
  null
)

await fillOpeningTx.signAndSend(ALICE)
console.log('Sent')

[...]

PR #3789, opened 2021-07-21
BTreeSet and BTreeMap - encoding and serialization fixes

Lezek123:
Resolves https://github.com/polkadot-js/api/issues/3636 + an additional issue discovered when attempting to re-construct `BTreeMap` with `Tuple` keys from json (this was failing due to `Tuple` key beeing serialized to string and then incorrectly deserialized from string)

Added tests to verify if `BTreeSet`s / `BTreeMap`s are sorted and serialized/deserialized as expected.

Sorting rules were determined by trying sample calls against the substrate runtime with multiple different `BTreeSet` / `BTreeMap` extrinsic arguments.

### DIFF
--- a/packages/types/src/codec/utils/sortValues.ts
+++ b/packages/types/src/codec/utils/sortValues.ts
@@ -5,15 +5,21 @@ import type { Codec } from '../../types';
 
 import { AbstractArray } from '../AbstractArray';
 import { AbstractInt } from '../AbstractInt';
+import { Enum } from '../Enum';
+import { Struct } from '../Struct';
 
 /**
 * Sort keys/values of BTreeSet/BTreeMap in acending order for encoding compatibility with Rust's BTreeSet/BTreeMap
 * (https://doc.rust-lang.org/stable/std/collections/struct.BTreeSet.html)
 * (https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html)
 */
-export function sortAsc<V extends Codec | number[] | number | Uint8Array = Codec> (a: V, b: V): number {
+export function sortAsc<V extends Codec | Codec[] | number[] | number | Uint8Array = Codec> (a: V, b: V): number {
   if (typeof a === 'number') {
     return a - (b as typeof a);
+  } else if (a instanceof Struct) {
+    return sortAsc(Array.from(a.values()), Array.from((b as Struct).values()));
+  } else if (a instanceof Enum) {
+    return sortAsc(a.index, (b as Enum).index) || sortAsc(a.value, (b as Enum).value);
   } else if (a instanceof AbstractArray || a instanceof Uint8Array || Array.isArray(a)) {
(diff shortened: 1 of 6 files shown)


jacogr (commented):
Thank you. There are some CI issues. `yarn lint` locally may auto-fix some, or at least point out what is wrong.

Lezek123:
Added support for `Struct` and `Enum` keys/values.

packages/types/src/codec/utils/sortValues.ts
```diff
@@ -0,0 +1,53 @@
...
+      sortRes = sortAsc(a[i], (b as typeof a)[i]);
+
+      if (sortRes !== 0) {
+        return sortRes;
+      }
+    }
+
+    return lenA - lenB;
+  } else if (a instanceof AbstractInt) {
+    return a.eq(b) ? 0 : a.lt(b as typeof a) ? -1 : 1;
```
jacogr, line 38:
```suggestion
    return a.cmp(b);
```

packages/types/src/codec/utils/sortValues.ts
```diff
@@ -0,0 +1,53 @@
...
+    throw new Error(`Attempting to sort unrecognized value: ${JSON.stringify(a)}`);
+  }
+}
+
+export function sortSet<V extends Codec = Codec> (set: Set<V>): Set<V> {
+  return new Set(Array.from(set).sort(sortAsc));
+}
+
+export function sortMap<K extends Codec = Codec, V extends Codec = Codec> (map: Map<K, V>): Map<K, V> {
+  return new Map(Array.from(map.entries()).sort(([keyA], [keyB]) => sortAsc(keyA, keyB)));
```
jacogr, line 52: Really wondering if this is not just a 

```js
return new Map(Array.from(map.entries()).sort(([keyA], [keyB]) => u8aCmp(keyA.toU8a(), keyB.toU8a())));
```

Or at least for most of them - the Codec numbers are the only issue since they are LE encoded.

Lezek123: Unfortunately it's more complicated.

For example a `Struct` with 2 `Text` fields should be sorted by first filed first and then by second field only in case first fields `a` and `b` are equal.

The example below shows that it's not possible to achieve this with `u8aCmp` (expected sorting is: `{ handle: 'a' }`, `{ handle: 'aa' }`, `{ handle: 'b' }`)

image

Then there is also an issue with nested types, like ints inside tuples inside structs etc.

jacogr: 100%, was wondering about the `Vec<T>`-like encoding where the length is encoded, makes sense that it is raw data only.

jacogr (approved):
LGTM. Thank you very much for this. 

I really appreciate the extensive test-cases here, including the checks against what works on the node and what doesn't.

Issue #3801, opened 2021-07-23
Westmint events at 565/577 can not be decoded


wliyongfeng:
Hi, @jacogr . As the title  mentioned, please check westmint block 565/577:

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fwestmint-rpc.polkadot.io#/explorer/query/565
https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fwestmint-rpc.polkadot.io#/explorer/query/577

```
Unable to retrieve the specified block details. Unable to decode storage system.events:: createType(Vec<EventRecord>):: decodeU8a: failed at 0x00000000000302af… on : {"phase":"Phase","event":"Event","topics":"Vec<Hash>"}:: decodeU8a: failed at 0x0302af1bece884b2… on event: {"index":"EventId","data":"Null"}:: findMetaEvent: Unable to find Event with index 0x0302/[3,2]
```

should we fix it?

FYI, check following links for the event details which we scan days ago with the old version @polkadot/api:
https://www.statescan.io/westmint/block/565
https://www.statescan.io/westmint/block/577

wliyongfeng:
And another thing, when I call `GenericBlock#toHex`, api#5.1.1 will produce different result with the one by the previous 5.0.1. So I guess there's no standard for the hex, and we just follow the latest one?

jacogr:
Will check. There has been some old types than needed an update for Westend, there may have been a wrong change - this was related to heartbeats. 

The toHex should be the same if the blocks are the same.

jacogr:
Ahhh... Westmint... interesting. (Not Westend, sorry, pre-coffee...) 

Ok, atm the API does not track the changes in types on Westmint/Statemint, so will need an addition.

wliyongfeng:
> Ahhh... Westmint... interesting. (Not Westend, sorry, pre-coffee...)
> 
> Ok, atm the API does not track the changes in types across upgrades on Westmint/Statemint, so will need an addition to track historic types for these chains, e.g. https://github.com/polkadot-js/api/tree/master/packages/types-known/src/spec.

Sorry, not quite understand, do you mean we should add the the types manually while current `@polkadot/api` will not support them?

jacogr:
So between versions the types change, see for instance Kusama (which is the chain tracked the longest) - https://github.com/polkadot-js/api/blob/master/packages/types-known/src/spec/kusama.ts#L37

[...]

PR #3820, opened 2021-07-28
Check and match both specName/specVersion for registries

jacogr:
Closes https://github.com/polkadot-js/api/issues/3801

The issue is the following - 

- to determine the registry, we only matched on specVersion
- however some chains, like Westmint, changed the specName (shell -> westmint) and now has a conflict since it matches the specVersion, but not the name (and thus uses the completely wrong registry for pre-name-change blocks)

This changes the registry logic to match both the specName & specVersion.


jacogr:
Problematic blocks as linked decoded -

image

image

Issue #3793, opened 2021-07-22
Uniques: Cannot query by account


Yuripetusko:
Trying to query Uniques by account always returns empty

`const myAssets = await polkadotUniquesApi.query.uniques.account(userAddressEncoded);`
Same with classId passed

`const myAssets = await polkadotUniquesApi.query.uniques.account(userAddressEncoded, classId);`

This can be confirmed in polkadot.js/apps too

![image (2)](https://user-images.githubusercontent.com/666169/126606168-5d29ee24-aacc-4109-9c77-3531c1140120.png)

jacogr:
All maps require Tuple arguments, i.e. pass `[userAddressEncoded, classId]`

(In this case I would now refer to the UI for NMap usage, it has never been tested so assuming there will be issues that should be addressed)

Yuripetusko:
I tried different ways, and only this way doesn't throw any errors but result is empty

1. Passing a map
<img width="947" alt="Screenshot 2021-07-23 at 10 13 45" src="https://user-images.githubusercontent.com/666169/126761846-dffa0ed0-7926-4cff-9fad-923917446828.png">

2. Using query entries
<img width="929" alt="Screenshot 2021-07-23 at 10 14 08" src="https://user-images.githubusercontent.com/666169/126761922-914060de-9b14-422d-b1c1-e8b748938e4e.png">

3. And finally like my example above passing map 3 arguments directly (no map) is fine but result is empty when it shouldn't be
`const myAssets = await polkadotUniquesApi.query.uniques.account(userAddressEncoded, 0, '')`

![image](https://user-images.githubusercontent.com/666169/126762042-0af1af24-f378-429b-aef7-aa4c63e20760.png)

Yuripetusko:
Actually last example (3) can return non-empty result but types are incorrect 🤔 

![image](https://user-images.githubusercontent.com/666169/126763209-b8585f56-8ad0-4eb9-bb83-a58aee2fef91.png)

Does it suppose to return `Account` or `assets` for account?

Yuripetusko:
Still getting `.query.uniques.account.entries is not a function` 

All other uniques queries are fine, but not the one with `.account`

Yuripetusko:
I am happy to report that indeed updating `typescript` from `4.0.5` to `4.3.5` solved all the issues

Yuripetusko:
[...]

PR #3858, opened 2021-08-08
enable keys(), entries() for NMaps

hamidra:
We need to address #3793 to enable the support for uniques NFTs, and also move on with the Kusama NFT campaign.
I found the methods are already implemented but the support for NMaps were disabled. If not for any other specific reason, I think we should be able to enable them to unblock NFT campaign. 
Sending this PR to enable iteration on NMaps. I tested for unique accounts and could verify that the iteration on NMap keys/entries works as expected.

let accountAssets = await api.query.uniques.account.keys(accountAddress , classId); 
  for (let key of accountAssets) {...}


jacogr (approved):
Thank you.

Issue #3865, opened 2021-08-10
Can we make second arg (ClassId) optional in uniques.account query


Yuripetusko:
I would like to query all assets by account, because I might not know all classIds where these assets are located

`const keys = await api.query.uniques.account.keys(accountId);`

But this throws an error that second argument is missing

The line in question for generated types where second argument is not Optional is here, but I am not sure how to make it optional as it's generated from onchain metadata
https://github.com/polkadot-js/api/blob/ba74ad0be9ea5b2640f3b2c0ba1b23c786bc190b/packages/api/src/augment/query.ts#L1347

Yuripetusko:
And follow up question. If ClassId becomes optional, will returned double map contain classId. Because I don't see any other way to get classId from queried asset/instance

jacogr:
When you query directly, you need to supply all arguments - as per the generated types. For item retrieval, unlike keys/entries, you need to be pin-point accurate.

Currently, for keys/entries -

- with 1 argument (Map), none should be supplied, N - 1
- with 2 arguments (DoubleMap), 1 should be supplied, N -1
- with N arguments, N -1 should be supplied

It should (untested, unknown) be possible to query with N -2 or no keys in all cases. This would probably bring the RPC nodes down since the keys queries are quite heavy (although at least now it is a paged RPC).

PRs and tests welcome to see if it could work - https://github.com/polkadot-js/api/blob/master/packages/types/src/metadata/decorate/storage/createFunction.ts#L106-L132

Yuripetusko:
I will try this in a week if no one else will, please don't close this ticket yet.

I understand there can be an issue of a query that is too heavy, but using paged query should be ok

jacogr:
Closed in #3886

Yuripetusko:
Thank you @hamidra

PR #3886, opened 2021-08-21
add support for iteration of k<N entries for NMaps

hamidra:
Currently polkadot.js NMap entries enforce the number of arguments be one less than the full arguments or no arguments provided , but the NMap entries are able to iterate through any arbitrary number of arguments as long as args.length<N.
These changes are made to add support for entries iterators with k arguments while k<N for NMaps.
This should address issue #3865 and let the API query the uniques NFTs for an account, regardless of the class of NFTs.
Was able to verified that the below scenarios work:

let accountAssets = await api.query.uniques.account.entries('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'); 
  accountAssets.forEach(([key, value]) => {
    console.log(key.toHuman())
    console.log(value.toHuman())
  })

let accountAssets = await api.query.uniques.account.keys('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY'); 
  accountAssets.forEach((key) => {
    console.log(key.toHuman())
  })
  
  Also should not impose any extra burden on rpcs since the current APIs support not passing any arguments which is a heavier query than passing partial arguments.

### DIFF
--- a/packages/types/src/metadata/decorate/storage/createFunction.ts
+++ b/packages/types/src/metadata/decorate/storage/createFunction.ts
@@ -108,21 +108,21 @@ function extendPrefixedMap (registry: Registry, itemFn: CreateItemFn, storageFn:
       (
         (args.length === 0) ||
         (type.isDoubleMap && args.length === 1) ||
-        (type.isNMap && args.length === (type.asNMap.hashers.length - 1))
+        (type.isNMap && args.length < (type.asNMap.hashers.length))
       ),
-      () => `Iteration ${stringCamelCase(section || 'unknown')}.${stringCamelCase(method || 'unknown')} needs arguments to be one less than the full arguments, found [${args.join(', ')}]`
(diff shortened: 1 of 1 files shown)


packages/types/src/metadata/decorate/storage/createFunction.ts
```diff
@@ -108,21 +108,21 @@ function extendPrefixedMap (registry: Registry, itemFn: CreateItemFn, storageFn:
...
-        const hashers = [...type.asNMap.hashers];
+        let keys = [...type.asNMap.keyVec];
+        let hashers = [...type.asNMap.hashers];
 
-        // remove the last entry
-        keys.pop();
-        hashers.pop();
+        // pick the first n entries where n = args.length which is already verified above to be less that the full arguments.
+        keys = keys.slice(0, args.length);
+        hashers = hashers.slice(0, args.length);
```
jacogr, line 125: We could simplify this with the above, i.e. `const keys = type.asNMap.keyVec.slice(0, args.length)`

jacogr (approved):
Thank you. Looks sane.

PR #3932, opened 2021-09-06
Version Junction and MultiLocation Types

apopiak:
The [`Parent` variant](https://github.com/paritytech/polkadot/blob/730c769505106d180f0febed3f050349d4502c20/xcm/src/v0/junction.rs#L110) in `v0::Junction` was removed. (See [`v1::Junction` here](https://github.com/paritytech/polkadot/blob/730c769505106d180f0febed3f050349d4502c20/xcm/src/v1/junction.rs#L29))


jacogr (approved):
Thank you, I really appreciate it.

PR #3949, opened 2021-09-14
HttpProvider support clone()

### LINKED ISSUE #3948: support api.clone() when provider is HttpProvider

ianhe8x:
Closes #3948

### DIFF
--- a/packages/rpc-provider/src/http/index.spec.ts
+++ b/packages/rpc-provider/src/http/index.spec.ts
@@ -29,6 +29,12 @@ describe('Http', (): void => {
     ).not.toThrow();
   });
 
+  it('allow clone', (): void => {
+    const clone = http.clone();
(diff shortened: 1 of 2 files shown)


packages/rpc-provider/src/http/index.spec.ts
```diff
@@ -29,6 +29,12 @@ describe('Http', (): void => {
     ).not.toThrow();
   });
 
+  it('allow clone', (): void => {
+    const clone = http.clone();
+    expect((clone as any)['#endpoint']).toEqual((http as any)['#endpoint']);
```
jacogr, line 34:
```suggestion
    const clone = http.clone();

    expect((clone as any)['#endpoint']).toEqual((http as any)['#endpoint']);
```

jacogr: Have not checked, but this I believe is the linting issue

jacogr (approved):

Issue #3961, opened 2021-09-16
Cannot compile the current beta for polkadot js api (5.9.2-7)


grbIzl:
After the metadata change TS cannot compile polkadot modules. A lot of errors like:

```
node_modules/@polkadot/api/augment/consts.d.ts:4:15 - error TS2305: Module '"@polkadot/types/lookup"' has no exported member 'FrameSupportPalletId'.

4 import type { FrameSupportPalletId, FrameSupportWeightsRuntimeDbWeight, FrameSupportWeightsWeightToFeeCoefficient, FrameSystemLimitsBlockLength, FrameSystemLimitsBlockWeights, PalletContractsSchedule, SpVersionRuntimeVersion } from '@polkadot/types/lookup';

```
The minimal TS app can be found here: https://github.com/grbIzl/ts_min_example
yarn install && tsc

PR #3964, opened 2021-09-17
Additional augmentation fixes

jacogr:
Also related to https://github.com/polkadot-js/api/issues/3961 (Since types are augmented in all lookup cases now, ensure we follow the augmentation path first)

### DIFF
--- a/packages/api/src/checkTypes.manual.ts
+++ b/packages/api/src/checkTypes.manual.ts
@@ -21,7 +21,7 @@ function consts (api: ApiPromise): void {
   console.log(
     api.consts.foo.bar,
     api.consts.balances.existentialDeposit.toNumber(),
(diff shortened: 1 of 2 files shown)


packages/api/src/checkTypes.manual.ts
```diff
@@ -21,7 +21,8 @@ function consts (api: ApiPromise): void {
   console.log(
     api.consts.foo.bar,
     api.consts.balances.existentialDeposit.toNumber(),
-    api.consts.balances.existentialDeposit.meta.docs.map((s): string => s.toString()).join('')
+    api.consts.balances.existentialDeposit.meta.docs.map((s): string => s.toString()).join(''),
```
jacogr, line 24:
```suggestion
    api.consts.balances.existentialDeposit.meta.docs.map((s) => s.toString()).join(''),
```

PR #4143, opened 2021-11-02
Performance tweaks

nazar-pc:
These optimizations allow to skip some work and optimize other kinds of work leading to a few times higher performance when dealing with large number of subscriptions to events.

These are some of the low-hanging fruits I found after a bit of profiling of the app which is CPU bound because of polkadot.js.

### DIFF
--- a/packages/api/src/util/filterEvents.ts
+++ b/packages/api/src/util/filterEvents.ts
@@ -7,24 +7,22 @@ import { l } from './logging';
 
 export function filterEvents (extHash: H256, { block: { extrinsics, header } }: SignedBlock, allEvents: EventRecord[], status: ExtrinsicStatus): EventRecord[] | undefined {
   // extrinsics to hashes
-  const myHash = extHash.toHex();
-  const allHashes = extrinsics.map((ext): string => ext.hash.toHex());
-
-  // find the index of our extrinsic in the block
-  const index = allHashes.indexOf(myHash);
-
-  // if we do get the block after finalized, it _should_ be there
-  if (index === -1) {
-    // only warn on filtering with isInBlock (finalization finalizes after)
-    if (status.isInBlock) {
-      l.warn(`block ${header.hash.toHex()}: Unable to find extrinsic ${myHash} inside ${allHashes.join(', ')}`);
+  for (const [index, extrinsic] of extrinsics.entries()) {
+    if (extrinsic.hash.eq(extHash)) {
+      return allEvents.filter(({ phase }) =>
+        // only ApplyExtrinsic has the extrinsic index
+        phase.isApplyExtrinsic && phase.asApplyExtrinsic.eqn(index)
+      );
     }
-
-    return;
   }
 
-  return allEvents.filter(({ phase }) =>
-    // only ApplyExtrinsic has the extrinsic index
-    phase.isApplyExtrinsic && phase.asApplyExtrinsic.eqn(index)
-  );
+  // if we do get the block after finalized, it _should_ be there
+  // only warn on filtering with isInBlock (finalization finalizes after)
+  if (status.isInBlock) {
+    const allHashes = extrinsics.map((ext): string => ext.hash.toHex());
+
+    l.warn(`block ${header.hash.toHex()}: Unable to find extrinsic ${extHash.toHex()} inside ${allHashes.join(', ')}`);
+  }
+
+  return undefined;
 }
--- a/packages/types/src/codec/Enum.ts
+++ b/packages/types/src/codec/Enum.ts
@@ -155,6 +155,42 @@ function decodeEnum (registry: Registry, def: TypesDef, value?: any, index?: num
   return decodeFromValue(registry, def, value);
 }
 
+const camelCaseMap = new Map<string, string>();
+
+function memoizedStringCamelCase (value: string): string {
+  {
+    const camelCaseValue = camelCaseMap.get(value);
+
+    if (camelCaseValue) {
+      return camelCaseValue;
+    }
+  }
+
+  const camelCaseValue = stringCamelCase(value);
+
+  camelCaseMap.set(value, camelCaseValue);
+
+  return camelCaseValue;
+}
+
+const propertySuffixMap = new Map<string, string>();
+
+function memoizedPropertySuffix (value: string): string {
+  {
+    const suffix = propertySuffixMap.get(value);
+
+    if (suffix) {
+      return suffix;
+    }
+  }
+
+  const suffix = stringUpperFirst(stringCamelCase(value.replace(' ', '_')));
+
+  propertySuffixMap.set(value, suffix);
+
+  return suffix;
+}
+
 /**
  * @name Enum
  * @description
@@ -206,7 +242,7 @@ export class Enum implements IEnum {
 
         for (let i = 0; i < keys.length; i++) {
           const key = keys[i];
-          const name = stringUpperFirst(stringCamelCase(key.replace(' ', '_')));
+          const name = memoizedPropertySuffix(key);
           const askey = `as${name}`;
           const iskey = `is${name}`;
 
@@ -351,7 +387,7 @@ export class Enum implements IEnum {
   public toJSON (): AnyJson {
     return this.#isBasic
       ? this.type
-      : { [stringCamelCase(this.type)]: this.#raw.toJSON() };
+      : { [memoizedStringCamelCase(this.type)]: this.#raw.toJSON() };
   }
 
   /**
--- a/packages/types/src/codec/Set.ts
+++ b/packages/types/src/codec/Set.ts
@@ -80,6 +80,24 @@ function decodeSet (setValues: SetValues, value: string[] | Set<string> | Uint8A
   return decodeSetNumber(setValues, value);
 }
 
+const iskeyMap = new Map<string, string>();
+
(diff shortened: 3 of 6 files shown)


packages/types/src/create/sanitize.ts
```diff
@@ -204,20 +205,26 @@ export function removePairOf (): Mapper {
 
 // remove the type traits
 export function removeTraits (): Mapper {
+  const from1 = /\s/g;
+  const from2 = /(T|Self)::/g;
+  const from3 = /<(T|Self)asTrait>::/g;
+  const from4 = /<Tas[a-z]+::Trait>::/g;
+  const from5 = /<LookupasStaticLookup>/g;
+  const from6 = /::Type/g;
```
jacogr, line 213: These could be moved to file-level, that is a small bump, i.e.doesn't need to be created each time in the function..

nazar-pc: They have pretty bad names that only make sense in a narrow scope.
Also this function is called just once, so will not make any difference in practice anyway.

jacogr: Cool, all good. At best it is a real micro-opt.

packages/api/src/util/filterEvents.ts
```diff
@@ -7,24 +7,21 @@ import { l } from './logging';
...
-  // find the index of our extrinsic in the block
-  const index = allHashes.indexOf(myHash);
-
-  // if we do get the block after finalized, it _should_ be there
-  if (index === -1) {
-    // only warn on filtering with isInBlock (finalization finalizes after)
-    if (status.isInBlock) {
-      l.warn(`block ${header.hash.toHex()}: Unable to find extrinsic ${myHash} inside ${allHashes.join(', ')}`);
+  for (const [index, extrinsic] of extrinsics.entries()) {
+    if (extHash === extrinsic.hash) {
```
jacogr, line 11:
```suggestion
    if (extrinsic.hash.eq(extHash) {
```

The objects may not be the same, so we compare the raw bytes - will bail on the first mismatch.

packages/types/src/create/sanitize.ts
```diff
@@ -68,14 +68,13 @@ export function findClosing (value: string, start: number): number {
 }
 
 export function alias (src: string, dest: string, withChecks = true): Mapper {
+  const from = new RegExp(`(^${src}|${BOX_PRECEDING.map((box) => `\\${box}${src}`).join('|')})`, 'g');
```
jacogr, line 71: +1 Nice catch.

packages/types/src/codec/Base.ts
```diff
@@ -42,7 +43,11 @@ export abstract class Base<T extends Codec> implements Codec {
    * @description returns a hash of the contents
    */
   public get hash (): CodecHash {
-    return this.registry.hash(this.toU8a());
+    if (!this.#hashCache) {
```
jacogr, line 46: This one is actually quite problematic, specifically for the case you want to optimize, e.g. extrinsics. The issue is the following - especially in the case of extrinsics, the values are mutable. Basically you can create a extrinsic via `api.tx....` which will have one hash and then after submission (signature attached), it will have a different hash.

So where this will fail -

```js
const tx = api.tx.balances.transfer(BOB, 12345);

console.log('pre signing hash', tx.hash.toHex());

await tx.signAndSend(...);

// in this case the object mutated, the hash should be is different, be here showed as the same
console.log('post signing hash', tx.hash.toHex());
```

jacogr: The way around this is probably to actually check the extrinsic differently - i.e. rather use the `.eq` on the extrinsic to see if all the details do match. (Once again, as soon as something mismatched is found, it will bail immediately)

This is probably a better approach since we won't serialize and hash, which is expensive.

nazar-pc: Hm, but we don't always have access to the full extrinsic to check, this is why extrinsics are checked by hash instead.

What if hash is cached in an extrinsic instead and cache is reset on signature, will that work?

jacogr: Yes, that works - it makes sense to move this logic to the extrinsic itself (since that is where you are trying to optimize).

packages/types/src/extrinsic/v4/Extrinsic.ts
```diff
@@ -24,6 +25,8 @@ export interface ExtrinsicValueV4 {
  * The third generation of compact extrinsics
  */
 export class GenericExtrinsicV4 extends Struct implements IExtrinsicImpl {
+  #hashCache?: CodecHash;
```
jacogr, line 28: This should only be needed on the upper layer instead of duplicating here? (Effectively the Extrinsic is an Enum, so the V4 structure should be be stand-alone)

packages/types/src/extrinsic/Extrinsic.ts
```diff
@@ -36,6 +37,8 @@ const VERSIONS = [
 export { EXTRINSIC_VERSION as LATEST_EXTRINSIC_VERSION } from './v4/Extrinsic';
 
 abstract class ExtrinsicBase<A extends AnyTuple> extends Base<ExtrinsicVx | ExtrinsicUnknown> {
+  protected hashCache?: CodecHash;
```
jacogr, line 40: Not 100% sure why this is protectected? Doesn't the `#hashCache` make more sense?

If `protected` is needed, just do it as `_hashCache`, just to indicate it is internal. (minor)

packages/types/src/extrinsic/v4/Extrinsic.ts
nazar-pc: I was not sure how inheritance works here, there was no clear path that this one extends any other ones, it seems to directly extend `Struct`, not `BaseExtrinsic`, so I ended up adding it to both. I can remove it from here if it is indeed covered by the base extrinsic.

packages/types/src/extrinsic/Extrinsic.ts
nazar-pc: This is `ExtrinsicBase`, it doesn't have signature-related methods on it, they are in `GenericExtrinsic`, which extends `ExtrinsicBase`, so I had to make property protected such that code in `GenericExtrinsic` can access it.

nazar-pc:
With PR applied as of right now applied I see ~4x increase in number of transactions processed by Node.js application

packages/types/src/extrinsic/v4/Extrinsic.ts
jacogr: Will bring it in as-is and cleanup. (It only needs to be on the base, i.e. type `Extrinsic` is created, but that acts like an enum over the V4)

packages/types/src/codec/Set.ts
```diff
@@ -80,6 +80,24 @@ function decodeSet (setValues: SetValues, value: string[] | Set<string> | Uint8A
...
+      return iskey;
+    }
+  }
+
+  const iskey = `is${stringUpperFirst(stringCamelCase(value))}`;
+
+  iskeyMap.set(value, iskey);
+
+  return iskey;
+}
```
jacogr, line 99: Will drop this - it only gets use inside `with` (and classes are cached, so it really doesn't make a difference, i.e. it is more micro-optimization that adds maintanence)

packages/types/src/codec/Enum.ts
```diff
@@ -206,7 +242,7 @@ export class Enum implements IEnum {
 
         for (let i = 0; i < keys.length; i++) {
           const key = keys[i];
-          const name = stringUpperFirst(stringCamelCase(key.replace(' ', '_')));
+          const name = memoizedPropertySuffix(key);
```
jacogr, line 245: I view this in the same way as the one from `Set`, so will revert this after merge.

(There is probably something to be said for a generalized solution with Struct, Set, Enum, ... where they all follow the same pattern, but as it stands it is micro in the lesser-used paths)

jacogr (approved):
Nice, thank you.

As mentioned there are still some cleanups around Extrinsic, will adapt once merged.

packages/types/src/codec/Enum.ts
nazar-pc: I found that this was an interesting case where it did make sense. `stringCamelCase` does a few regular expressions internally, but always act on the same inputs. It was showing as a significant contributor of CPU usage in the profiler before this change.

jacogr: If it calls it on the same enum, doing the same keys - there is another issue, i.e. the like-definitions are re-created instead of re-used. That should be solved. At this layer, for a specific definition, it _should_ only be created once. In that case the cache hits here should be close to 0.

(It is quite possible that that assumption doesn't hold anymore after the metadata v14 conversion, but that is still a bug/unintended consequence then.)

jacogr: Ok, this is not something new, but certainly something that should be done - so yes, indeed base aalsses are cached, e.g. `NodeRuntimeSessionKeys`. 

However what is not cached and re-created are stuff that doesn't appear in the lookup registry explicitly, e.g. `Vec<NodeRuntimeSessionKeys>` will be recreated each time it is used. (Same with embedded enums, etc.) So there is an opportunity to improve on a per-registry basis. The biggest overhead is indeed the dance without going from definition -> Class while parsing.

PR #4204, opened 2021-11-11
Change BeefySignedCommitment signatures type to EcdsaSignature

czl1378:
Maybe we should use EcdsaSignature instead of Signature, because we can recover the public key from this.

### DIFF
--- a/packages/types/src/interfaces/beefy/definitions.ts
+++ b/packages/types/src/interfaces/beefy/definitions.ts
@@ -28,7 +28,7 @@ export default {
     BeefyId: '[u8; 33]',
     BeefySignedCommitment: {
       commitment: 'BeefyCommitment',
-      signatures: 'Vec<Option<Signature>>'
+      signatures: 'Vec<Option<EcdsaSignature>>'
     },
     BeefyNextAuthoritySet: {
       id: 'u64',


jacogr:
Do you have a link to the Beefy source to just do a sanity check? (These are always tricky, so an extra pair of eyes, in this case me, is always good).

This should not be an issue with any recent version of Substrate, e.g. in metadata v14 these types are not used at all, rather all is specified by the runtime metadata.

czl1378:
We found that beefy uses the Signature struct of 65 bits:

`/// A signature (a 512-bit value, plus 8 bits for recovery ID).
#[derive(Encode, Decode, PassByInner, TypeInfo)]
pub struct Signature(pub [u8; 65]);`

Can you check this?

https://github.com/octopus-network/substrate/blob/4b40c2aa05930a0aae5a7eb0c50a596a44eb6fb1/primitives/core/src/ecdsa.rs#L239

I don't know why you say that: "this should not be an issue with metadata v14".

Actually I'm building on the Substrate of metadata v14 . When I tried to change the signatures type to `EcdsaSignature`, I got a signatures of 65 bits, and then I changed it's back to `Signature`, I only got a 64 bits signatures.

Thanks.

jacogr:
Where how did you change the signature? Which version of the API? So for any encoding/decoding the API (with metadata v14) uses the on-chain registry, it does not refer to these types at all. (If you do a manual decoding, using these string-named types, then yes, it will come into play). The on-chain runtime injected types are not these, they are named differently.

czl1378:
The API version is `6.7.2`, and here's my code to subscribe justifications: 

```
const { ApiPromise, WsProvider } = require('@polkadot/api');

const provider = new WsProvider('ws://localhost:9944');
const api = new ApiPromise({ provider });

api.on('connected', () => {
  console.log('connected');
});

api.on('ready', async () => {
  api.rpc.chain.subscribeNewHeads((lastHeader) => {
    console.log('header', lastHeader.toJSON());
  });

  api.rpc.beefy.subscribeJustifications((justifications) => {
    console.log('justifications.signatures', justifications.signatures.toString());
  });
});
```

I changed the signature type in my `node_modules/@polkadot/types/interfaces/beefy/definitions.cjs`
```
...
BeefySignedCommitment: {
  commitment: 'BeefyCommitment',
  signatures: 'Vec<Option<EcdsaSignature>>'
},
...
```

jacogr:
Ahhh, RPC - yes, sadly those don't (yet) get any information about the expected types from the metadata. (It is on the roadmap for implementation in Substrate, but no idea when it will be available)

jacogr (approved):

Issue #4213, opened 2021-11-15
Metadata v14 error during typegen


savudani8:
I'm using `@polkadot/api` v6.8.1 and a chain with metadata v14 (`polkadot-v0.9.11`) whose metadata I retrieved in `parachain.json`. Not sure how to tackle this error on `polkadot-types-from-chain`
```
$ ts-node --skip-project node_modules/.bin/polkadot-types-from-chain --package @interlay/interbtc/interfaces --endpoint ./src/json/parachain.json --output ./src/interfaces
Generating from metadata, 145,369 bytes

interbtc-api/node_modules/@polkadot/types/codec/utils/decodeU8a.cjs:41
      throw new Error(`decodeU8a: failed at ${(0, _util.u8aToHex)(u8a.subarray(offset).slice(0, 8))}… on ${keys[i] ? `${keys[i]}` : ''}${rawType ? `: ${rawType}` : ''}:: ${error.message}`);
            ^
Error: decodeU8a: failed at 0x0e6104000c1c7370… on metadata:: decodeU8a: failed at 0x741853797374656d… on pallets: Vec<PalletMetadataV14>:: decodeU8a: failed at 0x1853797374656d01… on : {"name":"Text","storage":"Option<PalletStorageMetadataV14>","calls":"Option<PalletCallMetadataV14>","events":"Option<PalletEventMetadataV14>","constants":"Vec<PalletConstantMetadataV14>","errors":"Option<PalletErrorMetadataV14>","index":"u8"}:: decodeU8a: failed at 0x011853797374656d… on storage: Option<PalletStorageMetadataV14>:: decodeU8a: failed at 0x401c4163636f756e… on items: Vec<StorageEntryMetadataV14>:: decodeU8a: failed at 0x04e8205468652066… on : {"name":"Text","modifier":"StorageEntryModifierV14","type":"StorageEntryTypeV14","fallback":"Bytes","docs":"Vec<Text>"}:: decodeU8a: failed at 0x68652066756c6c20… on modifier: {"_enum":{"Optional":0,"Default":1,"Required":2}}:: Unable to create Enum via index 104, in Optional, Default, Required
    at decodeU8a (interbtc-api/node_modules/@polkadot/types/codec/utils/decodeU8a.cjs:41:13)
    at decodeStruct (interbtc-api/node_modules/@polkadot/types/codec/Struct.cjs:101:41)
[...]

PR #4224, opened 2021-11-17
fix: Generate custom type definitions

savudani8:
A first attempt at fixing https://github.com/polkadot-js/api/issues/4213. I wasn't sure where exactly to output lookup types, so I modified `polkadot-types-from-def` as a proof-of-concept:
- the metadata file is input to the script using the same logic from `polkadot-types-from-chain`, so CLI logic is refactored
- generates the `definitions.ts` file from a v14 metadata file instead of using manual definitions. This means any manual type definitions would be overwritten
- leaves the type-generation logic (which takes `definitions.ts`) unchanged, and produces decorated types in `default/types.ts` as before

The resulting `definitions.ts` file looks like this:
```
// Auto-generated via `yarn polkadot-types-from-defs`, do not edit
/* eslint-disable */

/* eslint-disable sort-keys */

export default {
  rpc: {},
  types: {
    /**
     * Lookup3: frame_system::AccountInfo<Index, pallet_balances::AccountData<Balance>>
     **/
    FrameSystemAccountInfo: {
      nonce: 'u32',
      consumers: 'u32',
      providers: 'u32',
      sufficients: 'u32',
      data: 'PalletBalancesAccountData'
    },
    /**
     * Lookup5: pallet_balances::AccountData<Balance>
     **/
    PalletBalancesAccountData: {
      free: 'u128',
      reserved: 'u128',
      miscFrozen: 'u128',
      feeFrozen: 'u128'
    },
...
```

`polkadot-types-from-chain` is not changed because it is now able to find the types from `interfaces/default/types.ts` to resolve imports in the api interface files.

There is one error left in this approach, which is that in `interfaces/default/types.ts` pallet-specific types fail to be resolved for custom pallets. For instance, `OrmlTokensModuleCall`, which comes from an external pallet, is successfully resolved. I'm not very knowledgeable about Substrate but maybe this has to do with our pallets or runtime needing to make these types public somehow?

Screenshot from 2021-11-17 15-49-26

### DIFF
--- a/packages/typegen/src/templates/lookup/defs.hbs
+++ b/packages/typegen/src/templates/lookup/defs.hbs
@@ -2,8 +2,6 @@
 
 /* eslint-disable sort-keys */
 
-import type { Definitions } from '../../types';
-
 export default {
   rpc: {},
   types: {
@@ -14,5 +12,5 @@ export default {
     {{/each}}
     {{/each}}
   }
-} as Definitions;
+};
 {{> footer }}
--- a/packages/typegen/src/fromChain.ts
+++ b/packages/typegen/src/fromChain.ts
@@ -6,6 +6,7 @@ import yargs from 'yargs';
 
 import { formatNumber } from '@polkadot/util';
 import { WebSocket } from '@polkadot/x-ws';
+import { Definitions } from '@polkadot/types/types';
 
 import { generateDefaultConsts, generateDefaultErrors, generateDefaultEvents, generateDefaultQuery, generateDefaultRpc, generateDefaultTx } from './generate';
 import { HEADER, writeFile } from './util';
@@ -18,12 +19,21 @@ function generate (metaHex: string, pkg: string | undefined, output: string, isS
     ? { [pkg]: require(path.join(process.cwd(), output, 'definitions')) as Record<string, any> }
     : {};
 
-  generateDefaultConsts(path.join(process.cwd(), output, 'augment-api-consts.ts'), metaHex, extraTypes, isStrict);
+  let customLookupDefinitions = undefined;
+  try {
+    customLookupDefinitions = {
+      rpc: {},
+      types: require(path.join(process.cwd(), output, 'lookup.ts')).default
+    }  as Definitions;
+  } catch (error) {
+    console.log("No custom definitions found.");
+  }
+  generateDefaultConsts(path.join(process.cwd(), output, 'augment-api-consts.ts'), metaHex, extraTypes, isStrict, customLookupDefinitions);
   generateDefaultErrors(path.join(process.cwd(), output, 'augment-api-errors.ts'), metaHex, extraTypes, isStrict);
-  generateDefaultEvents(path.join(process.cwd(), output, 'augment-api-events.ts'), metaHex, extraTypes, isStrict);
-  generateDefaultQuery(path.join(process.cwd(), output, 'augment-api-query.ts'), metaHex, extraTypes, isStrict);
+  generateDefaultEvents(path.join(process.cwd(), output, 'augment-api-events.ts'), metaHex, extraTypes, isStrict, customLookupDefinitions);
+  generateDefaultQuery(path.join(process.cwd(), output, 'augment-api-query.ts'), metaHex, extraTypes, isStrict, customLookupDefinitions);
   generateDefaultRpc(path.join(process.cwd(), output, 'augment-api-rpc.ts'), extraTypes);
-  generateDefaultTx(path.join(process.cwd(), output, 'augment-api-tx.ts'), metaHex, extraTypes, isStrict);
+  generateDefaultTx(path.join(process.cwd(), output, 'augment-api-tx.ts'), metaHex, extraTypes, isStrict, customLookupDefinitions);
 
   writeFile(path.join(process.cwd(), output, 'augment-api.ts'), (): string =>
     [
--- a/packages/typegen/src/fromChain.ts
+++ b/packages/typegen/src/fromChain.ts
@@ -4,9 +4,9 @@
 import path from 'path';
 import yargs from 'yargs';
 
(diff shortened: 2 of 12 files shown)


packages/typegen/src/templates/lookup/defs.hbs
```diff
@@ -2,8 +2,6 @@
 
 /* eslint-disable sort-keys */
 
-import type { Definitions } from '../../types';
```
savudani8, line 5: I removed this import because the definitions are type-asserted on usage anyway: https://github.com/polkadot-js/api/blob/master/packages/typegen/src/fromDefs.ts#L29

jacogr:
Thank you.

So the missing imports needs to come from `@polkadot/types/lookup`.

So this means that this file generated should actually augment to that location (to share imports). (for old-style definitions the augmented path is `@polkadot/types/types/registry`, i.e. the `declare module at the top of the file)

savudani8:
> So where you see the above issues, if you just change to declare module '@polkadot/types/lookup' { at the file-top, the types should be available. (This uses the templates/lookup hbs template)

This PR generates the same type definitions in two files: 
- one with the `templates/lookup/types.hbs`(the one you suggested, which has `declare module '@polkadot/types/lookup' {` at the top)
- one with the `/templates/tsDef/moduleType.hbs` template

Both of these files have the same import errors seen in the screenshot above.

> So the missing imports needs to come from @polkadot/types/lookup.

I just checked `augment-api-query.ts` and it imports some types from  `'@interlay/interbtc/interfaces/default'` (generated from `/templates/tsDef/moduleType.hbs`) and others from `'@polkadot/types/lookup'` (generated from `templates/lookup/types.hbs`) -> these two `import` lines can be combined into a single import from `'@polkadot/types/lookup'`.

The errors are present in the types file itself, not in the files importing generated types.

jacogr:
Ok, it seems the `*Call` types re not being generated since they alias through to `Call`. The same should actually happen on the `*StandaloneCall`, so there is a mismatch. (However, have been thinking that even these aliased types should be generated, took a brief look 2 weeks ago, but didn't get to a resolution there)

savudani8:
I tried both and only rebasing off `jg-Call-Event-2` fixes the errors

jacogr:
Yes, RPCs are not yet with metadata definitions. So for those custom types are needed.

jacogr:
I took a brief look last night, I think it looks good. Will give a proper look-over and merge in a short-while.

savudani8:
There are still two issues to fix I think:
- (second bullet point [here](https://github.com/polkadot-js/api/pull/4224#issue-1056288524)) Since custom RPC definitions are not included in the metadata, `definitions.ts` (which exposes manual definitions) should not be overwritten as this PR does. The auto-generated types should be output somewhere else
- All custom-type imports should come from `'@polkadot/types/lookup'`. However currently there are two almost identical files for these types being generated, and imports are picked up from both (described [here](https://github.com/polkadot-js/api/pull/4224#issuecomment-971846782))

jacogr:
1. For the from metadata types, it probably makes sense to output a `lookup.ts` file that contains the definitions. Clear split with the old version still in-tact. (Have not looked as to the overhead/mangling that needs to be in place to make it happen). But we probably needs a `lookup.types.ts` as well that contains the module declaration then?

2. For the lookup imports, it should not come from the definitions, with the type augmentation straight from `@polkadot/types/lookup` should be fine. (The manual definitions are a bit different)

jacogr (commented):
Code looks spot-on. Would just adjust the lookup generated `definitions.ts` and `types.ts` naming to not have a conflict, but apart from that it is clean.

savudani8:
Generating separate lookup types was easy, but I've not had success finding a place to add their import. The new `types-lookup.ts` declares the module `'@polkadot/types/lookup'`, but it's not being picked up.

> For the lookup imports, it should not come from the definitions, with the type augmentation straight from @polkadot/types/lookup

[createImports](https://github.com/polkadot-js/api/blob/master/packages/typegen/src/util/imports.ts#L92) works based on the definitions, not the types. There's currently some types that are already being imported from `'@polkadot/types/lookup'`, but that's because they're specified in [lookupDefinitions](https://github.com/polkadot-js/api/blob/master/packages/typegen/src/generate/consts.ts#L22) and added [here](https://github.com/polkadot-js/api/blob/master/packages/typegen/src/generate/query.ts#L72).

In files like `generate/query.ts`, I tried something like
```
import * as customLookupTypes from '@polkadot/types/lookup';
...
const allTypes: ExtraTypes = {
  '@polkadot/types/augment': { lookup: lookupDefinitions },
  '@polkadot/types/interfaces': defaultDefs,
  ...extraTypes,
  ...customLookupTypes
};
```
but `customLookupTypes` is empty

jacogr:
At least for usage, something like this to tell TS to augment will be needed - https://polkadot.js.org/docs/api/examples/promise/typegen#typescript-config

not sure that moves you forward, will have to take a peek.

jacogr:
Ok, I see how I adjusted this - for instance `generate/consts.ts`, have this - '@polkadot/types/augment': { lookup: lookupDefinitions },

So it needs to be something like -

```js
'@polkadot/types/augment': { 
  lookup: { ...lookupDefinitions, ...useLookupDefinitions }
},
```

And then we need to pass in the user definitions.

savudani8:
It should be done now - no errors left after typegen is run. I'll run some integration tests on the generated interfaces tomorrow to make sure all is good.

After this is merged I'll update the CLI arguments in [this](https://polkadot.js.org/docs/api/examples/promise/typegen/#packages) snippet in the docs. Let me know if there are other places that need updating too.

jacogr (approved):
Thanks an absolute million for this - it has been at the back of my mind (since 6.0 went out the door), but just never had a gap to attack. 

This really helps all teams out there and (while I may have neglected this), it a critical piece.

PR #4304, opened 2021-12-08
Add storage deposit support

statictype:
An attempt to add support for storage deposits that were introduced in the contracts pallet with [this PR](https://github.com/paritytech/substrate/pull/10082)

Not sure what to do for backwards compatibility. 

This is the first step, a PR in `polkadot-js/apps` will follow where i implement the input for storage deposit.
I will rename endowment in a following commit here


packages/api-contract/src/base/Blueprint.ts
```diff
@@ -52,11 +52,12 @@ export class Blueprint<ApiType extends ApiTypes> extends Base<ApiType> {
     return this.#tx;
   }
 
-  #deploy = (constructorOrId: AbiConstructor | string | number, { gasLimit = BN_ZERO, salt, value = BN_ZERO }: BlueprintOptions, params: unknown[]): SubmittableExtrinsic<ApiType, BlueprintSubmittableResult<ApiType>> => {
+  #deploy = (constructorOrId: AbiConstructor | string | number, { gasLimit = BN_ZERO, storageDepositLimit = BN_ZERO, salt, value = BN_ZERO }: BlueprintOptions, params: unknown[]): SubmittableExtrinsic<ApiType, BlueprintSubmittableResult<ApiType>> => {
     return this.api.tx.contracts
       .instantiate(
         value,
         gasLimit,
+        storageDepositLimit,
```
jacogr, line 60: Backwards compat is always a HUGE issue. So, in this case (luckily) the number of paramaters changed, so it would be a case of -

```js
this.api.tx.contracts.instantiate.meta.arags.length === 6
  ? this.api.tx.contracts.instantiate(value, gasLimit, storageDepositLimit, codeHash, params, salt)
  : this.api.tx.contracts.instantiate(value, gasLimit, codeHash, params, salt)
```

statictype:
it seems that when i do `yarn build` locally some of my changes are reverted. 
i couldn't find entry point where all the interfaces are generated from. do I need to modify this too? https://github.com/polkadot-js/api/blob/master/packages/types-support/src/metadata/v14/substrate-types.json#L14629

statictype:
actually, i think something like https://github.com/polkadot-js/api/pull/4286 will do the trick. but don't know what the bump process is exactly

jacogr:
It is supposed to be documented here, but don't think it had much attention lately - 

https://github.com/polkadot-js/api/blob/master/packages/types/src/metadata/README.md

Small differences from the above - for substrate it is in v14/substrate.ts (instead of static.ts, kusama/polkadot dev now have their own as well)

Once you run `yarn builld:interfaces` it will re-generate all the TS stuff, the `yarn test:one packages/types/src/metadata/v14` will do the rest of the checks.

packages/api-contract/src/base/Blueprint.ts
statictype: is this change necessary in case api-contract version is different than api version? 
it would mean overloading the generated interfaces somehow.

statictype:
is there a way i can run local apps with local api to verify my changes?

packages/api-contract/src/base/Blueprint.ts
jacogr: IT is needed since it depends on the chain it is connected to - so it needs to be detected via on-chain metadata. You can have the most recent API with an ancient node.

jacogr:
To check local changes -

- assume you have a root folder, let's call it ROOT
- checkout ROOT/api and ROOT/apps
- cd api
- `yarn polkadot-dev-copy-to apps`

That will build and copy the changes to the apps UI, replacing what is there

- cd ../apps
- `yarn start`

packages/api-contract/src/base/Blueprint.ts
statictype: i see... tried a few things yesterday but unable to figure out how to do that. is there a PR where you did something similar that you can point me to?

jacogr: So, it is a bit of a nightmare, since these things are short-lived - so I generally allow 3 months leeway before ripping it out, so the examples that were here (contracts specific) have been removed.

Will dig around the older versions and find it. (There is a similar approach in the apps UI as well, which tries to also give around 3 months.... but generally they end up living a "bit" longer)

jacogr: Here is an example for the apps UI that does the same thing -

https://github.com/polkadot-js/apps/blob/master/packages/page-democracy/src/Overview/Fasttrack.tsx#L59-L63

(There are millions of them in there... sadly, mostly in the API it is always only contracts since there are convenience wrappers around... will find an old release having some)

statictype: got it now! you have to `ts-ignore` the old style

jacogr: Yes, it is slightly messy code-wise :(

… but better than dealing with the “it doesn’t work-anymore” flood of support…

statictype: I think i got everything but still can't call. I get this   rpc core error
https://github.com/polkadot-js/apps/pull/6657#issuecomment-991983593

There is also this contract query, but not sure if adding an extra param there would solve it and

statictype:
instantiating and calling contracts work now on old and new node
i had to hide `storage_deposit` and `storage_deposit_limit` completely from rpc type definitions because of back compatibility. 
when the type comes from metadata the value is hardcoded to null

it works, but the subtracted fee is not visible in the UI at all.
to properly support these new features more things need to be done, but this might work for now. what do you think? @jacogr

still to do
- getting the estimation for storage deposit (and gas limit) when instantiating - not a small issue
- fix InputBalance problem with accepting default values in some cases (works only in the upload modal) and actually showing the input

athei:
> getting the estimation for storage deposit (and gas limit) when instantiating - not a small issue

Can we do this in a follow up? This might be out of scope (we hadn't had this before). There is a separate issue for that.

statictype:
a follow-up was exactly what i was thinking

jacogr (approved):
Seems sane :) Thanks an absolute million, this is great.

PR #4363, opened 2021-12-20
don't throw error at `onSocketClose`

### LINKED ISSUE #4348: onSocketClose throw error
https://github.com/polkadot-js/api/blob/7b8a6e89d978b8dca679715ca70f2c515075ecee/packages/rpc-provider/src/ws/index.ts#L384-L385

When a ws connection was closed, all the hanging requests won't be sent and will be rejected(throw error), I think throw error is not expected in this function, because you handled error at 
https://github.com/polkadot-js/api/blob/7b8a6e89d978b8dca679715ca70f2c515075ecee/packages/rpc-provider/src/ws/index.ts#L388-L394

So I suggest changging this line to

```js
eraseRecord(this.#handlers, (h) => {
  h.callback(error, undefined)).catch(() => {
    // does not throw 
  });
}
```

It's hard to simulate hanging request, attach a picture about how I construct this error
image

xcaptain:
fixes #4348

### DIFF
--- a/packages/rpc-provider/src/ws/index.ts
+++ b/packages/rpc-provider/src/ws/index.ts
@@ -382,7 +382,14 @@ export class WsProvider implements ProviderInterface {
     this.#emit('disconnected');
 
     // reject all hanging requests
-    eraseRecord(this.#handlers, (h) => h.callback(error, undefined));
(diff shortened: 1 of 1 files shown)


jacogr (approved):
Thank you.

PR #4370, opened 2021-12-20
Storage deposit limit changes

kwingram25:
- Update blueprint and contract options w storageDepositLimit
- Update substrate contracts definitions


packages/types/src/interfaces/contracts/types.ts
```diff
@@ -436,6 +439,15 @@ export interface ScheduleTo264 extends Struct {
 /** @name SeedOf */
 export interface SeedOf extends Hash {}
 
+/** @name StorageDeposit */
+export interface StorageDeposit extends Enum {
+  readonly isRefund: boolean;
+  readonly asRefund: Balance;
+  readonly isCharge: boolean;
+  readonly asCharge: Balance;
+  readonly type: 'Refund' | 'Charge';
```
achimcc, line 448: Hey, I've a question here: Perhaps it is because of compatibility with legacy code. But why do we need Booleans `isRefund` and `asRefund` when we have `type` which can be either `'Refund'` or `'Charge'`? And why do we need two different Balances? From my intuition, I'd say wen only need the two fields `type: 'Refund' | 'Charge'` and `balance: Balance`?

statictype: these types are auto generated from substrate metadata

statictype:
not sure i understand what this PR fixes exactly.
the next logical step is to implement the `instantiate` RPC call, then we can update contract options so that people can use the thing properly.

jacogr (approved):
lgtm.

PR #4385, opened 2021-12-23
resolve promise on send failed

xcaptain:
Related to #4348 

If a socket connection was closed, the hanging promises should be resolved rather than rejected, because we didn't define `.catch` when execute this promise, and application code can't access this promise, make it a unhandledrejection.

### DIFF
--- a/packages/rpc-provider/src/ws/index.ts
+++ b/packages/rpc-provider/src/ws/index.ts
@@ -288,10 +288,9 @@ export class WsProvider implements ProviderInterface {
         const id = this.#coder.getId();
 
         const callback = (error?: Error | null, result?: T): void => {
-          if (error) {
-            this.#emit('error', error);
-          }
-          resolve(result as T);
+          error
+            ? reject(error)
+            : resolve(result as T);
         };
 
         l.debug(() => ['calling', method, json]);
--- a/packages/rpc-provider/src/ws/index.ts
+++ b/packages/rpc-provider/src/ws/index.ts
@@ -288,9 +288,10 @@ export class WsProvider implements ProviderInterface {
         const id = this.#coder.getId();
 
         const callback = (error?: Error | null, result?: T): void => {
-          error
-            ? reject(error)
-            : resolve(result as T);
+          if (error) {
+            this.#emit('error', error);
+          }
+          resolve(result as T);
         };
 
         l.debug(() => ['calling', method, json]);
@@ -384,7 +385,7 @@ export class WsProvider implements ProviderInterface {
     // reject all hanging requests
     eraseRecord(this.#handlers, (h) => {
       try {
-        h.callback(error, undefined);
+        h.callback(null, undefined);
       } catch (err) {
         // does not throw
         l.error(err);


jacogr:
When it fails, it needs to be rejected. We cannot resolve with invalid values here.

xcaptain:
But the purpose of `onSocketClose` is to handle connection disconnected error, in this error handler we produce error again, I think that's not good.
Errors rejected from this callback can't be catched at user side, the only way to capture this error is use `process.on(uncaughtException)` which I think is not good.

xcaptain:
@jacogr I updated my pr, reverted back to use reject, can you take a look again? `onSocketClose` should produce error so pass a `null` to callback.
I still recommend you to think twice about using `emit` to produce error here, because user can use `provider.on('error')` to handle the error.

jacogr:
You cannot resolve the promise with invalid values at all. So this is what now happens in the user code -

```js
const a = await api.something.somewhere(1223);

// now this needs to be applied everywhere to all user code ...
//
// a is either a 
//. - correct value, 
//. - undefined (e.g. raw queries) or 
// - the default value for the type
//
// assume only undefined, the simplest one
if (a === undefined) {
  // handle it
} else {
  console.log(a.toNumber();
}
```

In the case where it ends up being the default value, you are in a worse position, i.e. you query for block 123, but retrieve an empy header back, aka the default for that type.

So yes, in all cases the promise should reject, not return invalid/unexpected values.

xcaptain:
Understood, you are right, we should use reject here, we just need to make sure `this.#handlers` won't reject other promises. What do think of my current change, simply eat all errors at `onSocketClose`

xcaptain:
@jacogr Do you think we should put 
image
to the end of `onSocketClose`? Not sure whether this reject would trigger the `assert`
image

jacogr:
Not passing the error through means we still resolve with `undefined` which is not properly defined behavior. 

The user would have to do error handling when using the API, i.e.

```js
async main () {
  const api = await Api.create(...);
  ...
}

main().catch((error) => {
  console.error(error);
  process.exit(-1);
});
```

xcaptain:
Of course we can use `main.catch()`, but I think `main.catch`  means the process will exit, just clean some hanging request doesn't need to make the whole process exit. Making `onSocketClose` not throwing errors will be more useful, because people can do reconnect again, without restart the process.

jacogr:
Ok, so you need to handle promises - it has two states. Like.an old-style node callback has error & result, a promise has an error and result state. It needs to be handled by the user.

The `main.catch` is obviously a catch-all, generally you want finer control on a per-action level, i.e. some can fail and just be retried, others may be fatal.

Trying to hack the API to return invalid responses and hide the fact that there is a failure is not the right path.

xcaptain:
Ok, I see, In my application code, when I use `main.catch` and catched an error, I can choose to restart the process or ignore it if the error message contains `disconnected from`

xcaptain:
I think this pr can be closed, thank you anyway.

PR #4423, opened 2022-01-07
Instantiate contract RPC, storage deposit support, merge Code and Blueprint

kwingram25:
- Add instantiate rpc query to Blueprint.tsx
- Use Blueprint for new uploads and code hashes, remove Code
- Storage deposit related additions


packages/api-contract/src/base/types.ts
```diff
@@ -4,40 +4,58 @@
...
+  (origin: AccountId | string | Uint8Array, value: bigint | BN | string | number, gasLimit: bigint | BN | string | number, source: string | Record<string, unknown> | CodeSource | Uint8Array, ...params: unknown[]): ContractResult<ApiType, ContractInstantiateOutcome>;
+  (origin: AccountId | string | Uint8Array, value: bigint | BN | string | number, gasLimit: bigint | BN | string | number, storageDepositLimit: bigint | BN | string | number, source: string | Record<string, unknown> | CodeSource | Uint8Array, ...params: unknown[]): ContractResult<ApiType, ContractInstantiateOutcome>;
+}
+
+export interface BlueprintTx<ApiType extends ApiTypes> {
   (options: BlueprintOptions, ...params: unknown[]): SubmittableExtrinsic<ApiType>;
   (value: bigint | string | number | BN, gasLimit: bigint | string | number | BN, ...params: unknown[]): SubmittableExtrinsic<ApiType>;
+  (value: bigint | string | number | BN, gasLimit: bigint | string | number | BN, storageDepositLimit: bigint | BN | string | number, ...params: unknown[]): SubmittableExtrinsic<ApiType>;
```
jacogr, line 20: We don't want these, i.e. we don't extend these into the future - it is purely for backwards compat. We expect everybody to use the options flavor. (Additionally this breaks compat completely, we don't like breaking user-space)

TL;DR This flavor _should_ have been removed in a previous major bump.

packages/api-contract/src/base/types.ts
```diff
@@ -4,40 +4,58 @@
...
 }
 
 export interface ContractQuery<ApiType extends ApiTypes> {
-  (origin: AccountId | string | Uint8Array, options: ContractOptions, ...params: unknown[]): ContractCallResult<ApiType, ContractCallOutcome>;
-  (origin: AccountId | string | Uint8Array, value: bigint | BN | string | number, gasLimit: bigint | BN | string | number, ...params: unknown[]): ContractCallResult<ApiType, ContractCallOutcome>;
+  (origin: AccountId | string | Uint8Array, options: ContractOptions, ...params: unknown[]): ContractResult<ApiType, ContractCallOutcome>;
+  (origin: AccountId | string | Uint8Array, value: bigint | BN | string | number, gasLimit: bigint | BN | string | number, ...params: unknown[]): ContractResult<ApiType, ContractCallOutcome>;
+  (origin: AccountId | string | Uint8Array, value: bigint | BN | string | number, gasLimit: bigint | BN | string | number, storageDepositLimit: bigint | BN | string | number, ...params: unknown[]): ContractResult<ApiType, ContractCallOutcome>;
```
jacogr, line 26: As above and for the rest of the file.

packages/types-augment/src/registry.ts
```diff
@@ -514,9 +519,6 @@ declare module '@polkadot/types/types/registry' {
     InstanceDetails: InstanceDetails;
     InstanceId: InstanceId;
     InstanceMetadata: InstanceMetadata;
-    InstantiateRequest: InstantiateRequest;
-    InstantiateReturnValue: InstantiateReturnValue;
-    InstantiateReturnValueTo267: InstantiateReturnValueTo267;
```
jacogr, line 519: The issue with removing these is that we break all code of those using overrides. It is fine adding new, but we cannot just drop old aliasses.

packages/types/src/interfaces/contracts/rpc.ts
```diff
@@ -28,12 +28,12 @@ export default {
...
       {
         name: 'request',
-        type: 'InstantiateRequest'
+        type: 'ContractInstantiateRequest'
       },
       {
         name: 'at',
         type: 'BlockHash',
-        isHstoric: true,
+        isHistoric: true,
```
jacogr, line 36: +1

jacogr (commented):
As it stands it break compat completely, i.e. forces all users of the API to re-write their code. (This includes the apps UI).

So it will have to wait until a major release - I gave it some comments already on areas that should be addressed, but it will have to wait a bit for a full review until there is a gap to evaluate against what is breaking (and there is some gap to look at impacts on other polkadot-js code that use this)

packages/api-contract/src/base/types.ts
kwingram25: So assuming this is for next major release, it's ok to remove both and similar ones?

packages/types-augment/src/registry.ts
kwingram25: Would it be possible to just mention it as a breaking change when released?

jacogr:
Closing, cannot be merged as-is due to above comments.

PR #4488, opened 2022-01-22
Extract short type aliases for Portable types

jacogr:


packages/types/src/metadata/PortableRegistry/PortableRegistry.ts
```diff
@@ -202,25 +205,56 @@ function removeDuplicateNames (lookup: PortableRegistry, names: [number, string
...
+        const [,,, others] = list[j];
+
+        for (let a = 1; a < aliases.length; a++) {
+          const alias = aliases[a];
+
+          if (!dupes.includes(alias) && others.includes(alias)) {
+            dupes.push(alias);
+          }
+        }
+      }
```
spirinvladimir, line 239: x2 faster:

```suggestion
    for (let j = i + 1; j < list.length; j++) {
      const [,,, others] = list[j];

      for (let a = 1; a < aliases.length; a++) {
        const alias = aliases[a];

        if (!dupes.includes(alias) && others.includes(alias)) {
          dupes.push(alias);
        }
      }
```

packages/types/src/metadata/PortableRegistry/PortableRegistry.ts
```diff
@@ -202,25 +205,56 @@ function removeDuplicateNames (lookup: PortableRegistry, names: [number, string
...
+    }
+  }
+
+  return list.map(([lookupIndex, name, params, aliases]) => [
+    lookupIndex,
+    name,
+    params,
+    aliases.slice(1).filter((a) => !dupes.includes(a))
+  ]);
```
spirinvladimir, line 248: Less memory consuming by mutate array items:
```suggestion
  return list.map((item) => {
    item[3] = item[3].slice(1).filter((a) => !dupes.includes(a))
    return item
  });
```

kmk142789 (approved):

Issue #4493, opened 2022-01-24
When and how are the node types augmented


vikiival:
I found the amazing bug in  https://github.com/kodadot/nft-gallery/issues/2058 

When the user for the first time connects to statemine/westmint node the `uniques` module is undefined 
e.g
```js
console.log(api.query.uniques) // undefined
```

`api.isReady` also has no `uniques` module

When and how are the node types augmented and if it's possible to listen to this change?

Other:
We use `"@polkadot/api": "^7.4.1"` and type augmentation is `import '@polkadot/api-augment'`

jacogr:
I would suggest looking at the 7.x release notes. (Major semver bump)

jacogr:
See https://github.com/jacogr/api-test-4493 (logging both `api.query.uniques` & `api.tx.uniques`

vikiival:
I found the bug, it's related to how we handle endpoint connection

> Api is enhanced with the modules for the runtime on load, metadata retrieval.

but still it would be cool to hove
something
`api.on('enhance', updatedApi => {})`

PR #4497, opened 2022-01-25
Emit decorated event on API (incl. upgrades)

jacogr:
As per https://github.com/polkadot-js/api/issues/4493#issuecomment-1020224334


yangwao:
@vikiival 👀🧹

vikiival:
@jacogr is the best!

Issue #4501, opened 2022-01-25
Typegen Issue with `sudo` (v7.4.1)


monitz87:
Having the same issue as #4254 on a custom chain

Snippet:

```
const transaction = api.tx.sudo.sudo(
  api.tx.rewards.setItnRewardStatus(itnAddress.publicKey, status)
);
```

Error:
```
Argument of type 'SubmittableExtrinsic<"promise", ISubmittableResult>' is not assignable to parameter of type 'string | Uint8Array | PolymeshRuntimeCiRuntimeCall | { System: any; } | { Babe: any; } | { Timestamp: any; } | { Indices: any; } | { Authorship: any; } | ... 35 more ... | { ...; }'.
  Type 'SubmittableExtrinsic<"promise", ISubmittableResult>' is missing the following properties from type 'PolymeshRuntimeCiRuntimeCall': isSystem, asSystem, isBabe, asBabe, and 87 more.
```

Augmented type for the sudo extrinsic:

```
sudo: AugmentedSubmittable<(call: PolymeshRuntimeCiRuntimeCall | { System: any } | { Babe: any } | { Timestamp: any } | { Indices: any } | { Authorship: any } | { Balances: any } | { Identity: any } | { CddServiceProviders: any } | { PolymeshCommittee: any } | { CommitteeMembership: any } | { TechnicalCommittee: any } | { TechnicalCommitteeMembership: any } | { UpgradeCommittee: any } | { UpgradeCommitteeMembership: any } | { MultiSig: any } | { Bridge: any } | { Staking: any } | { Session: any } | { Grandpa: any } | { ImOnline: any } | { Sudo: any } | { Asset: any } | { CapitalDistribution: any } | { Checkpoint: any } | { ComplianceManager: any } | { CorporateAction: any } | { CorporateBallot: any } | { Pips: any } | { Portfolio: any } | { ProtocolFee: any } | { Scheduler: any } | { Settlement: any } | { Statistics: any } | { Sto: any } | { Treasury: any } | { Utility: any } | { Base: any } | { ExternalAgents: any } | { Relayer: any } | { Rewards: any } | { TestUtils: any } | string | Uint8Array) => SubmittableExtrinsic<ApiType>, [PolymeshRuntimeCiRuntimeCall]>;
```

I read through the linked issue but didn't find any solution there. Most of the links posted there lead to 404 error pages

[...]

PR #4502, opened 2022-01-25
Non-aliased CallEvent type extraction

jacogr:
Closes https://github.com/polkadot-js/api/issues/4501

Remove the alias extraction for Call/Event and replace it by extracting param types and known indexes. 

This _should_ be more robust... until we don't find one of  `sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic<..>` and `frame_system::EventRecord<..>` types and we are back to the drawing board again...

However this should take us slightly further... hopefully...

Tested against known metadata v14 parachains on Kusama & Polkadot.

### DIFF
--- a/packages/types/src/metadata/PortableRegistry/PortableRegistry.ts
+++ b/packages/types/src/metadata/PortableRegistry/PortableRegistry.ts
@@ -297,7 +297,7 @@ function extractAliases (params: Record<string, SiTypeParameter[]>, isContract?:
 
     alias[type.unwrap().toNumber()] = 'Call';
   } else if (hasParams && !isContract) {
-    l.warn('Unable to determine runtime Call type, unable to inspect sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic');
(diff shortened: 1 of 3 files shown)


packages/types/src/metadata/PortableRegistry/PortableRegistry.ts
```diff
@@ -290,6 +285,33 @@ function registerTypes (lookup: PortableRegistry, lookups: Record<string, string
...
+  const hasParams = Object.keys(params).length !== 0;
+  const alias: Record<number, string> = {};
+
+  if (params.SpRuntimeUncheckedExtrinsic) {
+    // Address, Call, Signature, Extra
+    const [, { type }] = params.SpRuntimeUncheckedExtrinsic;
+
+    alias[type.unwrap().toNumber()] = 'Call';
+  } else if (hasParams && !isContract) {
+    l.warn('Unable to determine runtime Call type, unable to inspect sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic');
```
jacogr, line 300:
```suggestion
    l.warn('Unable to determine runtime Call type, cannot inspect sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic');
```

PR #4505, opened 2022-01-26
Fix memory leak on raw rpc call

r0t0r-r0t0r:
While processing a raw rpc call an observable is removed from the wrong cache. It causes a memory leak.


jacogr (commented):
Good catch. I believe the same would happen for the subscription case in the next method.

jacogr (approved):
Thank you.

PR #4509, opened 2022-01-27
Add derive.assets.all

MiZiet:


packages/api-derive/src/assets/all.ts
```diff
@@ -0,0 +1,75 @@
...
+import type { FetchedAssetsEntries, FetchedAssetsIddEntries, FetchedAssetsMetadataEntries } from '@polkadot/api-derive/assets/types';
+import type { DeriveApi } from '@polkadot/api-derive/types';
+import type { StorageKey } from '@polkadot/types';
+import type { EventRecord, Hash } from '@polkadot/types/interfaces';
+import type { FrameSystemEventRecord, PalletAssetsAssetDetails, PalletAssetsAssetMetadata } from '@polkadot/types/lookup';
+import type { u32, Vec } from '@polkadot/types-codec';
+
+import { combineLatest, concat, EMPTY, map, Observable, of, switchMap, take } from 'rxjs';
+
+export interface Asset extends PalletAssetsAssetDetails, PalletAssetsAssetMetadata{
```
tt-krzysztof, line 13: Looking at other derives, I believe the convention here is to:
* name it `DeriveAsset`
* put it in `./assets/types`

packages/api-derive/src/assets/all.ts
```diff
@@ -0,0 +1,75 @@
...
+  return result;
+}
+
+/**
+ * @name all
+ * @returns An array containing all assets with metadata
+ */
+
+export function all (api: DeriveApi): Observable<Asset[]> {
+  const initBlockHash = api.rpc.chain.subscribeNewHeads()
```
tt-krzysztof, line 55:
```suggestion
  const initialBlockHash = api.rpc.chain.subscribeNewHeads()
```

packages/api-derive/src/assets/assets.spec.ts
```diff
@@ -0,0 +1,238 @@
...
+              phase: { ApplyExtrinsic: 1 },
+              topics: [] as unknown as Vec<Hash>
+            } as unknown as EventRecord] as unknown as Vec<EventRecord>, { createdAtHash: api.createType('Header', { number: new BN(2) }).hash })
+          ]).pipe(concatMap((val) => of(val).pipe(delay(100))))
+        }
+      }
+    } as unknown as DeriveApi;
+
+    // eslint-disable-next-line no-return-assign
+    sub = all(mockedApiWithNonAssetEvent).subscribe((value) => assets = value);
```
tt-krzysztof, line 226: With this change we won't need the eslint-disable.
```suggestion
    sub = all(mockedApiWithNonAssetEvent).subscribe((value) => { assets = value });
```

packages/api-derive/src/assets/all.ts
```diff
@@ -0,0 +1,75 @@
+// Copyright 2017-2022 @polkadot/api-derive authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import type { FetchedAssetsEntries, FetchedAssetsIddEntries, FetchedAssetsMetadataEntries } from '@polkadot/api-derive/assets/types';
+import type { DeriveApi } from '@polkadot/api-derive/types';
```
tt-krzysztof, line 5: Prefer local import

krzysztof-jelski, line 5: Prefer local import

packages/api-derive/src/assets/all.ts
krzysztof-jelski, line 13: Looking at other derives, I believe the convention here is to:
* name it `DeriveAsset`
* put it in `./assets/types`.

Also, perhaps some types can be converted to js types instead of scale/substrte. Eg. string for asset name.

packages/api-derive/src/assets/all.ts
krzysztof-jelski, line 55:
```suggestion
  const initialBlockHash = api.rpc.chain.subscribeNewHeads()
```

packages/api-derive/src/assets/types.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2017-2022 @polkadot/api-derive authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import type { Option, StorageKey, u32 } from '@polkadot/types';
+import type { AssetId } from '@polkadot/types/interfaces';
+import type { PalletAssetsAssetDetails, PalletAssetsAssetMetadata } from '@polkadot/types/lookup';
+
+export type FetchedAssetsEntries = [StorageKey<[AssetId]>, Option<PalletAssetsAssetDetails>][];
+export type FetchedAssetsMetadataEntries = [StorageKey<[AssetId]>, PalletAssetsAssetMetadata][];
+export type FetchedAssetsIddEntries = StorageKey<[u32]>[]
```
krzysztof-jelski, line 10: Would be useful if we could keep those types local to the module and export only the `Derive...` types.

packages/api-derive/src/assets/all.ts
```diff
@@ -0,0 +1,70 @@
...
+    api.query.system.events()
+      .pipe(switchMap((events: Vec<EventRecord>) => extractAssetEventsHash(events)))
+  )
+
+    .pipe(
+      switchMap((blockHash: Hash) =>
+        combineLatest([
+          api.query.assets.asset.entriesAt(blockHash),
+          api.query.assets.metadata.entriesAt(blockHash),
+          api.query.assets.asset.keysAt(blockHash)
```
jacogr, line 65: The `entriesAt` 2 lines above includes this.

In the UI why the 2 are split - the keys is a (relative) fast operation, the entries is quite a lot of overhead.

packages/api-derive/src/assets/all.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * @name all
+ * @returns An array containing all assets with metadata
+ */
+
+export function all (api: DeriveApi): Observable<DeriveAsset[]> {
+  const initialBlockHash = api.rpc.chain.subscribeNewHeads()
+    .pipe(take(1))
+    .pipe((val) => val.pipe(map(({ hash }) => hash)));
```
jacogr, line 52: `rpc.chain.getHeader` returns the latest, no sub.

MiZiet: Do you mean changing to:
`const initBlockHash = api.rpc.chain.getHeader().pipe(val => val.pipe(map(({ hash }) => hash)))`?

packages/api-derive/src/assets/all.ts
MiZiet: ✅ Removed `keysAt` call

packages/api-derive/src/assets/all.ts
MiZiet: ✅

MiZiet:
By the way we've plugged this into useAssets hook in our [app](https://github.com/EthWorks/statemine-asset-creator), it seems to work as expected. 
Here's the [setup](https://github.com/EthWorks/statemine-asset-creator/tree/main/tools/local-statemint-launch) that we use for local relay chain and parachain testing, where the problem with observing metadata changes occured

packages/api-derive/src/assets/all.ts
jacogr: Indeed. `api.rpc.chain.getHeader().pipe(map(({ hash }) => hash))`

packages/api-derive/src/assets/all.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  return concat(
+    initialBlockHash,
+    api.query.system.events()
```
jacogr, line 72: The issue here is that you have no idea that the events are actually at that block. You are not getting the actual hash for the events, which is in `createdAt`

So basically what you are doing is monitoring 2 different subscriptions, that could come in any order and then assuming that they match.

jacogr: Ignore, initial concat...

packages/api-derive/src/assets/all.ts
```diff
@@ -0,0 +1,85 @@
...
+      .pipe(switchMap((events: Vec<EventRecord>) => extractAssetEventsHash(events)))
+  )
+
+    .pipe(
+      switchMap((blockHash: Hash) =>
+        combineLatest([
+          api.query.assets.asset.entriesAt(blockHash),
+          api.query.assets.metadata.entriesAt(blockHash)
+        ])
+      )
```
jacogr, line 82: This is a very, very heave operation. Basically you refresh everything. The events actually have the ids of those changed, so should pin-point to those and only update those.

(crowdloans is an example of this - it tries to not get everything each time something has changed, rather it only refreshes based on the events received. As the state grows, this approach would just become slower and slower...)

jacogr (changes requested):
As commented - we need a scalpel, not a hammer :)

packages/api-derive/src/assets/all.ts
krzysztof-jelski: Understood. Though still not sure how to solve it. Reason we use `entriesAt` is initially that `multi` did not get the updates from latest (non-finalized) block (https://github.com/polkadot-js/apps/issues/6874#issuecomment-1022038369). As there is no `entriesAt(blockHash, keys)`, I understand we need to load all, and then update state based on data from events (asset and metadata changes)?

jacogr: So if we can get a minimal reproducable sample with API code, can actually log that on Substrate. (There could obviously be an issue with the apps UI implementation itself)

jacogr: Take a look at this issue, which is also parachain-related - https://github.com/paritytech/substrate/issues/10743

Would love a sample on a single connection where -

(a) event is available
(b) query for that item is not updated

I honestly cannot replicate this on a stand-alone node.

jacogr:
Closing as-is. (However, this _may_ not be an issue anymore in the apps UI, I know it had some significant changes on asset handling a while ago, which _could_ have benefited here)

PR #4539, opened 2022-02-07
Add `isHistorical` field to relevant eth rpc requests

stwiname:
Adds `isHistorical` to `BlockNumber` parameters for eth rpc requests that support querying at a point in time.


jacogr (approved):
Thank you! 

Not sure why these were missed initially, possibly due to the naming (aka it is named "at" elsewhere, so could be the search that misfired...)

PR #4540, opened 2022-02-07
Fix `isHistorical` to `isHistoric`

stwiname:
Too early in the morning for me. Fixes my mistake in https://github.com/polkadot-js/api/pull/4539


jacogr (approved):

jacogr:
Need to. look at these... the TS checked should have picked it up, but since `objectSpread` is used, it doesn't. There must be some magic to apply to that utility so it works.

PR #4541, opened 2022-02-07
Update contract types and rpc

kwingram25:
Subset of #4423, only updating contract type definitions and RPC call

Not sure what work needs to accompany this for apps. The instantiate request struct fails on older Substrate chains as it doesn't expect storageDepositLimit.


kwingram25:
@jacogr Noticed that in some cases the rpc for contracts.instantiate returns an ExitError (ie OutOfGas), how would this be represented in the definitions?

jacogr:
Umpf... that is problematic, since generally the RPCs should not break, i.e. they are always backwards compatible. 

I guess in this case it should be up to the users to specify the correct (older) types.

kwingram25:
Anything more to do on our end for this?

packages/types/src/interfaces/contracts/definitions.ts
```diff
@@ -249,22 +255,35 @@ export default {
...
+      result: 'InstantiateReturnValue'
+    },
     InstantiateReturnValueTo267: {
       result: 'ExecReturnValue',
       accountId: 'AccountId',
       rentProjection: 'Option<RentProjection>'
     },
-    InstantiateReturnValue: {
-      _fallback: 'InstantiateReturnValueTo267',
+    InstantiateReturnValueOk: {
```
jacogr, line 277: There is no previous type definition for this one, it is just replaced?

packages/types/src/interfaces/contracts/definitions.ts
```diff
@@ -239,7 +245,7 @@ export default {
       value: 'Balance',
       gasLimit: 'Gas',
       storageDepositLimit: 'Option<Balance>',
-      code: 'Bytes',
+      code: 'CodeSource',
```
jacogr, line 248: Also needs a previous version here.

jacogr (commented):
Would just have older versions for the replaced types as well.

packages/types/src/interfaces/contracts/definitions.ts
kwingram25: I believe it was always supposed to be an enum, and Bytes was just a placeholder? Anyway, it seems like enum definitions can't have fallbacks at the moment, can they? I tried and it just made a struct with _enum and _fallback fields

kwingram25:
Added fallback support for enums and sets. Any I missed?

jacogr (approved):
Brilliant!

PR #4571, opened 2022-02-17
Fix getting block registry when historical type isn't BlockHash

stwiname:
In relation to https://github.com/polkadot-js/api/pull/4539, I was unaware of this part of the code.

Alternative solution would be the ability to get the registry via block number but that would require an extra RPC request. Happy to change the implementation if necessary


jacogr:
The isHistoric flag is just meant for areas where the at is a blockhash. So the correct solution is to remove where it is applied incorrectly.

stwiname:
Understood, I will remove the `isHistoric` on the incorrect fields for the time being and create a new PR for proper BlockNumber support

jacogr:
Thank you. Yes, it would be great if these can indeed work correctly registry decoration with the correct blockHash.

jacogr (approved):

PR #4574, opened 2022-02-18
Add support for historical BlockNumber rpc methods

stwiname:
Hopefully this is right this time.

I'm not sure if there's any need to cache the result of getting block hash from block number.

### DIFF
--- a/packages/rpc-core/src/bundle.ts
+++ b/packages/rpc-core/src/bundle.ts
@@ -4,7 +4,7 @@
 import type { Observer } from 'rxjs';
 import type { ProviderInterface, ProviderInterfaceCallback } from '@polkadot/rpc-provider/types';
 import type { StorageKey, Vec } from '@polkadot/types';
-import type { BlockNumber, Hash } from '@polkadot/types/interfaces';
+import type { Hash } from '@polkadot/types/interfaces';
 import type { AnyJson, AnyNumber, Codec, DefinitionRpc, DefinitionRpcExt, DefinitionRpcSub, Registry } from '@polkadot/types/types';
 import type { Memoized } from '@polkadot/util/types';
 import type { RpcInterfaceMethod } from './types';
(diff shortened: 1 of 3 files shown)


packages/rpc-core/src/bundle.ts
```diff
@@ -4,8 +4,8 @@
 import type { Observer } from 'rxjs';
 import type { ProviderInterface, ProviderInterfaceCallback } from '@polkadot/rpc-provider/types';
 import type { StorageKey, Vec } from '@polkadot/types';
-import type { Hash } from '@polkadot/types/interfaces';
-import type { AnyJson, Codec, DefinitionRpc, DefinitionRpcExt, DefinitionRpcSub, Registry } from '@polkadot/types/types';
+import type { BlockNumber, Hash } from '@polkadot/types/interfaces';
```
jacogr, line 7:
```suggestion
import type { Hash } from '@polkadot/types/interfaces';
```

As per next comment.

packages/rpc-core/src/bundle.ts
```diff
@@ -85,6 +85,7 @@ export class RpcCore {
   #registryDefault: Registry;
 
   #getBlockRegistry?: (blockHash: Uint8Array) => Promise<{ registry: Registry }>;
+  #getBlockHash?: (blockNumber?: BlockNumber | AnyNumber | undefined) => Promise<Uint8Array | string | null | undefined>;
```
jacogr, line 88:
```suggestion
  #getBlockHash?: (blockNumber: AnyNumber) => Promise<Uint8Array>;
```

`AnyNumber` will cater for it. Additionally this should always return a value. (It not, it will throw)

jacogr (commented):
Thank you! Looks great, some tiny comments.

packages/rpc-core/src/bundle.ts
```diff
@@ -146,6 +147,15 @@ export class RpcCore {
     });
   }
 
+  /**
+   * @description Sets a function to resolve block hash from block number
+   */
+  public setResolveBlockHash (resolveBlockHash: (blockNumber: BlockNumber | AnyNumber | undefined) => Promise<Uint8Array | string | null | undefined>): void {
```
jacogr, line 153:
```suggestion
  public setResolveBlockHash (resolveBlockHash: (blockNumber: AnyNumber) => Promise<Uint8Array>): void {
```

jacogr (approved):

PR #4637, opened 2022-03-08
Add rpc error fields

### LINKED ISSUE #4623: Fields of RPC error are not exposed
On our node's rpc endpoint, we'd like to provide some context along with the error code:
```rust
RpcError {
    code: ErrorCode::InternalError,
    message: format!("Transaction failed: {}", e),
    data: ErrorContext { should_retry: false }.into_value(),
}
```

However it appears these fields are not accessible to client-side user code. Instead they are [merged ](https://github.com/polkadot-js/api/blob/master/packages/rpc-provider/src/coder/index.ts#L68) into the error message [immediately](https://github.com/polkadot-js/api/blob/master/packages/rpc-provider/src/coder/index.ts#L36). Is this the intended behavior? If I'm missing something here please let me know.

Otherwise I'd be happy to open a PR.

intendednull:
Re-adds the error extension logic that allows access to rpc error fields - https://github.com/polkadot-js/common/tree/fbc38aabc2b6c554d616d62ae7c84916eeac63e3/packages/util/src/ext

fixes #4623

### DIFF
--- a/packages/rpc-provider/src/coder/index.ts
+++ b/packages/rpc-provider/src/coder/index.ts
@@ -3,8 +3,9 @@
 
 import type { JsonRpcRequest, JsonRpcResponse, JsonRpcResponseBaseError } from '../types';
 
-import RpcError from '@polkadot/rpc-core/error';
 import { assert, isNumber, isString, isUndefined, stringify } from '@polkadot/util';
+
+import RpcError from './error';
 
 function formatErrorData (data?: string | number): string {
   if (isUndefined(data)) {
@@ -69,8 +70,9 @@ export class RpcCoder {
   private _checkError (error?: JsonRpcResponseBaseError): void {
     if (error) {
       const { code, data, message } = error;
+      const fmtMessage = `${code}: ${message}${formatErrorData(data)}`;
 
-      throw new RpcError(message, code, data);
+      throw new RpcError(fmtMessage, code, data);
     }
   }
 }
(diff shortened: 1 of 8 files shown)


intendednull:
@jacogr got it started, however unsure how you want this organized. You mentioned it should be in `rpc-core`, but it seems like the error needs to be thrown in `rpc-provider`. Adding a dependency to `rpc-core` from `rpc-provider` causes a dependency loop. 

Wouldn't it make sense to define the error in `rpc-provider`? Otherwise coder will be generic over error type?

jacogr:
Place it where it is actually used, e.g. provider would make sense.

packages/rpc-provider/src/coder/index.ts
```diff
@@ -68,8 +70,9 @@ export class RpcCoder {
   private _checkError (error?: JsonRpcResponseBaseError): void {
     if (error) {
       const { code, data, message } = error;
+      const fmtMessage = `${code}: ${message}${formatErrorData(data)}`;
```
intendednull, line 73: Does it make sense to keep the formatted message? If we're making the other fields available separately, maybe we shouldn't change the message.

jacogr: I prefer the message formatted. This maintains full compatibility, ie. if you don’t yet use the new fields you have exactly the same information you had previously.

jacogr (approved):
Thank you! Seems spot-on from what I can see.

PR #4662, opened 2022-03-17
Add basic provider stats

jacogr:


jacogr:
cc @wirednkod - I'll make this optional (instead of required in the `ProviderInterface` as per the initial test round), but this could also be useful for the `ScProvider` to expose (obviously once it is available over the weekend). If there are other things that do pop up that seem useful as "provider-stats", do shout.

packages/rpc-provider/src/types.ts
```diff
@@ -57,6 +57,27 @@ export interface ProviderInterface {
...
   unsubscribe (type: string, method: string, id: number | string): Promise<boolean>;
 }
+
+export interface ProviderStatsCount {
+  errors: number;
+  requests: number;
+  subscriptions: number;
+  timeout: number;
+}
+
```
jacogr, line 71:
```suggestion
```

Unused.

PR #4665, opened 2022-03-18
Allow generic Data param at RpcError

MOZGIII:


jacogr (approved):
Great. You also left the code in a much better space, i.e. `extend` is now actually type safe as well.

MOZGIII:
I've actually found an issue just now: at the `extend`, the `value` should be `value: RpcError<Data>[K]` instead of `value?: RpcError<Data>[K]`.

I minor nit, but I think it deserves a fix

MOZGIII:
See #4665

PR #4682, opened 2022-03-24
Support BTreeMap/BTreeSet in typegen

CertainLach:


jacogr (approved):
I'm hoping it has the desired effect here :) The reason why is was not added by default is because I don't have any explcit testcases for this in either Substrate, Polkadot or Kusama.

Having said that, it _should_ work with these changes, if it doesn't some extra elbow grease may be required.

PR #4700, opened 2022-03-28
fix: export substrate-connect `WellKnownChain`

josepot:
Hi @jacogr ! I'm terribly sorry about this, but I forgot to export the `WellKnownChain` enum, which IMO it should be re-exported from the `ScProvider` as it's part of the public API of its constructor.

Also, FYI [this is coming](https://github.com/paritytech/substrate-connect/pull/909) and it's looking good :relaxed:. So thanks a lot! :bow:


jacogr:
Makes sense.

jacogr (approved):

PR #4702, opened 2022-03-30
Fix types for maxExtrinsic

HackFisher:
Ref: https://github.com/paritytech/substrate/blob/41567c30da77a2546a90c1049042ebd7b0b09bc4/frame/system/src/limits.rs#L99-L102


jacogr (commented):
Generally we don't want to maintain the types, however in this case it seems like an ancient changes predating metadata v14, so would make sense to adjust, i.e. it is wrong in the v13 version as well.

jacogr (approved):

Issue #4707, opened 2022-03-31
The performance will be very poor when using apiAt heavily.


ntduan:
This is the test code：

```javascript
const { WsProvider, ApiPromise } = require("@polkadot/api");

(async () => {
  const api = new ApiPromise({
    provider: new WsProvider("wss://rpc.polkadot.io/"),
  });

  await api.isReady;

  const blocks = 70;

  const blockHashes = await Promise.all(
    [...new Array(blocks)].map((_, index) =>
      api.rpc.chain.getBlockHash(index + 1)
    )
  );

  console.time("apiAt:");
  await Promise.all(blockHashes.map((blockHash) => api.at(blockHash)));
  console.timeEnd("apiAt:");

  process.exit(0);
})();
```

For 70 consecutive blocks, each block seems to fetch the metadata and create a new api object. Here's the result I got：

```
2022-03-31 16:00:52        RPC-CORE: getMetadata(at?: BlockHash): Metadata:: No response received from RPC endpoint in 30s
2022-03-31 16:00:52        RPC-CORE: getMetadata(at?: BlockHash): Metadata:: No response received from RPC endpoint in 30s
2022-03-31 16:00:52        RPC-CORE: getMetadata(at?: BlockHash): Metadata:: No response received from RPC endpoint in 30s
2022-03-31 16:00:52        RPC-CORE: getMetadata(at?: BlockHash): Metadata:: No response received from RPC endpoint in 30s
2022-03-31 16:00:52        RPC-CORE: getMetadata(at?: BlockHash): Metadata:: No response received from RPC endpoint in 30s
2022-03-31 16:00:52        RPC-CORE: getMetadata(at?: BlockHash): Metadata:: No response received from RPC endpoint in 30s
2022-03-31 16:00:52        RPC-CORE: getMetadata(at?: BlockHash): Metadata:: No response received from RPC endpoint in 30s
```

Even if I reduce the number of concurrent requests, it will take several seconds.

Although I know that getting the metadata and decoding the metadata can be very time taking. But the fact is that the metadata for these blocks should be the same.

Wouldn't it be better to cache the api according to runtimeversion?

jacogr:
It should check the runtime version first and find a match of exiting metadata. If not there is an issue. 

https://github.com/polkadot-js/api/blob/a98e9543b85c46a9cc56fe6f874e34d5ee3d69b4/packages/api/src/base/Init.ts#L156

jacogr:
[...]

PR #4708, opened 2022-03-31
Promise cache for metadata from version (improves parallel)

jacogr:
Closes https://github.com/polkadot-js/api/issues/4707

As per https://github.com/polkadot-js/api/issues/4707#issuecomment-1084101377 - basically we already have an in-flight Promise cache for same-blockhash metadata retrievals, introduce one for parallel same-version retrievals

- [x] tested existing logic as working
- [x] tested as working for parallel

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Contributed:
 Changes:
 
 - Cater for `DispatchError` with `error` as `[u8; 4]`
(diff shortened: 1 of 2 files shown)


jacogr:
So the 70-block test in the linked issue, the as-is code takes around 4.5s when it completes. In this version the time is down to 850ms.

PR #4720, opened 2022-04-06
fix: typegen supports vectors of slices

zannis:
This PR fixes an issue where @polkadot/typegen would throw an error when  calling `getSimilarTypes` with vectors of slices such as `Vec<[u8; 4]>`.

In the existing codebase, when parsing a vector of byte slices, the `subType.info` property is only checked to be `Plain` or `Tuple`, resulting in an exception thrown, even though the expected type `Vec<U8aFixed>` has been resolved fine.

Also added a couple tests that fail without the proposed change.

I don't have the complete context of the `getSimilarTypes` function, so you feel this needs further handling feel free to advise.

### DIFF
--- a/packages/typegen/src/util/derived.ts
+++ b/packages/typegen/src/util/derived.ts
@@ -58,7 +58,9 @@ export function getSimilarTypes (registry: Registry, definitions: Record<string,
         );
 
         possibleTypes.push(`([${subs.join(', ')}])[]`);
-      } else if (subDef.info !== TypeDefInfo.VecFixed) {
+      } else if (subDef.info === TypeDefInfo.VecFixed) {
+        // TODO: Add possibleTypes so imports work
+      } else {
         throw new Error(`Unhandled subtype in Vec, ${stringify(subDef)}`);
       }
     }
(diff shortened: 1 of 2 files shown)


packages/typegen/src/util/derived.ts
```diff
@@ -58,7 +58,7 @@ export function getSimilarTypes (registry: Registry, definitions: Record<string,
         );
 
         possibleTypes.push(`([${subs.join(', ')}])[]`);
-      } else {
+      } else if (subDef.info !== TypeDefInfo.VecFixed) {
```
jacogr, line 61:
```suggestion
      } else if (subDef.info === TypeDefInfo.VecFixed) {
        // TODO: Add possibleTypes so imports work
      } else {
```

jacogr (changes requested):
One small adjustment

packages/typegen/src/util/derived.ts
jacogr: This will need to be looked at specifically - the `possibleTypes` contains the imports as well. So if already imported, it will be all ok. Additionally it adjust the signature. So for now, just marked as a TODO, since it probably will need some love.

jacogr (approved):
Thank you. Apart from the now-future-TODO it looks great.

PR #4743, opened 2022-04-14
fix: Typegen handles structs and vectors of structs without erroring

zannis:
Another minor PR to @polkadot/typegen that fixes vectors of structs erroring out when typegen'd.

### Context
In 7.15.x, vectors of structs were wrongly parsed as vectors of 'plain' objects, leading to successful type generation. Since 8.x.x, they are now correctly parsed as vectors of structs leading to an exception. This should fix this behavior until further handling is implemented.


jacogr:
The change in generation was a weird - since it wasn’t touched (only the VecFixed PR went in). Anyway, this does look sane, will run through and merge a bit later.

jacogr (approved):

Issue #4744, opened 2022-04-14
uniques entriesPaged  throws error 


Yuripetusko:
I used to be able to do this:

```
const nftClasses = await polkadotUniquesApi.query.uniques.class.entriesPaged({
      pageSize,
      args: [],
});
```

To query paged entries from all classes, but now it throws an error

```
RPC-CORE: getKeysPaged(key: StorageKey, count: u32, startKey?: StorageKey, at?: BlockHash): Vec<StorageKey>:: -32603: Unknown error occurred: Client error: UnknownBlock: Header was not found in the database: 0x0000000000000000000000000000000000000000000000000000000000000000
DRR: -32603: Unknown error occurred: Client error: UnknownBlock: Header was not found in the database: 0x0000000000000000000000000000000000000000000000000000000000000000
```

empty `entries` call (non-paged) still works. Is this a change on polkadot.js or something changed on statemine itself?

Yuripetusko:
Actually adding class id to args doesn't help either. So something happened with paged queries

Yuripetusko:
I wonder if this is related to this PR https://github.com/polkadot-js/api/commit/39b1de326cd27d104c4c67a49cddcc67a25273d5

@xlc would you be able to chime in?

jacogr:
Yes, the above doesn't look correct - it shouldn't use `0x000...000`, so there is a bug somewhere.

PR #4745, opened 2022-04-14
Adjust entriesPages w/ & w/o at

jacogr:
Closes https://github.com/polkadot-js/api/issues/4744

Probably needs a patch release, i.e. 8.0.3 (Having said that, it is close enough to Sat/Sun where the normal releases are made, so a day or two is probably not bad... as horrible as it is...)

### DIFF
--- a/packages/api/src/base/Decorate.ts
+++ b/packages/api/src/base/Decorate.ts
@@ -722,8 +722,8 @@ export abstract class Decorate<ApiType extends ApiTypes> extends Events {
 
   private _retrieveMapEntries (entry: StorageEntry, at: Hash | Uint8Array | string | null, args: unknown[]): Observable<[StorageKey, Codec][]> {
     const queryStorageAt = at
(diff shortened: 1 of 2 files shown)


packages/api/src/base/Decorate.ts
```diff
@@ -737,10 +739,14 @@ export abstract class Decorate<ApiType extends ApiTypes> extends Events {
   }
 
   private _retrieveMapEntriesPaged (entry: StorageEntry, at: Hash | Uint8Array | string | undefined, opts: PaginationOptions): Observable<[StorageKey, Codec][]> {
+    const queryStorageAt = at
+      ? (keyset: StorageKey[]) => this._rpcCore.state.queryStorageAt(keyset, at)
+      : (keyset: StorageKey[]) => this._rpcCore.state.queryStorageAt(keyset);
```
jacogr, line 744:
```suggestion
      ? (keys: StorageKey[]) => this._rpcCore.state.queryStorageAt(keys, at)
      : (keys: StorageKey[]) => this._rpcCore.state.queryStorageAt(keys);
```

PR #4747, opened 2022-04-16
Update file.ts for esm support

drewstone:
This was needed to get our API compiling. Not sure what the implications are but we're unable to fine `__dirname` without it.


jacogr:
Thank you. Replacing it with https://github.com/polkadot-js/api/pull/4748 (which works inside the API repo, i.e. no CI failures, as well as in both CJS and ESM modes)

jacogr:
PS: Added you to the CHANGELOG in the API PR, although yours is not used entirely as-is

PR #4751, opened 2022-04-17
Fix for when esmodule parameter exists

drewstone:
In our API system, it seems that our types targeting ESM come with an additional parameter `__esModule: true` which doesn't get processed cleanly by the PolkadotJS API. These changes add basic null checks to ensure that the types that are processed are only processed if they exist.

Not sure if this type of change is what is wanted in this API but I'm not sure otherwise how to get our API building.


jacogr:
I need to see an example of this, with the usage. I know that commonjs outputs (from Babel at least), does add it. Not sure this would protect against that, i.e. here we do `Object.entries(...).forEach(([, { types }]) => ...`, here `{ types }` would fail since you cannot destructure `true`. 

Similar is `Object.values(...).forEach((type) => if (types) ...` since `types === true`, so it will always pass.

PR #4773, opened 2022-04-26
fetch metadata via websocket in polkadot-types-from-defs

alex-nax:
add ability to fetch metadata from websocket connection as stated in usage: https://github.com/polkadot-js/api/blob/master/packages/typegen/src/fromDefs.ts\#L20


jacogr (approved):
Thank you.

PR #4796, opened 2022-05-05
Bump @substrate/connect in rpc-provider to 0.7.4

wirednkod:
This PR bumps version on rpc-provider and replaces rococo_v2_1 with rococo_v2_2
<del>This update brings a new optional parameter in `ScProvider`'s `connect()`.
This parameter is of type [` Config`](https://github.com/paritytech/substrate-connect/blob/main/packages/connect/src/connector/smoldot-light.ts#L117-L135) that passes the desired max log level to smoldot, allowing better debug-ability.</del>

[substrate/connect's Changelog](https://github.com/paritytech/substrate-connect/blob/main/packages/connect/CHANGELOG.md#074---2022-05-05)


tomaka:
It didn't trigger in my brain, but this config is completely not correct. The `createScClient` that you're calling is not the one that has the configuration, and for good reasons since the configuration only applies to the inline smoldot.

jacogr (approved):

jacogr (commented):
Happy. Just not sure how to read Pierre's last comment - it is pre-coffee here...

tomaka:
Let's just update substrate/connect without adding this config

PR #5086, opened 2022-07-19
Update pallet uniques v1 type definitions according to latest changes in substrate repo

Yuripetusko:
The only part that is not released yet is `ItemPrice` but it's merged to master and soon will be released on statemine. I think it can already be merged to polkadot.js API as runtime release should follow shortly and this is a non-breaking change.

https://github.com/paritytech/substrate/blob/9dad17f77b8a56ccf0d289ae410ee1e61d9b12a1/frame/uniques/src/types.rs


Yuripetusko:
Actually I see that tests are failing as these new types are not in metadata yet. so this PR can wait until new runtime is in I guess

jacogr:
This should not really be needed - the types in the definitions are only used when the client is not on V14 metadata. (And since these are used for old chains, pre metadata-V14, should stay in-tact for the old chains.)

The only areas where this _may_ need an update is if it is used in `state_call` or any RPCs. (Neither of these are sadly decorated by metadata)

Yuripetusko:
> This should not really be needed - the types in the definitions are only used when the client is not on V14 metadata. (And since these are used for old chains, pre metadata-V14, should stay in-tact for the old chains.)
> 
> The only areas where this _may_ need an update is if it is used in `state_call` or any RPCs. (Neither of these are sadly decorated by metadata)

So once runtime is upgraded the API will be decorated automatically without the need to manually define definitions?

PR #5089, opened 2022-07-20
Re-introduce getters for types-codec

jacogr:


jacogr:
Closing. Adding more getters has around a 10-15% overhead on object creation.

PR #5128, opened 2022-08-04
feat: allow objects to be supplied as map keys

CertainLach:
Allows to simplify construction of complex objects from

```ts
api.tx.some.extrinsic(new Map([
  [JSON.stringify(key), value],
]))
```

To

```ts

api.tx.some.extrinsic(new Map([
  [key, value],
]))
```

### DIFF
--- a/packages/types-codec/src/extended/Map.ts
+++ b/packages/types-codec/src/extended/Map.ts
@@ -46,7 +46,7 @@ function decodeMapFromMap<K extends Codec, V extends Codec> (registry: Registry,
       output.set(
         key instanceof KeyClass
           ? key
-          : new KeyClass(registry, isComplex ? JSON.parse(key as string) : key),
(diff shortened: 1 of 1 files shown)


packages/types-codec/src/extended/Map.ts
```diff
@@ -46,7 +46,7 @@ function decodeMapFromMap<K extends Codec, V extends Codec> (registry: Registry,
       output.set(
         key instanceof KeyClass
           ? key
-          : new KeyClass(registry, isComplex ? JSON.parse(key as string) : key),
+          : new KeyClass(registry, isComplex && typeof key === 'string' ? JSON.parse(key as string) : key),
```
jacogr, line 49: Weird that `as string` is needed if the typeof check has been done.

CertainLach: Yep, it isn't needed

PR #5149, opened 2022-08-11
Add rpc provider in the list of 'to-be-created' bundles

wirednkod:
This is highly connected to [this PR](https://github.com/polkadot-js/dev/pull/772) of dev

### DIFF
--- a/rollup.config.mjs
+++ b/rollup.config.mjs
@@ -8,6 +8,7 @@ import { createBundle } from '@polkadot/dev/config/rollup';
 const pkgs = [
   '@polkadot/api',
   '@polkadot/api-contract',
+  '@polkadot/rpc-provider',
   '@polkadot/types'
 ];
 


jacogr:
Ummm, not sure about this. We try to minimize the bundles (although judging by the number it is still way too many) - the provider is not much use without the API, hence being combined. 

This is obviously a major release if it has to go in since it breaks everything for people using bundles.

wirednkod:
> Ummm, not sure about this. We try to minimize the bundles (although judging by the number it is still way too many) - the provider is not much use without the API, hence being combined.
> 
> This is obviously a major release if it has to go in since it breaks everything for people using bundles.

I am not sure I understand why this breaks everything. Can you please elaborate?

jacogr:
The providers are already exposed in the API, e.g. the recommended approach is always

```js
import { ApiPromise, WsProvider } from '@polkadot/api';

...
```

So if ScProvider is not available there, it should be added. (Which means it is available on the `globalThis.polkadotApi` as exposed by the bundle)

This breaks everything since -

- the api bundle now requires the provider bundle
- nobody using bundles has it
- which means that code changes for everybody

Hence a major api-breaking release. Unless there is a very good reason to do so, we generally try not to.

wirednkod:
Should this PR close in favor of #5150?

Issue #5161, opened 2022-08-15
[Bug] apiAt.call.transactionPaymentApi not decorated when calls are available in the runtime


TarikGul:
* **I'm submitting a ...**

  - [X] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

### Current Behavior 

When calling `apiAt.call.transactionPaymentApi` for a runtime where `queryInfo`, and `queryFeeDetails` is available returns undefined. 

### Expected Behavior 

`apiAt.call.transactionPaymentApi` should be decorated with the runtimes that have those methods available. 

### Example Code

```javascript
// Network: Polkadot
// BlockNumber: 4092619
const extrinsic = '0x4102840012852bea5af210b7c409dadac3e2146857f7d2d63695cedcb8534ebff83a7ed3001a084779ccd11a3bfbaea5b81346c3943164e789fe367d4e39069bb4e321d248e90ad394307947b9e21c80fdf5d08d267bd211430d34c5abb7d828ddfe5aed0cb50100000503003b127a1acc2dd2c90ccbbd8d608bd58a527c0a700ab9bc323032414a85f81f1207009217b122';
const previousBlockHash = '0x3051684c90d92f234ed2ddeb434e4a93fc4a115600731a1805175498cd4f6359';
const currentBlockHash = '0xcebeb96ddf6a64d97e916ad1accf2e93f749df33d3a225f6fdf3981c1b678b3c';

const apiAt = await api.at(currentBlockHash);

const paymentCallExists = apiAt.call.transactionPaymentApi;

// Returns undefined
console.log('TransactionPaymentApi: ', paymentCallExists)

const queryInfo = await api.rpc.payment.queryInfo(extrinsic, previousBlockHash);
const { inclusionFee } = await api.rpc.payment.queryFeeDetails(extrinsic, previousBlockHash);

// returns valid responses
console.log('QueryInfo call: ', queryInfo.toJSON());
console.log('QueryFeeDetails call: ', inclusionFee.isSome);
```

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**

Note: I tested this locally, and also in the developer tab of polkadot-js apps.

  - Version: `@polkadot/api` v9.2.1
  - Environment:

    - [X] Node.js (v17.1.0)
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
[...]

PR #5163, opened 2022-08-15
Ensure historic runtime interfaces are decorated

jacogr:
Closes https://github.com/polkadot-js/api/issues/5161

Not merging yet - we would need to expand the historic list (a bit of a bitch), see https://github.com/polkadot-js/api/blob/9e4db51dd71773f12f47200d90a34ab9336b8f70/packages/types-known/src/upgrades/polkadot.ts#L6 to include all runtime apis as well. 

Until we do, historic queries on Westend/Kusama/Polkadot would be slightly slower. Probably will merge since it is a fix, however not quite 100% yet on how to handle the above lists, getting it updated initially and then keeping it updated... (Overall the shortcut is _quite_ useful)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Changes:
 
 - Replace `rpc.payment.queryInfo` usage with `call.transactionPaymentApi.queryInfo`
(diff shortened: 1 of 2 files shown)


jacogr:
Closing in favor of https://github.com/polkadot-js/api/pull/5164

PR #5236, opened 2022-09-25
Fix code source type

statictype:
When trying to call `api.call.contractsApi.instantiate` there was always an error decoding the `Code` param, which is an enum variant not `Bytes`


jacogr (approved):
Thank you. 

(Checked it against the Rust codebase as well, certainly seems like an oversight here, aka it have never been `Vec<u8>`)

Issue #4197, opened 2021-11-10
Extract types for signedExtensions from v14 metadata 


jacogr:
We should encode with those specific types, just in case of customisation. (No per-string type overrides anymore)

PR #5239, opened 2022-09-26
add signed extensions for encointer

clangenb:


packages/types/src/extrinsic/signedExtensions/encointer.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import type { ExtDef } from './types';
+
+export const encointer: ExtDef = {
+  ChargeAssetTxPayment: {
+    extrinsic: {
+      tip: 'Compact<Balance>',
+      // eslint-disable-next-line sort-keys
+      assetId: 'Option<CommunityIdentifier>'
+    },
```
clangenb, line 12: Here I am uncertain if it matters that this type does not exist when we connect to a chain that does not have this type defined.

packages/types/src/extrinsic/signedExtensions/index.ts
```diff
@@ -9,13 +9,14 @@ import { polkadot } from './polkadot';
...
 import { substrate } from './substrate';
+import { encointer } from './encointer';
 
 // A mapping of the known signed extensions to the extra fields that they
 // contain. Unlike in the actual extensions, we define the extra fields not
 // as a Tuple, but rather as a struct so they can be named. These will be
 // expanded into the various fields when added to the payload (we only
 // support V4 onwards with these, V3 and earlier are deemed fixed))
-export const allExtensions: ExtDef = objectSpread({}, substrate, polkadot, shell, statemint);
+export const allExtensions: ExtDef = objectSpread({}, substrate, polkadot, shell, statemint, encointer);
```
jacogr, line 19: In this case, this extension would override the existing `ChargeAssetTxPayment` for statemint, which means it will break usage on statemint.

The idea with these are that they map to the default pallets for Substrate & Polkadot. Anything chain-specific needs to be supplied as part of the chain types. For instance as used in the apps UI.

clangenb: I see, thanks for explaining. However, I would like to request #4197 once again because this makes many the polkadot-js/api powered products hard to use for us, e.g. subquery because it can't interpret our extrinsics, and it does not offer this kind of customization without entering the source code.

jacogr: The linked issue is in the queue - so yea, will be done at some point.

clangenb: Ok, thanks. I will close the PR then.

PR #5276, opened 2022-10-18
super small code polish

shunjizhan:


jacogr:
There is actually method behind this madness - with the `?.` babel generates horrible code for environments aka "a complete overkill for method availability checks", so generally we try and stay away from the operator. This is the actual compiled output after the change is made -

```js
      var _api$rpc$system;

      if (!((_api$rpc$system = api.rpc.system) !== null && _api$rpc$system !== void 0 && _api$rpc$system.dryRun)) {
        throw new Error('The system.dryRun RPC call is not available in your environment');
      }
```

This applies to at least in the api and the utils, there has actually been a conscious effort to not use it - however it certainly have not been consistently removed all-over polkadot-js repos. If left to the engine in all cases, it would be less messy, but we are not quite at that point with JS support everywhere as of yet.

wirednkod:
Did **NOT** know about that.. Thank you @jacogr

shunjizhan:
There is also similar concern about not using `for ... of` since it transpiles to generators. I guess there is always a tradeoff between better performance and cleaner source code. 

Personally I am more into cleaner code, because of two reasons:
- we don't really need to read the prod es5 code, so its ok to be messy
- the "performance downgrade" introduced by generator or more messy generated code is trivial for an app like this, where internet traffic takes much longer than code running

jacogr:
There are no eslint (nor ts-eslint) rules for these. Also as mentioned, in UI-code, due to what you mentioned above, it has not been a "rule" and is certainly used a lot. (The API and util is very different from "UI code")

In general, these are not that bad in the PR form, where it becomes hairy and worth screaming is when doing `a?.b?.c?.d`. So each time I use one of these, my brain just screams at me since it know what it causes :)

However, it is actually fixable with a babel `== null` or `!= null` output with [assumptions](https://github.com/polkadot-js/dev/blob/0e6291ea8e2fad36851c5ae39618669ac0ddee83/packages/dev/config/babel-general.cjs#L5-L8), specifically `noDocumentAll: true` - that would remove any "damn, I'm writing really bad stuff now" concerns that causes minor brain explosions.

shunjizhan:
sounds great, so I am closing this PR as it's not the desired syntax

jacogr:
Feel free to re-open, adjusted the babel outputs slightly with https://github.com/polkadot-js/dev/pull/827/files#diff-e46f7bc448773ef1bc430c11187663a5431bc8d7a67aba417675b2e31f3e8cf1R6-R7 

The above now certainly removes my issues with the outputs, especially on those pesky longer items, it is no no worse than doing it "manually".

Issue #5282, opened 2022-10-21
Error Call needs 2 arguments, when 2 arguments are given


crystalin:
When querying moonbeam:
```
const storageKey = await apiAt.query.parachainStaking.atStake.key(
  round,
  candidate
);
```
(round is `u32` and candidate is `AccountId20`)

I get:
```
ERR! Error: Call to parachainStaking.atStake needs 2 arguments, found [7752,0x5e583614F17D3eBe3bD72653F28e2C01D247419A]
    at assertArgs (.../projects/moonbeam-tools/node_modules/@polkadot/types/cjs/metadata/decorate/storage/createFunction.js:46:11)
    at createKey (.../projects/moonbeam-tools/node_modules/@polkadot/types/cjs/metadata/decorate/storage/createFunction.js:110:3)
    at .../projects/moonbeam-tools/node_modules/@polkadot/types/cjs/metadata/decorate/storage/createFunction.js:188:10
    at Function.decorated.key (.../projects/moonbeam-tools/node_modules/@polkadot/api/cjs/base/Decorate.js:787:64)
```

PR #5283, opened 2022-10-21
Ensure apiAt.<section>.<method>.key(...) passes through the correct args

jacogr:
Closes https://github.com/polkadot-js/api/issues/5282

### DIFF
--- a/packages/types/src/metadata/decorate/storage/index.spec.ts
+++ b/packages/types/src/metadata/decorate/storage/index.spec.ts
@@ -36,6 +36,10 @@ describe('decorateStorage', (): void => {
     expect(() => query.timestamp.now()).not.toThrow();
   });
 
+  it('should return a value if the storage function has the correct arguments', (): void => {
+    expect(() => query.staking.erasStakers(1, keyring.alice.address)).not.toThrow();
+  });
+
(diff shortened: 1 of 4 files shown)


packages/types/src/metadata/decorate/storage/index.spec.ts
```diff
@@ -36,6 +36,10 @@ describe('decorateStorage', (): void => {
     expect(() => query.timestamp.now()).not.toThrow();
   });
 
+  it('should return a value if the storage function has the correct arguments', (): void => {
+    expect(() => query.staking.erasStakers(1, keyring.alice.address)).not.toThrow();
```
crystalin, line 40: Shouldn't it also verify that calling the key function is working properly ?

jacogr: This is on the storage decorator (named `creator` in the API), the responsibility of the method is to actually return the key. (This is slightly different than in the API itself, here it returns a prefixed `Uint8Array`, hence the slight hoops in the API to convert to hex key)

crystalin: Thank you for the explanations and the quick fix

PR #5299, opened 2022-10-27
support unknown rpc

xlc:
I want to have some tests as I have no idea if this works or not but not sure where should I add tests for this

Background:
I am working on https://github.com/AcalaNetwork/chopsticks as a test client for Substrate. In addition to Substrate RPC, it also implements some more development RPCs. I would like to use polkadot.js apps to call those RPCs somehow.

The RPC parameters & returns are just JSON. There is no reason why polkadot.js can't do some sensible default for unknown RPC methods.

### DIFF
--- a/packages/rpc-core/src/bundle.ts
+++ b/packages/rpc-core/src/bundle.ts
@@ -29,6 +29,10 @@ type MemoizedRpcInterfaceMethod = Memoized<RpcInterfaceMethod> & {
   meta: DefinitionRpc;
 }
 
+type DefinitionRpcWithUnknown = DefinitionRpc & {
+  isUnknown?: true;
+};
+
 const l = logger('rpc-core');
 
 const EMPTY_META = {
@@ -121,6 +125,11 @@ export class RpcCore {
 
     // decorate all interfaces, defined and user on this instance
     this.addUserInterfaces(userRpc);
+
+    // decorate unknown interfaces (i.e. not in the definitions)
+    this.addUnknownRpc().catch((error): void => {
+      l.error('Unable to add unknown RPC methods:', error);
+    });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/rpc-core/src/bundle.ts
```diff
@@ -356,7 +389,13 @@ export class RpcCore {
     return memoized;
   }
 
-  private _formatInputs (registry: Registry, blockHash: Uint8Array | string | null | undefined, def: DefinitionRpc, inputs: unknown[]): Codec[] {
+  private _formatInputs (registry: Registry, blockHash: Uint8Array | string | null | undefined, def: DefinitionRpcWithUnknown, inputs: unknown[]): Codec[] {
+    if (def.isUnknown) {
+      return inputs.map((input): Codec =>
+        registry.createTypeUnsafe('Text', [input], { blockHash })
```
xlc, line 395: Not sure if there is a better type for this.

packages/rpc-core/src/bundle.ts
```diff
@@ -371,7 +410,11 @@ export class RpcCore {
     );
   }
 
-  private _formatOutput (registry: Registry, blockHash: Uint8Array | string | null | undefined, method: string, rpc: DefinitionRpc, params: Codec[], result?: unknown): Codec | Codec[] {
+  private _formatOutput (registry: Registry, blockHash: Uint8Array | string | null | undefined, method: string, rpc: DefinitionRpcWithUnknown, params: Codec[], result?: unknown): Codec | Codec[] {
+    if (rpc.isUnknown) {
+      return registry.createTypeUnsafe('Text', [result], { blockHash });
```
jacogr, line 415: This won't be correct - there is a JSON type, which should rather be used.

jacogr:
The issue with this is that the API would have no idea if there are params or not. If there are params, it also has no idea how to encode them. So basically this would work in the cases where there are no params and create overhead in support in those where there are params.

For those reasons, it is not something that would probably make it in. It is very limited in scope and won't really move forward.

xlc:
Why any encoding is required? The params are just JSON array.
I really want to have the polkadot apps RPC section to display those, and just present a text input for me to enter JSON string and it will be pretty useful

jacogr:
The focus should be on getting Substrate to return info to describe these. On that layer: RPCs won't happen (that battle seems lost), but maybe the runtime calls may happen at some point down the line. 

I feel your pain, since I have to deal with this on a daily basis (hunting all these down) between RPCs and runtime calls. However, I don't believe that putting on yet more layers to hide the unknowns will help in getting this fixed. And not at the expense of different behavior for typed vs untyped either.

xlc:
I guess I will just ask people to use postman for now.

jacogr:
I do think there is another (similar) approach that can be taken here. Been thinking about a lower-level interface for a while, but have not had a gap to explore. So basically my thinking is that, like we have `api.tx(...)` that can instantiate based on the inputs, we could do the same at rpc/call/query level. For the first one -

```js
const json = await api.rpc('exec_storageGet', 123, '0x12345678', '0x12345678');
```

So in this case would always return the raw JSON result. Same would apply to call/query where it just takes what you have as-is and passes it through, with basically no "jumping in the middle".

This is quite close to the approach you have taken here, but what it does do is be explicit in providing a lower-level interface (no confusion in typed vs untyped interfaces) where the dev is responsible for encoding. Same would apply to the others mentioned (they probably may get some massaging in places).

PR #5305, opened 2022-10-31
able to send raw rpc methods

xlc:
#5299 take two

so we can send undecorated rpc calls
```js
const json = await api.rpc('exec_storageGet', 123, '0x12345678', '0x12345678');
```

### DIFF
--- a/packages/api-base/src/types/rpc.ts
+++ b/packages/api-base/src/types/rpc.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { Observable } from 'rxjs';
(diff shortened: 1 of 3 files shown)


xlc:
The typing could be improved but not sure what's the better way
I would like to write some tests but can't find a good place for that and therefore have no idea if this actually works or not

jacogr:
Brilliant. Sorry for not looking yet, the intent is indeed to get it into the next release.

jacogr (commented):
One small comment, the rest looks sane.

jacogr (approved):
Thank you.

PR #5324, opened 2022-11-08
Update contract' gas to Weight type

jasl:
I don't have confidence I'm doing right.

Starting from `polkadot v0.9.31`, the gas relates fields has changed to `Weight` (V2, https://github.com/paritytech/substrate/blob/polkadot-v0.9.31/frame/contracts/primitives/src/lib.rs#L35-L71 ),
but it looks polkadot.js still define them `u64`


packages/api-contract/src/types.ts
```diff
@@ -55,8 +55,8 @@ export interface InterfaceContractCalls {
 
 export interface ContractCallOutcome {
   debugMessage: Text;
-  gasConsumed: u64;
-  gasRequired: u64;
+  gasConsumed: WeightV2;
```
jacogr, line 58: The next one is correct - the Weight type would be detected on-chain, so using “Weight” will result in the correct on-chain type (which could be v1, v1.5 or v2)

this is important using the detect here since different chains would have different requirements (based on their upgrade cycle)

So the best approach is to change all the different definitions, where weight is involved, to “Weight”.

jasl: Changed to `Weight` and force pushed

jasl: Should this line https://github.com/polkadot-js/api/blob/master/packages/api-contract/src/types.ts#L76 change to `Weight` too?

jasl: https://github.com/polkadot-js/api/blob/master/packages/api-contract/src/types.ts#L87
Should I need to add `proof_size` for `v2Weight` ?

jacogr: The interface (first question) is fine. It detects what is on-chain and then passes the correct construct through, either for v1, v1.5 or v2

jacogr: For the second - no. That is an internal interface that just manages the shape for compat reasons.

jasl: I changed https://github.com/polkadot-js/api/blob/master/packages/api-contract/src/types.ts#L76 to `Weight` and don't touch `AllWeight`

packages/api-contract/src/types.ts
```diff
@@ -73,7 +73,7 @@ export interface DecodedMessage {
 }
 
 export interface ContractOptions {
-  gasLimit?: bigint | string | number | BN | WeightV2;
+  gasLimit?: bigint | string | number | BN | Weight;
```
jacogr, line 76: This one should indeed be `WeightV2`. There is internal detection from the options to cater for both old/new (aka if a `number` is passed, it would convert to `WeightV2` before using.

jasl: Sorry I misunderstand your means, I've reverted this change

statictype:
this is great! thanks @jasl 
the [runtime call](https://github.com/polkadot-js/api/blob/5c81e4fd3b0ff8b5b8aea24fad12992a7ccbedb9/packages/api-contract/src/base/Contract.ts#L145) needs to be adjusted too. similar to [exec](https://github.com/polkadot-js/api/blob/5c81e4fd3b0ff8b5b8aea24fad12992a7ccbedb9/packages/api-contract/src/base/Contract.ts#L110)

jasl:
> Make Contracts Runtime Calls support WeightV2

I make a new commit for this, check https://github.com/polkadot-js/api/pull/5324/commits/ca51a65086dcf2b75d5d6a3ffe6e83a20e59466d

jasl:
@jacogr what does the `Gas` type definition do? should I convert all `gas` relates fields (`gasLimit`, `gasRequired`) to `Gas` type?

jasl:
ops it seems I force pushed too many times the CI refuse to run ...
unforuntately my local environment fails on some random parts...

statictype:
no worrries. the CI doesn't run automatically for first time contributions. it's still a bit more work, unfortunately. 
i'm testing your changes locally with apps. 
@jacogr is [this](https://github.com/polkadot-js/api/blob/5c81e4fd3b0ff8b5b8aea24fad12992a7ccbedb9/packages/api-contract/src/base/Contract.ts#L152) necessary? Can we return gasRequired as is?

jasl:
https://github.com/polkadot-js/api/pull/5324/files#diff-e923378b0f9524ff3bcbc37be8010d0f81c271e98fceabc6cc9641adb08de09bR58 
the `Weight` here, are pointed to `WeightV1` 
https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/runtime/types.ts#L392
so that's why CI failed because
`error: TS2345 [ERROR]: Argument of type '{ refTime: BN; }' is not assignable to parameter of type 'AnyNumber | u64'.`
I guess

jasl:
> this file

understood, thank you, now the latest commit should pass `yarn build` without error

jasl:
CI PASSED !!!

packages/api-contract/src/types.ts
```diff
@@ -55,8 +55,8 @@ export interface InterfaceContractCalls {
 
 export interface ContractCallOutcome {
   debugMessage: Text;
-  gasConsumed: u64;
-  gasRequired: u64;
+  gasConsumed: bigint | string | number | BN | WeightV2;
```
statictype, line 58: i think here it should be `u64 | WeightV2`

statictype: the `ContractCallOutcome` interface represents what is returned from the runtime call that executes on-chain. `gasConsumed` and `gasRequired` can't ever be returned as string.

on the other hand `ContractOptions` represents what can be passed in the runtime call by users. i suppose the api converts it before sending and it's more convenient to have looser typing here.

take a look at how it is on master. they are not the same.

jasl: Thank you for clarify! I have changed as your suggest

packages/types/src/interfaces/contracts/runtime.ts
```diff
@@ -24,7 +24,7 @@ export const runtime: DefinitionsCall = {
             },
             {
               name: 'gasLimit',
-              type: 'u64'
+              type: 'Weight'
```
athei, line 27: This is an option: https://github.com/paritytech/substrate/blob/78f8f708781da6fd340ef2a1fc288060a35295ad/frame/contracts/src/lib.rs#L1215

packages/types/src/interfaces/contracts/runtime.ts
```diff
@@ -64,7 +64,7 @@ export const runtime: DefinitionsCall = {
...
```
athei, line 67: This is an option: https://github.com/paritytech/substrate/blob/78f8f708781da6fd340ef2a1fc288060a35295ad/frame/contracts/src/lib.rs#L1226

packages/types/src/interfaces/contracts/definitions.ts
```diff
@@ -45,7 +45,7 @@ export default {
       origin: 'AccountId',
       dest: 'AccountId',
       value: 'Balance',
-      gasLimit: 'u64',
+      gasLimit: 'Weight',
```
athei, line 48: Is optional.

athei (commented):
The `Weight` is optional in runtime APIs now. It allows you to let the runtime choose a default value in case you don't care (you usually don't care if you use it to estimate `gas_required`).

packages/types/src/interfaces/contracts/runtime.ts
jasl: Changed to `Option<Weight>`

packages/types/src/interfaces/contracts/definitions.ts
jasl: Changed to `Option<Weight>`

jasl:
> The `Weight` is optional in runtime APIs now. It allows you to let the runtime choose a default value in case you don't care (you usually don't care if you use it to estimate `gas_required`).

Changed your mentioned fields to `Option<Weight>`

athei:
Don't you also need to adapt some code in the contracts api that uses those, now changed, signatures?

jasl:
> Don't you also need to adapt some code in the contracts api that uses those, now changed, signatures?

Sorry, I'm not familier with `polkadot-js` codebase, but if could give me some directions, I think I can work out

jasl:
I do unit test in local, they're all passed, you mean I need to add unit tests for WeightV2?

packages/api-contract/src/types.ts
```diff
@@ -55,8 +55,8 @@ export interface InterfaceContractCalls {
 
 export interface ContractCallOutcome {
   debugMessage: Text;
-  gasConsumed: u64;
-  gasRequired: u64;
+  gasConsumed: u64 | WeightV2;
```
jacogr, line 58: This should just be `Weight` - this is a bit tricky, but the runtime will inject the correct type.

packages/types/src/interfaces/contracts/definitions.ts
```diff
@@ -45,7 +45,7 @@ export default {
       origin: 'AccountId',
       dest: 'AccountId',
       value: 'Balance',
-      gasLimit: 'u64',
+      gasLimit: 'Option<Weight>',
```
jacogr, line 48: This is going to be problematic - there is now old and new, we need to cater for both.

packages/types/src/interfaces/contracts/runtime.ts
```diff
@@ -24,7 +24,7 @@ export const runtime: DefinitionsCall = {
             },
             {
               name: 'gasLimit',
-              type: 'u64'
+              type: 'Option<Weight>'
```
jacogr, line 27: This can't be - the runtime is supposed to be versioned, i.e. there should not be a breaking change without updating versions. (This is the same as above, this is breaking for users)

jacogr (commented):
I am literally not sure about these `Option<Weight>` changes - there seems to ahve been a breaking update to the runtime interfaces without a version update. This is hugely problematic, there is a reason why the runtime interfaces have versions.

EDIT: Indeed. This PR broke the world without updating the API version. https://github.com/paritytech/substrate/pull/12429 - not sure how that passed reviews. It is creating huge (unsolvable) issues here now.

packages/types/src/interfaces/contracts/definitions.ts
jasl: does `u64 | Option<Weight>` work?

jasl:
> EDIT: Indeed. This PR broke the world without updating the runtime api version. https://github.com/paritytech/substrate/pull/12429 - not sure how that passed reviews. It is creating huge (unsolvable) issues here now.

Do you have idea @athei ?

packages/types/src/interfaces/contracts/runtime.ts
jasl: IMO, `Option<Weight>` can be treated as `0u64`? and `0u64` can be treated as `None` ?

jacogr: There is a difference in SCALE encoding `Option<u64>` vs `u64`. The former would be 9 bytes in length, the later would be 8 bytes. So it breaks the structure decoding if wrong.

athei:
> EDIT: Indeed. This PR broke the world without updating the runtime api version. https://github.com/paritytech/substrate/pull/12429 - not sure how that passed reviews. It is creating huge (unsolvable) issues here now.

What exactly do you mean by "broke the world"? Do you mean existing code which didn't update to a new polkadot.js version or is new code also affected?

jacogr:
The issue is that the `state_call` interface has been changed without updating the runtime API version. The API has no way to detect which version is in use - generally (with all other APIs), the runtime version gives this indication, i.e. for version 1 (the default if not specified), use X, for version 2, use Y.

This is the great thing about using `state_call`, it is versioned, unlike RPCs which are always problematic in terms of params and results. 

The nightmare here is that the result encoding has completely changed - since none of this information is in the metadata (there is a Substrate issue to expose it), there is no way to know _how_ to decode it, i.e. what is the format. This would have been solved with adding the `version 2` attribute to the APIs.

athei:
So in order to unblock this PR it would be fine if I just flag the new version as `version 2` but don't supply a `version 1` for backwards compatibility? This will still break old polkadot.js code (which isn't aware of a new version) but we are okayish with that?

> The JS API has no way to detect which version is in use - generally (with all other runtime APIs), the runtime api version gives this indication, i.e. for version 1 (the default if not specified), use X, for version 2, use Y.

How do you detect the version (given that I add the version attribute in substrate)? There is no indication in metadata.

jasl:
So we need to do these things:

- add `#[api_version(2)]` to `pallet_contracts` RuntimeAPI
- backport above change to `polkadot-v0.9.31` branch
- refactor this PR use `Weight` for V2 API

`polkadot-v0.9.31` release just few days ago, I think it's not late to append the fix.

Am I understanding correct?

jacogr:
`state_getRuntimeVersion` RPC returns all the used APIs and their versions.

jacogr:
@jasl Yes, with the version bump, we are all ok, since we can add versioned definitions.

athei:
> @athei do you think my plan work? shall I make a PR for `pallet_contracts`?

Yes and yes. Just need another approval.

jasl:
@jacogr I refactor contracts API interface with versioning style, but the result of `yarn build` seems not correct

packages/api-contract/src/types.ts
jasl: Changed

jasl:
the Substrate PR has merged, so the `gasLimit: 'Option<Weight>',` should not be a problem?
Anything else I need to do?

athei:
> backport above change to `polkadot-v0.9.31` and `polkadot-v0.9.32` branch

Can you do that? Just cherry-pick and create a PR against those branches.

jasl:
> > backport above change to `polkadot-v0.9.31` and `polkadot-v0.9.32` branch
> 
> Can you do that? Just cherry-pick and create a PR against those branches.

No problem, I'll do now

jasl:
> > backport above change to `polkadot-v0.9.31` and `polkadot-v0.9.32` branch
> 
> Can you do that? Just cherry-pick and create a PR against those branches.

Since 0.9.31 has replaced by 0.9.32, I pushed backport PRs for 0.9.32 and 0.9.33

packages/types/src/interfaces/contracts/definitions.ts
jacogr: For his one we need a runtime result for V1 as well - that keeps as it. For v2, we have a new version that gets returned. So basically rename the existing to `ContractCallRequestV1` (to be returned for v1 runtime) and the change the above version for `Option<Weight>`.

Basically (until we finally have all this info on the metadata), we need to have old and new types available. Yes, it is a bitch.

packages/api-contract/src/base/Contract.ts
```diff
@@ -141,8 +120,10 @@ export class Contract<ApiType extends ApiTypes> extends Base<ApiType> {
           origin,
           this.address,
           value,
-          // the runtime interface still used u64 inputs
-          this.#getGas(gasLimit, true).v1Weight,
+          this._isOldWeight
```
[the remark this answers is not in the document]
jasl: Sorry I don't understand your means, do you mean I need to revert this part change?

jacogr:
From a TS perspective, this is slightly tricky. So with the runtime API in Substrate master, we first need to add the runtime function and ensure it generates to that version (This basically means a metadata bump, i.e. the interfaces are generated based on what is in the metadata). 

So would suggest making a small split -

1. PR for only the v2 + v1 runtime changes (the rest will still be broken)
2. Jaco to update the static metadata (can be done once the above is in)
3. The rest of the changes in here to ensure V1/V2 gets used as detected

jasl:
> From a TS perspective, this is slightly tricky. So with the runtime API in Substrate master, we first need to add the runtime function and ensure it generates to that version (This basically means a metadata bump, i.e. the interfaces are generated based on what is in the metadata).
> 
> So would suggest making a small split -
> 
> 1. PR for only the v2 + v1 runtime changes (the rest will still be broken)
> 2. Jaco to update the static metadata (can be done once the above is in, will then generate correct call definitions)
> 3. The rest of the changes in here to ensure V1/V2 gets used as detected

For 1 I'll try to make a commit, but I met some trouble `yarn build` generate incorrect augument code, anyway, I'll try

jasl:
Sorry I think I need more clear instructions, `PR for only the v2 + v1 runtime changes` do you mean I should only push 
`packages/types/src/interfaces/contracts/runtime.ts`
and
`packages/types/src/interfaces/contracts/types.ts`
?

jasl:
I rebased master

jasl:
Try my best to understanding your means... I'm dizzy, sorry 😓

jacogr:
`yarn build` will only yield updated code when the static Substrate metadata has been updated. Before that, the only known to the generation is basically a week old. It should generate for the Contract V2 APIs now after you pulled-in master which has current metadata.

(Basically on the generated types it uses the chain metadata - this means that all chains supply their metadata and it tweaks the generated types based on that)

jasl:
> `yarn build` will only yield updated code when the static Substrate metadata has been updated. Before that, the only known to the generation is basically a week old. It should generate for the Contract V2 APIs now after you pulled-in master which has current metadata.
> 
> (Basically on the generated types it uses the chain metadata - this means that all chains supply their metadata and it tweaks the generated types based on that)

Yeah, I rebased master and specify deprecated Contracts RPC to use V1 struct, then `yarn build`, I did `yarn test` in my local it seems passed

packages/api-contract/src/base/Contract.ts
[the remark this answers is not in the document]
jacogr: Yes. It should rather be adapted to al check for the new stuff. You can do this by detecting which version is used, i.e something like -

```js
...
value,
this.api.tx.contracts.call.meta.version === 1
  ? this.#getGas(gasLimit, true).v1Weight
  : this.#getGas(gasLimit, true).v2Weight
...
```

This is important since the getGas also allows the developer to pass a number - since we don't like breaking compat, we have not changed the signatures.

packages/rpc-augment/src/augment/jsonrpc.ts
```diff
@@ -145,7 +145,7 @@ declare module '@polkadot/rpc-core/types/jsonrpc' {
        * @deprecated Use the runtime interface `api.call.contractsApi.call` instead
        * Executes a call to a contract
        **/
-      call: AugmentedRpc<(callRequest: ContractCallRequest | { origin?: any; dest?: any; value?: any; gasLimit?: any; storageDepositLimit?: any; inputData?: any } | string | Uint8Array, at?: BlockHash | string | Uint8Array) => Observable<ContractExecResult>>;
+      call: AugmentedRpc<(callRequest: ContractCallRequestV1 | { origin?: any; dest?: any; value?: any; gasLimit?: any; storageDepositLimit?: any; inputData?: any } | string | Uint8Array, at?: BlockHash | string | Uint8Array) => Observable<ContractExecResult>>;
```
jacogr, line 148: If you run `yarn build:metadata` now, it should actually pull in the latest V2 stuff (assuming there is a version: 2 definition available in contracts/runtime.ts)

packages/types/src/interfaces/contracts/definitions.ts
jasl: I already created `ContractCallRequestV1`, on line 44

packages/rpc-augment/src/augment/jsonrpc.ts
jasl: I ran `yarn build:metadata` seems no file changes

packages/api-contract/src/base/Contract.ts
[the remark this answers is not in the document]
jasl: Changed as your suggest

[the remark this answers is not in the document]
jasl:
```
packages/api-contract/src/base/Contract.ts:144:43 - error TS2339: Property 'version' does not exist on type 'FunctionMetadataLatest'.

144           this.api.tx.contracts.call.meta.version === 1
```

I'll dig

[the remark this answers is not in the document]
jasl: seems `FunctionMetadataLatest` doesn't have `version` field?

packages/rpc-augment/src/augment/jsonrpc.ts
jacogr: Ensure you have master merged in - just did a quick (5 mins, literally), `ContractsApi/2` in here and it generates the new interfaces - https://github.com/polkadot-js/api/pull/5330

(No idea if it is fully correct in regards to `ContractsApi` definitions)

jasl: I rebased master and it has the new metadata commit...
I see you open a new PR for Contracts V2, do you wanna take it?

jacogr: Happy if you take it - you have done the hard work. But either way ok. (Just wanted to show that the runtime apis _does_ get updated, once defined, on master, in the above PR)

jacogr: I believe your issues (reading below), is that your u64 -> Option<Weight> changes are on the v1 APIs, which are not being generated. My PR adds for v2 (and keeps v1 as-is, with some small historic-types-now updates), hence being generated.

jasl: I miss the head of the file `yarn polkadot-types-from-chain`, so I should build a latest Substrate node, and run `yarn polkadot-types-from-chain --endpoint ws://127.0.0.1:9944 --output packages/rpc-augment/src/augment/jsonrpc.ts`

packages/api-contract/src/base/Contract.ts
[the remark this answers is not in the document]
jasl: @jacogr I still don't got this one

packages/rpc-augment/src/augment/jsonrpc.ts
jacogr: We don include the latest [static metadata](https://github.com/polkadot-js/api/blob/master/packages/types-support/src/metadata/v14/substrate-hex.ts), so `yarn build:metadata` will pick it up. (Well, "latest" as of this morning - it gets updated once weekly)

jasl: I port your PR's changes, and rerun `yarn build:metadata`, Doe it look correct now?

jasl:
Just made another batch of changes, ran `yarn build` and `yarn build:metadata` no new changes, `yarn test` pass

jacogr:
Nice. I can see the v2 apis popping up now. Will get it in in the morning.

jasl:
> Nice. I can see the v2 apis popping up now. Will get it in in the morning.

Thank you for your patience mentoring, if possible, could you make a new release after this PR merged? our team is blocked by the issue

jacogr:
Releases go on weekends. (Generally Sundays, each week, but I have been known to push on Saturday as well) The next is scheduled for tomorrow.

jacogr (approved):
Thank you. Seems spot-on.

PR #5325, opened 2022-11-10
fix unkown rpc

### LINKED ISSUE #5321: api.rpc(method, ...params) has additional encoding
This is the result -

image

I believe we are most probably doing double stringify, aka 

https://github.com/polkadot-js/api/blob/5c81e4fd3b0ff8b5b8aea24fad12992a7ccbedb9/packages/api/src/base/Decorate.ts#L416

This then gets `JSON.stringify-ed` again when passing the stuff via e.g. websocket, aka 

https://github.com/polkadot-js/api/blob/5c81e4fd3b0ff8b5b8aea24fad12992a7ccbedb9/packages/rpc-provider/src/coder/index.ts#L72-L76

This is where tests would have been good. This is certainly not the fault of the PR pulling this in, these e2e tests are completely lacking in here, so there was just no place to put them. 

(I should have also tested manually against the apps UI, like I did now with a bump, but ran out of time there before cutting the release, it was the intent to do so. The good news is that param-less calls do work exactly as expected and just removing the `stringify` above, which means that stuff needs to be in the format RPC expects, would fix it. TL;DR I should have tested manually since it was known there "cannot be (atm)" automated tests.)

cc @xlc

xlc:
closes #5321
and have some test

### DIFF
--- a/packages/api/src/base/Decorate.ts
+++ b/packages/api/src/base/Decorate.ts
@@ -413,7 +413,7 @@ export abstract class Decorate<ApiType extends ApiTypes> extends Events {
 
   private _rpcSubmitter<ApiType extends ApiTypes> (decorateMethod: DecorateMethod<ApiType>): DecoratedRpc<ApiType, RpcInterface> {
     const method = (method: string, ...params: any[]) => {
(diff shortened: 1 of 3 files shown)


jacogr (approved):
Thank you.

PR #5329, opened 2022-11-11
Bump Substrate metadata

jacogr:
As per https://github.com/polkadot-js/api/pull/5324#issuecomment-1311365091

Generation here reports "Unable to find matching version for runtime ContractsApi, expected 2" - this is expected since we don't have definitions for those at this point. Since unknown/not found in the type generation, it is removed - https://github.com/polkadot-js/api/pull/5329/files#diff-79eb57055ccb3c449430b1af803a0d0a29cbdc1b625b6bf8a25ff0b7ba329ea0L108


jacogr:
This seems merge-able without having the `ContractsApi` V2 defined yet, so will get it in.

Issue #5350, opened 2022-11-21
[BUG] Deno cannot compile since polkadot@0.2.12


Revolution1:
* **I'm submitting a ...**
  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Here is a simple script to test this issue from version 0.2.11 to 0.2.17

```bash
for v in {11..17}; do
    file="polkadot.0.2.$v.ts"
    cat <<EOF >$file
import { WsProvider } from "https://deno.land/x/polkadot@0.2.$v/api/mod.ts";
await (new WsProvider("wss://rpc.polkadot.io")).disconnect();
console.log("OK");
EOF
    echo "try running 'deno run -A $file'"
    deno run -A $file

    echo "try compile and run with 'deno compile -A -o test.0.2.$v $file && ./test.0.2.$v'"
    deno compile -A -o "test.0.2.$v" $file && ./"test.0.2.$v"
    echo
done
```

Here is the output in my machine

```bash
try running 'deno run -A polkadot.0.2.11.ts'
OK
try compile and run with 'deno compile -A -o test.0.2.11 polkadot.0.2.11.ts && ./test.0.2.11'
Compile file:///tmp/test/polkadot.0.2.11.ts
Emit test.0.2.11
OK

try running 'deno run -A polkadot.0.2.12.ts'
error: Uncaught SyntaxError: The requested module 'https://deno.land/std@0.153.0/node/process.ts' does not provide an export named 'hrtime'
    at <anonymous> (https://esm.sh/v96/@substrate/smoldot-light@0.7.2/deno/smoldot-light.js:4:5277)
try compile and run with 'deno compile -A -o test.0.2.12 polkadot.0.2.12.ts && ./test.0.2.12'
Check file:///tmp/test/polkadot.0.2.12.ts
Compile file:///tmp/test/polkadot.0.2.12.ts
Emit test.0.2.12
error: Module not found

try running 'deno run -A polkadot.0.2.13.ts'
error: Uncaught SyntaxError: The requested module 'https://deno.land/std@0.153.0/node/process.ts' does not provide an export named 'hrtime'
    at <anonymous> (https://esm.sh/v96/@substrate/smoldot-light@0.7.2/deno/smoldot-light.js:4:5277)
try compile and run with 'deno compile -A -o test.0.2.13 polkadot.0.2.13.ts && ./test.0.2.13'
[...]

PR #5351, opened 2022-11-23
For Deno don't re-export ScProvider

jacogr:
This is a nuclear option for https://github.com/polkadot-js/api/issues/5350 - don't re-export ScProvider at all for Deno bundles. 

The caveats are that this is (hopefully) short-term and certainly creates duplication and non-conformity between the outputs. At the same time, not sure if this covers _all_ of the usages, however it _should_ hopefully do as much.

### DIFF
--- a/packages/api/src/bundle.ts
+++ b/packages/api/src/bundle.ts
@@ -1,6 +1,10 @@
 // Copyright 2017-2022 @polkadot/api authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+// NOTE: Any export changes here also needs to be applied to mod.ts
+// (generally that file _should_ only `export * from './index'`, however
+// for the time being we are not exporting sc-provider directly there)
+
 import '@polkadot/rpc-augment';
 
 export { Keyring } from '@polkadot/keyring';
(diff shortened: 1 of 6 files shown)


wirednkod:
Can you please elaborate a bit more on this? Why these changes would resolve issue #5350?

jacogr:
So basically all this does is that for Deno the `substrate-connect` path would not be include in the default.

If you look at the log, it basically goes

api/mod.ts
rpc-provider/mod.ts
rpc-provider/substrate-connect/index.ts
substrate-connect

So with this the mod.ts files does not reference substrate-connect at all, so you won't run into those issues. If you use something non-compiled, it does mean you can still use the provider, you would just need to explicitly import it.

So it does "solve" the specific issue. (It is only a `deno compile ...` issue, not a `deno run ...` issue) Having said that, I don't like this PR - I made it, it _should_ solve the issue, but I don't particularly like the feature/imports discrepancy now and neither having to think around mod.ts and index.ts. It should only be a very, very last resort, so I would prefer not having this at all.

wirednkod:
> Having said that, I don't like this PR - I made it, it _should_ solve the issue, but I don't particularly like the feature/imports discrepancy now and neither having to think around mod.ts and index.ts.

I agree with that.. I will try to figure out if there is any workaround

PR #5371, opened 2022-12-08
Revert removal of WellKnownChains export

wirednkod:
Reverting the removal of WellknownChains. This enum needs to be exported, for users of substrate-connect. They need it in order to define which chain of the WellknownChains should be passed as parameter to ScProvider.

### DIFF
--- a/packages/rpc-provider/src/substrate-connect/index.ts
+++ b/packages/rpc-provider/src/substrate-connect/index.ts
@@ -37,6 +37,8 @@ const subscriptionUnsubscriptionMethods = new Map<string, string>([
 const wellKnownChains = new Set(Object.values(WellKnownChain));
 const scClients = new WeakMap<ScProvider, ScClient>();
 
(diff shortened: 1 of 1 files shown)


jacogr:
It is exposed on `ScProvider`. It has been due for removal since 9.2.1 [as per the CHANGELOG](https://github.com/polkadot-js/api/releases/tag/v9.2.1).

aka 

```js
// or an import from @polkadot/api
import { ScProvider } from '@polkadot/rpc-provider';

const provider = new ScProvider(ScProvider.WellKnownChain.polkadot);

await provider.connect();

const version = await provider.send('chain_getBlockHash', []);
```

The above is actually from the [README](https://github.com/polkadot-js/api/tree/master/packages/rpc-provider#usage). I suppose what didn't happen really well was the explicit reference of "if you have not swapped, you should do so right now".

wirednkod:
Thank you for the response Jaco. Will close that as resolved

PR #5376, opened 2022-12-13
Deprecate initialU8aLength/encodedLength (replace w/ $*)

jacogr:


jacogr:
This is too much. We just just take a step back and _only_ add for `$initialU8aLength` & `$blockHash`... then look at the rest at a later point.

PR #5394, opened 2023-01-01
Expose `blockNumber` via filterEvents function

### LINKED ISSUE #5393: Expose block height on extrinsic results
* **I'm submitting a ...**

  - [ ] Bug report
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

> Currently, on extrinsic result, we can get `blockHash`, `txHash` and `txIndex` but not `blockHeight`

> In our apps, we log the extrinsic's block height for auditing and analytics purposes. With the current setup, we would have to do another RPC call `api.rpc.chain.getBlock` to retrieve the block height. If you are happy, we can submit a PR following the same principle as https://github.com/polkadot-js/api/pull/4556 to add `blockHeight` to the extrinsic results

  - Version: 9.9.1
  - Environment:

    - [x] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript (include tsc --version)
    - [ ] Other

ken-centrality:
Closes #5393

### DIFF
--- a/packages/api/src/util/filterEvents.ts
+++ b/packages/api/src/util/filterEvents.ts
@@ -3,6 +3,8 @@
 
 import type { BlockNumber, EventRecord, ExtrinsicStatus, H256, SignedBlock } from '@polkadot/types/interfaces';
 
+import { isCompact } from '@polkadot/util';
+
 import { l } from './logging';
 
 export function filterEvents (txHash: H256, { block: { extrinsics, header } }: SignedBlock, allEvents: EventRecord[], status: ExtrinsicStatus): { events?: EventRecord[], txIndex?: number, blockNumber?: BlockNumber } {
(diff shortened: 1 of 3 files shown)


packages/api/src/util/filterEvents.ts
```diff
@@ -14,7 +14,8 @@ export function filterEvents (txHash: H256, { block: { extrinsics, header } }: S
           phase.isApplyExtrinsic &&
           phase.asApplyExtrinsic.eqn(txIndex)
         ),
-        txIndex
+        txIndex,
+        blockNumber: header.number.unwrap()
```
jacogr, line 18: As per the linting comments, we just need to move this up.

I'm just a bit concerned that some chains may be configured with `Compact<BlockNumber>` and others with `BlkockNumber`. So just add a bit more protection around this, here is an example used in the derives - https://github.com/polkadot-js/api/blob/f9df7a5c0df68b688b151296967599d8b0b595af/packages/api-derive/src/util/blockNumber.ts#L12-L14

ken-centrality: Ah I see, thank you. I've pushed a new commit with eslint issues addressed plus adding the guard.

Btw, I wasn't sure if it was a good idea to import the `api-derive/unwrapBlockNumber` function in there so I used the `isCompact` directly instead. Let me know if you want me change it around

jacogr: Happy as-is with the way you structured it.

jacogr (approved):
Thank you.

Issue #5383, opened 2022-12-19
Type generation error for nested tuples: `Unable to extract location of ';'`


sander2:
We have a storage type that has a key of type `((CurrencyId, CurrencyId), AccountId)`, i.e. it is a nested tuple. Type generation fails here: https://github.com/polkadot-js/api/blob/5cb6cf0a0034fccc992b182b8ebc7bc82a6b7eac/packages/types-create/src/util/getTypeDef.ts#L148

What seems to be happening is that [this code](https://github.com/polkadot-js/api/blob/7b5b8745a2af68f365987caccd922d01c813304f/packages/typegen/src/generate/query.ts#L44) creates the following `similarTypes`:
```
similarTypes[0][0]:
'ITuple<[ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]>, AccountId32]>'
similarTypes[0][1]:
'[ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]> | [InterbtcPrimitivesCurrencyId | { Token: any } | { ForeignAsset: any } | { LendToken: any } | { LpToken: any } | { StableLpToken: any } | string | Uint8Array, InterbtcPrimitivesCurrencyId | { Token: any } | { ForeignAsset: any } | { LendToken: any } | { LpToken: any } | { StableLpToken: any } | string | Uint8Array], AccountId32 | string | Uint8Array]'
```

Note the second item, which has the form `[value1, value2, value3]`. When [setImports](https://github.com/polkadot-js/api/blob/7b5b8745a2af68f365987caccd922d01c813304f/packages/typegen/src/generate/query.ts#L48-L56) is then called, `getTypeDef` gets called on that type, which calls `decodeFixedVec`, which expects something of the form `[type; len]` so the error above is thrown (note the `;` instead of a `,`).

bvotteler:
> What seems to be happening is that [this code](https://github.com/polkadot-js/api/blob/7b5b8745a2af68f365987caccd922d01c813304f/packages/typegen/src/generate/query.ts#L44) creates the following `similarTypes`

To add some context, it seems like the [recursive handling](https://github.com/polkadot-js/api/blob/5c6294b71ca758749419dfe3456f1fca5502246f/packages/typegen/src/util/derived.ts#L118-L128) inside `api/derived.ts` `getSimilarTypes` method might play a significant factor.

bvotteler:
[...]

PR #5395, opened 2023-01-02
fix: type generation for nested tuples

sander2:
Addresses https://github.com/polkadot-js/api/issues/5383 .

Without this, `setImports` fails for the input `"[ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]> | [InterbtcPrimitivesCurrencyId | { Token: any } | { ForeignAsset: any } | { LendToken: any } | { LpToken: any } | { StableLpToken: any } | string | Uint8Array, InterbtcPrimitivesCurrencyId | { Token: any } | { ForeignAsset: any } | { LendToken: any } | { LpToken: any } | { StableLpToken: any } | string | Uint8Array], AccountId32 | string | Uint8Array]"`.

I _think_ this example can be simplified to `"[ITuple<[u32, u32]>, u32]"`. Due to the `<` of the tuple, the wrong branch of `setImports` was taken. Swapping the order of the branches causes the right branch to be taken, but the regex-based splitting then fails -- the regex would return `ITuple<[u32, u32` due to the nested `[]`. As an aside, the regex also seemed to be incorrect for the `[a|b, c|d]` type: it would return `['a', 'b,c', 'd']`, i.e. not splitting `b,c`. The `splitAlternatives` function is meant to address these issues. 

When reviewing, please also note the change of `type.startsWith('[')` to `type.includes('[')`. All in all I am not 100% confident of the correctness of this pr. The type generation no longer throws an error, but I don't know whether or not the output is correct, and whether the change works for all possible types.

### DIFF
--- a/packages/typegen/src/util/imports.ts
+++ b/packages/typegen/src/util/imports.ts
@@ -33,10 +33,10 @@ export interface TypeImports {
   typeToModule: Record<string, string>;
 }
 
-// returns the top-level types in the alternatives list, taking into account nested [], <> and () items 
+// returns the top-level types in the alternatives list, taking into account nested [], <> and () items
 // E.g. for the string '[a<b,c>, d | e]' we return the array ['a<b,c>', 'd', 'e']
-function splitAlternatives(type: string): string[] {
-  let alternatives = [];
+function splitAlternatives (type: string): string[] {
+  const alternatives = [];
   let beginOfAlternative = 1;
   let level = 0;
 
@@ -47,8 +47,7 @@ function splitAlternatives(type: string): string[] {
         case ']':
         case ',':
         case '|':
-          const sub = type.substring(beginOfAlternative, i);
-          alternatives.push(sub.trim());
+          alternatives.push(type.substring(beginOfAlternative, i).trim());
           beginOfAlternative = i + 1;
           break;
       }
@@ -94,6 +93,7 @@ export function setImports (allDefs: Record<string, ModuleTypes>, imports: TypeI
       primitiveTypes[type] = true;
     } else if (type.startsWith('[') && type.includes('|')) {
       const splitTypes = splitAlternatives(type);
+
       setImports(allDefs, imports, splitTypes);
     } else if (type.includes('<') || type.includes('(') || type.includes('[')) {
       // If the type is a bit special (tuple, fixed u8, nested type...), then we
--- a/packages/typegen/src/util/imports.ts
+++ b/packages/typegen/src/util/imports.ts
@@ -33,6 +33,44 @@ export interface TypeImports {
   typeToModule: Record<string, string>;
 }
 
+// returns the top-level types in the alternatives list, taking into account nested [], <> and () items 
+// E.g. for the string '[a<b,c>, d | e]' we return the array ['a<b,c>', 'd', 'e']
+function splitAlternatives(type: string): string[] {
+  let alternatives = [];
+  let beginOfAlternative = 1;
+  let level = 0;
+
+  // we assume that the string starts with '['
+  for (let i = 1; i < type.length; i++) {
+    if (level === 0) {
+      switch (type[i]) {
+        case ']':
+        case ',':
+        case '|':
+          const sub = type.substring(beginOfAlternative, i);
+          alternatives.push(sub.trim());
+          beginOfAlternative = i + 1;
+          break;
+      }
+    }
+
+    switch (type[i]) {
+      case '[':
+      case '(':
+      case '<':
+        level++;
+        break;
+      case ']':
+      case ')':
+      case '>':
+        level--;
+        break;
+    }
+  }
+
+  return alternatives;
+}
+
 // Maps the types as found to the source location. This is used to generate the
 // imports in the output file, dep-duped and sorted
 /** @internal */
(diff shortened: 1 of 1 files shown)


jacogr:
Thank you. I believe if 

(a) it works on the problematic metadata,
(b) it yields the same output when running `yarn build:metadata` (so no changes)

then we should be ok. Will dive through the logic. (There are some small linting issues CI picked up, it is a bit pedantic...)

sander2:
Regarding (a), this is the output it generates:
```
      bootstrapPersonalSupply: AugmentedQuery<ApiType, (arg: ITuple<[ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]>, AccountId32]> | [ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]> | [InterbtcPrimitivesCurrencyId | { Token: any } | { ForeignAsset: any } | { LendToken: any } | { LpToken: any } | { StableLpToken: any } | string | Uint8Array, InterbtcPrimitivesCurrencyId | { Token: any } | { ForeignAsset: any } | { LendToken: any } | { LpToken: any } | { StableLpToken: any } | string | Uint8Array], AccountId32 | string | Uint8Array]) => Observable<ITuple<[u128, u128]>>, [ITuple<[ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]>, AccountId32]>]> & QueryableStorageEntry<ApiType, [ITuple<[ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]>, AccountId32]>]>;
```

Focusing on the `arg` type, and substituting the expanded currencyId for brevity, the generated type is:
```
ITuple<[ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]>, AccountId32]> | 
    [
        ITuple<[InterbtcPrimitivesCurrencyId, InterbtcPrimitivesCurrencyId]> | 
            [EXPANDED_CURRENCY_ID, EXPANDED_CURRENCY_ID], 
        AccountId32 | string | Uint8Array
    ]
```

So we have...
- Tuple(Tuple(CurrencyId), AccountId)
- Array(Tuple(CurrencyId), ExpandedAccountId)
- Array(Array(ExpandedCurrencyId), ExpandedAccountId)

But we don't have:
- Tuple(Array(ExpandedCurrencyId), AccountId)
- Tuple(Array(CurrencyId), AccountId)
- Tuple(Array(ExpandedCurrencyId), ExpandedAccountId)
- Tuple(Array(CurrencyId), ExpandedAccountId)
- Tuple(Tuple(ExpandedCurrencyId), ExpandedAccountId)
- Tuple(Tuple(CurrencyId), ExpandedAccountId)
- Array(Tuple(ExpandedCurrencyId), AccountId)
- Array(Tuple(CurrencyId), AccountId)
- Array(Array(ExpandedCurrencyId), AccountId)
- Array(Array(CurrencyId), AccountId)
- Tuple(Tuple(CurrencyId), AccountId)
- Tuple(Tuple(ExpandedCurrencyId), AccountId)
- Array(Tuple(CurrencyId), ExpandedAccountId)
- Array(Tuple(ExpandedCurrencyId), ExpandedAccountId)
- Array(Array(CurrencyId), ExpandedAccountId)

... That's a lot of alternatives.. Would these be expected to be present in the generated type?

jacogr:
Where the alternatives help. Think about a simple `enum` type, let'd define it as 

```rust
enum Sample {
  Alice(AccountId32),
  Bob(u32)
}
```

When it generates, it should look more-or-less like `Sample | { Alice: any } | { Bob: any } | Uint8Array` (I will get to my pet-peeve `any` right at the end).

So basically when using the argument, you can pass either -

- an actual `Sample` instance (crated via `createType`)
- an object of the form `{ Bob: 123 }` which the API then will convert via `createType`
- an already-encoded SCALE Uint8Array

The second option is what makes things easier to use since the API will convert to the correct type. Without having that the user always has to manually do the `createType` magic.

(On to any: it should in this case be something like `{ Bob: number | BN | Bbigint | string | Uint8Array }` but the typegen doesn't go that deep. At some point it really should, but will probably make the signatures even more unwieldly, but with the benefit of not having `any` anywhere)

jacogr:
So summarize my answer above - it would be great if the alternatives are supplied, i.e. users can have a mixture of stuff-already-created and pure JS objects for all. Alternatives are all to make it easier to use with TS complaining about type mismatches. (Since the API does convert)

sander2:
> So summarize my answer above - it would be great if the alternatives are supplied, i.e. users can have a mixture of stuff-already-created and pure JS objects for all. Alternatives are all to make it easier to use with TS complaining about type mismatches. (Since the API does convert from anything _reasonable_ to the correct type)

Right, that makes sense. I just checked that this PR creates the same output as the unmodified code, when run on a metadata that does not contain nested tuples. Given that the output for nested tuples seems to be correct (even though incomplete), and that the failing type generation is blocking us, would you be open to merging this PR and opening an issue for generating the alternatives?

jacogr:
Happy to merge as-is since it does move use into a much better spot, although it is incomplete.

jacogr (changes requested):
Happy with the output as-is. (Also tested the changes with `yarn build:metadata` and it is equivalent) 

Just needs some small adjustments to make CI happy.

sander2:
Thanks! I pushed a fix for the formatting and made an issue for the missing alternatives: https://github.com/polkadot-js/api/issues/5399

jacogr (approved):
Thank you.

Issue #490, opened 2018-12-18
Use TS bigint instead of BN


amaurymartiny:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#bigint

This is a big refactor.

jacogr:
Closing. Will re-visit when -

- the tooling is stable
- people aren't scared of it
- we have a Polkadot network out

Realistically, not in 2019.

PR #5449, opened 2023-01-26
Add BigInt composite

jacogr:
Needs https://github.com/polkadot-js/common/pull/1716

Last piece for issues such as https://github.com/polkadot-js/api/issues/5706, https://github.com/polkadot-js/api/issues/3046,  https://github.com/polkadot-js/api/issues/490

TODO:

- We may want to convert the standard `Int` to also use a wrapper (like here) - it is obviously a major breaking change, however does prepare the road for swap in the future

### DIFF
--- a/packages/types-codec/src/abstract/IntBigInt.ts
+++ b/packages/types-codec/src/abstract/IntBigInt.ts
@@ -75,7 +75,7 @@ function decodeBigInt (value: Exclude<AnyNumber, Uint8Array> | Record<string, st
 }
 
 /**
- * @name AbstractInt
+ * @name AbstractBigInt
  * @ignore
  * @noInheritDoc
  */


packages/types-codec/src/abstract/IntBigInt.ts
```diff
@@ -0,0 +1,272 @@
...
+    this.isUnsigned = !isSigned;
+
+    const isNegative = this.$ < _0n;
+    const maxBits = bitLength - (isSigned && !isNegative ? 1 : 0);
+
+    if (isNegative && !isSigned) {
+      throw new Error(`${this.toRawType()}: Negative number passed to unsigned type`);
+    } else if (this.#bitLengthInitial > maxBits) {
+      throw new Error(`${this.toRawType()}: Input too large. Found input with ${this.#bitLengthInitial} bits, expected ${maxBits}`);
+    }
```
spirinvladimir, line 111: It would be a bit more efficient to place validation before creating instance of a class because in case invalid state no extra operations for fill instance properties.

```suggestion
    if (this.$ < _0n && !isSigned) {
      throw new Error(`${this.toRawType()}: Negative number passed to unsigned type`);
    }
    this.#bitLengthInitial = this.$.toString(2).length;
    if (this.#bitLengthInitial >= maxBits) {
      throw new Error(`${this.toRawType()}: Input too large. Found input with ${this.#bitLengthInitial} bits, greater ${maxBits}`);
    }
    this.#bitLength = bitLength;
    this.#encodedLength = this.#bitLength / 8;
    this.isUnsigned = !isSigned;
```

kmk142789 (approved):

PR #5467, opened 2023-02-02
General WsProvider extensibility improvements

### LINKED ISSUE #5408: Ease WsProvider extensibility
- [ ] Bug report
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

`WsProvider` has a built-in non-extensible retry logic. While useful on its own, some more advanced scenario could be enabled if `WsProvider` extensibility was enhanced.

I am working on implementing a more advanced reconnection mechanism for `WsProvider`. Ideally such an extended WsProvider could track current performance and initiate a connection to a new endpoint of its choosing.
It appears to me that the `WsProvider` class could be made a little more accessible to facilitate that task.

Would you be open for a PR going in that direction?

Specifically, some ideas I have in mind:

* expose the `endpoint` currently used
* allow subclasses/external entities to consume per endpoint stats (via emitter -perf impact?- or user provided interface))
* make the `endpoint` selection mechanism overrideable (and have default implementation be the current one)

I don't expect that this would bring any breaking changes.

jeluard:
Allow to extend WsProvider so that informed logic can trigger endpoint selection and reconnection.

Fixes #5408

### DIFF
--- a/packages/rpc-provider/src/ws/index.ts
+++ b/packages/rpc-provider/src/ws/index.ts
@@ -181,6 +181,10 @@ export class WsProvider implements ProviderInterface {
     return new WsProvider(this.#endpoints);
   }
 
+  protected selectEndpointIndex (endpoints: string[]): number {
+    return (this.#endpointIndex + 1) % endpoints.length;
+  }
+
   /**
    * @summary Manually connect
    * @description The [[WsProvider]] connects automatically by default, however if you decided otherwise, you may
@@ -189,7 +193,7 @@ export class WsProvider implements ProviderInterface {
   // eslint-disable-next-line @typescript-eslint/require-await
   public async connect (): Promise<void> {
     try {
-      this.#endpointIndex = (this.#endpointIndex + 1) % this.#endpoints.length;
+      this.#endpointIndex = this.selectEndpointIndex(this.#endpoints);
 
       // the as typeof WebSocket here is Deno-specific - not available on the globalThis
       this.#websocket = typeof xglobal.WebSocket !== 'undefined' && isChildClass(xglobal.WebSocket as typeof WebSocket, WebSocket)
--- a/packages/rpc-provider/src/ws/index.ts
+++ b/packages/rpc-provider/src/ws/index.ts
@@ -3,7 +3,7 @@
 
 /* eslint-disable camelcase */
 
-import type { JsonRpcResponse, ProviderInterface, ProviderInterfaceCallback, ProviderInterfaceEmitCb, ProviderInterfaceEmitted, ProviderStats } from '../types';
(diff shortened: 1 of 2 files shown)


jeluard:
@jacogr Looking into this more, it looks to me total/active stats are not entirely aligned across providers (e.g. `requests`).
I think it makes sense to offer more stats for `active` though, do you have any recommendation on how you would see this being implemented?

packages/rpc-provider/src/ws/index.ts
```diff
@@ -268,13 +280,7 @@ export class WsProvider implements ProviderInterface {
    * @description Returns the connection stats
    */
   public get stats (): ProviderStats {
-    return {
-      active: {
-        requests: Object.keys(this.#handlers).length,
-        subscriptions: Object.keys(this.#subscriptions).length
-      },
```
jacogr, line 275: We lose this information - which is actually the only useful as-is information, current connections. (This is due to the rework of the structure above). We are actually interested in this, not just historic.

jacogr (commented):
I would suggest addition an additional field, not breaking active.

What you actually want is a `endpoint` entry in `ProviderStats`, not adjusting active.

jeluard:
@jacogr Is this acceptable to you?

packages/rpc-provider/src/ws/index.ts
```diff
@@ -174,13 +180,21 @@ export class WsProvider implements ProviderInterface {
...
+  }
+
   /**
    * @description Returns a clone of the object
    */
   public clone (): WsProvider {
     return new WsProvider(this.#endpoints);
   }
 
+  protected selectEndpointIndex (endpoints: string[]): number {
```
jacogr, line 194: Not sure why we are passing through the param here from the usage, i.e. line 206 below - should we not just use `this.#endpoints` here?

Additionally, I'm actually unsure about the utility here of having this function. Since it is only used in the one place (like 206 below), not sure why it is extracted since it doesn't actually set any values itself. So by the definition here -

1. it is external to the class - which means it is callable on the public interface
2. calling it external actually no no impact, i.e. doesn't set anything?

So circling back, I'm either missing something of don't see why this is now on the public interface here since it is a best (if not called below) a no-op.

jacogr (commented):
Once small comment - the stats + endpoint expose looks sane, just unsied about the select utility.

packages/rpc-provider/src/ws/index.ts
jeluard: The general idea is to allow subclasses to provide their own logic for endpoint selection (e.g. the one with lowest ping latency, the one with least previous errors per unit of time, ..).
It takes `endpoints` as parameters as it looks to me that having it as a field is an implementation detail that could evolve over time. If it was ok to break compatibility, I probably would have introduced some separate interface to model this.

jacogr (approved):

PR #5479, opened 2023-02-09
update MMR API

### LINKED ISSUE #5478: mmr_generateProof params is not correct
https://github.com/paritytech/substrate/blob/38c3e5ee10be760115a77a115ef0ac6ba381542b/client/merkle-mountain-range/rpc/src/lib.rs#L98 

this rpc method params need to update.

* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

  - Version:
  - Environment:

    - [ ] Node.js
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

Lederstrumpf:
fixes #5478


packages/types/src/interfaces/mmr/runtime.ts
```diff
@@ -64,38 +58,6 @@ export const runtime: DefinitionsCall = {
           ],
           type: 'Result<(), MmrError>'
         },
-        verify_proof: {
```
jacogr, line 67: The previous versions should not be removed. They can be available on older nodes.

The API decorates based on the `rpc_methods` call and will inject what is available on the RPC node.

Lederstrumpf: Taking account of your comment over here too:
https://github.com/polkadot-js/api/issues/5484#issuecomment-1430851882

The breaking change is deliberate since we've also reset MMR storage here: https://github.com/paritytech/devops/issues/2287
Thus, any proofs generated or supported by the old API will not bear meaning for the reset MMR.
I fully agree we should version the MMR RPC methods henceforth, but given the MMR reset, I reckon it's superfluous for this first iteration.

Lederstrumpf:
I've tested the API methods here: https://github.com/Lederstrumpf/polkadot-js-api-ts-template/blob/test-mmr-rpc/src/index.ts

I'm also fine to remove the `mmr_root` method here to reduce the API's surface since it can equivalently be performed with `historic_api.query.mmr.rootHash()`; I added for completeness since it's provided by RPC nodes.

packages/types/src/interfaces/mmr/runtime.ts
jacogr: The issue is that there are existing chains that could uses these RPCs - we cannot just remove what is there since it is not in master.

Old RPCs should stay since they are decorated on un-updated chains.

jacogr (dismissed):
Really appreciate all the effort here: thank you. 

Could you please just merge in master, GH shows some conflicts.

Lederstrumpf:
> Really appreciate all the effort here: thank you.

gladly!

> Could you please just merge in master, GH shows some conflicts.

Done. As a followup to the merge, I've fully decoupled the MMR runtime methods wrt versioning in https://github.com/polkadot-js/api/pull/5479/commits/ff9976410e8ada9f0ffe2ef2cf8f02658d1aa548  (previously `MMR_V1_V2` on fresh master). Since all of the methods' signatures changed from v1 to v2, there's no intersection between those versions. So this struck me as the most correct approach under the circumstances, but please let me know if I'm wrong here or you prefer it differently.

jacogr:
It is quite possible I mucked up something when adding the MMR V1/V2 split. Was also not quite comfortable with it anyway since it seemed that methods were added and then the version was only bumped afterwards - this could quite possibly just be the update schedule followed when these were adjusted.

Anyway, I'm much happier when it comes "from the source of knowledge" on these since the intent is much clearer.

packages/types/src/interfaces/mmr/rpc.ts
```diff
@@ -35,5 +37,31 @@ export const rpc: DefinitionsRpc = {
...
-  }
+  },
+
+  verifyProof: {
+    description: 'Verify an MMR proof',
+    params: [
+      {
+        name: 'proof',
+        type: 'MmrLeafBatchProof'
+      },
```
jacogr, line 48:
```suggestion
      }
```

packages/types/src/interfaces/mmr/rpc.ts
```diff
@@ -35,5 +37,31 @@ export const rpc: DefinitionsRpc = {
...
+    description: 'Verify an MMR proof statelessly given an mmr_root',
+    params: [
+      {
+        name: 'root',
+        type: 'MmrHash'
+      },
+      {
+        name: 'proof',
+        type: 'MmrLeafBatchProof'
+      },
```
jacogr, line 63:
```suggestion
      }
```

packages/types/src/interfaces/mmr/rpc.ts
```diff
@@ -35,5 +37,31 @@ export const rpc: DefinitionsRpc = {
...
+    ],
+    type: 'bool'
+  },
```
jacogr, line 66:
```suggestion
  }
```

packages/types/src/interfaces/mmr/runtime.ts
```diff
@@ -3,7 +3,61 @@
...
+    description: 'Generate MMR proof for the given block numbers.',
+    params: [
+      {
+        name: 'blockNumbers',
+        type: 'Vec<BlockNumber>'
+      },
+      {
+        name: 'bestKnownBlockNumber',
+        type: 'Option<BlockNumber>'
+      },
```
jacogr, line 17:
```suggestion
      }
```

packages/types/src/interfaces/mmr/runtime.ts
```diff
@@ -3,7 +3,61 @@
...
+        name: 'leaves',
+        type: 'Vec<MmrEncodableOpaqueLeaf>'
+      },
+      {
+        name: 'proof',
+        type: 'MmrBatchProof'
+      }
+    ],
+    type: 'Result<(), MmrError>'
+  },
```
jacogr, line 57:
```suggestion
  }
```

jacogr (changes requested):
Some small linting fixes suggested (as per CI failures) - there is however still an ordering issue, couldn't do that is the GH interface without mucking up.

Lederstrumpf:
> It is quite possible I mucked up something when adding the MMR V1/V2 split. Was also not quite comfortable with it anyway since it seemed that methods were added and then the version was only bumped afterwards - this could quite possibly just be the update schedule followed when these were adjusted.

Indeed - I post-bumped the versions here https://github.com/paritytech/substrate/pull/13509 after your comment regarding this (https://github.com/polkadot-js/api/pull/5479#discussion_r1118324350).

I acknowledge this compromise unfortunately leaves us with the inconsistency that nodes after ~v0.9.31 (iirc, the release where the first of these new/changed methods were introduced) but prior to v0.9.40 (where the version bump occurred) will report their API as being V1. Not bumping the API version would have avoided that, but begotten issues for earlier releases, hence the compromise.

> Some small linting fixes suggested (as per CI failures) - there is however still an ordering issue, couldn't do that is the GH interface without mucking up.

Should be fixed now, if you rerun the workflow.

jacogr (approved):
Thank you.

(Will be included in the next release over the weekend)

Issue #5483, opened 2023-02-14
Unexpected values received for types i8 and i16 from ink! contract


sharma66mahesh:
* **I'm submitting a ...**

  - [x] Bug report

* **What is the current behavior and expected behavior?**
Upon uploading [this ink! contract](https://pastebin.com/z2sehh14), I was expecting to receive the values of the variables `my_i8` and `my_i16` to be `127` and `32767` as it was set in the ink! contract respectively.

However, I received `-129` and `-32769` as the values for `my_i8` and `my_i16` respectively.

The following is the output of querying the contract's `get` method:
```
{"msg":"ICE/SNOW Network","u8Arr":"0x010203","value":"0x7fffffffffffffffffffffffffffffff","isTrue":true,"myAccount":"npRm3oLNUahbPgUnwjPYw9oLWdEigk6aHzwtvj4nibfAJMYVo","myBalance":123123123123,"myHash":"0xe0d83c067d9abf593a8089ef1f21fc30fafb02a8dd67a862f8ca47eb158735b9","myVec":[-65535,-65534,-65533,-65532,-65531,-65530,-65529,-65528,-65527,-65526,-65525],"myStruct":{"id":1,"status":"Invalid","strArr":["Str1","Str2"]},"myIntStruct":{"myU8":255,"myU16":65535,"myU32":4294967295,"myU64":"0xffffffffffffffff","myU128":"0xffffffffffffffffffffffffffffffff","myI8":-129,"myI16":-32769,"myI32":2147483647,"myI64":"0x7fffffffffffffff","myI128":"0x7fffffffffffffffffffffffffffffff"}}
```

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**
  - Version: [polkadot-v0.9.36](https://github.com/paritytech/substrate/tree/polkadot-v0.9.36/frame/contracts), @polkadot-js/api v9.13.6, @polkadot-js/api-contract v9.13.6
  - Environment:

    - [x] Node.js
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [x] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

cmichi:
@statictype @DoubleOTheven Any idea?

DoubleOTheven:
Hey @sharma66mahesh, would you please share the metadata for that contract?

I would try updating polkadotjs to the latest version in the mean time

sharma66mahesh:
[...]

PR #5506, opened 2023-02-23
Adapt to ink! 4.0

statictype:
According to the [migration guide ](https://use.ink/faq/migrating-from-ink-3-to-4#new-entrance-ink-crate) there is a new top-level crate where everything should be used from, so the types overrides should be updated.  
Might be the reason for the unexpected values in https://github.com/polkadot-js/api/issues/5483
Haven't really tested this, don't know how to run the [weight v2 contracts apps](https://polkadotjs-apps.web.app/#/contracts) locally

### DIFF
--- a/packages/types/src/metadata/PortableRegistry/PortableRegistry.ts
+++ b/packages/types/src/metadata/PortableRegistry/PortableRegistry.ts
@@ -70,6 +70,8 @@ const PATHS_ALIAS = splitNamespace([
   '*_runtime::RuntimeCall',
   '*_runtime::RuntimeEvent',
   // ink!
+  'ink::env::types::*',
+  'ink::primitives::types::*',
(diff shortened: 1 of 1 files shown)


jacogr (commented):
Thank you. Will take a look, but probably not soon-ish (i.e. within the next couple of days). Down with a bad case of covid.

TL:DR Hang tight.

jacogr (approved):
Seems innocent enough.

jacogr:
As for testing - a whole suite of v4 contracts would need to be added here https://github.com/polkadot-js/api/tree/master/packages/api-contract/src/test/contracts/ink/v4 and it should update itself when running locally.

There is 1 contract, from a alpha - which should probably be replaced with final.

statictype:
get well soon @jacogr! 🍀

PR #5531, opened 2023-03-09
Fix inaccessible yargs instance error

conr2d:
> In ES6, yargs is no longer a singleton, so it needs to call yargs() to get an instance of yargs. [[link]](https://github.com/yargs/yargs/issues/1854#issuecomment-787509517)

This issue breaks the [type generation example](https://github.com/polkadot-js/docs/tree/master/docs/api/examples/promise/typegen) in polkadot.js documentation, when executing it with the latest dependencies. This PR fixes the issue.

### DIFF
--- a/packages/typegen/src/extractChain.ts
+++ b/packages/typegen/src/extractChain.ts
@@ -6,6 +6,7 @@
 
 import process from 'process';
 import yargs from 'yargs';
+import { hideBin } from 'yargs/helpers';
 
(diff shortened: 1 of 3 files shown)


jacogr (approved):
Thank you. (Looks sane, aka what needed to be done elsewhere as well)

PS: Actually have the same in https://github.com/polkadot-js/api/pull/5530 (but that would be some time before it gets in + smaller PRs are best)

Issue #5540, opened 2023-03-15
Typegen Usage which worked with 9.14.2 doesn't work with 10.1.2


rgoulter:
* **I'm submitting a ...**

  - [ ] Bug report
  - [ ] Feature request
  - [X] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Code which worked with `@polkadot/api` `v9.14.2` doesn't work with `v10.1.2`. -- Likely same as https://github.com/polkadot-js/api/issues/5524#issuecomment-1468321718

Some clarification on what changes are needed would be useful.

Example code (& the commands used) can be found here: https://github.com/rgoulter/polkadot-js-typegen-efinity -- It's a pretty bare-bones usage of `typegen`, to generate the code, and includes a small `example` TypeScript program which uses the generated types.

The latest commit there updates the `@polkadot/typegen` to `10.1.2` (and in `example/package.json`, its `@polkadot/api` to `10.1.2`). 

I'm having trouble figuring out what changes I need to make to get it to work with `10.1.2`.

### What goes wrong:

When it runs `polkadot-types-from-defs`, it has problems loading the `definitions.ts`:

```
> ts-node --skip-project node_modules/.bin/polkadot-types-from-defs --package @efinity/api-augment/interfaces --input ./src/interfaces --endpoint ./metadata.json

ERROR: Unable to load user definitions: Unknown file extension ".ts" for /home/rgoulter/scratch/efinity-api-augment/src/interfaces/definitions.ts
```

When running `polkadot-types-from-chain`, some `.ts` code is generated, but the generated code then leads to errors when trying to build it.

```
> ts-node --skip-project node_modules/.bin/polkadot-types-from-chain --endpoint ./metadata.json --output ./src/interfaces

Generating from metadata, 254,086 bytes
ERROR: No lookup definitions found: Unknown file extension ".ts" for /home/rgoulter/scratch/efinity-api-augment/src/interfaces/lookup.ts
2023-03-15 14:41:52        REGISTRY: Unknown signed extensions CheckFuelTank found, treating them as no-effect
/home/rgoulter/scratch/efinity-api-augment/src/interfaces/augment-api-consts.ts
        Generating
        Writing

...
```

Sample errors when trying to run `tsc` in `example`:

```
node_modules/@efinity/api-augment/src/interfaces/augment-api-consts.ts:166:17 - error TS2304: Cannot find name 'FrameSupportPalletId'.

[...]

PR #5557, opened 2023-03-22
Resolve/reject once in getWsData for typegen

jacogr:
As mentioned in https://github.com/polkadot-js/api/issues/5540#issuecomment-1479522597

Here we add a promise tracker to ensure we resolve/reject once and don't "just exit" from typegen retrieval

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Contributed:
 
 Changes:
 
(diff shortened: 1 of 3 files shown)


jacogr:
Final test for this would be when bumping the metadata in the API - looks sane otherwise and should solve the root issue reported.

PR #5142, opened 2022-08-09
Extend getAuthorDetails to handle Nimbus for the MantaNetwork usecase

Garandor:
We over at [Manta & Calamari Network](https://www.manta.network/) will be using PureStake/nimbus in conjunction with `pallet_session` for ConsensusKey<->AuthorID mapping instead of `pallet_author_mapping`.
Eventually we'll upstream these changes if PureStake so wish.

I have extended the `getAuthorDetails` util to handle both cases and extended the ConsensusEngineId as well to contain an entry for the nimbus consensus engine.

This PR is WIP, I pushed it looking for comments and hints on improving this, as this is my first run-in with polkadot.js code-wise.

Signed-off-by: Adam Reif <Garandor@manta.network>

PR #5559, opened 2023-03-23
support manta author lookup

grenade:
manta and calamari use nimbus consensus which is similar to moonbeam's implementation of the same.

this pr adds support for mapping nimbus session keys to the collator account it is bound to when getAuthorDetails() is called for chains using nimbus consensus without pallet_author_mapping.

### DIFF
--- a/packages/api-derive/src/chain/util.ts
+++ b/packages/api-derive/src/chain/util.ts
@@ -24,34 +24,30 @@ export function createBlockNumberDerive <T extends { number: Compact<BlockNumber
 }
 
 export function getAuthorDetails (header: Header, queryAt: QueryableStorage<'rxjs'>): Observable<[Header, Vec<AccountId> | null, AccountId | null]> {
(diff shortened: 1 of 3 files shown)


grenade:
looks like there's more work to do here. still missing the actual mapping. current implementation just shows the nimbus session account in ss58 format rather than the collator account.

jacogr:
There was also this PR (which is still in draft as well) - https://github.com/polkadot-js/api/pull/5142

(Not sure if that should be closed)

grenade:
> There was also this PR (which is still in draft as well) - #5142
> 
> (Not sure if that should be closed)

feel free to close ( i'll also ask garandor to close). i'll resubmit a pr that contains the mapping shortly...

grenade:
i've created [pr 5561](https://github.com/polkadot-js/api/pull/5561) to address the mappings for manta/calamari

PR #5561, opened 2023-03-23
support manta author lookup

grenade:
manta and calamari use nimbus consensus which is similar to moonbeam's implementation of the same.

this pr adds support for mapping nimbus session keys to the collator account it is bound to when getAuthorDetails() is called for chains using nimbus consensus without pallet_author_mapping.

### DIFF
--- a/packages/api-derive/src/chain/util.ts
+++ b/packages/api-derive/src/chain/util.ts
@@ -36,7 +36,7 @@ export function getAuthorDetails (header: Header, queryAt: QueryableStorage<'rxj
 
   if (loggedAuthor) {
 
-    // use the author mapping pallet if available (ie: moonbeam, moonriver)
+    // use the author mapping pallet, if available (ie: moonbeam, moonriver), to map session (nimbus) key to author (collator/validator) key
     if (queryAt.authorMapping && queryAt.authorMapping.mappingWithDeposit) {
       return combineLatest([
         of(header),
@@ -46,7 +46,7 @@ export function getAuthorDetails (header: Header, queryAt: QueryableStorage<'rxj
       ]);
     }
 
-    // use the session pallet if available (ie: manta, calamari)
+    // fall back to session pallet, if available (ie: manta, calamari), to map session (nimbus) key to author (collator/validator) key
     if (queryAt.session && queryAt.session.queuedKeys) {
       return combineLatest([
         of(header),
--- a/packages/api-derive/src/chain/util.ts
+++ b/packages/api-derive/src/chain/util.ts
@@ -41,9 +41,9 @@ export function getAuthorDetails (header: Header, queryAt: QueryableStorage<'rxj
       ? queryAt.authorMapping.mappingWithDeposit<IOption<{ account: AccountId } & Codec>>(authorSessionKey).pipe(
           map(opt => opt.unwrapOr({ account: null }).account)
         )
-      // use the author session pallet if available (ie: manta, calamari)
+      // use the session pallet if available (ie: manta, calamari)
       : (queryAt.session && queryAt.session.queuedKeys)
-        ? queryAt.session.queuedKeys<Vec<(AccountId, { nimbus: Address })>>().pipe(
+        ? queryAt.session.queuedKeys<Vec<[AccountId, { nimbus: Address }]>>().pipe(
             catchError(() => of(null)), // handle scenarios where queuedKeys is not of the expected type
             map((queuedKeys) => queuedKeys.find(([_, { nimbus }]) => nimbus.toHex() === authorSessionKey.toHex())),
             map(([collator]) => collator || null)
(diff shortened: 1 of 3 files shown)


grenade:
i believe the earlier build failure was caused by an edge case in the testing where `authorSessionKey` had obtained a value other than null (should not be possible outside of nimbus), but session.queuedKeys did not conform to to the type expected in nimbus `<Vec<(AccountId, { nimbus: Address })>>`.
outside of the tests, i don't believe this scenario is possible, but i've put in a handler that reverts to omitting the author account if that happens.

jacogr:
Just run yarn lint locally 

… and yes, it is an issue with first time contributors (I end up being a bottleneck with having to click)

grenade:
thanks for the tip about yarn lint. i believe i've resolved the issues with this change.

grenade:
well that was a learning curve. i broke chatgpt several times. `yarn lint` and `yarn build` succeed locally. i don't know my way around deno so i'm just crossing my fingers that that part of the ci will be happy with the last update.

jacogr:
Deno sometimes just picks up extra TS issues. (It is however flakey, i.e. sometimes the actual deno.land/x gives 500's, so generally use it as guidance, if it does pickup extra issues, and not "law")

grenade:
sorry about that last push! i'm done now as long as the build is happy.

packages/api-derive/src/chain/util.ts
```diff
@@ -24,34 +31,32 @@ export function createBlockNumberDerive <T extends { number: Compact<BlockNumber
...
-    of(header),
-    queryAt.session
-      ? queryAt.session.validators()
-      : of(null),
-    of(null)
-  ]);
+        : null
+    : null;
+
+  return combineLatest([of(header), of(validators), of(mappedAuthor)]) as Observable<[Header, Vec<AccountId> | null, AccountId | null]>;
```
jacogr, line 61: This cast should not be needed - `combineLatest` does actually return the correct types.

The `of(validators)` here seems suspect since it could be an `Observable<AccountId>` or `null` (as opposed to `Observable<null>` as per lines 41 to 43 above). 

So with this is seems to be `Observable<Observable<AccountId>>` or `Observable<null>` since we apply `of` on it in combineLatest.

packages/api-derive/src/chain/util.ts
```diff
@@ -38,9 +38,7 @@ export function getAuthorDetails (header: Header, queryAt: QueryableStorage<'rxj
     (log.isPreRuntime && log.asPreRuntime[0].isNimbus && log.asPreRuntime[1])
   )) || null;
 
-  const validators = (queryAt.session)
-    ? queryAt.session.validators()
-    : null;
+  const validators = (queryAt.session) ? queryAt.session.validators() : null;
```
jacogr, line 41: This is where the issue starts and why we need a cast below -

The types between these two are different (the one being an Observable, the other not), so this pushed through to the actual returntype. Originally we have `Observable<AccountId> | Observable<null>`, now it is `Observable<AccountId> | null`

Garandor: CI looks to have succeeded with the last run.
Does this work or do you prefer the cast-less version?

grenade: the version before my last commit didn't actually work (despite successful lint and build).

for whatever reason, having a single return statement at the end of the method and populating the final element of the combinelatest array arg with either a `mappedAuthor` (nimbus) or `of(null)` was refusing to lint and build.

my efforts to replace `combineLatest` were flawed and as it turns out, also unnecessary.

having three separate return paths (nimbus/moonbeam, nimbus/manta, generic) does both lint and build and also does not throw deprecation errors on `combineLatest`. i can't say i understand why, but i assume that only specific invocation signatures are deprecated.

earlier versions (of my changes in this pr) were also flawed in the type declaration of the return value for `session.queuedKeys()`, which in the case of manta/nimbus is as simple as `Array<[AccountId, { nimbus: Address }]>`. something i only learned through a bunch of trial and error here.

grenade:
### testing notes

to test these changes, i took the following steps:
- run `yarn lint`, verify absence of errors
- run `yarn build`, verify absence of errors
- clone https://github.com/polkadot-js/apps.git
- run `yarn` (inside the apps folder)
- copy `api/node_modules/@polkadot/api-derive/build/chain/util.*` to `apps/node_modules/@polkadot/api-derive/chain/` (overwriting what the earlier yarn install step had put in the target folder)
- run `yarn run start` (in the apps folder)
- visit the explorer pages of:
  - [polkadot](http://localhost:3000/?rpc=wss%3A%2F%2Frpc.polkadot.io#/explorer)
  - [kusama](http://localhost:3000/?rpc=wss%3A%2F%2Fkusama-rpc.polkadot.io#/explorer)
  - [moonbeam](http://localhost:3000/?rpc=wss%3A%2F%2Fwss.api.moonbeam.network#/explorer)
  - [moonriver](http://localhost:3000/?rpc=wss%3A%2F%2Fwss.api.moonriver.moonbeam.network#/explorer)
  - [manta](http://localhost:3000/?rpc=wss%3A%2F%2Fws.manta.systems#/explorer)
  - [calamari](http://localhost:3000/?rpc=wss%3A%2F%2Fws.calamari.systems%2F#/explorer)
  and verified that all show author address links

jacogr (approved):
Seems sane and easy to follow.

Thank you.

PR #5568, opened 2023-03-31
decorateMethod with better internal TS typing

jacogr:


jacogr:
We will have to revisit, this is now completely stale.

Issue #5577, opened 2023-04-04
Add new `FungiblesApi` runtime api for statemine/statemint/westmint


bkontur:
- [ ] Bug report
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

I would like to add new `FungiblesApi` runtime api - to be seen in `polkadot.js.org/apps` for statemine/statemint/westmint.
How should I do this?
I tried to push this, but I am not able to push to this repo, it should be like this:

`packages/types/src/interfaces/fungibles/runtime.ts`
```
// Copyright 2017-2023 @polkadot/types authors & contributors
// SPDX-License-Identifier: Apache-2.0

import type { DefinitionsCall } from '../../types/index.js';

export const runtime: DefinitionsCall = {
  FungiblesApi: [
    {
      methods: {
        query_account_balances: {
          description: 'Returns the list of all `MultiAsset` that an `AccountId` has',
          params: [
            {
              name: 'account',
              type: 'AccountId'
            }
          ],
          type: 'Result<Vec<MultiAsset>, FungiblesAccessError>'
        }
      },
      version: 1
    }
  ]
};
```

`packages/types/src/interfaces/fungibles/definitions.ts`
```
// Copyright 2017-2023 @polkadot/types authors & contributors
// SPDX-License-Identifier: Apache-2.0

// order important in structs... :)
/* eslint-disable sort-keys */

import type { Definitions } from '../../types/index.js';

import { runtime } from './runtime.js';

export default {
  rpc: {},
  runtime,
  types: {
    FungiblesAccessError: {
      _enum: ['AssetIdConversionFailed', 'AmountToBalanceConversionFailed']
    },
  }
} as Definitions;

```

bkontur:
@jacogr ok, thank you, I started PR https://github.com/polkadot-js/api/pull/5584,
but I have really no idea how to check/test/regenerate whatever to finish it,
can you, please, point me, at least how to "compile" it locally?

Issue #5489, opened 2023-02-16
Where's xcm v3?


gilescope:
Should https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/xcm/v3.ts exist? That would be great if xcm v3 was supported.

jacogr:
PRs welcome if you are in a rush. It is somewhere on my TODO list.

jacogr:
From metedata, as per the connected chain, e.g. for Polkadot - https://github.com/polkadot-js/api/blob/6ea6b90e0395819c1d7ded0a2774710130eba892/packages/types-augment/src/lookup/types-polkadot.ts#L113-L122

PR #5584, opened 2023-04-05
Fungibles runtime api for statemine/statemint/westmint

bkontur:
Fixes https://github.com/polkadot-js/api/issues/5577


packages/types/src/interfaces/fungibles/definitions.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { runtime } from './runtime.js';
+
+export default {
+  rpc: {},
+  runtime,
+  types: {
+    FungiblesAccessError: {
+      _enum: ['AssetIdConversionFailed', 'AmountToBalanceConversionFailed']
+    },
```
jacogr, line 17:
```suggestion
    }
```

jacogr (changes requested):
Thank you!

One small comment where CI complained, the rest looks spot-on.

jacogr (dismissed):
One small addition needed - we just need to add an import in this file - https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/definitions.ts

Basically something like -

```js
export { default as fungibles } from './fungibles/definitions.js';
```

Just so it is available externally. (The above just goes in the first substrate section, alphabetically)

Apart from that, seems all good.

bkontur:
> One small addition needed - we just need to add an import in this file - https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/definitions.ts
> 
> Basically something like -
> 
> ```js
> export { default as fungibles } from './fungibles/definitions.js';
> ```
> 
> Just so it is available externally. (The above just goes in the first substrate section, alphabetically)
> 
> Apart from that, seems all good.


Fixed

packages/types/src/interfaces/fungibles/runtime.ts
```diff
@@ -0,0 +1,24 @@
...
+      methods: {
+        query_account_balances: {
+          description: 'Returns the list of all `MultiAsset` that an `AccountId` has',
+          params: [
+            {
+              name: 'account',
+              type: 'AccountId'
+            }
+          ],
+          type: 'Result<Vec<MultiAsset>, FungiblesAccessError>'
```
bkontur, line 18: @jacogr 
and I am not sure here with `MultiAsset`, should I import it here?

also here is missing `v3`:  https://github.com/polkadot-js/api/tree/master/packages/types/src/interfaces/xcm
https://github.com/polkadot-js/api/issues/5489

jacogr: The v3 "missing" is expected - all the V3 types are retrieved from the runtime.

I do understand where you are coming from since fixing it to "MultiAsset" means it uses the defined type and not the on-chain type.

(Cannot wait until we have runtime calls inside metadata and all these manual definitions are not required anymore)

bkontur: aha, ok, cool, thank you,

and one more dumb question, is there a way, how could I build and run `polkadot-js.org/apps` with this `api` from this PR locally against my local node? Just to verify, that it works as expected, because I am not sure, if `Vec<MultiAsset>` will work with `XcmVersionedMultiAssets`, because I am also thinking about to change runtime api to use `VersionedMultiAssets`

jacogr: Yes, indeed, you can test locally -

1. Assume you have a folder `<ROOT>`
2. Get both repos into that, i.e. `<ROOT>/apps` and `<ROOT>/api`
3. Make all change to the api
4. Inside the api folder run `yarn polkadot-dev-copy-to apps`
5. It should show it copies stuff
6. Go to the apps folder and start it, `yarn start`
7. Access on `localhost:3000`

jacogr (approved):
Thank you.

PR #5590, opened 2023-04-10
nimbus (manta/calamari implementation) author mapping correction

grenade:
in pull request https://github.com/polkadot-js/api/pull/5561, we introduced a session key to author mapping that was flawed. it relied on `session.queuedKeys()` to map nimbus session keys to the bound collator account, which only works if the collator is queued to author in subsequent rounds. this is often the case, but not always. this resulted in the correct collator account being shown most of the time and occasionally falling back to showing the ss58 representation of the bound nimbus session key whenever a collator is not queued to author in subsequent rounds.

this change corrects this mapping logic to instead use:
- `parachainStaking.selectedCandidates()` to obtain the list of all collators
- `session.nextKeys.multi(selectedCandidates)`: to obtain each of the candidate's bound session keys
this results in the correct collator account id being displayed all of the time.

### DIFF
--- a/packages/api-derive/src/chain/util.ts
+++ b/packages/api-derive/src/chain/util.ts
@@ -4,11 +4,12 @@
 import type { Observable } from 'rxjs';
 import type { QueryableStorage } from '@polkadot/api-base/types';
 import type { Compact, Vec } from '@polkadot/types';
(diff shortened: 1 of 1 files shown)


jacogr (approved):
Thank you.

PR #5592, opened 2023-04-14
Fungibles runtime api for statemine/statemint/westmint

bkontur:
follow up of https://github.com/polkadot-js/api/pull/5584

finally I was able to test it and this is working:
image

### DIFF
--- a/packages/types/src/interfaces/fungibles/runtime.ts
+++ b/packages/types/src/interfaces/fungibles/runtime.ts
@@ -15,7 +15,7 @@ export const runtime: DefinitionsCall = {
               type: 'AccountId'
             }
           ],
-          type: 'Result<Vec<MultiAsset>, FungiblesAccessError>'
+          type: 'Result<Vec<XcmV3MultiAsset>, FungiblesAccessError>'
         }
       },
       version: 1


jacogr:
Ummm... (on the test failure)

It tries to construct a type that is only there at runtime - if this runtime API is there, the type will be there.

I would just suggest that we tweak it (the alternative is obviously duplicating the type)

For the "types" we have actual overrides - https://github.com/polkadot-js/api/blob/e33139919f980c5cae29c639871b955a957fa80a/packages/types/src/interfaces/definitions.spec.ts#L55-L67

I would suggest adding the same to the runtime check section for `Result<Vec<XcmV3MultiAsset>, FungiblesAccessError>`. Somewhere here - https://github.com/polkadot-js/api/blob/e33139919f980c5cae29c639871b955a957fa80a/packages/types/src/interfaces/definitions.spec.ts#L124-L130

If you don't want to hack-around with it, I'm happy to merge these changes and make a small PR for that skip. Either way, lemme know.

bkontur:
@jacogr it would be cool, if you could merge this hack for skip.
anyway I am going to work on changing that runtime api to use better VersionedMultiAssets directly, so then we could remove that hack, but it will take some time

jacogr (approved):

PR #5594, opened 2023-04-14
contract metadata v5

kziemianek:
Constructor and message metadata is going to contain new attribute named `default`. The main reason behind adding it is to allow UIs (ex. `contracts-ui`) determine which one constructor/message should be selected as default in lists.

This is follow-up pull request to https://github.com/paritytech/ink/issues/1703


kziemianek:
I'm not sure if this should go to v4 or v5, finally I've decided to create v5.

packages/types/src/interfaces/contractsAbi/definitions.ts
```diff
@@ -170,6 +184,16 @@ const spec = {
     returnType: 'Option<ContractTypeSpec>',
     docs: 'Vec<Text>'
   },
+  ContractMessageSpecV3: {
+    label: 'Text',
+    selector: 'ContractSelector',
+    mutates: 'bool',
+    payable: 'bool',
+    default: 'bool',
```
jacogr, line 192: This is not correct - we cannot introduce fields retrospectively to existing versions. In v3 “default” was not available, it is o lay there from v5+

kziemianek: Ok, thank you. I wasn't certain about versioning here.
Would something like this be ok ?
```
  ContractMessageSpecV3: 'ContractMessageSpecV2',
  ContractMessageSpecV4: 'ContractMessageSpecV2',
  ContractMessageSpecV5: {
    label: 'Text',
    selector: 'ContractSelector',
    mutates: 'bool',
    payable: 'bool',
    default: 'bool',
    args: 'Vec<ContractMessageParamSpecV2>',
    returnType: 'Option<ContractTypeSpec>',
    docs: 'Vec<Text>'
  },```

kziemianek: @jacogr what do you think about my proposition? If it's still not correct, I kindly ask you to explicitly point how to name these types 🙃

peetzweg:
Would it be smart to adapt the pr to include this change from ink 4.2.0 as well? Otherwise we bump the metadata type version yet again in a short amount of time. 🤔 

https://github.com/polkadot-js/api/issues/5637

PR #5617, opened 2023-04-27
Remove private #, replace with private __$$_

jacogr:
To be reverted at some point when https://github.com/polkadot-js/dev/issues/966 hits. (Which hopefully then is a simple `s/__$$_/#/` and removing `private` identifiers for the affected items)

The issue is that the es2021 target has a "slight" performance impact due to the get/set private field polyfill. 

This is actually a regression when swapping from babel to pure tsc - in Babel, albeit incorrectly done so (for the intended target), we had pass-throughs on `#private` fields - which is now not the case.


jacogr:
Closing. Replaced by a catch-all https://github.com/polkadot-js/dev/pull/1029

PR #5624, opened 2023-05-01
Update .nvmrc

TarikGul:
This updates the `.nvmrc` to be inline with the package.json `engine` field.


.nvmrc
```diff
@@ -1 +1 @@
-14
+v18.14
```
jacogr, line 1: The engines is the absolute minimum.

18.15 is much better (in the implementation of the `node:test` framework), than 18.14. So while I don't use `.nvmrc` myself, would suggest at least that version for an auto-selector.

TarikGul: Sounds great, thanks for the suggestion. Updated!

TarikGul: After briefly reading the LTS notes on 18.15 I definitely agree.

jacogr (approved):
Thank you.

PR #5628, opened 2023-05-02
metadataMd: use endpoint for rpc and runtime details too

FlorianFranzen:
Fixes two TODOs in the code. Tested with node template.


jacogr (approved):
Thank you.

PR #5649, opened 2023-05-18
Allow external signer to be unset

polymath-eric:
On the SDK I work on a user requested the ability to switch it into "read only" mode.

I set my signer onto the polkadot instance with this call, but was unable to set it without a type cast here. It seems like this field is already `Signer | undefined`.

### DIFF
--- a/packages/api/src/base/index.ts
+++ b/packages/api/src/base/index.ts
@@ -58,7 +58,7 @@ export abstract class ApiBase<ApiType extends ApiTypes> extends Getters<ApiType>
   /**
    * @description Set an external signer which will be used to sign extrinsic when account passed in is not KeyringPair
    */
-  public setSigner (signer: Signer): void {
+  public setSigner (signer: Signer | undefined): void {
     this._rx.signer = signer;
   }
 


packages/api/src/base/index.ts
```diff
@@ -58,7 +58,7 @@ export abstract class ApiBase<ApiType extends ApiTypes> extends Getters<ApiType>
   /**
    * @description Set an external signer which will be used to sign extrinsic when account passed in is not KeyringPair
    */
-  public setSigner (signer: Signer): void {
+  public setSigner (signer: Signer | undefined): void {
```
jacogr, line 61:
```suggestion
  public setSigner (signer?: Signer): void {
```

Just a small suggestion for, making `setSigner()` slightly more ergonomic

jacogr:
Generally the best approach for a signer is to set it on a per-transaction basis, i.e. as part of the options passed to `signAndSend`, instead of using the global on the class.

packages/api/src/base/index.ts
polymath-eric: To be clear, I am not against the `?` per se, but I did think using the union to make a user pass in an explicit arg was a cleaner API given the name and intention for this call.

jacogr: Have no strong feelings on this one either way, so ok with it as-is.

jacogr (approved):

PR #5650, opened 2023-05-19
Add suport for latest ink! metadata

statictype:
Adds latest ink! metadata updates to the `Abi` interface and type definitions.
Related to https://github.com/paritytech/contracts-ui/issues/469

Changes:

- [x] contract environment types 
- [x] "default" property on messages and constructors
- [x] return type for constructors

Contract with modified environment
[custom_env.json.zip](https://github.com/polkadot-js/api/files/11525001/custom_env.json.zip)


packages/api-contract/src/Abi/index.ts
```diff
@@ -82,27 +108,38 @@ export class Abi {
...
       isString(abiJson)
         ? JSON.parse(abiJson) as Record<string, unknown>
         : abiJson,
       chainProperties
     );
-    this.constructors = this.metadata.spec.constructors.map((spec: ContractConstructorSpecLatest, index) =>
-      this.#createMessage(spec, index, {
+    this.constructors = this.metadata.spec.constructors.map((spec: ContractConstructorSpecLatest, index) => {
+      const isDefault = 'default' in spec ? spec.default.isTrue : undefined;
+      const typeSpec = 'returnType' in spec ? spec.returnType.unwrapOr(null) : null;
```
jacogr, line 122: Since we are always dealing with the latest spec here, it will be available. (We always only work with latest, hence the conversions - this means we don't need specific logic for each intermediate version as added here)

(For V3 -> V4 conversions, it would be false, the default if not supplied)

packages/api-contract/src/Abi/index.ts
```diff
@@ -53,7 +53,23 @@ function getLatestMeta (registry: Registry, json: Record<string, unknown>): Cont
   return converter[1](registry, metadata[`as${converter[0]}`]);
 }
 
-function parseJson (json: Record<string, unknown>, chainProperties?: ChainProperties): [Record<string, unknown>, Registry, ContractMetadataLatest, ContractProjectInfo] {
+function getEnvTypes (env: ContractEnvironment | undefined, lookup: PortableRegistry) {
```
jacogr, line 56: This should be defined, with default values? (Since it is part of the latest definition).

Since it is there, it would mean that the type check below should not be needed.

jacogr (changes requested):
We seem to have version specific logic which is unsatainable the more there are. While we always convert from Vx -> V<latest> as part of the logic - this allows us to always deal with only the latest, without specific existence checks.

packages/api-contract/src/Abi/index.ts
statictype: I'm doing something wrong i think. I added those checks because of runtime errors when working with older metadata.  So no, they weren't always available.
Should I add the fields with `undefined` here? https://github.com/polkadot-js/api/blob/8280bed13b7ee2f9dc6fa11ca7b9772cfba6c695/packages/api-contract/src/Abi/toV4.ts#L10

And then try to read the values from spec?

jacogr: They don't need to be added as undefined - when you create and it is not supplied, it will fill in with the default values for the specified types. (aka it will be all-0 values).

An example with `default` - since it is not supplied, it will just be `false`.

You only need to supply them explicitly if you wish to supply specific values. (In this case just leaving them out and accepting the defaults should be fine)

statictype:
@jacogr I removed all unnecessary checks and decoded environment types. 
decoding is weird because the environment can contain both types and values so some checks are needed.
For example `maxEventTopics` is a value and the rest are types.
https://github.com/paritytech/ink/blob/8bb4b40d76adb28cf046f385cb5842bcd90b8787/crates/env/src/types.rs#L125

Btw, is [this](https://github.com/polkadot-js/api/blob/11560d184188f28e2e1c88de3180a64e14065e4c/packages/types/src/interfaces/contractsAbi/definitions.ts#L292) ok for usize?

I added ink! v4.2.0 contracts for testing and the tests still fail, not sure why.

jacogr:
`usize` should never be used. It acts differently on WASM vs native.

packages/api-contract/src/Abi/index.ts
```diff
@@ -103,13 +111,24 @@ export class Abi {
...
+
+    for (const [key, value] of this.metadata.spec.environment.entries()) {
+      const typeSpec = value.toPrimitive();
+
+      if (typeof typeSpec === 'object' && typeSpec !== null && 'type' in typeSpec) {
+        this.environment.set(key, this.registry.lookup.getTypeDef(typeSpec.type as number));
+      } else {
+        this.environment.set(key, typeSpec as number);
+      }
+    }
```
jacogr, line 131: The issue here is that because of the conversions, some types may be completely invalid since they point to defaults.

The suggested approach (and here is hoping it never changes from JSON), is to mark the types as `Option<...>` for instance, `accountId: 'Option<ContractTypeSpec>',`. That way when converting from old -> new, it will be `isNone`.

statictype: created `Option<ContractEnvironment>` for the entire thing, now returns an empty map for older versions.

packages/api-contract/src/Abi/toV4.ts
```diff
@@ -4,6 +4,8 @@
 import type { ContractMetadataV3, ContractMetadataV4 } from '@polkadot/types/interfaces';
 import type { Registry } from '@polkadot/types/types';
 
-export function v3ToV4 (_registry: Registry, v3: ContractMetadataV3): ContractMetadataV4 {
-  return v3;
+import { objectSpread } from '@polkadot/util';
+
+export function v3ToV4 (registry: Registry, v3: ContractMetadataV3): ContractMetadataV4 {
+  return registry.createType('ContractMetadataV4', objectSpread({}, v3));
```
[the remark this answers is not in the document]
jacogr: Weird. All ok then, will make a note to check the internals here, seems weird.

packages/api-contract/src/Abi/index.ts
jacogr: Thank you.

jacogr (approved):
Thank you.

PR #5665, opened 2023-06-01
Cleanup internal `decorateMethod` typings

jacogr:


jacogr:
This is (sadly) way behind at this point, rather use as a base and re-apply as needed.

PR #5686, opened 2023-07-06
Add isEthereum to ChainProperties

fgamundi:
Giving the growing number of Ethereum based substrate chains, the Tanssi team is [proposing to add](https://github.com/moondance-labs/tanssi/pull/110) `isEthereum` to the chain properties so apps (such as polkadotjs-apps) can easily distinguish if the chain they're connecting to is Ethereum-based or not without having to maintain and update a manual list.


jacogr:
This gets us half-way and it a great start. 

With the chain properties, we have to ensure it is correctly decorated on the registry as well (so users can just use `registry.chainProperties()` instead of via RPC, you no doubt saw that apps doesn't replicate all the logic again). 

This part is done in the registry (where it also combines existing on-chain constants - always taking precedence, hence not applying that in eg. apps), you can find it here -

https://github.com/polkadot-js/api/blob/47df39e02d81dd088bf762ed7a3a62c37a91e5f0/packages/types/src/create/registry.ts#L154-L167

And here is where it is called/used -

https://github.com/polkadot-js/api/blob/47df39e02d81dd088bf762ed7a3a62c37a91e5f0/packages/types/src/create/registry.ts#L611-L614

fgamundi:
Hi @jacogr. Thanks for the review and explanation! I've updated the registry to set `isEthereum`

girazoki:
@jacogr any news on this side? it would be important for us as we will start having chains that need addresses being represented in ethereum format

jacogr (commented):
Why are we using `Option<Bool>` instead of just `Bool`?

I looked through and cannot see specific benefit.

With `Option<Bool>`:

- if not specified it is `isNone` (which is equivalent to `false`)
- if specified, it returns `isTrue` or `isFalse`

With `Bool`:

- If not specified, it is `isFalse`
- if specified, it is either `isTrue` or `isFalse`

The benefit of using `Bool` is that we don't need to go through unwrapping magic on use, i.e. it is a straight flag check.

EDIT: It would also be useful to add a getter (like the other `chain*` retrievals) to the registry for `chainIsEthereum(): boolean` (This makes downstream usage aq walk in the park)

jacogr (approved):
Thank you.

PR #5694, opened 2023-07-18
Add AssetConversionApi

jsidorenko:
image

Closes https://github.com/paritytech/cumulus/issues/2871

### DIFF
--- a/packages/types/src/interfaces/assetConversion/runtime.ts
+++ b/packages/types/src/interfaces/assetConversion/runtime.ts
@@ -1,0 +1,72 @@
+// Copyright 2017-2023 @polkadot/types authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import type { DefinitionsCall } from '../../types/index.js';
+
+export const runtime: DefinitionsCall = {
+  AssetConversionApi: [
+    {
+      methods: {
+        get_reserves: {
+          description: 'Get pool reserves',
+          params: [
+            {
+              name: 'asset1',
+              type: 'MultiAssetId',
+            },
+            {
+              name: 'asset2',
+              type: 'MultiAssetId',
+            },
+          ],
+          type: 'Option<(Balance,Balance)>',
+        },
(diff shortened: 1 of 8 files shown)


jsidorenko:
It looks like we would need to use `Box<MultiLocation>` instead of `MultiAssetId` to query the AssetHub API. Or even the `VersionedMultiLocation` once this gets changed on the node. in this PR: https://github.com/paritytech/cumulus/pull/2860 . 

How could we support both `MultiAssetId` and `VersionedMultiLocation` versions?

packages/types/src/interfaces/assetConversion/runtime.ts
```diff
@@ -0,0 +1,72 @@
...
+export const runtime: DefinitionsCall = {
+  AssetConversionApi: [
+    {
+      methods: {
+        get_reserves: {
+          description: 'Get pool reserves',
+          params: [
+            {
+              name: 'asset1',
+              type: 'MultiAssetId'
```
bkontur, line 15: @jsidorenko  
I think, primary, we use asset-conversion on AssetHubs, so lets make them work at least,
so I would change it here to `Box<XcmV3MultiLocation>`  (please, try it locally, should work),

and then lets solve: https://github.com/paritytech/cumulus/issues/2915
and then lets find solution how to dynamically support: MultiAssetId (substrate node) vs MultiLocation (asset-hub),
because this could require some dynamic loading or some switch for node kind (substrate -> MultiAssetId, asset-hubs -> MultiLocation)

jacogr (approved):
Thank you.

jsidorenko:
@jacogr shall I change from MultiAssetId to Box<XcmV3MultiLocation> as on Westmint?

packages/types/src/interfaces/assetConversion/runtime.ts
jsidorenko: Just tried, the `Box<XcmV3MultiLocation>` works well

PR #5699, opened 2023-08-02
Update check for contract instantiation

statictype:
Replaced  the `isWasm` check on contract instantiation with `isU8a` in order to allow experimentation with other targets. 
For more info see https://github.com/paritytech/contracts-ui/issues/507

### DIFF
--- a/packages/api-contract/src/base/Code.ts
+++ b/packages/api-contract/src/base/Code.ts
@@ -34,13 +34,14 @@ export class CodeSubmittableResult<ApiType extends ApiTypes> extends Submittable
   }
 }
 
-function isRiscV(bytes: unknown) {
+function isRiscV (bytes: unknown) {
   const ELF_MAGIC = new Uint8Array([0x7f, 0x45, 0x4c, 0x46]); // ELF magic bytes: 0x7f, 'E', 'L', 'F'
+
   return isU8a(bytes) && u8aEq(bytes.subarray(0, 4), ELF_MAGIC);
 }
 
-function isValidCode(code: Uint8Array) {
-  return isWasm(code) || isRiscV(code)
+function isValidCode (code: Uint8Array) {
+  return isWasm(code) || isRiscV(code);
 }
 
 export class Code<ApiType extends ApiTypes> extends Base<ApiType> {
--- a/packages/api-contract/src/base/Code.ts
+++ b/packages/api-contract/src/base/Code.ts
@@ -40,7 +40,7 @@ function isRiscV (bytes: unknown): bytes is Uint8Array {
   return isU8a(bytes) && u8aEq(bytes.subarray(0, 4), ELF_MAGIC);
 }
 
-function isValidCode (code: Uint8Array) {
+function isValidCode (code: Uint8Array): boolean {
   return isWasm(code) || isRiscV(code);
 }
 
--- a/packages/api-contract/src/base/Code.ts
+++ b/packages/api-contract/src/base/Code.ts
@@ -34,7 +34,7 @@ export class CodeSubmittableResult<ApiType extends ApiTypes> extends Submittable
   }
 }
 
-function isRiscV (bytes: unknown) {
+function isRiscV (bytes: unknown): bytes is Uint8Array {
   const ELF_MAGIC = new Uint8Array([0x7f, 0x45, 0x4c, 0x46]); // ELF magic bytes: 0x7f, 'E', 'L', 'F'
 
   return isU8a(bytes) && u8aEq(bytes.subarray(0, 4), ELF_MAGIC);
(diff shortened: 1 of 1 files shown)


athei:
Can we get this merged? I can't use contracts UI with RISC-V contracts otherwise.

statictype:
this PR seems pretty innocent, can we get it merged @jacogr? 🙏🏻

jacogr (commented):
I'm not convinced this is "innocent enough".

Basically at this point we can throw any data into the mix as expect it to work, i.e. some JSON, a random piece of text, etc. Basically we end up with garbage that cannot be executed in any interpreter.

If there are additional formats that should be supported, the better approach is to check the magic flag for that specific format in addition to the `isWasm` check.

EDIT: The `is<NewFormat>()` check function can just be added locally to the specific file - it would make sense to also add it to [common](https://github.com/polkadot-js/common/tree/master/packages/util/src/is) (and when that is available in a release, the local version inside the file can be removed)

athei:
> Basically at this point we can throw any data into the mix as expect it to work, i.e. some JSON, a random piece of text, etc. Basically we end up with garbage that cannot be executed in any interpreter.

So what? `pallet_contracts` has to make absolutely sure that it can handle any input securely anyways. It will fail gracefully and report an error back in case of the data not being processable.

The verification also does only check the first bytes for the "wasm magic". There are still infinite many ways to send "garbage that cannot be executed in any interpreter" after those bytes.

I cannot see the value that this very shallow check provides. Users have to do error handling anyways since `pallet_contracts` can reject a module for all sorts of reasons. On the other hand having this check makes our lives more difficult for the stated reasons.

jacogr:
It is what it is :)

jacogr (dismissed):
As explained in https://github.com/polkadot-js/api/pull/5699#pullrequestreview-1604329805:

1. Add a `is<FormatName>()` check function (e.g. assuming the new format is `XXX` we add `isXXX` - used in the sample for the next points) to check the magic bytes for the specific additional type
2. Adjust the `isWasm` checks to do `isWasm(...) || isXXX(...)`
3. (Optional, but appreciated) Also add the `isXXX` to the common repo (will eventually replace the local version when that stuff is released)

That should more-or-less cover the same functionality. (Not 100% sure on the `source.wasm` usage, implementers would know better on that)

statictype:
Added a similar check for RISC-V @jacogr 
If it's ok i can add it to the common repo and import it from there in a follow-up PR

packages/api-contract/src/base/Code.ts
```diff
@@ -34,6 +34,16 @@ export class CodeSubmittableResult<ApiType extends ApiTypes> extends Submittable
   }
 }
 
+function isRiscV (bytes: unknown) {
```
jacogr, line 37:
```suggestion
function isRiscV (bytes: unknown): bytes is Uint8Array {
```

This just gets the typings right as well. Every now and again TS has trouble with "auto" (although here it _should_ certainly detect the type as `Uint8Array`), but either way it is preferable being more explicit)

packages/api-contract/src/base/Code.ts
```diff
@@ -42,12 +52,12 @@ export class Code<ApiType extends ApiTypes> extends Base<ApiType> {
   constructor (api: ApiBase<ApiType>, abi: string | Record<string, unknown> | Abi, wasm: Uint8Array | string | Buffer | null | undefined, decorateMethod: DecorateMethod<ApiType>) {
     super(api, abi, decorateMethod);
 
-    this.code = isWasm(this.abi.info.source.wasm)
+    this.code = isValidCode(this.abi.info.source.wasm)
```
jacogr, line 55: _Slightly_ weird that a different format may come through in `source.wasm`, but either way, all ok.

packages/api-contract/src/base/Code.ts
```diff
@@ -34,6 +34,16 @@ export class CodeSubmittableResult<ApiType extends ApiTypes> extends Submittable
   }
 }
 
+function isRiscV (bytes: unknown) {
+  const ELF_MAGIC = new Uint8Array([0x7f, 0x45, 0x4c, 0x46]); // ELF magic bytes: 0x7f, 'E', 'L', 'F'
+
+  return isU8a(bytes) && u8aEq(bytes.subarray(0, 4), ELF_MAGIC);
+}
+
+function isValidCode (code: Uint8Array) {
```
jacogr, line 43:
```suggestion
function isValidCode (code: Uint8Array): boolean {
```

Slightly more explicit, it always helps down the line.

jacogr (commented):
Thank you, looks like it fits the bill.

Two small comment added just to be slightly more pedantic/explicit on return types/guards. (It always helps when we come back years from now)

packages/api-contract/src/base/Code.ts
statictype: the field in the contract metadata will most likely be renamed once the support for RISC-V is released. not something we should address now imo.

jacogr (approved):

Issue #5702, opened 2023-08-17
Incorrect decoding of signed integers `i8`/`i16` 


peetzweg:
The decoding of some scale encoded signed integers values result in the wrong representation in js. 
Generally scale encoded `i8` and `i16` value seem to work fine, however the edge cases, the value of `1` and `i8::MAX` and `i16::MAX` are decoded to the wrong representation in javascript.

This issue bubbles up in one form or another, see related issue section. How can we resolve this once and for all?

[I've created a PR adding testcases for the broken decoding cases #5703.](https://github.com/polkadot-js/api/pull/5703)

* **I'm submitting a ...**

  - [X] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

```ts
// Scale encoded signed integers of value `1`
const bytes_1i8 = new Uint8Array([1]);
const bytes_1i16 = new Uint8Array([1, 0]);
const bytes_1i32 = new Uint8Array([1, 0, 0, 0]);
const bytes_1i128 = new Uint8Array([
    1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
  ]);


api.createType("i8", bytes_1i8).toNumber(); // expected 1, actual -255 ❌
api.createType("i16", bytes_1i16).toNumber(); // expected 1, actual -65535 ❌
api.createType("i32", bytes_1i32).toNumber(); // expected 1, actual 1 ✅
api.createType("i128", bytes_1i128).toNumber(); // expected 1, actual 1 ✅


// Decoding MAX values for signed integers
const bytes_maxi8 = new Uint8Array([127]);
const bytes_maxi16 = new Uint8Array([255, 127]);
api.createType("i8", bytes_maxi8).toNumber(); // expected 127, actual -129 ❌
api.createType("i16", bytes_maxi16).toNumber(); // expected 32767, actual -32769 ❌


// SCALE encoded `0` value of `i8` and `i16`
const bytes_0i8 = new Uint8Array([0]);
const bytes_0i16 = new Uint8Array([0, 0]);
api.createType("i8", bytes_0i8).toNumber(); // expected 0, actual `Input too large. Found input with 9 bits, expected 8`  ❌
api.createType("i16", bytes_0i16).toNumber(); // expected 0, actual `Input too large. Found input with 17 bits, expected 16` ❌
```

[...]

PR #5703, opened 2023-08-17
Adds tests cases for issue #5702

peetzweg:
See #5702 for further details.


jacogr (changes requested):
Always appreciate this. Obviously it cannot be merged but useful for somebody that wishes to look into the logged issue - if CI fails the PR is not merge-able, but somebody can certainly copy this/parts of it into their effort.

One comment:

If anything fails you need to delve through an it statement to find the place - so each `it` should have 1 expectation (maybe more, if they are very closely related and follow-ups of each other and easy to find)

In some cases it is possible, in others it would be impossible to find which one creates issues, for instance

```js
// Zero
    expect(
      new Int(registry, new Uint8Array([0]), 8).toNumber()
    ).toEqual(0);

    expect(
      new Int(registry, new Uint8Array([0, 0]), 16).toNumber()
    ).toEqual(0);
```

TL;DR Rather use a `describe` block with specific `it` + (one) `expect` for each.

peetzweg:
Thanks @jacogr for pointing out where these errors might root from. Tried looking into it a bit already, will continue from there.

> In some cases it is possible, in others it would be impossible to find which one creates issues, for instance

Yes, agree it's not as transparent which one failed, however the stacktrace reveals which one it actually failed i.e.
```
  at TestContext.<anonymous> (/Users/xxx/git/pjs-api/packages/types-codec/src/base/Int.spec.ts:113:7)
``` 
  
Copy pasting almost identical `it`s is quite cumbersome. Unfortunately we can't create [parameterized tests](https://vitest.dev/api/#test-each) with @polkadot/dev-test right?

There also seems to be an issue with decoding just the value `0` to a Number, which results in 
```
i8: Input too large. Found input with 9 bits, expected 8
```
https://github.com/peetzweg/api/blob/f33bbc8340d66f46b77da041bfc1467b3d33c450/packages/types-codec/src/base/Int.spec.ts#L111-L114

Which is a quite odd. 🤔

jacogr:
Could you merge in the master branch (it contains the latest common stuff). Hopefully it means the test cases here would start passing…

peetzweg:
@jacogr done! They succeed locally now. Great job! 🙌

jacogr (dismissed):

packages/types-codec/src/base/Int.spec.ts
```diff
@@ -41,6 +41,108 @@ describe('Int', (): void => {
...
+  it('decodes edge case to js number', (): void => {
+    // Zero
+    expect(
+      new Int(registry, new Uint8Array([0]), 8).toNumber()
+    ).toEqual(0);
+
+    expect(
+      new Int(registry, new Uint8Array([0, 0]), 16).toNumber()
+    ).toEqual(0);
+
```
jacogr, line 115:
```suggestion
```

packages/types-codec/src/base/Int.spec.ts
```diff
@@ -41,6 +41,108 @@ describe('Int', (): void => {
...
+    // MAX
+    expect(
+      new Int(registry, new Uint8Array([127]), 8).toNumber()
+    ).toEqual(127);
+
+    expect(
+      new Int(registry, new Uint8Array([255, 127]), 16).toNumber()
+    ).toEqual(32767);
+  });
+
```
jacogr, line 144:
```suggestion
```

jacogr (approved):
Thank you.

PR #5731, opened 2023-09-27
Fix sorting of `Option` values and error on duplicates from BTreeSet/BTreeMap.

Neopallium:
Sorting was using `toU8a(true)` for `Option` values.  The bare encoding wasn't working for `Enum` values in an `Option`.

Also added detection of duplicate values (BTreeSet) and keys (BTreeMap).


jacogr (changes requested):
I believe this looks great.

Would like to see some tests added for dupe checking as well as options - this just ensures that into the future the behavior stays as intended.

Neopallium:
Added tests for duplicates and Option<Enum> cases for both BTreeMap & BTreeSet.  Even checked that the tests would fail with the old logic.

jacogr (approved):
Brilliant, thank you.

PR #5745, opened 2023-10-29
fix cache key

ermalkaleci:
Cache is never read because we're using `body` as key which includes `id`.


ermalkaleci:
@jacogr can you check this?

TarikGul:
It would be nice if the CI could run against this again since none of the tests are coming through...

Looking into this

ermalkaleci:
> The PR looks fine, but I would like the workflows to run before we merge... Very odd indeed.

maybe this PR is very old. I will do a master pull

TarikGul:
> > The PR looks fine, but I would like the workflows to run before we merge... Very odd indeed.
> 
> maybe this PR is very old. I will do a master pull

Sounds perfect! Was just going to ask the same thing :) Thanks

Issue #5750, opened 2023-11-09
Statemine blocks 5753828, 5753849 can not be decoded


wliyongfeng:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

* **What is the motivation for changing the behavior?**

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fstatemine-rpc.dwellir.com#/explorer/query/5753828
https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fstatemine-rpc.dwellir.com#/explorer/query/5753849
![PcRhQicaTq](https://github.com/polkadot-js/api/assets/2264908/60c43301-4063-450e-bd5b-94dc0570f5d1)

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [ ] Node.js
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

wliyongfeng:
@jacogr Could you take a look at this issue and #5734? They are crucial to explorers who depend on polkadot api.

valentunn:
This seems to be related to `ChargeAssetTxPayment` signed extension being used in transactions. Our guess is that API doesn't infer types of signed extensions from runtime metadata correctly. For `ChargeAssetTxPayment`, its type is a struct with `tip` and `assetId`. Type for `assetId` got changed from `u32` to `MultiLocation` with most recent Kusama/Westmint upgrades
This introduces issues for our indexer as well

valentunn:
@wliyongfeng we found a workaround which uses typesBundle overrides. This snippet is for rococo but it works in the same way in Kusama Asset Hub, adjusting runtime versions and specName

``` ts
import { OverrideBundleDefinition } from "@polkadot/types/types";

const definitions: OverrideBundleDefinition = {
  types: [
    {
      minmax: [0, 10004],
      types: {
        NovaAssetId: "Option<AssetId>"
      },
    },
    {
      minmax: [10005, null],
      types: {
        NovaAssetId: "Option<MultiLocation>"
      },
    },
  ],
  signedExtensions: {
    ChargeAssetTxPayment: {
      extrinsic: {
        tip: 'Compact<Balance>',
        // eslint-disable-next-line sort-keys
        assetId: 'NovaAssetId'
[...]

Issue #5734, opened 2023-10-05
Westmint block #5482732 can not be decoded


wliyongfeng:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fsys.ibp.network%2Fwestmint#/explorer/query/5482732
![PXyyRhbBw5](https://github.com/polkadot-js/api/assets/2264908/e2d0a76a-41f5-4ff5-ae96-692d1051e81f)

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [x] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [x] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

Imod7:
I can confirm that I see the same error message when requesting 
- the same block (`5482732`) 
- and while also connected to `Westend Asset Hub` 

either by : 

- using [polkadot js apps](https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fwestmint-rpc.dwellir.com#/rpc) (RPC call `chain` > [getBlock](https://polkadot.js.org/docs/substrate/rpc/#getblockhash-blockhash-signedblock))
- or Sidecar (endpoint `blocks/5482732`)

bee344:
This should be closed by #5752

jacogr:
Closing on the back of the comments. Feel free to re-open if there is another issue not solved by the PRs.

PR #5752, opened 2023-11-14
fix: ChargeAssetTxPayment to work with Asset Conversion on Westend's and Kusama's Asset Hubs

bee344:
Since the addition of the new `pallet-asset-conversion` to `Kusama Asset Hub` and `Westend Asset Hub`, the `ChargeAssetTxPayment` mutated to accept a `MultiLocation` as the ID of the asset used to pay the tx fees, which in turn broke the decoding of the blocks in which it was present. What this PR does is fix that by modifying the signed extension for the cases of KAH and WAH, by integrating an alias that modifies the SE to use `AssetId` or `MultiLocation` depending on the runtime version. At the same time, as the introduction of the pallet was at different times on KAH and WAH and was not added in Polkadot Asset Hub, it adds `type` override files for each of them, and leaves the `statemint` one without change.
It also modifies `SignerOptions` to accept an `object` for the SE `assetId`, so that a `MultiLocation` can be passed and the SE be used in a straightforward manner.

Block retrieved with substrate-sidecar pre-fix:

Block retrieved with substrate-sidecar after fix:

Closes [Issue 5750](https://github.com/polkadot-js/api/issues/5750#) and [5734](https://github.com/polkadot-js/api/issues/5734) (probably).

### DIFF
--- a/packages/api-base/src/types/submittable.ts
+++ b/packages/api-base/src/types/submittable.ts
@@ -16,7 +16,7 @@ export interface SignerOptions {
   nonce: AnyNumber | Codec;
   signer?: Signer;
   tip?: AnyNumber;
-  assetId?: AnyNumber;
+  assetId?: AnyNumber | object;
 }
 
 export type SubmittableDryRunResult<ApiType extends ApiTypes> =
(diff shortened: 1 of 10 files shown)


TarikGul:
I can confirm this fixes the decoding issue we've been seeing with the assetId being as a multilocation.

jacogr (approved):
Seems sane, thank you.

jacogr (changes requested):
Some small nigglies identified by CI, see https://github.com/polkadot-js/api/actions/runs/6870177012/job/18703510196?pr=5752#step:5:1294

It basically just needs a default type definition for the added type - generally we probably want to point this to the latest definition, which also means that the higher-version specs don't need the override.

So basically -

1. add a default to point to `Option<MultiLocation>`
2. keep the override for lower versions (as required)
3. remove the override for current versions (meaning we don't need to keep adding them into the future)

bee344:
@jacogr I added the default types as you suggested and it solved the issues, but I also had to modify the Signer test to account for the assetId being a MultiLocation

jacogr:
@bee344 All good, I appreciate your persistence :)

jacogr (approved):

PR #5763, opened 2023-11-21
fix: updated remaining assetId for ChargeAssetTxPayment

bee344:
This is complementary to #5752 , and it updates two remaining `assetId`s so that we can build an extrinsic correctly also passing an object as the `assetId`. We are updating `txwrapper-core` and found out we need this. Sorry for not covering that earlier.


jacogr (approved):
Thank you.

Issue #5765, opened 2023-11-28
polkadot-types-from-chain removes imports which are used later in the same file


atodorov:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [x] Other

* **What is the current behavior and expected behavior?**

Rebuilding type definitions removes imports which are actually being used, which in turn results to build issues with my package later on.

I am expecting that these imports will not be removed. 

* **Please tell us about your environment:**

  - Version: `@polkadot/typegen": "^10.10.1",`, `"@polkadot/api": "^10.10.1",`
  - Environment: Linux

    - [X] Node.js - v18.18.2
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript (include tsc --version: 5.3.2)
    - [ ] Other

** Steps to reproduce **

```
$ git clone https://github.com/gluwa/creditcoin3
$ cd creditcoin3/
$ cargo build --release
   Compiling creditcoin3-node v3.6.3 (/home/gluwa/creditcoin-3.0/creditcoin3/node)
    Finished release [optimized] target(s) in 53.77s

$ ./target/release/creditcoin3-node --dev --validator --pruning archive  --node-key d182d503b7dd97e7c055f33438c7717145840fd66b2a055284ee8d768241a463             --base-path ./alice-data


$ cd cli/
$ yarn install
$ ./get-metadata.sh

$ git status
On branch testing/csub-862-only-update-typedefs
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   creditcoin.json


$ yarn build:types
yarn run v1.22.19
$ yarn generate:defs && yarn generate:meta
$ ts-node --skip-project node_modules/.bin/polkadot-types-from-defs --package creditcoin3/lib/interfaces --input ./src/lib/interfaces --endpoint ./creditcoin.json
ERROR: Unable to load user definitions: /home/gluwa/creditcoin-3.0/creditcoin3/cli/src/lib/interfaces/definitions.ts is not a file
/home/gluwa/creditcoin-3.0/creditcoin3/cli/src/lib/interfaces/types.ts
	Generating
	Writing

/home/gluwa/creditcoin-3.0/creditcoin3/cli/src/lib/interfaces/augment-types.ts
	Generating
	Writing

[...]

PR #5769, opened 2023-12-10
fix: add lookupTypes to typegen imports

Muhammad-Altabba:
## Description

Fixes: #5765

The `lookupTypes` are missed from the list used to extract imports for the generated types. 
This PR will add it back.

### DIFF
--- a/packages/typegen/src/util/formatting.ts
+++ b/packages/typegen/src/util/formatting.ts
@@ -37,7 +37,8 @@ function extractImports ({ imports, types }: This): string[] {
     ...Object.keys(imports.extrinsicTypes),
     ...Object.keys(imports.genericTypes),
     ...Object.keys(imports.metadataTypes),
-    ...Object.keys(imports.primitiveTypes)
+    ...Object.keys(imports.primitiveTypes),
+    ...Object.keys(imports.lookupTypes)
   ];
 
   return [


Muhammad-Altabba:
The MR also resolves: https://github.com/polkadot-js/docs/issues/252

jacogr (changes requested):
The build fails because of this. On the top-level imports it now tries to include the lookup types from `@polkadot/types` since it was added to the top-level definition.

https://github.com/polkadot-js/api/actions/runs/7158255406/job/19733755190?pr=5769

It has these dupes since in the next section we explicitly import from lookup on line 52.

Muhammad-Altabba:
> The build fails because of this. On the top-level imports it now tries to include the lookup types from `@polkadot/types` since it was added to the top-level definition.
> 
> https://github.com/polkadot-js/api/actions/runs/7158255406/job/19733755190?pr=5769
> 
> It has these dupes since in the next section we explicitly import from lookup on line 52.

So what is your suggestion to fix the issue?

And did you try to generate the files locally and saw that there are missed imports for lookups? Or you where able to generate with no issues?

Thanks,

jacogr:
CI generates code for - Substrate, Kusama and Polkadot on each build. So it gets used each time there is a PR and each time there is a build - it literally runs as part of the API build process.

valentinfernandez1:
Closing outdated PR

PR #5776, opened 2024-01-02
feat: upgrade to substrate-connect v0.8.4

kratico:
Upgrade `@substrate/connect` to `v0.8.4` and use the new [`addChain`](https://github.com/paritytech/substrate-connect/pull/1604) API.

The `addChain` API uses the relay chain to instantiate new parachains, for example

```ts
const client = createScClient()
const relayChain = await client.addWellKnownChain(input, onMessageForRelay)
const parachain = await relayChain.addChain(input, onMessageForParachain)
```

Given that the `ScProvider` is injected a relay chain provider, no changes are needed in the `ScProvider` public API and it works as before, for example

```ts
import { ApiPromise } from "@polkadot/api";
import { ScProvider } from "@polkadot/rpc-provider";
import * as Sc from "@substrate/connect";

import jsonParachainSpec from './myParaChainSpec.json';

const parachainSpec = JSON.stringify(jsonParachainSpec);
const relayProvider = new ScProvider(Sc, Sc.WellKnownChain.westend2);
const provider = new ScProvider(Sc, parachainSpec, relayProvider);

await provider.connect();

// ...

```

Note: given that `ScProvider` receives `createScClient` from `@substrate/connect` as input to the constructor and that the `@substrate/connect` API has a breaking change, this is also a breaking change for `@polkadot/rpc-provider`.


packages/rpc-provider/src/substrate-connect/index.ts
```diff
@@ -154,9 +154,11 @@ export class ScProvider implements ProviderInterface {
...
     };
 
-    const addChain = this.#wellKnownChains.has(this.#spec as ScType.WellKnownChain)
-      ? client.addWellKnownChain
-      : client.addChain;
+    const addChain = this.#sharedSandbox
+      ? (async (...args) => (await this.#sharedSandbox!.#chain)!.addChain(...args)) as ScType.AddChain
+      : this.#wellKnownChains.has(this.#spec as ScType.WellKnownChain)
+        ? client.addWellKnownChain
+        : client.addChain;
```
kratico, line 161: The `this.#sharedSandbox!.#chain` is not transpiled properly see
image

I think this is related to this `polkadot-js/dev` script
https://github.com/polkadot-js/dev/blob/1d7db82109f666dd013d5db04ef2bd6a45f4d74a/packages/dev/scripts/polkadot-dev-build-ts.mjs#L74-L84

kratico: as an alternative that is coupled to a `polkadot-dev-build-ts.mjs` implementation detail, this can be re-written to
```ts
    const addChain = this.#sharedSandbox
      ? (async (...args) => {
          const source = this.#sharedSandbox!;
          return (await source.#chain)!.addChain(...args);
        }) as ScType.AddChain
      : this.#wellKnownChains.has(this.#spec as ScType.WellKnownChain)
        ? client.addWellKnownChain
        : client.addChain;
```
and it will transpile to
```js
        const addChain = this.__internal__sharedSandbox
            ? (async (...args) => {
                const source = this.__internal__sharedSandbox;
                return (await source.__internal__chain).addChain(...args);
            })
            : this.__internal__wellKnownChains.has(this.__internal__spec)
                ? client.addWellKnownChain
                : client.addChain;
```

jacogr: So the dev script implementation can actually be dropped (after we dropped Node 16 last year). The issue is that TSC compiled everything `#<id>` to WeakMaps for Node 16 compat and it had severe implications for performance. So at best it is/was a horrible workaround.

So the adjustment will be dropped _sometime_, but no idea when that _sometime_ will be - needs some checks all over when done.

kratico: thx for the info @jacogr 

I'll re-write it as suggested previously so it's more friendly with this script

kratico: re-written in https://github.com/polkadot-js/api/pull/5776/commits/7df6a77ecee56e57a9fc019a167c201999c14b3a

jacogr (changes requested):
Some small CI linting issues - should not be a massive issue to resolve.

packages/rpc-provider/src/substrate-connect/index.ts
```diff
@@ -154,9 +154,17 @@ export class ScProvider implements ProviderInterface {
       callback?.(decodedResponse);
     };
 
-    const addChain = this.#wellKnownChains.has(this.#spec as ScType.WellKnownChain)
-      ? client.addWellKnownChain
-      : client.addChain;
+    const addChain = this.#sharedSandbox
+      ? (async (...args) => {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const source = this.#sharedSandbox!;
```
kratico, line 160: `eslint-disable` because it's checked on line [157](https://github.com/polkadot-js/api/pull/5776/files#diff-66ebd813ed4a22dc4419b70f5b5771ac5f597162674c041aad2ecec74717fd0dR157)

packages/rpc-provider/src/substrate-connect/index.ts
```diff
@@ -154,9 +154,17 @@ export class ScProvider implements ProviderInterface {
...
+
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        return (await source.#chain)!.addChain(...args);
```
kratico, line 163: `eslint-disable` because `this.#sharedSandbox.connect` is invoked on line [113](https://github.com/polkadot-js/api/pull/5776/files#diff-66ebd813ed4a22dc4419b70f5b5771ac5f597162674c041aad2ecec74717fd0dR113) and `.connect` creates the a chain instance on `.#chain` on line [169](https://github.com/polkadot-js/api/pull/5776/files#diff-66ebd813ed4a22dc4419b70f5b5771ac5f597162674c041aad2ecec74717fd0dR169)

Note: all these changes are inside the `.connect` method.

jacogr (approved):

jacogr (dismissed):
Seems spot-on. Please just merge master to resolve the `yarn.lock` conflicts.

josepot (approved):
Thanks @kratico ! I will try to find out how to create a release.

PR #5778, opened 2024-01-11
feat(rpc-provider): allow cacheCapacity option for WsProvider

TarikGul:
## Summary:

This PR aims to allow configuring the cacheCapacity for the LRUCache used in the WsProvider. It will still default to 128 if no option is passed in. 

## Concerns

This PR will be a breaking change for users that use the `timeout` option. I did think about creating an options object instead and including some of the args in there but I thought I would leave it up to discussion. 

Feel free to close the PR if there is no intention to allow configuration of the capacity size of the LRU cache.

### DIFF
--- a/packages/rpc-provider/src/ws/index.ts
+++ b/packages/rpc-provider/src/ws/index.ts
@@ -108,7 +108,7 @@ export class WsProvider implements ProviderInterface {
    * @param {Record<string, string>} headers The headers provided to the underlying WebSocket
    * @param {number} [timeout] Custom timeout value used per request . Defaults to `DEFAULT_TIMEOUT_MS`
    */
-  constructor (endpoint: string | string[] = defaults.WS_URL, autoConnectMs: number | false = RETRY_DELAY, headers: Record<string, string> = {}, cacheCapacity: number = DEFAULT_CAPACITY, timeout?: number) {
+  constructor (endpoint: string | string[] = defaults.WS_URL, autoConnectMs: number | false = RETRY_DELAY, headers: Record<string, string> = {}, timeout?: number, cacheCapacity?: number) {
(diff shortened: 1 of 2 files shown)


packages/rpc-provider/src/ws/index.ts
```diff
@@ -108,7 +108,7 @@ export class WsProvider implements ProviderInterface {
    * @param {Record<string, string>} headers The headers provided to the underlying WebSocket
    * @param {number} [timeout] Custom timeout value used per request . Defaults to `DEFAULT_TIMEOUT_MS`
    */
-  constructor (endpoint: string | string[] = defaults.WS_URL, autoConnectMs: number | false = RETRY_DELAY, headers: Record<string, string> = {}, timeout?: number) {
+  constructor (endpoint: string | string[] = defaults.WS_URL, autoConnectMs: number | false = RETRY_DELAY, headers: Record<string, string> = {}, cacheCapacity: number = DEFAULT_CAPACITY, timeout?: number) {
```
jacogr, line 111: Rather move this as the last parameter. (No breaking change - we really don't like this at all, unless very much required)

On use can do `this.#callCache = new LRUCache(cacheCapacity || DEFAULT_CAPACITY);` to apply defaults.

jacogr (changes requested):

packages/rpc-provider/src/ws/index.ts
TarikGul: Thanks for the feedback. I just applied the change you recommended and I agree it makes more sense.

TarikGul:
@jacogr Apologies for pinging you, the PR is ready for another review whenever you have a chance. Thanks again for everything.

jacogr (approved):

PR #5791, opened 2024-02-01
feat: ink! v5

### LINKED ISSUE #5789: Support new `ink!` contract events
**I'm submitting a ...**

  - [ ] Bug report
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

`ink!` events are changing in version `5.0`. See https://github.com/paritytech/ink/pull/1827, and https://use.ink/5.x/basics/events/.

The most important change is that there is a new field on each event definition in the metadata: `signature_topic` which must now be used to identify the type of event being received. Previously the index of the event i.e its position in the json array was used to determine which event to decode.

Now when handling an event, the first topic of the substrate event is matched to that signature topic in the metadata. If the event is anonymous, there is no signature topic in that case the raw hex bytes of the event can be returned and leave it up to the user what to do there.

Event decoding here: https://github.com/polkadot-js/api/blob/3bdf49b0428a62f16b3222b9a31bfefa43c1ca55/packages/api-contract/src/Abi/index.ts#L236

peetzweg:
Trying to make your life, @jacogr, as easy as possible to resolve #5789. I left a few comments about parts I'm not sure what is the correct way to do here and would appreciate your input. 

Will look into writing a test, making sure both decoding methods work as intended for v4 & v5 contracts.


packages/api-contract/src/Abi/toV5.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { objectSpread } from '@polkadot/util';
+
+export function v4ToV5 (registry: Registry, v4: ContractMetadataV4): ContractMetadataV5 {
+  return registry.createType('ContractMetadataV5', objectSpread({}, v4, {
+    spec: objectSpread({}, v4.spec, {
+      events: v4.spec.events.map((e) =>
+        registry.createType('ContractEventSpecV3', objectSpread({
+          module_path: undefined,
+          signature_topic: undefined
```
peetzweg, line 15: Or do we consider this a breaking change and don't allow converting from v4 to v5?

packages/types/src/interfaces/contractsAbi/types.ts
```diff
@@ -161,6 +170,15 @@ export interface ContractEventSpecV2 extends Struct {
...
 }
 
+/** @name ContractEventSpecV3 */
+export interface ContractEventSpecV3 extends Struct {
+  readonly label: Text;
+  readonly args: Vec<ContractEventParamSpecV2>;
+  readonly docs: Vec<Text>;
+  readonly module_path: Text;
+  readonly signature_topic: U8aFixed;
+}
```
peetzweg, line 180: Not sure if these are the correct types to be used here. 🤔 
`signature_topic` being a hash and `module_path` a `::` path.

packages/api-contract/src/Abi/index.ts
```diff
@@ -235,15 +244,17 @@ export class Abi {
...
       docs: spec.docs.map((d) => d.toString()),
       fromU8a: (data: Uint8Array): DecodedEvent => ({
         args: this.#decodeArgs(args, data),
         event
       }),
-      identifier: spec.label.toString(),
-      index
+      identifier: [spec.module_path, spec.label.toString()].join('::'),
+      index,
+      signatureTopic: spec.signature_topic.toHex() || undefined
```
peetzweg, line 257: If we allow for upgrades from `v4` this could be `undefined`.

packages/api-contract/src/base/Contract.ts
```diff
@@ -115,9 +115,9 @@ export class Contract<ApiType extends ApiTypes> extends Base<ApiType> {
       // ContractEmitted is the current generation, ContractExecution is the previous generation
       new ContractSubmittableResult(result, applyOnEvent(result, ['ContractEmitted', 'ContractExecution'], (records: EventRecord[]) =>
         records
-          .map(({ event: { data: [, data] } }): DecodedEvent | null => {
+          .map(({ event: { data: [, data] }, topics }): DecodedEvent | null => {
             try {
-              return this.abi.decodeEvent(data as Bytes);
+              return this.abi.decodeEvent(data as Bytes, topics[0]);
```
peetzweg, line 120: Or should we rather pass in the whole `EventRecord`?

packages/api-contract/src/Abi/index.ts
```diff
@@ -162,9 +162,18 @@ export class Abi {
   /**
    * Warning: Unstable API, bound to change
    */
-  public decodeEvent (data: Bytes | Uint8Array): DecodedEvent {
+  public decodeEvent (data: Bytes | Uint8Array, topic: EventRecord['topics'][0]): DecodedEvent {
+    // try to find a topic signature match - ink! v5 upwards
+    let event = this.events.find((e) => e.signatureTopic === topic.toHex());
```
peetzweg, line 167: Not the prettiest but we would need support both way to identify the Event to be decoded, right?

packages/api-contract/src/Abi/toV5.ts
cmichi: I'm not sure about the implications, but here's how we describe the change:

>Previously the order of the events in the `events` array was significant (i.e. the first one had an implied index of `0`), and this index could be used to determine which event to decode. Now that is replaced by the `signature_topic`, and the order of the events in the metadata no longer has any significance.

cmichi:
@jacogr Could you take a look at this PR? That would be very helpful for us, as we would like to release ink! 5 with a compatible `polkadot-js/api`.

packages/api-contract/src/Abi/index.ts
```diff
@@ -162,9 +162,18 @@ export class Abi {
   /**
    * Warning: Unstable API, bound to change
    */
-  public decodeEvent (data: Bytes | Uint8Array): DecodedEvent {
+  public decodeEvent (data: Bytes | Uint8Array, topic: EventRecord['topics'][0]): DecodedEvent {
```
TarikGul, line 165: The type for `EventRecord['topics'][0]` should be a `Hash`?

packages/types/src/interfaces/contractsAbi/types.ts
TarikGul: In this case the API's types for `Hash` end up resolving to a U8aFixed: https://github.com/polkadot-js/api/blob/master/packages/types/src/interfaces/runtime/types.ts#L158-L170

So I would imagine in this case `signature_topic` being set as `U8aFixed` is correct.

packages/api-contract/src/Abi/index.ts
peetzweg: Afaik, yes it's a hash. Both should work here. But yes if the structure of the `topic` field of an `EventRecord` gets more complex this breaks. Probably it won't.

Using `Hash` here and renaming the parameter `signature_topic` would make it more atomic and clearer what's happening. Will adapt the code. 👍

packages/api-contract/src/Abi/toV5.ts
TarikGul: > However, this time there is newly added data we can't create for older version of the metadata.

I am getting myself familiar with the correlation between Ink v5 and the conversion process above, and found this on the following `signatureTopic`:

```rust
/// # The Signature Topic
///
/// By default, the [`ink::Event::SIGNATURE_TOPIC`] is calculated as follows:
///
/// `blake2b("EventStructName(field1_type_name,field2_type_name)")`
/// The hashing of the topic is done at codegen time in the derive macro, and as such only has
/// access to the **names** of the field types as they appear in the code. As such, if the
/// name of a field of a struct changes, the signature topic will change too, even if the
/// concrete type itself has not changed. This can happen with type aliases, generics, or a
/// change in the use of a `path::to::Type` qualification.
```

1. Would it be invalid to construct the signatureTopic for a v4 event when converting to v5?

2. In this case do we not have the valid info/data to construct it if the conversion can be considered valid?

cmichi: Not sure if relevant here, but the topic calculation changed in general from v4 to v5, so it's a breaking change for any clients that used topics to filter for events. 

We have a migration guide for the events changes [here](https://use.ink/5.x/faq/migrating-from-ink-4-to-5/#events-20).

TarikGul:
Definitely think we should get this in as soon as we can, I'll be looking into the metadata conversion to see some viable solutions: Specifically for `toV5`.

peetzweg:
> Definitely think we should get this in as soon as we can, I'll be looking into the metadata conversion to see some viable solutions: Specifically for `toV5`.

Just asked @ascjones a bunch of questions regarding the changes from events 1.0 to events 2.0.

Although it's possible to derive a v5 version of the metadata. [As it's just the hash of the Event label and it's types.](https://github.com/paritytech/ink/blob/master/crates/env/src/event.rs#L203-L209)
As example the `signature_topic` of this `Transfer` event would be the result of 

`blake2b256("Transfer(Option<AccountId>,Option<AccountId>,Balance)")`

```rust
    #[ink(event)]
    pub struct Transfer {
        #[ink(topic)]
        from: Option<AccountId>,
        #[ink(topic)]
        to: Option<AccountId>,
        value: Balance,
    }
```

All of this data would be available in v4 metadata.

However it would not make too much sense to migrate v4 metadata `toV5` as the payload of the `EventRecord` is different.

In v4 the first byte of the event data payload is the [index of the event type in the event list of the metadata](https://github.com/polkadot-js/api/blob/master/packages/api-contract/src/Abi/index.ts#L162-L174). Potentially emitted `topics` are ignored for decoding.

```
deployed v4 contract call emits event like this => (data: 0x01_abc..., topics: [?]) 
// topics empty? At least ignored for matching event with metadata, 
// but first byte of `data` is `1` the index of the event in the metadata v4 events array, 
// payload of event is `0xabc`
```

Important to note here, is everything compiled and deployed with ink4 and prior will still work this way.

In v5 however the `topic` of an `EventRecord` contains the newly added `signature_topic` topic value for matching. And the data of a event is not prefix with the event index anymore.

```
deployed v5 contract call emits event like this => (data: 0xabc, topic: [0xsignature_topic]) 
// => match event by signature_data in v5 metadata, 
// payload of event is `0xabc`
```

So as v4 and v5 emit completely different data in the raw event, I think it might be best to make the actual `Abi` instance aware the metadata version used to instantiate it instead of having one Abi for all versions, deciding which path to take by checking which values are present or not. 
[This is how I done it in this pr for now.](https://github.com/peetzweg/pjs-api/blob/cc946da2e904c7114aaa9914aee0e864f140f770/packages/api-contract/src/Abi/index.ts#L162-L171)

Not sure how much work this rework might imply. 🤔

TarikGul:
> So as v4 and v5 emit completely different data in the raw event, I think it might be best to make the actual Abi instance aware the metadata version used to instantiate it instead of having one Abi for all versions, deciding which path to take by checking which values are present or not.

This is definitely a valid solution. I think the common concern here is this would introduce a large change in the Abi. In terms of work I think it would be a large amount. That being said it is probably one of the best options. 

I'm trying wrap my head around the current solution to find some viability (setting `signature_topic`, and `module_path` to undefined and or if we set them to an Option).

Have we tried to run the current solution against a v4 contract to see the sideaffects?

TarikGul:
As far as I can tell the conversion is there to allow the Abi to work with a single version of the metadata. Correct me if I am wrong but if a v4 contract is being used and the abi transforms the metadata `toLatest` which would be v5 the general structure would still be valid for the v4 structure?

peetzweg:
@TarikGul, so the conversion from v4 to v5 is possible. However, I'm not sure if the metadata ink v5 spits out still adheres to the order of events as it was in v4. The order was important to decode the event in v4.

And as the Abi class right now always just works with the ["latest"](https://github.com/peetzweg/pjs-api/blob/cc946da2e904c7114aaa9914aee0e864f140f770/packages/api-contract/src/Abi/index.ts#L65) of the metadata, it's a bit ugly to determine if an event should be decoded in v4 style or v5 style.

TarikGul:
> I think it would be best to work with concrete metadata version in the Abi Class and narrow implementations based on that version type.

Considering your findings I agree. Better to take care of this now, and do it correctly than have a hacky less than ideal situation. It will also help decouple each version implementation in a way where it'll be easier to debug and create new versions in the future. Im on board, love it :)

> Jaco already added a comment about the decodeEvent(data: Bytes) method being unstable. This might become a breaking change, as we don't know who is using it directly. However, I think this method should just receive the full EventRecord instead of the destructured event.data.

Good point. For now I guess we can aim to just do a MINOR bump for this feature, and note the internal changes in the PR description.

peetzweg:
@TarikGul changed the method signature as discussed above in this commit: https://github.com/polkadot-js/api/pull/5791/commits/013ea59f70b265c7038936dbf8fa85729559a4a2

Furthermore, I drafted a version of the Abi which is aware of it's version in this commit: https://github.com/polkadot-js/api/pull/5791/commits/a028c554ee4260af08084c1c4818053a076e0577

Would appreciate a review of it. I'm not yet super happy with it, got a bit ugly with externalizing the `version` field to the `Abi` class. Maybe this could be done a bit better if we decide to move a `version` field into the `ContractMetadataVX`, however seems maybe we can leverage this `ContractMetadata` enum?

As of now I'm somewhat fine with the implementation. I haven't removed any of the previously used code yet, first and foremost `toLatest` => `toLatestCompatible`.

TarikGul:
> @TarikGul changed the method signature as discussed above in this commit: [013ea59](https://github.com/polkadot-js/api/commit/013ea59f70b265c7038936dbf8fa85729559a4a2)
> 
> Furthermore, I drafted a version of the Abi which is aware of it's version in this commit: [a028c55](https://github.com/polkadot-js/api/commit/a028c554ee4260af08084c1c4818053a076e0577)
> 
> Would appreciate a review of it. I'm not yet super happy with it, got a bit ugly with externalizing the `version` field to the `Abi` class. Maybe this could be done a bit better if we decide to move a `version` field into the `ContractMetadataVX`, however seems maybe we can leverage this `ContractMetadata` enum?
> 
> As of now I'm somewhat fine with the implementation. I haven't removed any of the previously used code yet, first and foremost `toLatest` => `toLatestCompatible`.

So I just had a look, did a moreso soft review, but It looks really good to me and feels like its going in the right direction.

peetzweg:
@TarikGul Okay I think this is ready for a proper review. For me this is the way to go for now. 🙌 

Once this is merged I will make sure the metadata in here is up to date and maybe adapt minor stuff as the "version" with this pr no longer `Text` but a `u64`.

https://github.com/paritytech/ink/pull/2126

packages/api-contract/src/Abi/Abi.spec.ts
```diff
@@ -122,4 +124,78 @@ describe('Abi', (): void => {
...
+      const expectedEvent = {
+        from: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY',
+        to: '5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty',
+        value: '123.4567 MUnit'
+      };
+
+      expect(decodedEventHuman).toEqual(expectedEvent);
+    });
+
+    it('decoding <=ink!v5 events', (): void => {
```
TarikGul, line 169: Do you mean `>=ink!v5` here?

packages/api-contract/src/Abi/index.ts
```diff
@@ -2,22 +2,24 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { Bytes } from '@polkadot/types';
-import type { ChainProperties, ContractConstructorSpecLatest, ContractEventSpecLatest, ContractMessageParamSpecLatest, ContractMessageSpecLatest, ContractMetadata, ContractMetadataLatest, ContractProjectInfo, ContractTypeSpec } from '@polkadot/types/interfaces';
 import type { Codec, Registry, TypeDef } from '@polkadot/types/types';
 import type { AbiConstructor, AbiEvent, AbiMessage, AbiParam, DecodedEvent, DecodedMessage } from '../types.js';
 
 import { Option, TypeRegistry } from '@polkadot/types';
+import { type ChainProperties, type ContractConstructorSpecLatest, type ContractMessageParamSpecLatest, type ContractMessageSpecLatest, type ContractMetadata, type ContractMetadataV4, type ContractMetadataV5, type ContractProjectInfo, type ContractTypeSpec, type EventRecord } from '@polkadot/types/interfaces';
```
TarikGul, line 9: I think this can just be `import type { ... }`.

packages/api-contract/src/Abi/index.ts
```diff
@@ -162,7 +167,31 @@ export class Abi {
...
+      // earlier version are hoisted to v4
+      case '4':
+        return this.#decodeEventV4(record);
+      // Latest
+      default:
+        return this.#decodeEventV5(record);
+    }
+  }
+
+  #decodeEventV5 = (record: EventRecord): DecodedEvent => {
```
TarikGul, line 181: Super clean. Nice job with `decodeEvent`, and `createEvent`.

packages/api-contract/src/Abi/index.ts
```diff
@@ -233,7 +262,36 @@ export class Abi {
...
+    // which breaks the type inference.
+    switch (this.metadata.version.toString()) {
+      case '4':
+        return this.#createEventV4((this.metadata as ContractMetadataV4).spec.events[index], index);
+      default:
+        return this.#createEventV5((this.metadata as ContractMetadataV5).spec.events[index], index);
+    }
+  };
+
+  #createEventV5 = (spec: ContractMetadataV5['spec']['events'][number], index: number): AbiEvent => {
```
TarikGul, line 277: The type for `spec` is a bit confusing, I think we can narrow it down to `ContractEventSpecV3`

TarikGul: instead of `ContractMetadataV5['spec']['events'][number]`

packages/api-contract/src/Abi/index.ts
```diff
@@ -233,7 +262,36 @@ export class Abi {
...
+      }),
+      identifier: [spec.module_path, spec.label].join('::'),
+      index,
+      signatureTopic: spec.signature_topic.toHex()
+    };
+
+    return event;
+  };
+
+  #createEventV4 = (spec: ContractMetadataV4['spec']['events'][number], index: number): AbiEvent => {
```
TarikGul, line 294: Same with the type here for `spec`.

TarikGul (changes requested):
Overall, excellent job! Looks really solid. Just a few nits.

Some thoughts on the decision for the specific versioning in the ABI. I think moving forward its probably the best choice, and allows for cleaner abstraction of logic. Things were a bit coupled earlier which caused our main issue of metadata conversion. Regardless, if there is some type of breaking change in the future it will always be difficult, but now there is a pretty solid track for multiple version support.

All that being said, very pleased with how it turned out! Nice job! When those few nits are resolved, I think we can give this a ✅ .

packages/api-contract/src/Abi/index.ts
peetzweg: I used a local type helper for this to make it less confusing. Otherwise it would be 

```ts
  #createEventV5 = (spec: ContractEventSpecV3): AbiEvent = {}
```

Method stating V5 and expects a V3. Using the helper it's always referring to a `V5`

```ts
  #createEventV5 = (spec: EventOf<ContractMetadataV5>): AbiEvent = {}
```

Internally we have version for some of the contract metadatas fields, like `ContractEventsV1/V2/V3`, `ContractMessageSpecV1/V2/V3` but only the actual top level metadata is properly version. IMHO this was confusing to me. Therefore I like route with the helper here more. Let me know if you disagree. ✌️

peetzweg:
@TarikGul tackled all requested changes. However, I would wait for a new [inkv5.rc2](https://github.com/paritytech/ink/releases) candidate to make sure the contract metadata has the expected shape as it was changed again here: https://github.com/paritytech/ink/pull/2126.

I manually patched the included metadata here https://github.com/polkadot-js/api/pull/5791/commits/f67a88af4d80dccb3566c683210836592b0bd98b, but just to be on the save side. ✌️

TarikGul:
> @TarikGul tackled all requested changes. ~However, I would wait for a new [inkv5.rc2](https://github.com/paritytech/ink/releases) candidate to make sure the contract metadata has the expected shape as it was changed again here: [paritytech/ink#2126](https://github.com/paritytech/ink/pull/2126).~
> 
> ~I manually patched the included metadata here [f67a88a](https://github.com/polkadot-js/api/commit/f67a88af4d80dccb3566c683210836592b0bd98b), but just to be on the save side. ✌️~
> 
> Just added updated the contract metadata with the most recent `master` branch of `ink!`, [27d6b46](https://github.com/polkadot-js/api/commit/27d6b46afaded8f0aabc2e570afc38664abd0f03).
> 
> So ready when you are. 🚀

Just a quick question given [27d6b46](https://github.com/polkadot-js/api/commit/27d6b46afaded8f0aabc2e570afc38664abd0f03), was it suppose to be set to 5.0.0-rc2?

peetzweg:
> Just a quick question given [27d6b46](https://github.com/polkadot-js/api/commit/27d6b46afaded8f0aabc2e570afc38664abd0f03), was it suppose to be set to 5.0.0-rc2?

It's bit tricky right now. There is no ink! 5.rc2 out just yet. So the contract metadata included in this PR are from these development version of ink!v5 aka master branch. 

The proper way would be probably to wait for the release of ink!v5 and then one final time 
refresh the contract metadata with output from the release version of ink!v5.

How imminent do you think the release of ink!v5 is, @cmichi?

Afaik the metadata is not about to change anymore for v5, so the only difference will be in the language field. But I'm fine also waiting for the actual release. 🤷  
```json
{
...
"language": "ink! 5.0.0-rc.1",
"language": "ink! 5.0.0",
...
}
```

cmichi:
We'll release either end of this week or first thing next week. So you can already update to "ink! 5.0.0" and merge with that. There'll be no more metadata changes.

@TarikGul It would be great if following the merge of this PR you could then issue a new release of `polkadot-js/api`. Then we can say that ink! 5 is compatible with the latest `polkadot-js/api` release, which would be a big thing.

TarikGul:
> > Just a quick question given [27d6b46](https://github.com/polkadot-js/api/commit/27d6b46afaded8f0aabc2e570afc38664abd0f03), was it suppose to be set to 5.0.0-rc2?
> 
> It's bit tricky right now. There is no ink! 5.rc2 out just yet. So the contract metadata included in this PR are from these development version of ink!v5 aka master branch.
> 
> The proper way would be probably to wait for the release of ink!v5 and then one final time refresh the contract metadata with output from the release version of ink!v5.
> 
> How imminent do you think the release of ink!v5 is, @cmichi?
> 
> Afaik the metadata is not about to change anymore for v5, so the only difference will be in the language field. But I'm fine also waiting for the actual release. 🤷
> 
> ```json
> {
> ...
> "language": "ink! 5.0.0-rc.1",
> "language": "ink! 5.0.0",
> ...
> }
> ```

Ahh okay thanks for the clear explanation, this should be fine!

TarikGul:
> We'll release either end of this week or first thing next week. So you can already update to "ink! 5.0.0" and merge with that. There'll be no more metadata changes.
> 
> @TarikGul It would be great if following the merge of this PR you could then issue a new release of `polkadot-js/api`. Then we can say that ink! 5 is compatible with the latest `polkadot-js/api` release, which would be a big thing.

@cmichi Would it be okay if we did the `polkadot-js/api` release on Monday to keep in cadence with the release schedule/process. Today, the extension, ui, tools, and phishing will be released, and the next day apps.

TarikGul (approved):
🚀

cmichi:
> @cmichi Would it be okay if we did the `polkadot-js/api` release on Monday to keep in cadence with the release schedule/process. Today, the extension, ui, tools, and phishing will be released, and the next day apps.

@TarikGul Yup, that's fine! What will be the version number of that release?

packages/api-contract/src/Abi/index.ts
```diff
@@ -162,7 +169,31 @@ export class Abi {
...
+      // Latest
+      default:
+        return this.#decodeEventV5(record);
+    }
+  }
+
+  #decodeEventV5 = (record: EventRecord): DecodedEvent => {
+    const data = record.event.data[1] as Bytes;
+    const signatureTopic = record.topics[0];
+    const event = this.events.find((e) => e.signatureTopic !== undefined && e.signatureTopic === signatureTopic.toHex());
```
ascjones, line 186: Some events can be "anonymous", i.e. they have no signature topic. 

In this case we can either:

1. Don't attempt to decode the event and just return the raw bytes, leave it up to the user to do the decoding.
2. Best effort attempt to decode, i.e. try to decode into all anonymous event definitions from the metadata.

TarikGul:
> > @cmichi Would it be okay if we did the `polkadot-js/api` release on Monday to keep in cadence with the release schedule/process. Today, the extension, ui, tools, and phishing will be released, and the next day apps.
> 
> @TarikGul Yup, that's fine! What will be the version number of that release?

Thanks so much! Version will be - `10.12.0` to reflect a feature change.

TarikGul:
@peetzweg Good for me to merge it in?

packages/api-contract/src/Abi/Abi.spec.ts
```diff
@@ -122,4 +124,78 @@ describe('Abi', (): void => {
...
+      }, {});
+
+      const expectedEvent = {
+        from: '5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY',
+        to: '5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty',
+        value: '123.4567 MUnit'
+      };
+
+      expect(decodedEventHuman).toEqual(expectedEvent);
+    });
```
ascjones, line 199: Could add a test here for an "anonymous" event

ascjones (commented):
Looks good! I think we need to handle anonymous events though https://use.ink/5.x/basics/events#anonymous-events

ascjones:
> Should we than just omit the Label of the event and use "Anonymous", log a warning or call it MaybeTransfer

To be safe for the initial version we can just say that if we can't find a matching event for the first topic **and** there exists an anonymous event in the metadata then we just label it as "Anonymous" and provide the raw bytes.

peetzweg:
@TarikGul @ascjones in this commit https://github.com/polkadot-js/api/pull/5791/commits/0e039e795131214aeb9cd0ebcb50ab70e72e608f I've implemented handling and trying to determine anonymous events. 

If only a single anonymous event exists in the contract, than this works just fine.

[When we have two or more anonymous events it tries to distinguish them by the amount of indexed parameters.](https://github.com/peetzweg/pjs-api/blob/3b489409ae38cb8b08e06a8eb7f72a1fcac5ead9/packages/api-contract/src/Abi/index.ts#L205-L210) Which could help separating two or more anonymous events if `[#[ink(topic)]](https://use.ink/macros-attributes/topic)` was used.

FYI in the referenced erc20 example contract we have two events, `Transfer` and `Approval`. I made the "Transfer" anonymous.

Afaik trying to best effort decoding the for all the anonymous events does not work with the current implementation of pjs, as it does not fail if the data is to long, so it could lead to false positive cases.

The other options of returning an `UndecodedEvent` with the plain bytes of the event is possible but takes a lot of refactoring the types up the hierarchy. I'm not sure if I can deliver it in time for the planned release on monday.

Furthermore this PR is quite big already and I suggest to add it in another PR, wdyt?

packages/api-contract/src/Abi/index.ts
```diff
@@ -181,15 +181,43 @@ export class Abi {
...
+
+      // If all conditions met, it's a potential event
+      return true;
+    });
+
+    if (potentialEvents.length === 1) {
+      return potentialEvents[0].fromU8a(data);
+    }
+
+    throw new Error('Unable to determine event');
```
ascjones, line 220: The other case to consider is if a cross contract call occurs, which in turn raises an event. Since the event is raised from another contract we don't have the metadata here...so we might not want to raise an error and just give back the raw bytes instead of attempting to decode? 

This scenario might also lead to a false positive in the heuristic for above for determining anon events, if the foreign event has the same number of topics :thinking:

ascjones:
> The other options of returning an UndecodedEvent with the plain bytes of the event is possible but takes a lot of refactoring the types up the hierarchy. I'm not sure if I can deliver it in time for the planned release on monday.

In that case we can do it in a follow up, better to have some support for the new event format - with the caveat that it will error if it encounters an unknown event (raised from another contract)

ascjones (approved):
LGTM

More work required to figure out the best way to handle anon and foreign events, but this is a great start.

packages/api-contract/src/Abi/index.ts
peetzweg: Oh okay, so if contract A calls a function on B which emits an event, the event will be emitted by A and not B? 

Feels like it makes sense as users might not know anything about B so can only listen to A. 

As of now I have to little knowledge on what's happening above this little 'Abi.decodeEvent world' to have an idea how to handle this the best, yet.

TarikGul:
Once this is in by EOD, i will most likely do the release Sunday evening in order to stay ahead of CET!

packages/api-contract/src/Abi/index.ts
ascjones: > Oh okay, so if contract A calls a function on B which emits an event, the event will be emitted by A and not B?

It depends, using `forward_call` (contract A calling contract B), the account id topic will be for contract B so would be easy to distinguish 

The other case is using `delegate_call` where the **code** of contract B is invoked from contract A, in this case the account id topic would be of contract A.

TarikGul:
The release should be out in 10.12.1 in a few moments.

To my surprise I tried to do a release last night after checking everything and woke up to the CI failing. Ends up you can't do a release in PJS with a `0` as the patch version. It always needs to end in `> 0`...

PR #5800, opened 2024-02-29
fix: Validate api.derive.staking instead of api.query.staking

frankli-dev:


TarikGul:
What is this exactly fixing? The code as is does what is intended, it is ensuring `api.query.staking` exists before `nextElected` checks if `api.query.staking.eraStakers` exists.

frankli-dev:
> What is this exactly fixing? The code as is does what is intended, it is ensuring `api.query.staking` exists before `nextElected` checks if `api.query.staking.eraStakers` exists.

I thought nextElected is under `api.derive.staking.nextElected()` so it would check `api.derive.staking`.
If it was intended, happy to close this.
Thanks!

TarikGul:
> > What is this exactly fixing? The code as is does what is intended, it is ensuring `api.query.staking` exists before `nextElected` checks if `api.query.staking.eraStakers` exists.
> 
> I thought nextElected is under `api.derive.staking.nextElected()` so it would check `api.derive.staking`. If it was intended, happy to close this. Thanks!

No worries, thank you for the PR regardless :) You are correct that `nextElected` is under `api.derive.staking.nextElected()`. It can be a bit confusing at first but the derive functions will always exist due to the fact that the `ApiDerive` auto generates them from the API itself. Therefore we wont need to check for `api.derive`. 

This can be seen [here](https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/bundle.ts#L137), where `derive` contains all the functions inside of `api-derive`.

In any case, I will close the PR, but thank you again for submitting it just in case. Better safe than sorry :)

frankli-dev:
> > > What is this exactly fixing? The code as is does what is intended, it is ensuring `api.query.staking` exists before `nextElected` checks if `api.query.staking.eraStakers` exists.
> > 
> > 
> > I thought nextElected is under `api.derive.staking.nextElected()` so it would check `api.derive.staking`. If it was intended, happy to close this. Thanks!
> 
> No worries, thank you for the PR regardless :) You are correct that `nextElected` is under `api.derive.staking.nextElected()`. It can be a bit confusing at first but the derive functions will always exist due to the fact that the `ApiDerive` auto generates them from the API itself. Therefore we wont need to check for `api.derive`.
> 
> This can be seen [here](https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/bundle.ts#L137), where `derive` contains all the functions inside of `api-derive`.
> 
> In any case, I will close the PR, but thank you again for submitting it just in case. Better safe than sorry :)

Thanks for your explanation!

Issue #5771, opened 2023-12-12
Underlying storage for exposures has changed


Ank4n:
* **I'm submitting a ...**
  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**
The `api.derive.staking._era_exposures(era)` under the hood calls `ErasStakersClipped`. This storage item would not be used anymore with the paged rewards changes ([PR link](https://github.com/paritytech/polkadot-sdk/pull/1189), [forum link](https://forum.polkadot.network/t/staking-update-paged-staking-reward-to-avoid-validator-oversubscription-issue-is-live-on-westend/5215)) and would be replaced with a new storage `ErasStakersPaged` which is a nmap of (Era, Validator, Page).

* **What is the motivation for changing the behavior?**
When runtime upgrades to [v1.4](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.4.0), this api would start breaking.

* **Suggest a motivation for the request or ideas how to implement the
addition or change**
There can be two apis, one to query how many pages of exposure exists which can be looked up by reading the storage item `ErasStakersOverview`. The current `_era_exposures` should then take a page index to read the exposure.

TarikGul:
I'll be working on this soon and getting this supported/fixed Going to ensure there is compatibility for previous runtime versions as well so this doesn't act as a breaking change for some folks.

TarikGul:
So i currently just made sure there is backwards and forward compatibility with `erasStakersPaged` for `_eraExposure`. I am going to create a tracking issue to remove `erasStakersClipped` once all relays have finally been updated to appropriate runtime.

Couple things:

> one to query how many pages of exposure exists which can be looked up by reading the storage item ErasStakersOverview

This already exists, the user can just use `api.query.staking.erasStakersOverview `

> The current _era_exposures should then take a page index to read the exposure.

With the current implementation this is not possible. Since `_eraExposure` is grabbing all the storageEntries using `.entries` the last storage entry arg is removed from the call. In this case that is `pageIndex`.

--------

On another note, https://github.com/polkadot-js/api/pull/5807 did augment the substrate types accordingly so the new storage calls are statically there for users.

PR #5813, opened 2024-03-08
Fix api.derive.staking.eraExposure [WIP]

TarikGul:
closes: https://github.com/polkadot-js/api/issues/5771

The following PR ensures to update the `api.derive.staking.eraExposure` to use the right storage queries. For now I ensure that is is backwards compatible. Once the transition between storage entries is completely finished with polkadot, `eraStakersClipped` can be completely removed. 

Once https://github.com/polkadot-js/api/pull/5812 goes in this will have to be updated again as the types are then removed from the augmented polkadot and kusama files.


TarikGul:
Few things need to be fixed in the derives that use eraExposure.

TarikGul:
Still need to test this against the APPS and also figure out how to get a validators own, and total since those are now deprecated from `PalletStakingExposure`

packages/api-derive/src/staking/stakerRewards.ts
```diff
@@ -48,8 +48,8 @@ function parseRewards (api: DeriveApi, stashId: AccountId, [erasPoints, erasPref
         const valCut = valComm.mul(avail).div(BN_BILLION);
         let staked: BN;
 
-        if (validatorId === stakerId) {
-          staked = exposure.own.unwrap();
+        if (validatorId === stakerId && (exposure as PalletStakingExposure).own) {
```
TarikGul, line 51: Not sure if this is valid. Need to double check

TarikGul: I need to add the storage call to `ErasStakersOverview `, and do this correctly so we don't need to come back to it.

TarikGul:
This PR brings up a good point for future derive work. What and how should compatibility be troubleshot and what is an acceptable change for upstream changes. Also how and when should they be released?

TarikGul:
It honestly might be better to just create a whole new derive called `eraExposurePaged` which users can then migrate too when the time comes

TarikGul:
closing in favor of https://github.com/polkadot-js/api/pull/5815

PR #5821, opened 2024-03-14
feat: collectives

ryanleecode:
Seems to nuke when trying to generate the xcm types


TarikGul:
Closing in favor of the solution we discussed to move all the logic directly in apps.

PR #5827, opened 2024-03-18
fix: check if minimumPeriod is zero

### LINKED ISSUE #5826: Support `minimumPeriod` of `0`
With the switch to async backing, parachains will set a `minimumPeriod` of `0`. (Not all, but some).

This currently breaks transaction signing here: https://github.com/polkadot-js/api/blob/8dbc373ac573e8c7aea580e8d17d634e40f01956/packages/api-derive/src/tx/signingInfo.ts#L88

We should check that if `minimumPeriod` is `0` and then just also use the fallback duration. On top of that we should use `SlotDuration` from Aura: https://github.com/paritytech/polkadot-sdk/pull/3732 as we use `ExpectedBlockTime` from Babe.

pepoviola:
Fix #5826 

Add check to ensure we don't use `0`.

### DIFF
--- a/packages/api-derive/src/tx/signingInfo.ts
+++ b/packages/api-derive/src/tx/signingInfo.ts
@@ -85,7 +85,7 @@ export function signingInfo (_instanceId: string, api: DeriveApi): (address: str
           MORTAL_PERIOD
             .div(
               api.consts.babe?.expectedBlockTime ||
-              (!api.consts.timestamp?.minimumPeriod.isZero() && api.consts.timestamp.minimumPeriod.muln(2)) ||
+              (!api.consts.aura?.slotDuration.isZero() && api.consts.aura.slotDuration) ||
               FALLBACK_PERIOD
             )
             .iadd(MAX_FINALITY_LAG)
--- a/packages/api-derive/src/tx/signingInfo.ts
+++ b/packages/api-derive/src/tx/signingInfo.ts
@@ -85,7 +85,7 @@ export function signingInfo (_instanceId: string, api: DeriveApi): (address: str
           MORTAL_PERIOD
             .div(
               api.consts.babe?.expectedBlockTime ||
-              api.consts.timestamp?.minimumPeriod.muln(2) ||
+              (!api.consts.timestamp?.minimumPeriod.isZero() && api.consts.timestamp.minimumPeriod.muln(2)) ||
               FALLBACK_PERIOD
             )
             .iadd(MAX_FINALITY_LAG)
--- a/packages/api-derive/src/tx/signingInfo.ts
+++ b/packages/api-derive/src/tx/signingInfo.ts
@@ -12,7 +12,7 @@ import { isNumber, isUndefined } from '@polkadot/util';
 
 import { unwrapBlockNumber } from '../util/index.js';
 import { FALLBACK_MAX_HASH_COUNT, FALLBACK_PERIOD, MAX_FINALITY_LAG, MORTAL_PERIOD } from './constants.js';
-import { AugmentedConsts } from '@polkadot/api-base/types';
+import type { AugmentedConsts } from '@polkadot/api-base/types';
 import { u64 } from '@polkadot/types';
 
 
--- a/packages/api-derive/src/tx/signingInfo.ts
+++ b/packages/api-derive/src/tx/signingInfo.ts
@@ -12,6 +12,9 @@ import { isNumber, isUndefined } from '@polkadot/util';
 
 import { unwrapBlockNumber } from '../util/index.js';
 import { FALLBACK_MAX_HASH_COUNT, FALLBACK_PERIOD, MAX_FINALITY_LAG, MORTAL_PERIOD } from './constants.js';
+import { AugmentedConsts } from '@polkadot/api-base/types';
+import { u64 } from '@polkadot/types';
+
 
 interface Result {
   header: Header | null;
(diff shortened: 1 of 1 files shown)


packages/api-derive/src/tx/signingInfo.ts
```diff
@@ -84,8 +84,8 @@ export function signingInfo (_instanceId: string, api: DeriveApi): (address: str
           api.consts.system?.blockHashCount?.toNumber() || FALLBACK_MAX_HASH_COUNT,
           MORTAL_PERIOD
             .div(
-              api.consts.babe?.expectedBlockTime ||
-              api.consts.timestamp?.minimumPeriod.muln(2) ||
+              (!api.consts.babe?.expectedBlockTime.isZero() && api.consts.babe?.expectedBlockTime) ||
+              (!api.consts.aura?.slotDuration.isZero() && api.consts.aura.slotDuration) ||
```
pepoviola, line 88: Maybe is better to use a helper function to check that is not `zero` for both (babe/aura) fields.

TarikGul: Yea this makes sense! WIll help clean it up a bit

TarikGul:
Since the augmented types dont have the aura consts yet, you will have to type case a custom type on `api.consts.aura` that sets slotDuration as a key value to `AugmentedConsts<"rxjs">.aura`. This can be all done locally inside of the api-derive.

pepoviola:
> Since the augmented types dont have the aura consts yet, you will have to type case a custom type on `api.consts.aura` that sets slotDuration as a key value to `AugmentedConsts<"rxjs">.aura`. This can be all done locally inside of the api-derive.

Oks, does make sense to use the old version

```ts
(!api.consts.timestamp?.minimumPeriod.isZero() && api.consts.timestamp.minimumPeriod.muln(2) ) ||
```

until https://github.com/paritytech/polkadot-sdk/pull/3732 is merged so we can use the aura const?

TarikGul:
> > Since the augmented types dont have the aura consts yet, you will have to type case a custom type on `api.consts.aura` that sets slotDuration as a key value to `AugmentedConsts<"rxjs">.aura`. This can be all done locally inside of the api-derive.
> 
> Oks, does make sense to use the old version
> 
> ```ts
> (!api.consts.timestamp?.minimumPeriod.isZero() && api.consts.timestamp.minimumPeriod.muln(2) ) ||
> ```
> 
> until [paritytech/polkadot-sdk#3732](https://github.com/paritytech/polkadot-sdk/pull/3732) is merged so we can use the aura const?

If we can maintain the same logic and patch with this then sure that is totally acceptable!

packages/api-derive/src/tx/signingInfo.ts
```diff
@@ -12,6 +12,9 @@ import { isNumber, isUndefined } from '@polkadot/util';
 
 import { unwrapBlockNumber } from '../util/index.js';
 import { FALLBACK_MAX_HASH_COUNT, FALLBACK_PERIOD, MAX_FINALITY_LAG, MORTAL_PERIOD } from './constants.js';
+import { AugmentedConsts } from '@polkadot/api-base/types';
```
TarikGul, line 15:
```suggestion
import type { AugmentedConsts } from '@polkadot/api-base/types';
```

packages/api-derive/src/tx/signingInfo.ts
```diff
@@ -63,6 +66,20 @@ function signingHeader (api: DeriveApi): Observable<Header> {
   );
 }
 
+interface Aura {
+  slotDuration: u64 & AugmentedConsts<'rxjs'>;
+}
+
+function babeOrAuraPeriod(api: DeriveApi) : AnyNumber|undefined {
```
TarikGul, line 73:
```suggestion
function babeOrAuraPeriod(api: DeriveApi) : BN | undefined {
```

This will also require the following

```typescript
import type { BN } from '@polkadot/util';
```

TarikGul (changes requested):
Few nits, but looks amazing. Once the suggestions are fixed this should be good to go. Also needs a `yarn lint` once the following is fixed.

pepoviola:
> Few nits, but looks amazing. Once the suggestions are fixed this should be good to go. Also needs a `yarn lint` once the following is fixed.

Thanks for review it, I made those changes 👍

TarikGul (approved):
LGTM 🚀

packages/api-derive/src/tx/signingInfo.ts
```diff
@@ -63,6 +66,19 @@ function signingHeader (api: DeriveApi): Observable<Header> {
...
 
+interface Aura {
+  slotDuration: u64 & AugmentedConsts<'rxjs'>;
+}
+
+function babeOrAuraPeriod (api: DeriveApi): BN | undefined {
+  const period = api.consts.babe?.expectedBlockTime ||
+    // this will be present ones https://github.com/paritytech/polkadot-sdk/pull/3732 is merged
+    (api.consts['aura'] as unknown as Aura)?.slotDuration ||
+    api.consts.timestamp?.minimumPeriod;
```
sinzii, line 77: Hi everyone, I notice that we do `x2` (`.muln(2)`) for the `minimumPeriod` previously, but now we removed it here. I didn't see we mentioned this change anywhere, any reason for this? I just want to confirm that this is intentional. Thank you!

pepoviola: Hey @sinzii, thanks for the heads up! This was an error, this [pr](https://github.com/polkadot-js/api/pull/5829) fix it.
Thx!

sinzii: Awesome, thanks for the quick fix!

PR #5831, opened 2024-03-20
Add `disableAscii` option for toHuman, and toPrimitive

TarikGul:
My attempt at resolving the upstream issue in apps: https://github.com/polkadot-js/apps/issues/10343

This adds the `disableAscii` option to the `toHuman` and `toPrimitive` property on Codec types. The problem came up when the UI was trying to decode the chainState for `broker.regions` in rococo coretime parachain. The type `PalletBrokerCoreMask` inside of `PalletBrokerRegionId` had the following bytes `[85, 85, 85, 85, 85, 85, 85, 85, 85, 85]` which can be also interpreted as valid ASCII, and when passed through the [Raw types toHuman](https://github.com/polkadot-js/api/blob/master/packages/types-codec/src/native/Raw.ts#L111-L138) it converts the following to ASCII which should not be the case. Instead a hex value here makes more sense. 

So the solution was to add `disableAscii` which should now work for any Codec type that might include a sub value that is interpreted to the `Raw` base type.

### DIFF
--- a/packages/types-codec/src/extended/Map.ts
+++ b/packages/types-codec/src/extended/Map.ts
@@ -182,7 +182,7 @@ export class CodecMap<K extends Codec = Codec, V extends Codec = Codec> extends 
 
     for (const [k, v] of this.entries()) {
       json[
-        k instanceof Raw && k.isAscii
+        k instanceof Raw && !disableAscii && k.isAscii
           ? k.toUtf8()
           : k.toString()
       ] = v.toHuman(isExtended, disableAscii);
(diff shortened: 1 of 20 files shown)


packages/types-codec/src/extended/Map.ts
```diff
@@ -177,15 +177,15 @@ export class CodecMap<K extends Codec = Codec, V extends Codec = Codec> extends
   /**
    * @description Converts the Object to to a human-friendly JSON, with additional fields, expansion and formatting of information
    */
-  public toHuman (isExtended?: boolean): Record<string, AnyJson> {
+  public toHuman (isExtended?: boolean, disableAscii?: boolean): Record<string, AnyJson> {
     const json: Record<string, AnyJson> = {};
 
     for (const [k, v] of this.entries()) {
       json[
         k instanceof Raw && k.isAscii
```
TarikGul, line 185: disableAscii should be used here as well

TarikGul: Fixed

Issue #5725, opened 2023-08-30
Expose runtime methods from metadata V15


jacogr:
Last remaining issue from https://github.com/polkadot-js/api/issues/5633

xlc:
I only remember one feature from metadata v15 and it is not supported 😢

ntn-x2:
Hello @TarikGul! How is this scored in priority compared to other tasks? When can we expect work to be completed on this?

TarikGul:
Hey @ntn-x2, this is definitely a high priority task. There has just been a big backlog (bugs, reviews etc.), and learning curve with certain workflows especially with all the packages. Things will really start to gain speed soon, and I will have this on the list for next week. 

Thanks again for your patience.

ntn-x2:
Great, thank you! I will check in with my co-workers if we can support this, since we have digged quite deep already into type augmentation, and having this in and released would greatly benefit us, so we could help out.

TarikGul:
Okay cool, I've started this and am making some ground. PR should be wrapped up next weekish

rflechtner:
> Okay cool, I've started this and am making some ground. PR should be wrapped up next weekish

Very cool! Will this be covering the typegen as well, so that runtime call interfaces are generated from the metadata without any additional input?

TarikGul:
> > Okay cool, I've started this and am making some ground. PR should be wrapped up next weekish
> 
> Very cool! Will this be covering the typegen as well, so that runtime call interfaces are generated from the metadata without any additional input?

So my thought to this is a 2 part solution, first the new runtime api's will be added via `./types/*/*/runtime.ts` since the runtimes default metadata is still v14.

The following runtime apis to be added:

```
ValidateStatement
AssetsApi
ContractsApi
AssetConversionApi
NftsApi
MMRApi
    - leafCount
```

AFAIU the next step will be changing the `typegen` to be able to read the `apis` field inside of the metadata itself. This will be a bigger change though as the structure has changed from v14 to v15.

PR #5837, opened 2024-03-27
Add Updated Runtime Defintions

TarikGul:
This is part 1 of updating the api to deal with new runtime definitions, and then add functionality to read it directly from the metadata.

- [x] ValidateStatement
- [X] MMRApi
    - [X] mmr_root
    - [X] mmr_leaf_count
- [x] Core v5
- [x] GenesisBuildApi
- [x] MixnetApi
- [x] Ensure all headers are set with copyright.

rel: https://github.com/polkadot-js/api/issues/5725


packages/types/src/interfaces/statement/definitions.ts
```diff
@@ -0,0 +1,56 @@
...
+import { runtime } from './runtime.js';
+
+export default {
+  rpc: {},
+  runtime,
+  types: {
+    SpStatementStoreStatementSource: {
+      _enum: ['Chain', 'Network', 'Local']
+    },
+    // SpStatementStoreStatement: {
```
TarikGul, line 18: This type already exists in the substrate defs, so we can get rid of it. But that begs the questions should we be naming these types similar to the typegen? I would say no so there is no conflicts in the future.

TarikGul:
Also need to ensure the versions are correct for all the runtime api definitions.

TarikGul:
runtime/core has a version bump from 4-5 that needs to be covered.

TarikGul:
The goal here is to add the runtime definitions and update them so we can get it out in the next release. Then we will update the actual typegen to use the following apis field in the metadata for V15

PR #5835, opened 2024-03-24
chore: fix typos

xiaoxianBoy:
fix a typo: `Unkown`->`Unknown`

PR #5855, opened 2024-04-19
Fix typos and grammatical errors.

Sunny-2077:


TarikGul:
@Sunny-2077 Because its spam. The repos have been getting flooded with typo PR's from private accounts with no github history, and no previous track record of open source contributions. Seems suspicious that the same behavior is happening from a bunch of different private accounts across a bunch of repos. 

https://github.com/polkadot-js/api/pull/5835 was an exception since it fixed an actual error message, but moving forward all these spam accounts will have their PR's closed.

PR #5862, opened 2024-04-23
Add claimedRewardsEra to api.derive.staking.query for compatibility with legacyClaimedRewards

TarikGul:
### `StakingQueryFlags`

Now takes in a new flag: `withClaimedRewardsEras`. This flag returns all the claimedRewards eras based on the new storage entry `ClaimedRewards`.

### `api.derive.staking.electedInfo`:

Adds a 3rd param for `page`.

### `api.derive.staking.query` && `api.derive.staking.queryMulti` 

Returns a new field `claimedRewardsEras` when `withClaimedRewardsEras` is true. This is for compatibility reasons for legacyClaimedRewards.

### Fix `api.derive.staking.stakingRewards`

Now takes into account claimedRewards storage key to ensure all eras are accounted for.

TODO:
- [x] Get the observable working in `query`
- [x] Fix  claimedRewards in `stakerRewards`
- [x] Fix `electedInfo`

### DIFF
--- a/packages/api-derive/src/staking/query.ts
+++ b/packages/api-derive/src/staking/query.ts
@@ -21,6 +21,7 @@ function rewardDestinationCompat (rewardDestination: PalletStakingRewardDestinat
 }
 
 function filterClaimedRewards(cl: number[]): number[] {
+  console.log('CL: ', cl);
   return cl.filter((c) => c !== -1)
 }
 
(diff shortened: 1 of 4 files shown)


packages/api-derive/src/staking/query.ts
```diff
@@ -84,17 +99,37 @@ function getStashInfo (api: DeriveApi, stashIds: AccountId[], activeEra: EraInde
...
+      ? combineLatest(stashIds.map((s) =>
+        combineLatest(
+          eras.map((e) => api.query.staking.claimedRewards(e, s))
+        ))
+      ).pipe(
+        map((r) => {
+          return r.map((stashClaimedEras) => {
+            // stashClaimedEras length will match the length of eras
+            return stashClaimedEras.map((claimedReward, idx) => {
+              if (claimedReward.length) {
```
TarikGul, line 113: This isn't completely accurate and would need to be verified against how many actual pages there are.

TarikGul: Updated this to verify against the overview pages

PR #5860, opened 2024-04-22
Update api.derive.staking calls with breaking changes

TarikGul:
This updates the staking derives with the updated breaking changes. More info here: [5771](https://github.com/polkadot-js/api/issues/5771) and https://github.com/polkadot-js/apps/issues/10004

### `api.derive.staking.ownExposure`:

Now takes in a 3rd param `page` to allow for paged calls, since it internally uses `erasStakersPaged`. If no page is passed in, it will default to 0.

The output has new additions. This includes `exposurePaged`, and `exposureMeta`. The old fields remain since `_ownExposures` internally is used by other calls that may query a historical range. 

```typescript
export interface DeriveOwnExposure {
  clipped: SpStakingExposure;
  exposurePaged: Option<SpStakingExposurePage>;
  era: EraIndex;
  exposure: SpStakingExposure;
  exposureMeta: Option<SpStakingPagedExposureMetadata>;
}
```

### `api.derive.staking.ownExposures`:

Now takes in a 4th param `page` to allow for paged calls, since it internally uses `erasStakersPaged`. If no page is passed in, it will default to 0.

The output has new additions. This includes `exposurePaged`, and `exposureMeta`. The old fields remain since `_ownExposures` internally is used by other calls that may query a historical range. 

```typescript
export interface DeriveOwnExposure {
  clipped: SpStakingExposure;
  exposurePaged: Option<SpStakingExposurePage>;
  era: EraIndex;
  exposure: SpStakingExposure;
  exposureMeta: Option<SpStakingPagedExposureMetadata>;
}
```

### `api.derive.staking.query`:

Now takes in a 3rd param `page` to allow for paged calls, since it internally uses `erasStakersPaged`. If no page is passed in, it will default to 0.

The output has additions and changes. This includes the addition of the `exposureMeta` field, and the change of `exposure` to `exposurePaged`.

```typescript
export interface DeriveStakingStash {
  controllerId: AccountId | null;
  exposurePaged: Option<SpStakingExposurePage>;
  exposureMeta: Option<SpStakingPagedExposureMetadata>;
  nominators: AccountId[];
  rewardDestination: PalletStakingRewardDestination | null;
  stashId: AccountId;
  validatorPrefs: PalletStakingValidatorPrefs;
}
export interface DeriveStakingQuery extends DeriveStakingStash {
  accountId: AccountId;
  stakingLedger: PalletStakingStakingLedger;
}
```

### `api.derive.staking.nextElected` && `api.derive.staking.validators`:

[...]

PR #5868, opened 2024-04-30
ApiDerive.staking backwards compatibility. 

TarikGul:
This PR works on giving the support needed for backwards compatibility. With the recent staking breaking changes these PR's introduced: https://github.com/polkadot-js/api/pull/5860 https://github.com/polkadot-js/api/pull/5862 It also ended up breaking compatibility for some parachains in the UI. 

This is part 1 of the staking changes saga, and once this goes in I can rework the UI to have support for the new and the old.

## Summary

- `ownExposure` && `ownExposures` will properly build safe types for historic calls such as `erasStakers`, and `erasStakersClipped`. It also ensures the calls exist or else will default to the empty type.

- `query` && `queryMulti` now take in a flag `withExposureErasStakersLegacy` and returns a new field called `exposureEraStakers`

- `nextElected` now has support for both `erasStakersPaged` and `erasStakers`.

TODO:

- [x] Test the following changes against apps to ensure it works for both old and new. 

NOTE: My one concern is that api.derive.staking.stakersRewards* uses `withClaimedErasRewards` as a default flag [here](https://github.com/polkadot-js/api/blob/master/packages/api-derive/src/staking/stakerRewards.ts#L190) which could potentially still act as a regression for the UI.

rel: https://github.com/polkadot-js/apps/issues/10512
rel :https://github.com/polkadot-js/apps/issues/10513
rel: https://github.com/polkadot-js/apps/issues/10505

### DIFF
--- a/packages/api-derive/src/staking/ownExposure.ts
+++ b/packages/api-derive/src/staking/ownExposure.ts
@@ -4,7 +4,7 @@
 import type { Observable } from 'rxjs';
 import type { Option, u32 } from '@polkadot/types';
 import type { EraIndex } from '@polkadot/types/interfaces';
-import type { SpStakingExposurePage } from '@polkadot/types/lookup';
+import type { SpStakingExposure, SpStakingExposurePage, SpStakingPagedExposureMetadata } from '@polkadot/types/lookup';
 import type { AnyNumber } from '@polkadot/types-codec/types';
 import type { DeriveApi, DeriveOwnExposure } from '../types.js';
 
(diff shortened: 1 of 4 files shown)


TarikGul:
It might be useful to add some throttling to `api.derive.staking.queryMulti` so that thousands of calls aren't emitted at the same time. (Not sure if this is the appropriate decision yet as I have not thought about it much.)

TarikGul:
Note: I wont be able to test this against Apps until I have some stable internet come Sunday as I am still OOO.

TarikGul:
I tested each call that was changed against polkadex to ensure the old calls are now available, and I also ensured everything is fault tolerant so it wont fail if the calls don't exist.

PR #5870, opened 2024-05-07
Fix claimedRewards in staking.derives

TarikGul:
This PR focuses on fixing the regression that was introduced to `api.derive.staking.queryMulti` in https://github.com/polkadot-js/api/pull/5862. 

Instead of creating thousands of subscriptions to get all the info we need we simple receive all the entries and transform the data to fit the existing structure.

Thanks to @josepot for the help and inspiration.


TarikGul:
closes: https://github.com/polkadot-js/api/pull/5870

PR #5914, opened 2024-06-21
Add `signedTransaction` to `SignerResult` for injected Signers in `signAndSend`

TarikGul:
## Summary

**Disclaimer**: (NO BREAKING CHANGES) The following additions do not change any existing logic unless a wallet opts in to using `signedTransaction`. More information below.

The following allows signers to modify fields in the payload passed into `signPayload`. This results in the return type of `SignerResult` to now be:

```typescript
export interface SignerResult {
  /**
   * @description The id for this request
   */
  id: number;

  /**
   * @description The resulting signature in hex
   */
  signature: HexString;

  /**
   * @description The payload constructed by the signer.
   */
  signedTransaction?: HexString | Uint8Array;
}
```

### `signedTransaction` field. (Optional)

If no `signedTransaction` is present in `SignerResult` everything will continue as it always has. When `signedTransaction` is present it will broadcast the transaction that is inputted.

### Additional Changes

`mode` - I added the ability to decode `mode` from the output of an `ExtrinsicBase`

### TODO:

- [x] Live testing via https://github.com/TarikGul/signer-test
- [X] Internal Docs

Edit: I have edited out the last summary and iteration which has been agreed upon as a bad iteration. It used `SignerPayloadJSON` as the response in the `SignerResult` which makes it hard for other library implementations. Moving forward there will be a standardization for the external facing API to make it better for wallets and lib maintainers.

### DIFF
--- a/packages/api/src/submittable/createClass.ts
+++ b/packages/api/src/submittable/createClass.ts
@@ -89,6 +89,11 @@ export function createClass <ApiType extends ApiTypes> ({ api, apiType, blockHas
 
   class Submittable extends ExtrinsicBase implements SubmittableExtrinsic<ApiType> {
     readonly #ignoreStatusCb: boolean;
+    /**
+     * For signers that leverage sending back a modified `SignerPayloadJson`, this
+     * ensures that only certain keys are allowed to be modified.
+     */
+    readonly #whitelistedKeys: string[];
 
     #transformResult = identity<ISubmittableResult>;
 
@@ -96,6 +101,7 @@ export function createClass <ApiType extends ApiTypes> ({ api, apiType, blockHas
       super(registry, extrinsic, { version: api.extrinsicType });
 
       this.#ignoreStatusCb = apiType === 'rxjs';
+      this.#whitelistedKeys = ['mode', 'metadataHash'];
     }
 
     public get hasDryRun (): boolean {
@@ -364,8 +370,6 @@ export function createClass <ApiType extends ApiTypes> ({ api, apiType, blockHas
     };
 
     #validateResults = (originalPayload: SignerPayload, newPayload: SignerPayload): void => {
-      // Whitelisted keys are free to be modified
-      const whitelistedKeys = ['mode', 'metadataHash'];
       const newPayloadJSON = newPayload.toPayload();
       const originalPayloadJSON = originalPayload.toPayload();
       const keys = Object.keys(newPayload);
@@ -373,7 +377,7 @@ export function createClass <ApiType extends ApiTypes> ({ api, apiType, blockHas
       // Validation process checks to make sure the type integrity from the original payload is maintianed.
       for (const key of keys) {
         // Whitelisted keys are free to be modified by the signer.
-        if (whitelistedKeys.includes(key)) {
+        if (this.#whitelistedKeys.includes(key)) {
           continue;
         }
 
--- a/packages/api/src/submittable/createClass.ts
+++ b/packages/api/src/submittable/createClass.ts
@@ -219,7 +219,16 @@ export function createClass <ApiType extends ApiTypes> ({ api, apiType, blockHas
     // signAndSend with options and a callback
     public signAndSend (account: AddressOrPair, partialOptions: Partial<SignerOptions>, statusCb?: Callback<ISubmittableResult>): SubmittableResultSubscription<ApiType>;
 
-    // signAndSend implementation for all 3 cases above
+    /**
+     * signAndSend implementation for all 3 cases above.
+     *
+     * @description Sign and broadcast the constructued transaction.
+     *
+     * Note for injected signers. The `SignerResult` type allows for the `signedTransaction` field
+     * for v12.0.1 and up. This allows the signer to input a signed transaction that will be directly broadcasted. This
+     * bypasses the api adding the signature to the payload. The api will ensure that the Call Data is not changed before it broadcasts the
+     * transaction.
+     */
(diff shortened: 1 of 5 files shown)


josepot:
Using this incredibly leaky interface as an output only makes the problem worse for other library authors.

The solution should **not** be to make these interfaces leakier, but rather: to start using a non-leaky interface.

For instance, a solution could be to add a new method like [the `signTransaction`](https://forum.polkadot.network/t/the-need-for-an-improved-signer-api/6466) that I described **4 months ago**). All what PJS would have to then is to check whether the `signTransaction` API is available... If it isn't, then it just keeps doing the same stuff that it's currently doing, and if it is available then it uses the non-leaky API (which is properly decoupled from PJS).

This would allow us to set a time-frame to wallets so that they can deprecate these old leaky methods that only make sense in to PJS.

It is incredibly frustrating and unfair for other library authors to have to witness how PJS keeps making the situation worse.

TarikGul:
@josepot I have changed the whole structure to outline the decisions we have discussed and agreed upon.

josepot:
> @josepot I have changed the whole structure to outline the decisions we have discussed and agreed upon.

Thank you very much, @TarikGul, for making this change more library-agnostic.

However, there's an important point to note: the new API allows the signer/extension to modify the original call-data and/or the values of other signed extensions.

I believe this is actually beneficial. To clarify, the DApp should merely suggest what it wants the user to sign, but the final decision lies with the user. Additionally, the extension must handle certain aspects no matter what, such as the nonce and the value of the new signed extension.

From the perspective of PJS DApp developers, there used to be a "guarantee" that if the extension signed data different from what was instructed (altered call-data, tip, nonce, mortality, etc.), PJS would ensure the transaction wasn't broadcasted from their DApp. The extension could still broadcast the transaction on its own, so the old API wasn't inherently more secure.

Some might argue that this PR alters the `signAndSend` (et al) behavior because the signed and sent data can now differ. Therefore, it might be wise for PJS to maintain consistency with the "old" behavior in `signAndSend`.

A practical approach could be to encode the [extra-data](https://spec.polkadot.network/id-extrinsics#defn-extra-data) from the `SignerPayloadJSON` excluding the "mode," then compare these bytes with those from the `signedTransaction` while removing the `CheckMetadataHash` bytes.

We could consider releasing this change as a new major version of PJS, ensuring backward compatibility. This version could introduce a new API like `signAndSubmitStrict`, which enforces the "old" behavior, while `signAndSubmit` would broadcast whatever the extension returns. Alternatively, we could maintain the current restrictive behavior of `signAndSubmit`... Another option could be to just release it as a major version with a change in behaviour on `signAndSend` et al. I think that all these options are actually reasonable.

TarikGul:
> Thank you very much, @TarikGul, for making this change more library-agnostic.

You're very welcome, and thank you for the write up!

> Another option could be to just release it as a major version with a change in behaviour on signAndSend et al. I think that all these options are actually reasonable.

I like all the options outlined, but I would align with this current one the most. As mentioned all have their trade offs and while I do think decoupling the stricter ("old") signAndSend from the new unrestricted one would be nice, internally it would feel like a nightmare IMO. Being that our main goal for the future is to create a more standardized `signTransaction` api that is well documented and is more generalized for all library authors making the current api (and the new `signedTransaction` field) as simple as possible to integrate with would be a big plus which I think the current implementation does. 

> Some might argue that this PR alters the signAndSend (et al) behavior because the signed and sent data can now differ. Therefore, it might be wise for PJS to maintain consistency with the "old" behavior in signAndSend.

Agreed, technically speaking yes some might argue that technically this is a "breaking change" since the signer can bypass the original payload being signed so it would be wise for us to do a major increase of the version.

TarikGul:
This is also obviously up for discussion, but I added validation to the `Extra Types` {nonce, era, tip}, but that being said I also included the `Call` data. 

`Mode` - and `Metadata` hash were left out for obvious reasons.

packages/api/src/submittable/createClass.ts
```diff
@@ -349,6 +377,32 @@ export function createClass <ApiType extends ApiTypes> ({ api, apiType, blockHas
         }
       }
     };
+
+    /**
+     * When a signer includes `signedTransaction` within the SignerResult this will validate
+     * specific fields within the signed extrinsic against the original payload that was passed
+     * to the signer.
+     */
+    #validateSignedTransaction = (signerPayload: SignerPayload, signedExt: Extrinsic): void => {
```
IkerAlus, line 386: Note that this validation logic only covers the current case of Polkadot / Kusama relay chains. Other chains (and future relay chain upgrades) may have other signed extensions which would be skipped if we go for this function (for example, the asset ID for fee payment in the Asset Hubs).

Therefore my suggestion is either to:
1. Do no validation at all before submission, and clearly document the behaviour of the`signedTransaction` mode,
2. Only validate the payload method, also documenting the implications, or
3. do a byte-level comparison of the `included_in_extrinsic` kind of signed extensions given by the relevant chain, this is, the "extra-data" present in the `signedTransaction`.

My suggestion is to go for 1 or 2 as discussed beforehand.

SBalaguer: I agree with what @IkerAlus proposes, however as discussed, I'm more of a proponent of option (2): I would validate that `payload.method` remains the same.

Even though adding this check does not inherently improves the security of the system as the signer can broadcast any transaction it wants, I believe it does help to take a more conservative approach until the ecosystem can move to a bigger, different structure.

TarikGul: Sounds good, I'll just ensure that the method is the same and that's it!

bee344 (approved):
Amazing work

PR #5915, opened 2024-06-24
fix: updated `nextNonce` to use `accountNonceApi`

bee344:
Updates `nextNonce` to use the runtimeApi instead of the rpc call, when available.

Related to [10641](https://github.com/polkadot-js/apps/issues/10641)


TarikGul (approved):
Thank you :) 👍

Issue #5845, opened 2024-04-10
Add decoration of runtimeApis


TarikGul:
rel: https://github.com/polkadot-js/api/issues/5725

The next step in getting runtimeApis from the metadata is to have them decorated at runtime via the metadata. This can be seen as similar to [`decorateExtrinsics`](https://github.com/polkadot-js/api/blob/master/packages/types/src/metadata/decorate/extrinsics/index.ts#L49). This should only be available for V15. 

Once this issue is complete we can then add it to the static typegen.

TarikGul:
This would actually replace `decorateCalls` inside of the api base decoration.

peetzweg:
@TarikGul Is there a way to manually add runtime API calls for now? 

I tried using the `runtime` option during `ApiPromise` creation like below. However, it's not even added to `ApiPromise.call` (`undefined is not an object`), so probably not the way how to do it in the first place. 🤔 

```ts
import { DefinitionsCall } from "@polkadot/types/types";
import { ApiPromise, WsProvider } from "@polkadot/api";

const wsProvider = new WsProvider(`wss://...`);

export const api = await ApiPromise.create({
  provider: wsProvider,
  runtime: {
    camelCasedPalletNameApi: [
      {
        version: 1,
        methods: {
          camelCasedRuntimeFn {
            description: "This is my function",
            params: [],
            type: "Balance",
          },
        },
      },
    ],
  } as DefinitionsCall,
});

const result = await pjsApi.call.camelCasedPalletNameApi.camelCasedRuntimeFn();
console.log({result})
```

decentration:
this should go in V16

josepot:
> this should go in V16

This is already available in V15... 🤷

PR #5941, opened 2024-07-19
Initialize the API with metadata V15

TarikGul:
[WIP]

rel: https://github.com/polkadot-js/api/issues/5725
rel: https://github.com/polkadot-js/api/issues/5845

This is Part 1 of a 2 part process.

### Summary

This will Allow the API to initialize with metadata 15. Once this is finished, then I will ensure the runtime calls are decorated via this metadata.

### DIFF
--- a/packages/api/src/base/Init.ts
+++ b/packages/api/src/base/Init.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { Observable, Subscription } from 'rxjs';
-import type { Text } from '@polkadot/types';
+import type { Text, u32 } from '@polkadot/types';
 import type { ExtDef } from '@polkadot/types/extrinsic/signedExtensions/types';
 import type { ChainProperties, Hash, HeaderPartial, RuntimeVersion, RuntimeVersionPartial } from '@polkadot/types/interfaces';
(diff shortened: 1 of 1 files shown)


TarikGul:
I got `api.at` to also work with the above which is great. Looks like the PR is about done, and part 2 will be using this ground work to decorate the runtime calls using the metadata

TarikGul:
I have found some issues with some historic runtimes that have the metadata api, but there is specific decoding issues

TarikGul:
Need to resolve this before it's ready:

```
packages/api/build/bundle.js → packages/api/build/bundle-polkadot-api.js...
(!) Unresolved dependencies
https://rollupjs.org/troubleshooting/#warning-treating-module-as-external-dependency
@polkadot/types/metadata/versions (imported by "packages/api/build/base/Init.js")
(!) Missing global variable name
https://rollupjs.org/configuration-options/#output-globals
Use "output.globals" to specify browser global variable 
```

TarikGul:
The current issue with v15 - the `new Metadata(registry, metadatav15Hex)` breaks all decoding of blocks so the Metadata structure and decoding needs to be fixed before this can go in.

TarikGul:
Registry is now fixed and decoding is good to go!

PR #5946, opened 2024-07-30
Revert "Alias for Noir types (#5678)"

conr2d:
This reverts commit d946a99a353f93920e506fa25ffc2a20dcecf4fe.

This PR removes the Noir-specific types from `@polkadot/types`.

In the early version of Noir, variable-length public keys are used as account id directly, but found that it is really painful to replace `AccountId32` with custom account id type `UniversalAddress`. (e.g. incompatible with existing wallets, etc.) Now Noir also uses `AccountId32`, so those changes for adding Noir-specific types become obsolete.


TarikGul (approved):
Thank you

Issue #5856, opened 2024-04-19
Fix transferable balance in `api.derive.balances.all` with new formula


TarikGul:
Currently the following code snippet shows that we are using an outdated balance formula for retrieving the `transferable` balance. Related issue: https://github.com/polkadot-js/apps/issues/9988 (Please see the issue for more detail).

https://github.com/polkadot-js/api/blob/6415eb6515e8dee0944fb162a2af30da395b0a1e/packages/api-derive/src/balances/all.ts#L32-L53

https://github.com/polkadot-js/api/blob/6415eb6515e8dee0944fb162a2af30da395b0a1e/packages/api-derive/src/balances/all.ts#L55-L64

The curent implementation needs to take into account:

https://github.com/paritytech/polkadot-sdk/blob/3069b0af398e5a0374802c31201637e97f4a530a/substrate/frame/balances/src/impl_fungible.rs#L44-L70

TarikGul:
rel: https://github.com/paritytech/polkadot-sdk/issues/1833

TarikGul:
This is currently what I am working with: `max(free - (frozen - on_hold), free - ED)`

PR #5955, opened 2024-08-18
Rework api.derive.balance.{account, all} to work with the new frame account data

TarikGul:
closes: https://github.com/polkadot-js/api/issues/5856

## Changes

`DeriveBalancesAllAccountData`: This now adds a new field `transferable` which has a type of `Balance | null`. This will only be applicable if the underlying runtime supports `FrameSystemAccountInfo` type.

```typescript
export interface DeriveBalancesAllAccountData extends DeriveBalancesAccountData {
  /**
   * Calculated available balance. This uses the formula: max(0, free - locked)
   * This is only correct when the return type of `api.query.system.account` is `AccountInfo` which was replaced by `FrameSystemAccountInfo`.
   * See `transferable` for the correct balance calculation.
   *
   * ref: https://github.com/paritytech/substrate/pull/12951
   */
  availableBalance: Balance;
  /**
   * The amount of balance locked away.
   */
  lockedBalance: Balance;
  /**
   * The breakdown of locked balances.
   */
  lockedBreakdown: (PalletBalancesBalanceLock | BalanceLockTo212)[];
  /**
   * Calculated transferable balance. This uses the formula: free - max(frozen - reserve, ed)
   * This is only correct when the return type of `api.query.system.account` is `FrameSystemAccountInfo`.
   * Which is the most up to date calulcation for transferrable balances.
   *
   * ref: https://github.com/paritytech/polkadot-sdk/issues/1833
   */
  transferable: Balance | null;
  /**
   * Amount locked in vesting.
   */
  vestingLocked: Balance;
}
```

`DeriveBalancesAccountData`: This now adds an optional field `frameSystemAccountInfo` which returns 2 fields (frozen, and flags) that were not present in older `AccountData` formats.

```typescript
export interface DeriveBalancesAccountData {
  frameSystemAccountInfo?: {
    frozen: Balance;
    flags: Balance;
  }
  freeBalance: Balance;
  frozenFee: Balance;
  frozenMisc: Balance;
  reservedBalance: Balance;
  votingBalance: Balance;
}
```

## Transferable Formula

`free - max(frozen - reserve, ed)`

### DIFF
--- a/packages/api-derive/src/balances/account.ts
+++ b/packages/api-derive/src/balances/account.ts
@@ -33,13 +33,13 @@ function getBalance (api: DeriveApi, [freeBalance, reservedBalance, frozenFeeOrF
 
   if (accType.isFrameAccountData) {
     return {
-      freeBalance,
-      frozenFee: api.registry.createType('Balance', 0),
-      frozenMisc: api.registry.createType('Balance', 0),
-      newFrameData: {
+      frameSystemAccountInfo: {
         flags: frozenMiscOrFlags,
         frozen: frozenFeeOrFrozen
       },
+      freeBalance,
+      frozenFee: api.registry.createType('Balance', 0),
+      frozenMisc: api.registry.createType('Balance', 0),
       reservedBalance,
       votingBalance
     };
@@ -49,7 +49,6 @@ function getBalance (api: DeriveApi, [freeBalance, reservedBalance, frozenFeeOrF
     freeBalance,
     frozenFee: frozenFeeOrFrozen,
     frozenMisc: frozenMiscOrFlags,
-    newFrameData: {},
     reservedBalance,
     votingBalance
   };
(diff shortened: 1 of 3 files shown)


joepetrowski:
This is nice since so many people recalculate this. Is it broadcast anywhere?

Nick-1979:
```
let transferable = null;

  if (data.frameSystemAccountInfo?.frozen) {
    const frozenReserveDiff = data.frameSystemAccountInfo.frozen.sub(data.reservedBalance);

    transferable = api.registry.createType('Balance', allLocked ? 0 : data?.freeBalance.sub(bnMax(frozenReserveDiff, api.consts.balances.existentialDeposit)));
  }
  
```
  Based on the above code which is already merged, if "Free" is 7, "Frozen" is 2, "Reserved" is 4 (i.e., reserved > frozen), and "ED" is 1, then the transferable amount should be 7. However, according to the code, it appears to be 6 which is incorrect!
  Am i Right?

TarikGul:
> This is nice since so many people recalculate this. Is it broadcast anywhere?

I haven't thought about broadcasting this, but totally can in channels.

The release should be out today then propagated downstream in the pjs libs to apps.

kianenigma:
> This is nice since so many people recalculate this. Is it broadcast anywhere?

Yes, but I'd actually say this code is a great example to be moved into a runtime api. The runtime knows end of the day how much is transferrable better than anyone else. 

This same code is likely not valid in a context of a runtime other than RCs.

TarikGul:
Yes, I totally agree this should be a runtime-api. I think back to the `TransactionPaymentApi::query_fee_fetails` as a similar example to how the runtime-api really had a 100x positive impact.

PR #5960, opened 2024-08-21
fix: account balance not showing

Imod7:
Closes: https://github.com/polkadot-js/apps/issues/10851

### Issue
When I run in debug mode and I try to load the Account's page in Kusama:

- the [calcShared](https://github.com/polkadot-js/api/blob/72d1326de6a813e2d5e096121fc0b99458bd0e08/packages/api-derive/src/balances/all.ts#L55) function is called multiple times.
- in some of the calls of this function, [data](https://github.com/polkadot-js/api/blob/72d1326de6a813e2d5e096121fc0b99458bd0e08/packages/api-derive/src/balances/all.ts#L59) is not defined.

### Suggested Solution
Add a check to see if `data` is defined and adjust the calculation of `availableBalance` to make use of `data` or not.


Imod7:
Closing in favor of https://github.com/polkadot-js/api/pull/5961

Issue #5965, opened 2024-08-29
Pack `assetId` as `HexString` in `SignerPayloadJSON`


carlosala:
* **I'm submitting a ...**

  - [ ] Bug report
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

The interaction between a decentralized application (dApp) and a Polkadot.js (PJS) wallet extension is facilitated through a [`SignerPayloadJSON`](https://github.com/polkadot-js/api/blob/25ab1d123975a5e25c93de2f2380fe9b321d66eb/packages/types/src/types/extrinsic.ts#L32-L113) object. This object specifies the call data (`method`) and the necessary signed extensions required to reconstruct the extrinsic within the wallet. However, this process presents several challenges:

- The interface of this object is heavily tailored to the conventions used by PJS. For instance, numbers are transmitted as `HexString` and encoded in big-endian format, whereas the broader Substrate ecosystem typically uses little-endian encoding (Simple Concatenated Aggregate Little Endian, or SCALE).
- It lacks flexibility for incorporating new signed extensions.

Given these factors, developing a wallet that does not use the PJS Registry but remains compatible with PJS-based dApps is challenging and involves significant friction. A major issue arises when `assetId` is included in a transaction; it is [formatted as a JSON object](https://github.com/polkadot-js/api/blob/25ab1d123975a5e25c93de2f2380fe9b321d66eb/packages/types/src/extrinsic/SignerPayload.ts#L162) according to PJS's specific structure.

Ideally, `assetId` should be transmitted as a `HexString` using the SCALE encoding expected by the chain. This approach is compatible with current extensions, as PJS Types can handle the encoded format when constructing the types, and would simplify massively the development of new wallets outside of the PJS Registry world.

NB: _ideally all signed extensions should be sent with its SCALE encoding defined in the metadata_

TarikGul:
> Ideally, assetId should be transmitted as a HexString using the SCALE encoding expected by the chain. This approach is compatible with current extensions, as PJS Types can handle the encoded format when constructing the types, and would simplify massively the development of new wallets outside of the PJS Registry world.

Works for me! This totally makes sense - I'll try to get something up soon so it can be released on Sunday into Monday!

carlosala:
Wow, such a fast response. Thanks for your hard word Tarik! 😉

PR #5967, opened 2024-08-31
Change assetId type in SignerPayloadJSON to HexString

TarikGul:
closes: https://github.com/polkadot-js/api/issues/5965

The following is a BREAKING CHANGE

The assetId within the `SignerPayloadJSON` will now be transmitted as a HexString.

### DIFF
--- a/packages/api/src/submittable/createClass.ts
+++ b/packages/api/src/submittable/createClass.ts
@@ -348,7 +348,7 @@ export function createClass <ApiType extends ApiTypes> ({ api, apiType, blockHas
           const ext = this.registry.createTypeUnsafe<Extrinsic>('Extrinsic', [result.signedTransaction]);
           const newSignerPayload = this.registry.createTypeUnsafe<SignerPayload>('SignerPayload', [objectSpread({}, {
             address,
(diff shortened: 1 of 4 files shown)


bee344 (approved):

carlosala:
Hi! Just saw this. I think it could be a great idea to make `asHex` directly with the option, without making `unwrap`. Then we have the exact hexstring that will make part of the final extrinsic. What do you think?

TarikGul:
There is one annoying thing that I am remembering now, its that the `ExtrinsicPayload` type in PJS doesn't know how to handle the AssetId when the Option type is passed in as a hex if one wants to reconstruct the extrinsic... I'll need to fix this today.

carlosala:
I mean, if the type is `Option<MultiLocation>` it should take the option-prepended hex, right?

TarikGul:
@carlosala got a working solution, its a bit hacky, but IMO its the safest option, returns the assetId as it should be (Option<TAssetConversion> = Option<MultiLocation | AssetId> for backwards compatibility), and doesn't break any major workflows other than the actual SignerPayloadJSON.

carlosala:
Sounds good

PR #5968, opened 2024-09-03
Keep assetID hex as Option for toPayload

TarikGul:
ref: https://github.com/polkadot-js/api/pull/5967

This ensures that the `HexString` that is part of the SignerPayloadJSON is an Option.

### DIFF
--- a/packages/types/src/extrinsic/ExtrinsicPayload.ts
+++ b/packages/types/src/extrinsic/ExtrinsicPayload.ts
@@ -36,6 +36,15 @@ function decodeExtrinsicPayload (registry: Registry, value?: GenericExtrinsicPay
     return value.unwrap();
   }
 
+  /**
+   * HACK: In order to change the assetId from `number | object` to HexString (While maintaining the true type ie Option<TAssetConversion>),
+   * to allow for easier generalization of the SignerPayloadJSON interface the below check is necessary. The ExtrinsicPayloadV4 class does not like
+   * a value passed in as an Option, and can't decode it properly. Therefore, we ensure to convert the following below, and then pass the option as a unwrapped
+   * JSON value.
+   *
+   * ref: https://github.com/polkadot-js/api/pull/5968
+   * ref: https://github.com/polkadot-js/api/pull/5967
+   */
   if (value && (value as ExtrinsicPayloadValue).assetId && isHex((value as ExtrinsicPayloadValue).assetId)) {
     const adjustedPayload = {
       ...(value as ExtrinsicPayloadValue),
--- a/packages/types/src/extrinsic/ExtrinsicPayload.ts
+++ b/packages/types/src/extrinsic/ExtrinsicPayload.ts
@@ -11,7 +11,7 @@ import type { ExtrinsicPayloadValue, ICompact, IKeyringPair, INumber, IOption } 
 import type { GenericExtrinsicEra } from './ExtrinsicEra.js';
 
 import { AbstractBase } from '@polkadot/types-codec';
-import { u8aToHex } from '@polkadot/util';
+import { hexToU8a, isHex, u8aToHex } from '@polkadot/util';
 
 import { DEFAULT_VERSION } from './constants.js';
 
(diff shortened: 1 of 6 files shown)


TarikGul:
So I tested this locally to send Assets over xcm with usdt as a asset to pay fees with and I am getting a `wasm trap` error:

```
2024-09-03 18:10:41        RPC-CORE: submitExtrinsic(extrinsic: Extrinsic): Hash:: 1002: Verification Error: Runtime error: Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
WASM backtrace:
error while executing at wasm backtrace:
    0: 0x3f78ac - <unknown>!rust_begin_unwind
    1: 0x2113 - <unknown>!core::panicking::panic_fmt::hbb5a6b42001bdfec
    2: 0x34d3c8 - <unknown>!TaggedTransactionQueue_validate_transaction: RuntimeApi, Execution failed: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\nerror while executing at wasm backtrace:\n    0: 0x3f78ac - <unknown>!rust_begin_unwind\n    1: 0x2113 - <unknown>!core::pa…
```

TarikGul:
The wasm trap issue above is fixed, but fees are not getting payed with usdt, but instead keeps defaulting to ksm...

carlosala (approved):
Very hacky, LGTM! 🚀

packages/types/src/extrinsic/SignerPayload.spec.ts
```diff
@@ -130,11 +134,11 @@ describe('SignerPayload', (): void => {
     expect(payload.blockHash.toHex()).toEqual(TEST.blockHash);
     expect(payload.nonce.eq(TEST.nonce)).toBe(true);
     expect(payload.tip.eq(TEST.tip)).toBe(true);
-    expect(payload.assetId.isSome && payload.assetId
-      ?.eq(registry.createType('MultiLocation', {
+    expect(u8aToHex(payload.assetId?.toU8a()))
```
bee344, line 137: We could just use `.toHex()` and make it simpler:
```
expect(payload.assetId?.toHex())
      .toEqual(registry.createType('Option<MultiLocation>', {
        // eslint-disable-next-line sort-keys
        parents: 0, interior: { X2: [{ palletInstance: 50 }, { generalIndex: 123 }] }
      }).toHex());
      ```

File: packages/types/src/extrinsic/ExtrinsicPayload.ts [hunk 2]
```diff
@@ -36,6 +36,24 @@ function decodeExtrinsicPayload (registry: Registry, value?: GenericExtrinsicPay
...
+   * a value passed in as an Option, and can't decode it properly. Therefore, we ensure to convert the following below, and then pass the option as a unwrapped
+   * JSON value.
+   *
+   * ref: https://github.com/polkadot-js/api/pull/5968
+   * ref: https://github.com/polkadot-js/api/pull/5967
+   */
+  if (value && (value as ExtrinsicPayloadValue).assetId && isHex((value as ExtrinsicPayloadValue).assetId)) {
+    const adjustedPayload = {
+      ...(value as ExtrinsicPayloadValue),
+      assetId: registry.createType('TAssetConversion', hexToU8a((value as ExtrinsicPayloadValue).assetId)).toJSON()
```
bee344, line 51: Is passing it as `JSON` necessary? I can't seem to find why can't we just use the `TAssetConversion`

packages/types/src/extrinsic/SignerPayload.spec.ts
TarikGul: This doesn't work -  `toHex` will unwrap the value if possible... therefore we have to do the above workaround.

packages/types/src/extrinsic/ExtrinsicPayload.ts
TarikGul: Let me check this one, yea we might be able to just pass in the main type!

packages/types/src/extrinsic/SignerPayload.spec.ts
bee344: You are right, it's in the definition `toHex()` in `Option`:
```
public toHex (): HexString {
    // This attempts to align with the JSON encoding - actually in this case
    // the isSome value is correct, however the `isNone` may be problematic
    return this.isNone
      ? '0x'
      : u8aToHex(this.toU8a().subarray(1));
  }
  ```

packages/types/src/extrinsic/ExtrinsicPayload.ts
TarikGul: Ahh yes, the reason I set it to JSON was because the `ExtrinsicPayloadValue` doesn't take any Codec values to begin with, so to align with the current interface, I used a decoded value!

bee344: Makes sense

bee344 (approved):
LGTM

PR #5987, opened 2024-09-19
Set static metadata to v15 and update typegen to generate runtime apis and use v15 metadata

TarikGul:
This PR updates the type generation (`typegen`) process to use the new v15 static metadata format. It also modifies the static runtime-generated calls to utilize v15 metadata when available, ensuring that all runtime TypeScript-augmented types are consistent with the metadata.

Additionally, the metadata generation scripts are now updated to rely on v15 metadata.

Finally, this PR includes an update to the Polkadot `1003003` types.


TarikGul:
Really that last part of this PR is to add any types that the runtimes require now as imports as well. 

AFAIU from the current code, the import type defs are separately parse and generated.

PR #5991, opened 2024-10-01
H160 address parsing issues 

piggydoughnut:
This should solve issues with parsing H160 addresses for Laos Network https://github.com/polkadot-js/apps/issues/10954

### DIFF
--- a/packages/api-derive/src/accounts/accountId.ts
+++ b/packages/api-derive/src/accounts/accountId.ts
@@ -24,7 +24,7 @@ export function accountId (instanceId: string, api: DeriveApi): (address?: Addre
       : decodeAddress((address || '').toString());
 
     if (decoded.length > 8) {
-      return of(api.registry.createType('AccountId', decoded));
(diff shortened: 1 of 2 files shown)


packages/api-derive/src/accounts/accountId.ts
```diff
@@ -24,7 +24,7 @@ export function accountId (instanceId: string, api: DeriveApi): (address?: Addre
       : decodeAddress((address || '').toString());
 
     if (decoded.length > 8) {
-      return of(api.registry.createType('AccountId', decoded));
+      return of(api.registry.createType(decoded.length === 20 ? 'AccountId20' : 'AccountId', decoded));
```
TarikGul, line 27: I wonder if this would be good to use `encodedLength` as well.

TarikGul (approved):

packages/api-derive/src/accounts/accountId.ts
TarikGul: Ahhh this actually doesn't make sense since `encodedLength` is only available for Codec classes

PR #5997, opened 2024-10-10
fix: lru ttl cache implementation

filvecchiato:
Implements a TTL in the LRU cache used to cache responses in http and ws providers. It allows to also bypass querying the cache in case the users sets the cache size to 0.


TarikGul:
I think the tests are hanging.

filvecchiato:
> I think the tests are hanging.

Checking them now

packages/rpc-core/src/bundle.ts
```diff
@@ -487,15 +490,15 @@ export class RpcCore {
       : u8aToU8a(value);
     const codec = this._newType(registry, blockHash, key, input, isEmpty, entryIndex);
 
-    // store the retrieved result - the only issue with this cache is that there is no
-    // clearing of it, so very long running processes (not just a couple of hours, longer)
-    // will increase memory beyond what is allowed.
-    this.#storageCache.set(hexKey, codec);
-    this.#storageCacheSize++;
```
TarikGul, line 494: Since cache size is being removed here it makes the `stats` call irelevant. this should be added in `-setToCache`

packages/rpc-provider/src/lru.ts
```diff
@@ -28,10 +41,29 @@ export class LRUCache {
   #length = 0;
   #head: LRUNode;
   #tail: LRUNode;
+  // TTL
```
TarikGul, line 44: I think we can remove this comment, the code is pretty self documenting

TarikGul (changes requested):
Some nits above. But overall really awesome. thank you 🚀

packages/rpc-core/src/bundle.ts
filvecchiato: I updated the stats returned using the LRU cache size

packages/rpc-provider/src/lru.ts
```diff
@@ -115,14 +144,39 @@ export class LRUCache {
         this.#length += 1;
       }
     }
+    
+    if (this.#ttl > 0 && !this.#ttlP) {
```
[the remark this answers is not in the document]
filvecchiato: this is a bit outdated, i renamed the variable #ttlTimerId to make it clearer

TarikGul (approved):
lgtm, great job 🚀

Issue #6010, opened 2024-10-28
LRU cache is not releasing thread on api.disconnect() if you've written to the chain.


sicco-moonbeam:
* **I'm submitting a ...**

  - [x ] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

In version 14.1.1, Node.js retains a thread after writing to the chain, even when api.disconnect() is called. While reading chain state behaves as expected, any write operation—such as a simple balances.transfer—prevents the Node.js process from terminating automatically. The attached WTFNode output shows that the hanging process originates in node_modules\@polkadot\rpc-provider\cjs\lru.js:115.

I'm not entirely sure when this was introduced as I was previously on 10.11.2 but I suspect it's fairly recent - I noticed references to changes to the LRU cache in the last few commits.

* **What is the motivation for changing the behavior?**

api.disconnect() should gracefully release all resources. In a long-running process, this would result in a memory leak.

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [x] Node.js - tested on latest stable version (v23.1.0) but can reproduce on older versions
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript (include tsc --version) - 5.6.3
    - [ ] Other

TarikGul:
Yes this would be related to https://github.com/polkadot-js/api/pull/5997

cc: @filvecchiato

xlc:
I just spent 2 hours debugging why chopsticks no longer exit cleanly... Please add some regression tests.

TarikGul:
> I just spent 2 hours debugging why chopsticks no longer exit cleanly... Please add some regression tests.

Yes, we are actively getting this fixed ASAP and adding tests!

TarikGul:
PR is in and a patch release will be up soon

PR #6013, opened 2024-10-31
fix: update LRUcache to not keep thread

filvecchiato:
removes the use of intervals to evict the cache and instead evicts on read/write operations. This should stop blocking the thread


TarikGul (approved):
Thank you :)

fixes: https://github.com/polkadot-js/api/issues/6010

Issue #5884, opened 2024-05-16
Augmented types for system parachains


TarikGul:
We are currently seeing the identity pallet migrated from the relay chains to the people (systems) parachain. This also means any new types in the identity pallet that are added to [people](https://github.com/paritytech/polkadot-sdk/tree/master/cumulus/parachains/runtimes/people) runtime, will not be part of the augmented types that are generated in the API for TS enviornments since these updated types wont be apart of the substrate metadata.

We should consider also potentially augmenting these types for system parachains. It could improve a lot of the UX for developing with these new workflows.

TarikGul:
As AH-Plaza comes to fruition this will be necessary as well. Adding this sooner than later will also help make the transition of all types seemless.

TarikGul:
- [ ] Metadata get scripts need to include AH
- [ ] Add types-support metadata for AH
- [ ] Write to file Tests that add to types-support
- [ ] Typegen from metadata

TarikGul:
This is now completed. There is no need to typegen the types into the api, since the substrate types will contain any base type information that is needed from frame.

PR #6023, opened 2024-11-08
Apply AssetHub type gen to augmentations

TarikGul:
rel: https://github.com/polkadot-js/api/issues/5884

This PR adds assetHub kusama and polkadot to the type and api augmentation.


TarikGul:
There are some certain types that are not making it into the imports which need to be added. This problem originates at the typegen level.

TarikGul:
I think this PR can hold off for a second as the initial PR (https://github.com/polkadot-js/api/pull/6022) has support for the types that will be available to add in the docs - but currently the types conflict with the substrate types since they are not inline runtime wise.

TarikGul:
Closing as the base substrate types will suffice.

Issue #6017, opened 2024-11-05
Tests fail for node v22 LTS


TarikGul:
Currently all tests will fail for node version v22 LTS. This was originally caught in the CI

PR #6024, opened 2024-11-08
Fix node v22 testing compatibility

TarikGul:
closes: https://github.com/polkadot-js/api/issues/6017


TarikGul:
This requires a fix in dev to modifiy the source to correctly set the json assertions for testing

Issue #6032, opened 2024-11-20
parsing assetId from a json payload fails since v13.0.1


0xKheops:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

From a user perspective, on https://kheopswap.xyz/#/polkadot/transfer the transaction is rejected if user selecting a fee asset (ex: USDT) to pay for fees, if using Talisman (or polkadot.js extension) as wallet. 

Note: we upgraded `@polkadot/*` dependencies in Talisman few weeks back.
With a build from before that upgrade, signing those transactions was working fine.

I identified that since `@polkadot/api` v13, the assetId value cannot be parsed from a signer json payload anymore. This was working fine up to version v12.4.2

 

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**

Kheopswap is developed with PAPI, which is used to generate the JSON payload used in the example below.
Tagging @josepot just in case.

I don't know any other dapp that allows selecting the fee asset so I've only tested it there.

I've prepared this repo to demonstrate the issue : https://github.com/0xKheops/issue-pjs-ah-signing
Based on the version of `@polkadot/api` the assetId can be parsed or not. 
v12.4.2 or below: OK
v13.0.1 or above: assetId can't be parsed

![image](https://github.com/user-attachments/assets/5a7bd30d-2ca7-4353-bc68-434fcaec78bf)

  - Version:
  - Environment:

    - [x] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript (include tsc --version)
    - [ ] Other

josepot:
I'd like to take this opportunity to point out -once again- that the `SignerPayloadJSON` interface has been causing significant challenges and is hindering the ecosystem's progress. It would be beneficial to replace this interface with a non-leaky, library-agnostic alternative, as suggested in this [GitHub comment](https://github.com/paritytech/substrate-connect/issues/2468#issuecomment-2393163037).

[...]

PR #6037, opened 2024-11-25
fix(ExtrinsicPayload): decode assetId with and without option

carlosala:
#5968 introduced an unexpected breaking change since, at the moment, the `ExtrinsicPayload` type can only parse hex assetId with the Option byte. This PR addresses it, making it backwards compatible, and now all ways (hex with and without option, PJS object, and number) are supported.
This should fix as well #6032.
I'm truly sorry about the confusion I might have created :(
Thanks!


TarikGul (commented):
Thanks for the PR - much appreciated.

Just one comment!

TarikGul (approved):
👍 Thanks so much 🚀

Issue #5810, opened 2024-03-07
`api.derive.accounts.identity` called with SuperIdentity doesn't return SubIdentities


wpank:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Calling `api.derive.accounts.identity` gives some slightly unexpected behaviour - when called with the super / parent identity, it doesn't return the sub accounts, although calling with with one of the sub identities it returns the full thing, with all the sub identities / parent account. I want to say it used to be the case that this call returned sub-identities

https://github.com/w3f/1k-validators-be/blob/master/packages/common/src/chaindata/queries/Identity.ts#L91

This is how the call is being used in the 1kv-backend - I needed to add in the additional sub identity query.

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [x] Node.js
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript (include tsc --version)
    - [ ] Other

TarikGul:
I did just add some small compatibility fixes with `api.derive.accounts.identity` when increasing the substrate metadata to the latest release (This will be included in next weeks release), but I need to see if there is any correlation or what the root of the issue is.

PR #6066, opened 2025-01-14
Allow api.derive.accounts.identity() to obtain SubIdentities

valentinfernandez1:
This PR addresses Issue #5810 by introducing a new subs field to the DeriveAccountRegistration structure, allowing for the retrieval of SubIdentities associated with an account.

### Retrieval of SubIdentities:
- If the account is a primary Identity, its SubIdentities are directly retrieved.
- If the account is a SubIdentity, the parent account's SubIdentities are retrieved, and the queried account is included in the resulting list.

### DIFF
--- a/packages/api-derive/src/accounts/types.ts
+++ b/packages/api-derive/src/accounts/types.ts
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { AccountId, AccountIndex, RegistrationJudgement } from '@polkadot/types/interfaces';
+import type { Vec } from '@polkadot/types-codec';
 
 export type AccountIdAndIndex = [AccountId | undefined, AccountIndex | undefined];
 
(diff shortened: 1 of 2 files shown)


packages/api-derive/src/accounts/types.ts
```diff
@@ -18,6 +18,7 @@ export interface DeriveAccountRegistration {
   matrix?: string | undefined;
   other?: Record<string, string> | undefined;
   parent?: AccountId | undefined;
+  subs?: AccountId[] | undefined;
```
TarikGul, line 21: Is the actual type `Vec<AccountId>` returned from `subsResponse[1];` or `AccountId[]`?

TarikGul (changes requested):
Overall, amazing job just one question

TarikGul (commented):
Needs a `yarn lint` as well.

TarikGul (approved):
LGTM 🚀

Issue #5697, opened 2023-07-27
API accepting undefined destination for transfer


bee344:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Currently when calling `api.tx.balances.transfer` and leaving the destination address as `undefined`, the funds are taken from the sender's account and sent to the address `111111111111111111111111111111111HC1`. In these cases the transaction should fail and alert the sender they should input the destination address.

* **What is the motivation for changing the behavior?**

Now the funds sent this way are being lost, since trying to retrieve them using `keyring.createFromUri("0x Hex")` to recreate the private key of the `undefined` address and sending the funds back to the original address returns:
```
"error": "1010: Invalid Transaction: Inability to pay some fees , e.g. account balance too low"
```
* **Please tell us about your environment:**

  - Version: api v10.9.1
  - Environment:

    - [x] Node.js
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript (include tsc --version) v4.9.4
    - [ ] Other

valentinfernandez1:
This behavior is intentional: setting the address as `undefined`, `""`, or `[]` all represent the same concept—the zero account (`0x` / `0x00...00`). The reason why it was sent to `111111111111111111111111111111111HC1` in your case is because that is how `0x000...000` is represented in Polkadot.

![Image](https://github.com/user-attachments/assets/ae3da16f-f5c2-4596-948c-97fc6691c3b0)

![Image](https://github.com/user-attachments/assets/90dfa006-1a3c-4e48-a045-09176d788283)

I explored the possibility of modifying this behavior to disallow undefined as a value. However, this change introduced significant breaking issues downstream, as many functionalities depend on the current implementation.

To make this clearer, we can expand the in-code documentation to explicitly describe this behavior.

valentinfernandez1:
You can actually verify this yourself with the following code:
```typescript
import { ApiPromise, WsProvider } from "@polkadot/api";
const CHAIN_ENDPOINT = "wss://polkadot-rpc.publicnode.com";

const main = async () => {
  const api = await ApiPromise.create({
    provider: new WsProvider(CHAIN_ENDPOINT),
  });

  await api.isReady;

  // -- Create Account --
  // using undefined:
  const accUndefined = api.registry.createType("AccountId", undefined);
  // using []:
[...]

PR #6072, opened 2025-01-21
Extend `GenericAccountId` description

valentinfernandez1:
When creating an Account using an empty value such as `undefined`, `[]`, `""`, etc. The resulting account is "0x00..000". 

This PR further clarifies it by adding this description to the `GenericAccountId` documentation to avoid confusion. Should address #5697


TarikGul (approved):
Amazing 👍

Issue #5977, opened 2024-09-09
Typegen adding invalid import


TarikGul:
rel: https://github.com/polkadot-js/api/pull/5972

Currently the typegen for 1.3.0 for Kusama is adding `StagingXcmV4Xcm` to the imports of `query.ts` when there is no use of the Import. This is causing typescript build issues. I am not able to find the source of the issue as the code is quite complicated. Due to this being a relatively harmless issue, I am adding a build rule to ignore it!

That being said this issue should be fixed, and the bug can be replicated by removing the `"noUnusedLocals": false` rules in the tsconfig compiler options for the builds.

TarikGul:
We found that the type assignment in the typegen has different lookupName vs type:

ex
```
 {
  info: 19,
  lookupIndex: 400,
  lookupName: 'StagingXcmV4Xcm',
  type: 'Vec<StagingXcmV4Instruction>',
  docs: [],
  namespace: '',
  sub: {
    docs: [],
    info: 15,
    lookupIndex: 402,
    lookupName: 'StagingXcmV4Instruction',
    type: 'Lookup402'
  },
  lookupNameRoot: undefined,
  typeName: 'Vec<Instruction>'
}
```

Which is causing different import vs assignment for `RecordXcm`

TarikGul:
You can replicate the above output by inserting:

```
      if (typeDef.lookupName === 'StagingXcmV4Xcm' || typeDef.type === 'StagingXcmV4Xcm') {
        console.log('TYPEDEF: ', lookup.getTypeDef(storageEntry.type.asPlain))
      }
```

in https://github.com/polkadot-js/api/blob/master/packages/typegen/src/generate/query.ts#L31

PR #6073, opened 2025-01-23
Patch typegen query to prevent it from adding invalid import

valentinfernandez1:
This PR implements a temporary workaround for issue #5977. Currently, there's a bug where the `StagingXcmV4Xcm` type is incorrectly assigned in `query.rs`. While the root cause needs investigation, this PR avoids import errors by temporarily ignoring the type. This allows us to renable `"noUnusedLocals": true` in `tsconfig.json`

**Note:** This is a temporary solution. A proper fix for #5977 should be implemented in a future PR.


TarikGul (approved):
Amazing! Thanks so much

Issue #6074, opened 2025-01-23
Decode of `assetId` hex string fails for signed payload because of `Option`


Chralt98:
```
Error: Struct: failed on assetId: Option<MultiLocationV3>:: decodeU8aStruct: failed at 0x010300a10f043205e514… on interior (index 2/2): {"_enum":{"Here":"Null","X1":"JunctionV3","X2":"(JunctionV3,JunctionV3)","X3":"(JunctionV3,JunctionV3,JunctionV3)","X4":"(JunctionV3,JunctionV3,JunctionV3,JunctionV3)","X5":"(JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3)","X6":"(JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3)","X7":"(JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3)","X8":"(JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3,JunctionV3)"}}:: decodeU8aStruct: failed at 0xa10f043205e514… on key (index 2/2): [u8;20]:: Expected input with 20 bytes (160 bits), found 7 bytes
```

In the above error it expects `AccountKey20` (derived from `[u8;20]`). The index of this is `3` in `JunctionV3`. The `assetId` hex string always starts with `0x0103`, in which `0x01` is used for `parent: 1` and `0x03` should be used for `X3` of `JunctionsV3`. However, the `0x03` occurs to be used not only for `JunctionsV3` but also for the first entry into the tuple of `X3`, which is `JunctionV3`. So, it seems to be like a mistaken overlap for the decode functionality.

![Image](https://github.com/user-attachments/assets/42d06956-1b91-4029-9ee2-15b489194be6)

I want to use the following `assetId` (which is `USDC` on AssetHub for our Zeitgeist Parachain):

```javascript
let parentsNumber = api.registry.createType('u8', 1);
let parachainCompactNumber = api.registry.createType('Compact<u32>', 1000);
// let parachain = api.registry.createType('JunctionV3', { Parachain: parachainCompactNumber });
let palletInstanceNumber = api.registry.createType('u8', 50);
// let palletInstance = api.registry.createType('JunctionV3', { PalletInstance: palletInstanceNumber });
let generalIndexNumber = api.registry.createType('Compact<u128>', 1337);
[...]

PR #6081, opened 2025-01-28
Update signer payload known types to include `TAssetConversion` for assetId

Chralt98:
Fixes https://github.com/polkadot-js/api/issues/6074

https://github.com/polkadot-js/api/blob/ad6f3d45313b7ecc3f7f55a855789d960cf59981/packages/types/src/interfaces/assetConversion/definitions.ts#L15

In the above `TAssetConversion` is `Option<MultiLocation>`, but should `assetId` be `Option<TAssetConversion>` like [here](https://github.com/polkadot-js/api/blob/ad6f3d45313b7ecc3f7f55a855789d960cf59981/packages/types/src/extrinsic/SignerPayload.ts#L32)? In this case we would have `Option<Option<MultiLocation>>` in the end, right?

### DIFF
--- a/packages/types/src/extrinsic/SignerPayload.spec.ts
+++ b/packages/types/src/extrinsic/SignerPayload.spec.ts
@@ -130,7 +130,7 @@ describe('SignerPayload', (): void => {
     expect(payload.nonce.eq(TEST.nonce)).toBe(true);
     expect(payload.tip.eq(TEST.tip)).toBe(true);
     expect(u8aToHex(payload.assetId?.toU8a()))
-      .toEqual(u8aToHex(registry.createType('Option<MultiLocation>', {
+      .toEqual(u8aToHex(registry.createType('MultiLocation', {
         // eslint-disable-next-line sort-keys
         parents: 0, interior: { X2: [{ palletInstance: 50 }, { generalIndex: 123 }] }
       }).toU8a()));
--- a/packages/types/src/extrinsic/SignerPayload.spec.ts
+++ b/packages/types/src/extrinsic/SignerPayload.spec.ts
@@ -130,7 +130,7 @@ describe('SignerPayload', (): void => {
     expect(payload.nonce.eq(TEST.nonce)).toBe(true);
     expect(payload.tip.eq(TEST.tip)).toBe(true);
     expect(u8aToHex(payload.assetId?.toU8a()))
-      .toEqual(u8aToHex(registry.createType('MultiLocation', {
+      .toEqual(u8aToHex(registry.createType('Option<MultiLocation>', {
         // eslint-disable-next-line sort-keys
         parents: 0, interior: { X2: [{ palletInstance: 50 }, { generalIndex: 123 }] }
       }).toU8a()));
--- a/packages/types/src/extrinsic/SignerPayload.spec.ts
+++ b/packages/types/src/extrinsic/SignerPayload.spec.ts
@@ -36,11 +36,6 @@ describe('SignerPayload', (): void => {
     withSignedTransaction: false
   };
 
-  const TEST_WITH_ASSETID_HEX = {
-    ...TEST,
-    assetId: '0x010002043205ed01'
-  };
-
   it('creates a valid JSON output', (): void => {
     expect(
       new SignerPayload(registry, {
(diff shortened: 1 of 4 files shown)


TarikGul:
So AFAIU, the reason TAssetConversion is wrapped in an Option is because the it's an optional input for the Extrinsic class in PJS. In addition to that the Option type that the MultiLocation is wrapped in is exactly what the chain expects there it needs to be wrapped in an option when converted to it byte value.

^ I need to double check the above as well though.

Chralt98:
What I find confusing though, is the `metadataHash` example.

It uses an `Option` too [here](https://github.com/polkadot-js/api/blob/ad6f3d45313b7ecc3f7f55a855789d960cf59981/packages/types/src/extrinsic/SignerPayload.ts#L37). But in the test, the `metadataHash` doesn't change through calling `toPayload` from [here](https://github.com/polkadot-js/api/blob/ad6f3d45313b7ecc3f7f55a855789d960cf59981/packages/types/src/extrinsic/SignerPayload.spec.ts#L54) to [here](https://github.com/polkadot-js/api/blob/ad6f3d45313b7ecc3f7f55a855789d960cf59981/packages/types/src/extrinsic/SignerPayload.spec.ts#L71). So, why should the hex value of `TAssetConversion` change?

Chralt98:
I mean [this override](https://github.com/polkadot-js/api/blob/ad6f3d45313b7ecc3f7f55a855789d960cf59981/packages/types/src/interfaces/assetConversion/definitions.ts#L15) alone will replace the `knownTypes` including [`TAssetConversion`](https://github.com/polkadot-js/api/blob/ad6f3d45313b7ecc3f7f55a855789d960cf59981/packages/types/src/extrinsic/SignerPayload.ts#L32). So, why not using the default override [`TAssetConversion`](https://github.com/polkadot-js/api/blob/ad6f3d45313b7ecc3f7f55a855789d960cf59981/packages/types/src/interfaces/assetConversion/definitions.ts#L15) to be just `MultiLocation` instead of `Option<MultiLocation>`?

Chralt98:
Ok, with the latest commit I get: `Bad input data provided to validate_transaction: Codec error` for `RPC-CORE: submitAndWatchExtrinsic(extrinsic: Extrinsic): ExtrinsicStatus:: 1011: Error: wasm `unreachable` instruction executed`.

TarikGul:
This `Option` being part of the hex is intended, in addition to that we don't want to add any breaking changes to the API, definitely not with the Signer.

Example that its working as intended:

```typescript
    const signer = new GenericSignerPayload(api.registry, {
        address: '5DTestUPts3kjeXSTMyerHihn1uwMfLj8vU8sqF7qYrFabHE',
        assetId: { parents: 0, interior: { x2: [{ palletInstance: 50 }, { generalIndex: 123 }] } },
        blockHash: '0xde8f69eeb5e065e18c6950ff708d7e551f68dc9bf59a07c52367c0280f805ec7',
        blockNumber: '0x231d30',
        era: api.registry.createType('ExtrinsicEra', { current: 2301232, period: 200 }),
        genesisHash: '0xdcd1346701ca8396496e52aa2785b1748deb6db09551b72159dcb3e08991025b',
        metadataHash: '0xdcd1346701ca8396496e52aa2785b1748deb6db09551b72159dcb3e08991025b',
        method: api.tx.system.remark('0x00').method,
        mode: 1,
        nonce: 0x1234,
        signedExtensions: ['CheckNonce'],
        tip: 0x5678,
        version: 4,
        withSignedTransaction: true
    });
    
    console.log(signer.assetId.toHex())
```

This returns:

```json
{
  specVersion: '0x00000000',
  transactionVersion: '0x00000000',
  address: 'DyGYCpGXE4gVJLtE4nhkF5hvcCBAL9ubJJtGVX5KM4kKpaW',
  assetId: '0x010002043205ed01',
  blockHash: '0xde8f69eeb5e065e18c6950ff708d7e551f68dc9bf59a07c52367c0280f805ec7',
  blockNumber: '0x00231d30',
  era: '0x0703',
  genesisHash: '0xdcd1346701ca8396496e52aa2785b1748deb6db09551b72159dcb3e08991025b',
  metadataHash: '0xdcd1346701ca8396496e52aa2785b1748deb6db09551b72159dcb3e08991025b',
  method: '0x00000400',
  mode: 1,
  nonce: '0x00001234',
  signedExtensions: [ 'CheckNonce' ],
  tip: '0x00000000000000000000000000005678',
  version: 4,
  withSignedTransaction: true
}
```

I'll comment further in the issue.

Chralt98:
Closed, because this PR https://github.com/polkadot-js/api/issues/6074#issuecomment-2624361378 fixed the issue.

PR #6082, opened 2025-01-29
Fix typegen adding invalid imports

valentinfernandez1:
This PR addresses #5977 by tracking the lookup import types on each file and then filtering out the ones that are unused. 

**Why?**

Previously, the type generation scripts could add unnecessary or invalid imports that were never referenced in the generated file. This change ensures that only the required lookup types remain in the `imports` object used for file generation


TarikGul (approved):
Nice job! 🚀

Issue #5746, opened 2023-11-01
missing known keys


xlc:
Not all the known keys are configured here

https://github.com/polkadot-js/api/blob/6fcd984205e302a60b3242a568bf791c452be869/packages/types/src/metadata/decorate/storage/substrate.ts#L20-L39

For a hopefully completed list, see this issue: https://github.com/paritytech/polkadot-sdk/issues/2126

TarikGul:
Addressed in #6084

PR #6084, opened 2025-01-30
Include missing known keys

valentinfernandez1:
Addresses #5746 by including some missing known substrate storage keys.


TarikGul (approved):
👍

PR #6089, opened 2025-02-07
Ensure rpc-core can correctly decode types, even if multiple types share the same name

martin-chainflip:
Currently, rpc-core tries to decode the results of a storage query by using the type name. However, it is perfectly valid to have multiple, different types in the type registry that share the same name. In this case rpc-core will fail to decode. This change uses the unique type id for decoding if available instead. 
This change fixes https://github.com/polkadot-js/apps/issues/10432

### DIFF
--- a/packages/rpc-core/src/bundle.ts
+++ b/packages/rpc-core/src/bundle.ts
@@ -13,6 +13,7 @@ import { Observable, publishReplay, refCount } from 'rxjs';
 
 import { LRUCache } from '@polkadot/rpc-provider';
 import { rpcDefinitions } from '@polkadot/types';
+import { unwrapStorageSi } from '@polkadot/types/util';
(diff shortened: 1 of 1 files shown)


TarikGul:
Hey thanks for the PR. Were you able to check in apps if it works?

The way you can link the api with your changes is by having apps share the same root directory as the api - and running `yarn polkadot-dev-copy-to apps`.

TarikGul:
I checked this blockhash: `0xf7d516bd72ee1d76903e692157fc2a5260788faa6c9b6a369c3006b19413b6d2` given in the issue and still received the same decoding error as earlier with the applied changes above.

martin-chainflip:
you are correct. this only fixes the issue that storage items with the same type name could not be decoded. for calls/extrinsics, more work is needed.

martin-chainflip:
would it still be possible to merge this fix?

TarikGul:
cc: @ap211unitech @valentinfernandez1

ap211unitech (approved):

PR #6092, opened 2025-02-11
Only identify Struct as empty when it has no keys.

### LINKED ISSUE #6090: ValidatorPrefs.isEmpty returns true for validator with 0% commission
* **I'm submitting a ...**
  - [ ] Bug report
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

Current behavior:
When querying validator preferences using `api.query.staking.validators`, the returned `ValidatorPrefs` struct incorrectly reports `isEmpty: true` when the validator has valid preferences set with 0% commission.

Reproduction:
```typescript
const blockNumber = 24599609;
const validatorAddress = '1ChAR84s9n6ZTYGSJqkKizzdnJBarK85bxficpxrd8CkGt4';

const apiAt = await api.at(await api.rpc.chain.getBlockHash(blockNumber));
const prefs = await apiAt.query.staking.validators(validatorAddress);

console.log('Debug output:', {
  rawPrefs: prefs.toHuman(),
  isEmpty: prefs.isEmpty,
  commissionRaw: prefs.commission.toNumber(),
  encodedLen: prefs.encodedLength,
});
```

Output shows the contradictory state:
```json
{
  "rawPrefs": {
    "commission": "0.00%",
    "blocked": false
  },
  "isEmpty": true,
  "commissionRaw": 0,
  "encodedLen": 2
}
```

The struct reports `isEmpty: true` despite:
- Having valid commission data (0%)
- Having a valid encodedLength of 2
- All fields being accessible with valid values

Expected behavior:
`isEmpty` should return `false` when the struct contains valid data, even if that data represents a 0% commission. These are two distinct states:
- No preferences set = validator hasn't registered
- 0% commission = validator is registered but chose to not take commission

We can work around this issue in our code by not relying on `isEmpty`, but I think current API behavior is misleading and incorrect from semantic perspective.

  - Version: @polkadot/api ^15.5.2
  - Environment:
    - [x] Node.js
    - [ ] Browser
    - [ ] Other

  - Language:
    - [ ] JavaScript
    - [x] TypeScript

valentinfernandez1:
Addresses #6090. If fixes an issue where Struct.isEmpty() would return true for nonEmpty objects. 

This changes makes it so its in line with the behavior with [Json.isEmpty()](https://github.com/polkadot-js/api/blob/7bd4c7afb0441cc427e41d896a5cd48f6ff70143/packages/types-codec/src/native/Json.ts#L58).

closes #6090

### DIFF
--- a/packages/types-codec/src/native/Struct.ts
+++ b/packages/types-codec/src/native/Struct.ts
@@ -154,16 +154,14 @@ export class Struct<
   }
 
   /**
-   * @description Checks if the value is an empty value
+   * @description Checks if the value is an empty value '{}'
    */
   public get isEmpty (): boolean {
-    for (const v of this.values()) {
-      if (!v.isEmpty) {
-        return false;
-      }
+    if (Array.from(this.values()).length === 0) {
+      return true;
     }
 
-    return true;
+    return false;
   }
 
   /**
--- a/packages/types-codec/src/native/Struct.ts
+++ b/packages/types-codec/src/native/Struct.ts
@@ -157,11 +157,7 @@ export class Struct<
    * @description Checks if the value is an empty value '{}'
    */
   public get isEmpty (): boolean {
-    if (Array.from(this.values()).length === 0) {
-      return true;
-    }
-
-    return false;
+    return [...this.keys()].length === 0;
   }
 
   /**


TarikGul (approved):
Nice 👍

PR #6093, opened 2025-02-14
fix: check for api.query.identity in getSubIdentities

ap211unitech:
## 📝 Description

This PR adds a check before fetching sub-identities for an account, ensuring that the required storage pallet exists to prevent failures.


valentinfernandez1 (approved):
lgtm

PR #6107, opened 2025-02-27
Fix derives base path in typegen

valentinfernandez1:
Fixes an issue where the `polkadot-types-internal-metadata` script would fail when executed outside the project directory without the project being present in the same root folder.

This should resolve problems on polkadot-js/docs CI

### DIFF
--- a/packages/typegen/src/metadataMd.ts
+++ b/packages/typegen/src/metadataMd.ts
@@ -13,6 +13,8 @@ import path from 'node:path';
 import process from 'node:process';
 import yargs from 'yargs';
 import { hideBin } from 'yargs/helpers';
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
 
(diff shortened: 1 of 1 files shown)


packages/typegen/src/metadataMd.ts
```diff
@@ -563,15 +563,15 @@ function addErrors (runtimeDesc: string, { lookup, pallets }: MetadataLatest): s
   });
 }
 
-const BASE_DERIVE_PATH = '../api/packages/api-derive/src/';
+const BASE_DERIVE_PATH = 'node_modules/@polkadot/api-derive/';
```
TarikGul, line 566: Better to maybe use the built in `node:path` here?

valentinfernandez1: done

TarikGul:
Just needs a lint

TarikGul (approved):
🚀

Issue #6111, opened 2025-03-03
Simplify Retrieval of Extrinsic Details


valentinfernandez1:
As pointed out on #6109, retrieving details about a specific extrinsic (e.g., given its transaction hash) is not straightforward in PolkadotJS API. While you can obtain this information by querying the block and filtering extrinsics manually, this process is inefficient and cumbersome.

**Problem:**
If a user has an extrinsic's transaction hash or is interested in querying extrinsic details at a known block, there is no simple API function to retrieve it directly. Instead, the current approach requires:

- Fetching the block using api.rpc.chain.getBlock(blockHash).
- Iterating over all extrinsics in that block.
- Matching the extrinsic hash manually.

The same process applies for a user that is looking to query all transactions that it has done in a specific block.

**Solution:**

This process could be simplified using dedicated API Derive functions under the `api.derives.tx` namespace. 
- `extrinsicInfo(blockHash, extrinsicHash)`: Returns information of a specific extrinsic (hash, status, emittedEvents, account).
- `accountTxs(blockHash, accountId)`: Returns information about every extrinsic executed by an account at a block or set of blocks.

valentinfernandez1:
closed by 6113

PR #6113, opened 2025-03-04
Add derives to obtain Extrinsic Information

valentinfernandez1:
This PR addresses issue #6111 by adding two new derives to the `api.derives.tx` namespace:
- `extrinsicInfo(at, transactionHash)`: Retrieves the extrinsic information and its events.
- `accountExtrinsics(at, accountId)`: Retrieves information about every extrinsic submitted by an account at a given block.

This greatly simplifies the process of obtaining extrinsic information as the related events are directly associated with the extrinsic. Allowing the user to easily check what happened with an extrinsic as well as whether it was successful or not


TarikGul (approved):
Looks really good 🚀

PR #6116, opened 2025-03-14
Throws detailed error information

zjb0807:
Abnormal exit, which is difficult to debug, so add detailed information logs
```
/app/node_modules/@polkadot/api/cjs/base/Init.js:152
            throw new Error('Unable to retrieve header and parent from supplied hash');                                                                                                                                                 ^
                                                                                                                                                                                                                      Error: Unable to retrieve header and parent from supplied hash
    at ApiPromise._getBlockRegistryViaHash (/app/node_modules/@polkadot/api/cjs/base/Init.js:152:19)                                                                                                                      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
                                                                                                                                                                                                                      Node.js v18.20.5
```


valentinfernandez1 (approved):
Amazing work, thank you for the contribution!! 🚀

valentinfernandez1:
Amazing work 🚀 , some checks are failing please resolve them and we should be able to merge. Make sure that `yarn build` runs successfully and run `yarn lint` to ensure proper code formatting.

TarikGul (changes requested):
In addition to the comment above, you should not add an object to a template literal.

Overall, I think this should be changed to a debug log, therefore not adding to bloat current logs.

something like:

```
l.debug('foo: ', bar);
```

zjb0807:
@TarikGul, please review again. Thank you

packages/api/src/base/Init.ts
```diff
@@ -200,6 +200,7 @@ export abstract class Init<ApiType extends ApiTypes> extends Decorate<ApiType> {
     );
 
     if (header.parentHash.isEmpty) {
+      l.warn(`Unable to retrieve header ${blockHash.toString()} and parent ${header.parentHash.toString()} from supplied hash`);
```
TarikGul, line 203:
```suggestion
      l.debug(`Unable to retrieve header ${blockHash.toString()} and parent ${header.parentHash.toString()} from supplied hash`);
```

The if you do `export DEBUG=api/init` it will show up.

TarikGul (approved):
Just a suggestion but if @valentinfernandez1 is okay with it, then I am.

valentinfernandez1 (approved):

PR #6121, opened 2025-03-24
Fix signAsync() docs

### LINKED ISSUE #6118: Misleading documentation for `signAsync()`
In [packages/api-base/src/types/submittable.ts:71](https://github.com/polkadot-js/api/blob/852cbeb12357a2036d67c1bf3cf9fa9483a27a91/packages/api-base/src/types/submittable.ts#L71)
```ts
  /**
   * @description Sign and broadcast the constructued transaction.
   *
   * Note for injected signers:
   * As of v12.0.2 and up the `SignerResult` return type for `signPayload` allows for the `signedTransaction` field.
   * This allows the signer to input a signed transaction that will modify the payload. This
   * The api will ensure that the Call Data is not changed. This allows for the signer to modify the payload to add
   * things like `mode`, and `metadataHash` for signedExtensions such as `CheckMetadataHash`.
   */
  signAsync (account: AddressOrPair, _options?: Partial<SignerOptions>): PromiseOrObs<ApiType, this>;
```

`signAsync()` does not in fact broadcast the transaction, as far as I understand it.

valentinfernandez1:
The `signAsync()` docs stated that the transaction was broadcasted after executing this method which is incorrect. Fixed the documentation to properly reflect the function's behavior.

closes #6118.

### DIFF
--- a/packages/api-base/src/types/submittable.ts
+++ b/packages/api-base/src/types/submittable.ts
@@ -68,13 +68,17 @@ export interface SubmittableExtrinsic<ApiType extends ApiTypes, R extends ISubmi
   send (statusCb: Callback<R>): SubmittableResultSubscription<ApiType, R>;
 
   /**
-   * @description Sign and broadcast the constructued transaction.
+   * @description Sign the constructed transaction asynchronously.
+   *
+   * The result is a signed extrinsic that is ready to be broadcast to the network via `.send()`, `rpc.author.submitExtrinsic()`, or
+   * any custom submission logic.
    *
    * Note for injected signers:
-   * As of v12.0.2 and up the `SignerResult` return type for `signPayload` allows for the `signedTransaction` field.
-   * This allows the signer to input a signed transaction that will modify the payload. This
-   * The api will ensure that the Call Data is not changed. This allows for the signer to modify the payload to add
-   * things like `mode`, and `metadataHash` for signedExtensions such as `CheckMetadataHash`.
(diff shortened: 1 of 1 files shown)


packages/api-base/src/types/submittable.ts
```diff
@@ -68,13 +68,17 @@ export interface SubmittableExtrinsic<ApiType extends ApiTypes, R extends ISubmi
...
+   * The result is a signed extrinsic that is ready to be broadcast to the network via `.send()`, `rpc.author.submitExtrinsic()`, or
+   * any custom submission logic.
    *
    * Note for injected signers:
-   * As of v12.0.2 and up the `SignerResult` return type for `signPayload` allows for the `signedTransaction` field.
-   * This allows the signer to input a signed transaction that will modify the payload. This
-   * The api will ensure that the Call Data is not changed. This allows for the signer to modify the payload to add
-   * things like `mode`, and `metadataHash` for signedExtensions such as `CheckMetadataHash`.
+   * As of v12.0.1 and up the `SignerResult` return type for `signPayload` allows for the `signedTransaction` field.
+   * This allows the signer to input a signed transaction that will be directly broadcasted. This
```
TarikGul, line 78: >    * This allows the signer to input a signed transaction that will be directly broadcasted. This
   * bypasses the api adding the signature to the payload. The api will ensure that the Call Data is not changed before it broadcasts the
   * transaction.
   
   This still needs to be adjusted right?

Issue #6133, opened 2025-05-06
PalletViewFunction type generation


valentinfernandez1:
With the introduction of Metadata V16 there is a new field part of the PalletMetadata, that is  [`view_functions`](https://github.com/paritytech/frame-metadata/blob/f0742f8449233421d4db60e22175e26abaf68762/frame-metadata/src/v16.rs#L258C26-L258C52) which allows pallets to expose specific query functions that don't change the storage. 

They are fairly similar to RuntimeAPI in a sense with the core difference that they are defined directly on the pallet without the need of added the same boiler palte code to `impl_runtime_apis!` every time we want to add the pallet to a Runtime.

There needs to be a way to easily call such functions from the DAPP developer's perspective, that is why I propose for a new namespace `view` (e.g. api.**view**.templatePallet.some_query()) to be added to easily call view functions.

demoncoder-crypto:
I have done a rough implementation of the issue do le me know if tis upto expecation

PR #6130, opened 2025-04-17
Metadata v16

valentinfernandez1:
This PR adds support for Substrate Metadata V16. The main structure and conversion logic are in place, so the core of the feature is done.

There are still a few things to clean up and polish on the edges, so I’m opening this as a Draft to track progress and get early feedback.

PR #6135, opened 2025-05-07
feat: Add initial support for Metadata V16 view_functions

demoncoder-crypto:
Addresses #6133


valentinfernandez1:
Thanks for putting this together, I appreciate the effort here.

That said, I think the approach might need to be rethought a bit. A lot of the files touched in this PR are auto-generated (they explicitly state **// Auto-generated via `yarn polkadot-types-from-defs`, do not edit**) and so editing them directly isn’t the right path.

Also, some of what’s being added here overlaps with the upcoming [Metadata V16 PR](https://github.com/polkadot-js/api/pull/6130) so it probably makes more sense to wait until that lands, then build on top of it, otherwise we risk redoing work or introducing conflicts. Not saying that this effort isn't valuable but it might be a good idea to hold off until the Metadata 16 stuff is merged

demoncoder-crypto:
Understood I will wait till metadata V16 Pr, also please tell on what changes or what tangent do you want me to implement the changes I am very willing to make the desired changes. Thanks :). @valentinfernandez1

valentinfernandez1:
Closing as this PR is outdated

PR #6137, opened 2025-05-09
Update Substrate types for `polkadot-sdk` v2503-2-rc1

rajk93:


valentinfernandez1 (approved):
Amazing work 🚀

TarikGul:
Question, why are we updating the types for rc1? We should only be updating the types for stable versions.

valentinfernandez1:
I checked this they are the same
image

TarikGul (approved):
Works for me! 👍 Thank you

Issue #6138, opened 2025-05-09
Update Chain types to 1.5.0


valentinfernandez1:
This issue is meant to track the upgrades of types for the different chains to the latest[ polkadot-fellows/runtime](https://github.com/polkadot-fellows/runtimes) release [1.5.0](https://github.com/polkadot-fellows/runtimes/releases/tag/v1.5.0).

- [x] Kusama 1.5.0 - Official realease date: May 9th
- [x] Polkadot  1.5.0 - Official realease date: May 14th
- [x] AH 1.5.1 -  Official realease date: May 30th

Release date can be found in the [Polkadot Technical Fellowship calendar](https://forum.polkadot.network/t/polkadot-technical-fellowship-calendar/11598)

PR #6139, opened 2025-05-09
Upgrade Kusama types to `1.5.0`

valentinfernandez1:
In alignment with issue #6138 this PR upgrades the Kusama related types to the latest polkadot-fellow/runtimes release v1.5.0.

I had to add another override for `StagingXcmV5Junction` as it was previously done for `StagingXcmV4Junction` to properly use the type instead of the lookup when generating the file.


valentinfernandez1:
As this PR contains the override for `StagingXcmV5Junction` it must be merged before merging the upcoming upgrade to the Polkadot types as this override will also be required there.

TarikGul (approved):
👍

PR #6142, opened 2025-05-12
Improve XcmV5 support

valentinfernandez1:
This PR adds the corresponding definitions for Xcm V5 allowing us to generate the types in types-augment


TarikGul (approved):
Clutch, thanks 🚀

PR #6145, opened 2025-05-13
Fix StagingXcmV5Junctions generation

valentinfernandez1:
The applied patch in #6139 was incorrectly generating the StagingXcmV5Junctions types as
```typescript
readonly asX1: StagingXcmV5Junction;
```
instead of
```
readonly asX1: Vec<StagingXcmV5Junction>;
```
This PR addresses this issue.


valentinfernandez1:
As XcmV5 support was a big part of the last release, I will immediately do a followup patch release after this PR is merged to fix XcmV5 support

TarikGul (approved):
🚀  Nice catch. That HACK i wrote a bit ago is really ugly, but I am not sure how else we can fix it. Glad you found it for StagingXcmV5Junction as well.

valentinfernandez1:
Yeap not a huge fan of this Hack, I already spotted the root cause in the type generation code. Will address it later in the week but for not this should do.

PR #6147, opened 2025-05-15
fix(types-codec): correct BitVec encoded length calculation

### LINKED ISSUE #5886: somthing wrong in bitvec codec
In the BitVec [codec](https://github.com/polkadot-js/api/blob/master/packages/types-codec/src/extended/BitVec.ts), I see the following comment:

```
 * @name BitVec
 * @description
 * A BitVec that represents an array of bits. The bits are however stored encoded. The difference between this
 * and a normal Bytes would be that the length prefix indicates the number of bits encoded, not the bytes
 */
```

However, in the decoding section, we have:
```
function decodeBitVecU8a (value?: Uint8Array): [number, Uint8Array] {
  if (!value?.length) {
    return [0, new Uint8Array()];
  }

  // handle all other Uint8Array inputs, these do have a length prefix which is the number of bits encoded
  const [offset, length] = compactFromU8aLim(value);
  const total = offset + Math.ceil(length / 8);

  if (total > value.length) {
    throw new Error(`BitVec: required length less than remainder, expected at least ${total}, found ${value.length}`);
  }

  return [length, value.subarray(offset, total)];
}
```

Can you please double-check the encoding and decoding logic for BitVec? Thanks!

rajk93:
## Description

Fixes #5886

Corrects BitVec's encoded length calculation to consider bits instead of bytes, aligning with the class documentation that states "the length prefix indicates the number of bits encoded, not the bytes".

## Changes

- Modified [encodedLength](cci:1://file:///Users/raj/code/blockdeep/api/packages/types-codec/src/extended/BitVec.ts:59:2-64:3) getter to multiply decoded length by 8 when passing to `compactToU8a()`

### DIFF
--- a/packages/types-codec/src/extended/BitVec.ts
+++ b/packages/types-codec/src/extended/BitVec.ts
@@ -61,7 +61,7 @@ export class BitVec extends Raw {
    * @description The length of the value when encoded as a Uint8Array
    */
   public override get encodedLength (): number {
-    return this.length + compactToU8a(this.#decodedLength).length;
+    return this.length + compactToU8a(this.#decodedLength * 8).length;
   }
 
   /**


rajk93:
After further review regarding the `BitVec.encodedLength` method, I’ve realized that the changes proposed in this PR are unnecessary and would introduce incorrect behavior.

The original concern was whether the length prefix was correctly calculated based on the number of bits. It is now clear that:

1. The SCALE specification requires the prefix to be the compact-encoded **number of bits**.
2. The `BitVec` class already stores this value accurately in `this.#decodedLength`.
3. The original `encodedLength` implementation:

```js
return this.length + compactToU8a(this.#decodedLength).length;
```

correctly adds the byte length of the compact-encoded bit count to the byte length of the actual bit data. This results in the correct total SCALE-encoded length.

I’m closing this PR to avoid introducing an incorrect change.

Apologies

PR #6148, opened 2025-05-19
Add type augmentation for AssetHub

### LINKED ISSUE #6115: Add Asset Hub Augmentation to types-augmentation and api-augmentation
Since a majority of the pallets are moving from the relay chain to asset-hub, it would be good to also provide augmented static types for those chains as well.

This is where it would be imported for types-augment:

https://github.com/polkadot-js/api/blob/852cbeb12357a2036d67c1bf3cf9fa9483a27a91/packages/types-augment/src/lookup/types.ts#L4-L6

The main goal here is to understand the underlying architecture and user experience behind `api-augment`, and see how we can easily add the asset-hub types here as well.

valentinfernandez1:
This PR adds Assehub Kusama and Assethub Polkadot to type augmentation.

This PR is part of the AssetHub migration that moves a lot of the relay chain functionality into Assethub meaning that for users to get access to those types we need to have them in the augmentation packages.

closes #6115


TarikGul (approved):
Really awesome, amazing job!

PR #6152, opened 2025-05-21
fix: correct BitVec length calculation from bytes to bits in codec

rajk93:
This PR corrects how `BitVec` calculates its internal bit length (`#decodedLength`) when constructed.

Previously, as discussed in [issue](https://github.com/polkadot-js/api/issues/5886) if such an input resulted in a `Uint8Array` of `N` bytes, the bit length was calculated as `N / 8`. This has been changed to `N * 8`.

This ensures accurate initialization of the bit length, leading to correct subsequent SCALE encoding.


valentinfernandez1 (approved):
@rajk93 LGTM. Changes are per our discussion. 

I would however like a second pair of eyes here so I will hold back on merging until we get a review from @TarikGul

PR #6164, opened 2025-06-04
Use ExtrinsicV4 by default

valentinfernandez1:
[An issue](https://github.com/polkadot-js/apps/issues/11602) on apps ighlighted that while `ExtrinsicV5` has been implemented, not all of its functionality is fully supported yet.

As some chains begin adopting `MetadataV16` (which supports `ExtrinsicV5`), they’re running into issues when trying to sign transactions using this version. Since resolving this will take some time, we’ve decided to temporarily disable `ExtrinsicV5` and fall back to `ExtrinsicV4` as the default until full support is in place.


TarikGul (approved):
🚀

PR #6168, opened 2025-06-13
feat(types-codec): Add maximum depth limit in `decodeU8a` functions

rajk93:
## Description
Adds a depth limit to the decoding logic to prevent issues with excessively nested complex types.


TarikGul:
~~Is there an example where there are really complex nested types? Would this break any current workflows?~~

Answered my own question :) Thanks for the PR.

PR #6183, opened 2025-07-07
Refactor signing functions to use blake2AsU8a for hashing instead of registry.hash in conditional sign of extrinsic payload

zerox-toml:
Polkadot api is signing extrinsic payload with registry.hash if payload size is bigger than 256 byte.
But in substrate it is hard coded as blake2 for conditional payload hash.
So when I sign big payload extrinsic with UI, I am getting bad signature message(when small size payload, it is working well).

https://github.com/paritytech/polkadot-sdk/blob/436b4935b52562f79a83b6ecadeac7dcbc1c2367/substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs#L559-L574
```rs
impl<Call, Extension> Encode for SignedPayload<Call, Extension>
where
	Call: Encode + Dispatchable,
	Extension: TransactionExtension<Call>,
{
	/// Get an encoded version of this `blake2_256`-hashed payload.
	fn using_encoded<R, F: FnOnce(&[u8]) -> R>(&self, f: F) -> R {
		self.0.using_encoded(|payload| {
			if payload.len() > 256 {
				f(&blake2_256(payload)[..])
			} else {
				f(payload)
			}
		})
	}
}
```

**Context**
I was working on new blockchain with using SHA3 as hash function. So I set sha3 hasher into api instance in frontend, so everything is working well, but for big payload extrinsic, I am getting bad signature message.
image
After changing api with same changes of this PR, I was able to get success messsage.

### DIFF
--- a/packages/types/src/extrinsic/util.ts
+++ b/packages/types/src/extrinsic/util.ts
@@ -4,18 +4,19 @@
 import type { SignOptions } from '@polkadot/keyring/types';
 import type { Registry } from '@polkadot/types-codec/types';
 import type { IKeyringPair } from '../types/index.js';
+import { blake2AsU8a } from '@polkadot/util-crypto';
 
 // a helper function for both types of payloads, Raw and metadata-known
 export function sign (registry: Registry, signerPair: IKeyringPair, u8a: Uint8Array, options?: SignOptions): Uint8Array {
   const encoded = u8a.length > 256
-    ? registry.hash(u8a)
+    ? blake2AsU8a(u8a)
     : u8a;
 
   return signerPair.sign(encoded, options);
 }
 
 export function signGeneral (registry: Registry, u8a: Uint8Array): Uint8Array {
-  const encoded = registry.hash(u8a);
+  const encoded = blake2AsU8a(u8a);
 
   return encoded;
 }


ap211unitech:
The CI checks are currently failing. Please review the errors and address the issues.

zerox-toml:
> The CI checks are currently failing. Please review the errors and address the issues.

Thanks, it seems lint error.
I will update it and push.

zerox-toml:
`signGeneral` function is not used for payload signing, so I changed it to `registry.hash` again.

ap211unitech:
`yarn lint` failing

zerox-toml:
> `yarn lint` failing

sorry, I missed place empty line.
just fixed it, please review it.

ap211unitech (commented):
Hey! I'm not sure this change is really needed. The current setup already uses `registry.hash`, and if no custom hasher is provided, it falls back to the default — which is [`blake2AsU8a(256)`](https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/blake2/asU8a.ts#L25).

So from what I see, this PR might not actually change anything in practice. In fact, it could introduce new issues by always forcing `blake2_256`, even when a different hasher was intentionally set at the API level.

https://github.com/polkadot-js/api/blob/master/packages/types/src/create/registry.ts#L469

```ts
  #hasher: (data: Uint8Array) => Uint8Array = blake2AsU8a;
  public hash (data: Uint8Array): IU8a {
    return this.createType('CodecHash', this.#hasher(data));
  }
```

CC: @TarikGul

ap211unitech:
Hey! Sooo... I got myself confused there — my bad 😅  The node just hardcodes it to `blake2_256` anyway no matter what hasher logic we pick, so this PR makes sense actually.

zerox-toml:
Okay, then this is available to merge to master branch?

ap211unitech:
> Okay, then this is available to merge to master branch?

I'll test it soon locally.

ap211unitech (approved):
Just tested it out — looks good to me. Let @TarikGul review it as well.

TarikGul (changes requested):
There is a greater underlying issue here and not the hasher itself.

`registry.hash` is using [`blake2AsU8a` as its default hasher](https://github.com/polkadot-js/api/blob/master/packages/types/src/create/registry.ts#L531-L533) unless a different hasher is passed in. This PR breaks critical signing functionality because it no longer allows the TypeRegistry to `setHasher`. It's important with PR's like this to look out from a wider lense and see how it can affect multiple facets. In this case substrate chains that use a different hasher will inevitably fail.

> I was working on new blockchain with using SHA3 as hash function. So I set sha3 hasher into api instance in frontend, so everything is working well, but for big payload extrinsic, I am getting bad signature message.

I presume the issue is you are setting the hasher to SHA3 but want blake2 for when the payload is above 256 bytes. In this case that doesn't fit the design of PJS since it only has support for a single hasher per instance.

zerox-toml:
Thanks for sharing your insight.
But as I mentioned in my first message, in substrate it is using blake2 hash, even if pass any different hash to registry.
We were discussing about it with @ap211unitech in here https://github.com/polkadot-js/api/pull/6183#pullrequestreview-3000453171
You can check hard coded hash in substrate side.
https://github.com/paritytech/polkadot-sdk/blob/436b4935b52562f79a83b6ecadeac7dcbc1c2367/substrate/primitives/runtime/src/generic/unchecked_extrinsic.rs#L559-L574

> registry.hash is using blake2AsU8a as its default hasher unless a different hasher is passed in.  
  
Substrate is always using blake2 for payload hashing, so we always have to use blake2 for payload hashing. And this PR doesnt disturb other hash logic(This PR only changes payload hash).

> This PR breaks critical signing functionality because it no longer allows the TypeRegistry to setHasher.

This is only applied into payload signing, so other hash logic will not be changed.

> I presume the issue is you are setting the hasher to SHA3 but want blake2 for when the payload is above 256 bytes.

This is not my desire, substrate is using blake2 with hard coded code, so we have to hard code for payload hash.

zerox-toml:
@TarikGul 
When could you check this?

TarikGul:
Soon, I had a chat with @ap211unitech today, im inclined to approve and merge in, he's just doing some due diligence with the core team.

zerox-toml:
Hi @TarikGul 
Did you have any chance to consider this request in your busy time?

josepot (approved):
This change is most definitely correct (for extrinsic v4 signed-transactions, that is, ofc).

Even frontier-based chains always use Blake2_256 for this "inner" hashing... Yes, even with ECDSA signatures, the "outer" hasher for frontier-based-chains is Keccak, but this "inner" hasher is always Blake2_256.

TarikGul:
@josepot Thanks for confirming!

TarikGul (approved):

zerox-toml:
Thanks for your support team.

zerox-toml:
Hi, team.
I was wondering if you could kindly share when the new node version was released for some urgent reason. 
I'd really appreciate any details you can provide.

ap211unitech:
Hi @zerox-toml , We'll release it next tuesday probably i.e. 29th July

PR #6191, opened 2025-07-25
support mldsa signatures

grenade:
mldsa is a post-quantum digital signature scheme standardized by NIST in FIPS 204. ML-DSA is designed to be secure against attacks from both classical and quantum computers, making it an important addition to the existing signature schemes (Ed25519, Sr25519, and ECDSA).

context:
- https://github.com/polkadot-js/common/issues/1997
- https://github.com/polkadot-js/common/pull/1998
- https://github.com/polkadot-js/api/issues/6192


josepot (changes requested):
Thanks for the contribution. Unfortunately, I don't think it makes sense to merge this PR.

While it's true that PJS aims to work out-of-the-box with chains built using the Polkadot SDK, there are a few foundational assumptions that must be respected. One of these is compliance with the canonical [MultiSignature variants](https://spec.polkadot.network/id-extrinsics#defn-extrinsic-signature).

As mentioned in the code-comment, the upstream `MultiSigner` enum is in the process of being extended with a new `Eth` variant. The changes proposed here would conflict with that [upcoming addition](https://github.com/paritytech/polkadot-sdk/pull/8757/files#diff-15823f4d311ab7b169b6bff56d5fbe9033496871d19b08f8b0abdc1ab6318fc5R369-R375).

We have to be cautious about opening the door to every chain-specific signing scheme. You surely must understand that we can't have every single chain which decides to use different signing-schemas to come here and fight for the `Enum` that works best for them, right?

Additionally, PJS currently makes [specific assumptions about signing behaviors](https://github.com/polkadot-js/api/pull/6183/files#diff-b6b1d2f12edc620bd0694f39e4d511ef959885159b725435b512763eae48efc0R11-R14) that likely don't hold true for these new signing schemes.

Given these constraints, the viable options would be:

1) Maintain a fork of PJS tailored to your chain's needs.
2) Consider using a more modular library with fewer built-in assumptions around signature formats.

grenade:
no worries. understood. every ecosystem has its priorities. ours is security that stands a chance in a pq world. i wouldn't have felt right forking without at least offering to contribute the pq support back upstream, but if it isn't wanted, we'll fork.

josepot:
> We have to be cautious about opening the door to every chain-specific signing scheme. You surely must understand that we can't have every single chain which decides to use different signing-schemas to come here and fight for the `Enum` that works best for them, right?

I mean, have you considered using extrinsic v5 in your chain?

It should be 100% possible, using extrinsic v5, to have transaction-extensions that use pq signing-schemes.

grenade:
in all honesty it will be less overhead for us to fork and deprecate support for non-pq schemes which are irrelevant to any chain that cares about pq safety. we won't have the baggage of caring about whether we break something for non-pq schemes. i think we needed to make the pr as an acknowledgement of the shoulders being stood on, but i can understand completely if this library's focus is on maintaining support for legacy schemes in a pre quantum world.

josepot:
> in all honesty it will be less overhead for us to fork and deprecate support for non-pq schemes which are irrelevant to any chain that cares about pq safety.

With extrinsic v5 you wouldn't even have to deprecate anything, b/c you could just use an extension that uses the signing schemes that your chain cares about.

I'm not trying to convince not to fork PJS. I'm just trying to help you! Please, do whatever works for you, ofc! Also, the fact that I'm pointing you towards extrinsic v5 should be a pretty convincing indicator of that, because this library doesn't currently support it and I have a **very hard time** imagining how or when it will be able to **_fully_** support it.

That being said: if you want to fork, you are most welcome to do so! However, please be aware of 2 things:
1) This library has a ton of technical debt, and it's bloated and tightly coupled to stuff that you probably don't need. 
2) There are newer, leaner and more performant [alternatives](https://papi.how/) that are perhaps could be better suited for your needs.

> but i can understand completely if this library's focus is on maintaining support for legacy schemes in a pre quantum world.

No, that's not it. It's just that b/c the way in which this library was designed years ago, it's not really feasible to accommodate for custom signed-schemes **on extrinsic v4**.

One of the reasons why extrinsic v5 exists is, in fact, to be able to account for a post-quantum world 🙂.

PR #6218, opened 2025-11-05
Upgrade Kusama and Kusama Asset Hub types for 1009003

ap211unitech:
Since chains are already live now, I simply followed these steps:

I first changed the versions in: `scripts/metadata-get.mjs` to be V15:

```
const META_VERSION_HEX = '0x0F000000';
const META_VERSION = 15;
```

then

```
node ./scripts/metadata-get.mjs --url https://kusama-rpc.n.dwellir.com # for kusama
node ./scripts/metadata-get.mjs --url https://kusama-asset-hub-rpc.polkadot.io # for kusama asset hub
```

and

```
yarn build:interfaces
yarn test
```


TarikGul:
LGTM

PR #6226, opened 2025-11-14
Initialize all `createType` and `createTypeUnsafe` with some default values

ap211unitech:
## 📝 Description

This PR initializes all `createType` and `createTypeUnsafe` values explicitly. While the API normally handles these defaults automatically, Firefox v145.0 introduces stricter behavior during type decoding, which causes the automatic defaulting to fail. Explicit initialization ensures the API works consistently across browsers.


ap211unitech:
Closing since issue was originally with Firefox itself and team released a patch version, so all good now.

PR #6270, opened 2026-07-21
fix(rpc-provider): repair LRU list corruption when the tail entry is …

### LINKED ISSUE #6269: LRUCache: get() on the LRU entry corrupts the list and leaks memory
## Description

The `LRUCache` used for the RPC provider call cache ([packages/rpc-provider/src/lru.ts](https://github.com/polkadot-js/api/blob/v16.5.6/packages/rpc-provider/src/lru.ts)) corrupts its internal double-linked list whenever `get()` returns the least-recently-used entry that is also its `#tail`.

The core defect is in `#toHead()`: it runs the generic unlink (`ref.prev.next = ref.next; ref.next.prev = ref.prev`) even when `ref` is the tail, and never retargets `#tail`. The node is moved to the head, but `#tail` is left pointing at the same node, so `#head === #tail` while `length > 1`.

Once `#head === #tail` with `length > 1`, the cache is permanently corrupted:

- `keys()` / `entries()` truncate (the head-to-tail walk terminates immediately);
- capacity eviction (`#data.delete(this.#tail.key)`) deletes the **most**-recently-used entry instead of the least;
- the maps and the list diverge: entries whose nodes have fallen off the walkable list can never become the tail, so neither capacity nor TTL eviction can ever remove them, while the nodes of already-evicted keys stay pinned through the stale `next`/`prev` chains of nodes the cache still holds.

Standalone repro:

```bash
npm i @polkadot/rpc-provider@16.5.6
```

```js
import { LRUCache } from '@polkadot/rpc-provider/lru';

const lru = new LRUCache(4);

for (const k of ['1', '2', '3', '4']) {
  lru.set(k, `${k}${k}${k}`);
}

console.log('keys before get :', lru.keys());

lru.get('1'); // cache hit on the least-recently-used entry (the list tail)

console.log('keys after get  :', lru.keys());
console.log('lengthData      :', lru.lengthData);

lru.set('5', '555'); // at capacity - should evict '2', the LRU entry

console.log("get('1')        :", lru.get('1'));
console.log("get('2')        :", lru.get('2'));
```

Expected behavior: a `get()` hit on any entry, including the tail, moves it to the head, keeps every entry reachable, and preserves LRU eviction order. Expected output (produced by re-running the same script with the fix from the linked PR applied):

```
keys before get : [ '4', '3', '2', '1' ]
keys after get  : [ '1', '4', '3', '2' ]
lengthData      : 4
get('1')        : 111
get('2')        : null
```

Actual output on 16.5.6:

```
keys before get : [ '4', '3', '2', '1' ]
keys after get  : [ '1' ]
lengthData      : 4
get('1')        : null
get('2')        : 222
```

That is: after the `get('1')` tail hit, 3 of the 4 entries are no longer visible via `keys()`/`entries()` while the backing map still holds all 4; the subsequent `set('5')` then evicted `'1'` (the most-recently-used entry) while `'2'` (the least-recently-used, which should have been evicted) survived.

Note the trigger is routine API usage: any cache hit on the oldest cached call within the TTL window (e.g. repeated `api.at()` reads of the same hashes within the 30s default TTL). Reducing `cacheCapacity` does not avoid it (any capacity >= 2 is affected).

## Environment
  - Version: 16.5.6 (latest stable; still present on current master)
  - First observed on Bun, but reproduced on Node.js

carpntr:
Fixes #6269

## Problem

A `get()` hit on the entry currently at the internal list tail corrupts `LRUCache`'s doubly-linked list.  See #6269 for more.

## Fix

Handle the tail case explicitly in `#toHead()`: retarget `#tail` to `ref.prev` and close the list behind the node being moved to the head (`this.#tail.next = ref`). Note the generic unlink cannot simply be reused with a retarget added: `tail.next` is written but never maintained (the first node inserted even keeps its constructor self-loop, `next = prev = this`), so dereferencing it via `ref.next.prev = ref.prev` can write into an arbitrary stale node.

## Tests

Three new unit tests in `lru.spec.ts`, all failing on master before the fix:

1. `retains all entries when the oldest entry is retrieved` — after a `get()` on the tail entry, `keys()` collapses from 4 entries to 1 on master.
2. `evicts the least-recently-used entry after the oldest entry is retrieved` — at capacity, master evicts the just-used entry instead of the LRU one.
3. `keeps the list and data in sync over repeated oldest-entry retrievals` — sustained set/get cycling; on master `keys().length` diverges from `length`/`lengthData`/`lengthRefs` on the second iteration.

`yarn test:one packages/rpc-provider/src`:  123 pass / 0 fail (1 pre-existing skip). 
Full monorepo `yarn test`: 12,866 pass / 0 fail.

### DIFF
--- a/packages/rpc-provider/src/lru.spec.ts
+++ b/packages/rpc-provider/src/lru.spec.ts
@@ -59,6 +59,49 @@ describe('LRUCache', (): void => {
     expect(lru?.length === lru?.lengthData && lru?.length === lru?.lengthRefs).toBe(true);
   });
 
+  it('retains all entries when the oldest entry is retrieved', (): void => {
+    const keys = ['1', '2', '3', '4'];
+
+    keys.forEach((k) => lru?.set(k, `${k}${k}${k}`));
+
+    // '1' is the least-recently-used entry, i.e. the internal list tail
+    expect(lru?.get('1')).toBe('111');
+
+    expect(lru?.keys().length).toBe(lru?.lengthData);
+    expect(lru?.entries()).toEqual([['1', '111'], ['4', '444'], ['3', '333'], ['2', '222']]);
+    expect(lru?.length === lru?.lengthData && lru?.length === lru?.lengthRefs).toBe(true);
+  });
+
+  it('evicts the least-recently-used entry after the oldest entry is retrieved', (): void => {
+    const keys = ['1', '2', '3', '4'];
+
+    keys.forEach((k) => lru?.set(k, `${k}${k}${k}`));
+
+    lru?.get('1');
+    lru?.set('5', '555');
+
+    // the get('1') made '2' the least-recently-used entry, so it is the
+    // one dropped by the set('5') above, with '1' itself retained
+    expect(lru?.get('2')).toBe(null);
+    expect(lru?.get('1')).toBe('111');
+    expect(lru?.entries()).toEqual([['1', '111'], ['5', '555'], ['4', '444'], ['3', '333']]);
+    expect(lru?.length === lru?.lengthData && lru?.length === lru?.lengthRefs).toBe(true);
+  });
+
+  it('keeps the list and data in sync over repeated oldest-entry retrievals', (): void => {
+    for (let i = 0; i < 32; i++) {
+      lru?.set(`${i}`, `value${i}`);
+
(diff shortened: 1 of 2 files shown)


TarikGul (approved):
Thanks so much for the fix, confirmed, and everything looks good. 🚀

PR #6274, opened 2026-08-10
fix(rpc-provider): don't error-log a normal (1000) disconnect of internal subscriptions

SnowingFox:
### Problem

`disconnect()` error-logs the API's own internal `subscribeRuntimeVersion` subscription on a normal (`1000`) close:

```
RPC-CORE: subscribeRuntimeVersion(): RuntimeVersion:: disconnected from wss://.../: 1000:: Normal Closure
```

`1000` is a deliberate, user-initiated close. The line goes through `l.error`, so it is not gated by `DEBUG` and cannot be suppressed. In the common "load metadata then release the socket" pattern it shows up twice per disconnect (one rejected pending subscribe request per registry/path set up during init). Cosmetic — the disconnect itself works — but it is spurious error noise for a normal teardown.

### Root cause

`disconnect()` (`packages/api/src/base/index.ts`) tears down the internal subs and then closes the socket with `close(1000)`. `WsProvider.disconnect()` already sets `#autoConnectMs = 0` before closing, which suppresses the *provider's* own disconnect log (`packages/rpc-provider/src/ws/index.ts`). That awareness does not reach the handler-rejection loop: `#onSocketClose` still rejects every pending handler with a `disconnected from ...: 1000:: Normal Closure` error, and `rpc-core`'s `logErrorMessage` (`packages/rpc-core/src/bundle.ts`) error-logs that rejection because it has no way to know the close was deliberate.

### Fix

Two small, additive changes that let the provider's "this close was normal" knowledge reach `rpc-core`:

- `packages/rpc-provider/src/ws/index.ts`: on a normal (`1000`) close, mark the disconnect error with `error.isNormalClose = true`.
- `packages/rpc-core/src/bundle.ts`: `logErrorMessage` skips the `l.error` when the error carries `isNormalClose`.

The error is still propagated to subscribers (pending promises/observables still reject/error), so nothing hangs — only the spurious error-level log is removed.

### Tests

- `packages/rpc-provider/src/ws/disconnect.spec.ts`: a real `WsProvider` + mock socket asserts that a pending request rejected by a `1000` close receives the normal-close marker (this failed before the provider change).
- `packages/rpc-core/src/disconnect.spec.ts`: asserts no error-level log is emitted when a pending subscription is torn down by a normal (`1000`) close, and that an abnormal close is still error-logged — guarding against hiding genuine errors.

`yarn test:one rpc-core` and `yarn test:one rpc-provider` both pass.

### Scope note

This is a cosmetic logging-level fix, scoped strictly to close code `1000` ("Normal Closure"). All other close codes still error-log as before. The Paseo runtime-gadgets/transaction-failure concern raised in the issue comments is a separate, deeper runtime issue and is intentionally not addressed here.


flipchan:
did you try this on paseo assethub(`wss://sys.turboflakes.io/asset-hub-paseo`)?

 does it patch the following?


```js
provider: wsProvider,
config: {
chain: CHAINSPEC.properties || { ss58Format: 42, tokenDecimals: 10, token
ymbol: "PAS" },
},
});
console.log("✅ Connected to", (await api.rpc.system.chain()).toString());
const srKeyring = new Keyring({ type: "sr25519", ss58Format: 42 });
const account1 = srKeyring.addFromSeed(ACCOUNT_1_SEED);
const account2 = srKeyring.createFromUri("//Alice");
console.log("\n📝 From:", account1.address);
console.log("📝 To:", account2.address);
const balance1 = await api.query.system.account(account1.address);
console.log("💰 Balance:", Number(balance1.data.free) / 1e10, "PAS");
// Create the transfer
const transfer = api.tx.balances.transferKeepAlive(account2.address, 1_000_00_000n);`
```


all my substrate transactions are failing today: 
```
1002: Verification Error: Runtime error: Execution failed: Execution aborted due to trap: wasm trap: wasm unreachable instruction executed WASM backtrace: error while executing at wasm backtrace: 0: 0x9c9bf8 - asset_hub_paseo_runtime.wasm!__rustc[d131491b17107b07]::rust_begin_unwind 1: 0x5350 - asset_hub_paseo_runtime.wasm!core::panicking::panic_fmt::h076e68bde4b88b39 2: 0x92ed6b - asset_hub_paseo_runtime.wasm!TaggedTransactionQueue_validate_transaction: RuntimeApi, Execution failed: Execution aborted due to trap: wasm trap: wasm unreachable instruction executed\nWASM backtrace:\nerror 
```

PR #6278, opened 2026-09-20
fix(api-derive): resolve staking exposure at the era the stash was elected for

### LINKED ISSUE #6277: `derive.staking.electedInfo` returns next-era validators but resolves their exposure at `activeEra`
`electedInfo` builds its account list from `nextElected`, which is read from `currentEra`, then fetches exposure for that list at `activeEra`. For validators entering the set at the next era the two eras disagree, so their exposure is `None` and they come back with no stake.

- packages/api-derive/src/staking/validators.ts:  nextElected() returns keys of staking.erasStakersOverview(currentEra)
- packages/api-derive/src/staking/electedInfo.ts:  combineAccounts(nextElected, validators) unions that with the current session's validators
- packages/api-derive/src/staking/query.ts:  queryMulti() takes activeEra from session.indexes() and queries erasStakersOverview(activeEra, s) / erasStakersPaged(activeEra, s, page)

Evidence on chain: Polkadot Asset Hub, activeEra 2298, currentEra 2299, 10 validators entering:

```js
const { info, nextElected } = await api.derive.staking.electedInfo({
  withController: true, withExposure: true, withExposureMeta: true, withPrefs: true, withLedger: true
});
// for each of the 10 accounts in nextElected but not in session.validators():
//   exposureMeta.isNone === true
// while staking.erasStakersOverview(2299, acc) has total 1.33M to 1.67M DOT
```

**NOTE**: The bug is only visible while `currentEra > activeEra`.

Downstream effect: Consumers treat zero exposure as "not active" and substitute the ledger. In polkadot-js/apps this surfaces on the `Targets` tab see  packages/page-staking/src/useSortedTargets.ts:143-151 where we set `bondTotal = bondOwn = stakingLedger.total`, and packages/page-staking/src/Targets/Validator.tsx:133-134 renders those as "total stake" and "own stake". 

Entering validators therefore display their self-bond as their backing, identical in both columns. On Polkadot AH at era 2299, that reads as 10,001 DOT and 100,000 DOT for validators actually backed by 1.33M to 1.67M DOT.

sigurpol:
Closes #6277

### Bug
`electedInfo` lists accounts from `currentEra` but resolved their exposure at `activeEra`. On Asset Hub `CurrentEra` is bumped at the first session of an era, so the two differ for most of it. Stashes newly elected into the planned era have no active-era exposure and came back unexposed. Callers read that as "not backed" and fall back to the ledger.
 
At era 2298/2299 on Polkadot AH, ten **new** validators rendered as a 10,001-100,000 DOT self-bond against 1.33M-1.67M real backing.

### Fix
- `queryMulti(accountIds, flags, page?, era?)` resolves exposure at `era` (default: active) and reports it as `DeriveStakingQuery.era`.
-  `electedInfo` reads each stash at the era it was elected for: those listed in `erasStakersOverview(currentEra)` at the planned era, and the current validators missing from that list at the active era. 
- The claimed-rewards window stays anchored on the active era.

### Behavioural change

Re-elected validators now report the stake they were elected with rather than their active-era backing, so the elected set ( and the aggregates computed over it) come from one era. 

`DeriveStakingQuery.era` is optional, so no type break.


sigurpol:
@TarikGul , @kianenigma , @ank4n

TarikGul (approved):
👍  Looks solid, thanks!

TarikGul:
The api is in need of a release, so let me look into if the npm keys are still valid tomorrow.

sigurpol:
> The api is in need of a release, so let me look into if the npm keys are still valid tomorrow.

Thanks for looking into this :bow:

